PR #6, opened 2017-05-15
Coq extraction script + simple ocaml translation

bkase:
Ported this code from `hazel_semantics.ml`

```ocaml
  let matched_arrow ty = match ty with
    | Arrow (ty1, ty2) (* MAArr *) -> Some (ty1, ty2)
    | Hole (* MAHole *) -> Some (Hole, Hole)
    | _ -> None
```

The coq definition extracts to this:

```ocaml
  (** val matched_arrow : t -> (t, t)  option **)

  let matched_arrow = function
  | Arrow (ty1, ty2) -> Some (ty1, ty2)
  | Hole -> Some (Hole, Hole)
  | _ -> None
 end
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ setup.log
 *.swo
 *.swp
 
+# coq
+*.glob
+*.vo
+*.vo.aux
+core/ml/*
--- a/core/extract-coq.sh
+++ b/core/extract-coq.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -eou pipefail
+
+SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+mkdir -p "$SCRIPTPATH/ml"
+coqtop -compile "$SCRIPTPATH/hazel_core" > "$SCRIPTPATH/ml/hazel_core.ml"
+
--- a/core/hazel_core.v
+++ b/core/hazel_core.v
@@ -95,11 +95,19 @@ Module HTyp.
      simpl; 
      reflexivity.
    Qed.
-  
+
+   Definition matched_arrow (ty : t) : option (t * t) :=
+     match ty with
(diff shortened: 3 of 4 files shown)


bkase:
I think travis fails because this branch is too old, I'll get it up-to-date after this lands

ivoysey:
It's interesting to me that this uses options. I'm not super familiar with Coq syntax so I might be totally wrong, but it looks like you're not defining a datatype or a judgement just because you have to specify what happens when it "doesn't work". It's not just a strictly positive characterization like you'd see in the on-paper mathematics.

Or am I missing something entirely?

The corresponding snippet of the Agda mechanization is:

```agda
  data _▸arr_ : τ̇ → τ̇ → Set where
    MAHole : ⦇⦈ ▸arr (⦇⦈ ==> ⦇⦈)
    MAArr  : {t1 t2 : τ̇} → (t1 ==> t2) ▸arr (t1 ==> t2)
```

cyrus-:
What Brandon is doing is defining a Coq implementation of that spec, corresponding directly to the extracted OCaml. What we can and should do is _also_ define the spec as a dependent type, exactly as in Agda, and then prove the implementation sound and complete with respect to the spec.

(The most magic would be to extract the implementation directly from the spec, using tactics or something, but I don't know how to do that...)

ivoysey:
Ah! That makes more sense. I guess I'd assumed that Coq to OCaml extraction was exactly that magic---that is, that it would take the spec as a dependent type and produce a function from if it had mode (∀, ∃!).

cyrus-:
I mean if you want to investigate how to do that, that would definitely save effort. Maybe Adam C.'s book has some insight. But it's not trivial. For now, at least we have an implementation in a language where we can prove theorems about it if we want to.

bkase:
I think you could encode something like this a bit more safely in Ocaml with GADTs but it looks like Coq doesn't know how to extract to GADTs

PR #7, opened 2017-05-16
Extract Ocaml hazel_semantics to Coq

bkase:
Used [coq-of-ocaml](https://github.com/clarus/coq-of-ocaml) to extract
Ocaml to Coq as a baseline. See [paper](https://ocaml.org/meetings/ocaml/2014/ocaml2014_15.pdf)

Currently the extracted Ocaml from this code is garbage, but it does
compile properly in Coq.

Will go through and fixup all the effectful code that was generated so
that the Coq -> Ocaml extraction works properly. At least we know the
data types are consistent with the Ocaml hazel_semantics.ml now.

I wanted to see if Ocaml -> Coq (f) and then Coq -> Ocaml
(g) behaved such that (f o g) is equivalent to (f o g o f o g)
(this means the two extractions would form an adjunction right?) but the
Coq -> Ocaml extraction fails on the second pass.

### DIFF
--- a/core/hazel_core.v
+++ b/core/hazel_core.v
@@ -1,4 +1,9 @@
 Require Coq.Bool.Bool. Open Scope bool.
+Require Import OCaml.OCaml.
+
+Local Open Scope Z_scope.
+Local Open Scope type_scope.
+Import ListNotations.
 
 Lemma and_proj_1 : forall b1 b2 : bool,
   (b1 && b2) = true -> b1 = true.
@@ -13,29 +18,29 @@ Proof.
 Qed.
 
 Module HTyp.
-
-  Inductive t : Set := 
-    Num : t
+  Inductive t : Type :=
+  | Num : t
   | Arrow : t -> t -> t
   | Sum : t -> t -> t
-  | Hole.
-  
-  Fixpoint eq (x y : t) : bool := 
-    match (x, y) with 
-    | (Num, Num) => true 
-    | (Arrow x' y', Arrow x'' y'') 
-    | (Sum x' y', Sum x'' y'') => 
-      (eq x' x'') && (eq y' y'')
+  | Hole : t.
+
+  Fixpoint eq (ty1 : t) (ty2 : t) : bool :=
+    match (ty1, ty2) with
+    | (Num, Num) => true
+    | (Arrow ty1_left ty1_right, Arrow ty2_left ty2_right) =>
+      andb (eq ty1_left ty2_left) (eq ty1_right ty2_right)
+    | (Sum ty1_left ty1_right, Sum ty2_left ty2_right) =>
+      andb (eq ty1_left ty2_left) (eq ty1_right ty2_right)
     | (Hole, Hole) => true
     | _ => false
     end.
-  
+
   Theorem eq_refl : forall (x : t),
     eq x x = true.
   Proof.
     induction x; (simpl; auto with *).
   Qed.
-  
+
   Theorem eq_sound : forall x y : t,
     x = y -> (eq x y = true).
   Proof.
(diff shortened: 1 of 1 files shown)


ivoysey:
I don't know about adjoint, but certainly self-inverse. It's interesting that it fails that test -- what specifically goes wrong? Maybe someone needs to formalize coq2ocaml and ocaml2coq in coq and prove the round trips are stable, and then extract the implemen... oh. 😄

bkase:
Haha, I did suspect it wouldn't be a simple inverse because coq has a much richer type system. The free functor would be ocaml to coq then the forgetful functor would be coq to ocaml. Then it's an adjunction if the round tripping worked after a few cycles! Of course this also doesn't work 🤷‍♂️

cyrus-:
Do you want me to merge this or?

bkase:
@cyrus- I requested a review, I can merge when you accept. Github has reviewing built-in: The NUX is not great I guess (I had to ask someone how to do it my first time reviewing). Here's how it works:

See this thing on the bottom of this conversation view:

review requested

That means that I requested a review (you can also see your name on the top-right). On the "files changed" tab on this page, there's a "Review Changes" button on the top right:

Review button

You can check the box that approves the PR there. Or you can check "Request Changes" if you want me to change something (and you can also comment on specific lines directly).

See https://help.github.com/articles/requesting-a-pull-request-review/ for more.

Hope that helps.

Industry has made me love code reviews.

core/hazel_core.v
```diff
@@ -72,42 +77,469 @@ Module HTyp.
...
+    end.
 End HTyp.
 
+Module Var.
+  Definition t := Z.
+
+  Definition equal (x : t) (y : t) : bool := equiv_decb x y.
+End Var.
+
+Module Ctx.
```
cyrus-: what happened to the signature? pretty sure Coq supports abstract types in modules right?

core/hazel_core.v
```diff
@@ -72,42 +77,469 @@ Module HTyp.
...
+  | Case : t -> (Var.t * t) -> (Var.t * t) -> t
+  | EmptyHole : t
+  | NonEmptyHole : t -> t.
+
+  Definition IllTyped := Effect.make unit unit.
+
+  Definition raise_IllTyped {A : Type} (x : unit) : M [ IllTyped ] A :=
+    fun s => (inr (inl x), s).
+
+  Fixpoint syn_rec (counter : nat) (ctx : Ctx.t) (e : t)
```
cyrus-: hopefully we don't actually need to use fuel like this after we get rid of all this effect stuff and just use option. (is there not an option in the conversion script to use options?)

bkase: This is an option in the conversion to not use fuel and to rely on Coq's totality checker, but not an option to rewrite exceptions to an option return value. I attempted that "auto-totality check" on all the recursive functions, but it failed on a few (the mutually recursive syn and ana in this case).

I'm hoping that when it's refactored to use options instead of exceptions that the totality checker will be happy.

core/hazel_core.v
bkase: Good catch, looks like the coq_of_ocaml missed that here.
Any idea why this fails to build at the "Parameter extend" line:

```
Module Type Ctx.
  Parameter t : Type.
  Parameter empty : t.
  Parameter extend : t => Var.t * HTyp.t => t.
  Parameter lookup : t => Var.t => HTyp.t option.
End Ctx.
```

Error is:
```
File "/Users/bkase/hazel/core/hazel_core.v", line 149, characters 23-25:
Syntax error: '.' expected after [vernac:gallina] (in [vernac_aux]).
```

I literally could not find one example on the internet of someone using Module Type in Coq to declare a "function signature" or (theorem signature? whatever it's called in proof land)

bkase:
I'm going to merge this for now. Will make sure to fix the module signatures that got erased

PR #8, opened 2017-05-25
Attempt to rewrite syn/ana with options

bkase:
Subsumption case confuses coq, not sure how to workaround it

### DIFF
--- a/core/hazel_core.v
+++ b/core/hazel_core.v
@@ -143,18 +143,24 @@ Module Var.
   Definition equal (x : t) (y : t) : bool := equiv_decb x y.
 End Var.
 
-Module Ctx.
+Module Type CTX.
+  Parameter t : Type.
+  Parameter empty : t.
+  Parameter extend : t -> Var.t * HTyp.t -> t.
+  Parameter lookup : t -> Var.t -> option HTyp.t.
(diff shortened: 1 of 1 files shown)


ivoysey:
@cyrus- I'm pretty sure the issues @bkase is having with subsumption are exactly because (under the iso into proof theory) our bidirectional stuff is doing chaining not really full focusing. Taken as written, you can get stuck in cycles of applying subsumption with these rules because the intent was that subsumption is a last-resort but that is not enforced in any way -- it's purely metatheoretic.

The last time this manifested itself was with the proof of determinism, that we solved with `aasubsume-min`. That was a little different because it was on the rule in the action semantics induced by the typing subsumption rule. We only didn't have to deal with it in both places as a quirk of the structure of the theorems we chose to prove.

cyrus-:
No I don't think so -- you can't get stuck in cycles of applying subsumption repeatedly given the functional implementation, which does in fact apply subsumption last.

I think the problem is just that the termination argument is a little subtle -- calls to ana from syn obey the sub-formula property, but the call to syn in ana does not. The termination checker might not be able to understand that it's a round-trip that has a strictly decreasing metric.

PR #18, opened 2017-06-22
Port to ReasonML

bkase:
Sucks to put this into the coq_core branch, but given that the hazel_semantics is going to be replaced anyway, it's probably okay.


cyrus-:
looks fine, but let's get this stuff merged into master soon!

PR #19, opened 2017-06-25
performSyn (not performAna) ported

bkase:
For some reason this part got lost in the ocaml->coq extraction. This is the first half of it. Coq is happy up to the point where it wants performAna to invoke performSyn (typechecking passes on  the performSyn part)


cyrus-:
Yeah I think we'll have to do the fuel trick here too.

PR #20, opened 2017-06-28
New version install steps

asolove:
Things that tripped me up while installing for the first time:

- Remember to change directory context after done pinning reason
- Need a few more opam dependencies


cyrus-:
Thanks!

PR #59, opened 2019-03-01
Keywords

dmoon1221:
Makes `let` and `case` reserved keywords. Variables in both pattern and expression position called `let` or `case` are unbound and given a new variable error status indicating that they are reserved keywords. Typing out the variables `let` or `case` and then hitting space creates a let line (and possibly an empty hole expression body if there is no line already following the let line) or a case expression, respectively.

A future refinement would include variable splitting. If I have `1|2` and hit plus, then I would get `1 + 2`. In the case that I type out a keyword, I would construct the corresponding expression, e.g., pressing space on `let|1` gives me `let _ = 1`.


dmoon1221:
Additional changes:
- Construction of let and case expressions are now limited to the beginning of lines and parenthesized expressions.
- Constructing a let or case expression through keyword is equivalent to constructing through the single key command.
- Construction of case expressions consumes subsequent lines into the first rule clause.
- Case expressions are no longer considered to be bidelimited -- this is to be consistent with the requirement that constructing a case in the middle of an opseq requires that you first parenthesize the desired scrutinee.

dmoon1221:
@cyrus- ready for review

dmoon1221:
Gonna enable keyword expression construction in the middle of opseqs after all. However, this turns out to be kinda tricky for `case` expressions because a `case` expression requires a type ascription in synthetic position, so the case expression needs to be constructed differently depending on whether it's in analytic or synthetic position within the opseq. https://github.com/hazelgrove/hazel/pull/63 lays the groundwork for this analytic/synthetic-dependent case expression construction. Merging these changes in the meantime, then will circle back to implement keyword expression construction in the middle of opseqs once https://github.com/hazelgrove/hazel/pull/63 is merged.

PR #62, opened 2019-03-09
Use ppx_sexp_conv for serialization/deserialization

dmoon1221:
Given frequent changes to the semantics, it is challenging to maintain a custom serializer and parser for a readable text syntax. Use [ppx_sexp_conv](https://github.com/janestreet/ppx_sexp_conv) instead until semantics stabilize.


nickcollins:
looks good

PR #64, opened 2019-03-25
Fix all warnings emitted with "dev" dune profile.

nmsmith:
The fixes are mostly a combination of removing unused imports, removing unnecessary rec keywords, and removing/underscoring unused variables.


src/semantics/ZExp.re
```diff
@@ -73,15 +73,15 @@ let rec make_inconsistent = (u_gen: MetaVarGen.t, ze: t): (t, MetaVarGen.t) =>
...
+  | Deeper(NotInHole, _)
+  | Deeper(InHole(WrongLength, _), _) =>
     let (u, u_gen) = MetaVarGen.next(u_gen);
     let ze' = set_err_status(InHole(TypeInconsistent, u), ze);
     (ze', u_gen);
   | Deeper(InHole(TypeInconsistent, _), _) => (ze, u_gen)
   | ParenthesizedZ(ze1) =>
     let (ze1', u_gen) = make_inconsistent(u_gen, ze1);
-    (ParenthesizedZ(ze1), u_gen);
+    (ParenthesizedZ(ze1'), u_gen);
```
nmsmith: I think I fixed a bug here? Someone needs to confirm this.

src/semantics/ZPat.re
```diff
@@ -68,7 +68,7 @@ and erase' = (zp': t'): UHPat.t' =>
     OpSeq(skel, OperatorSeq.opseq_of_exp_and_surround(p1, surround));
   };
 
-let is_after = (zp: t): bool =>
+let _is_after = (zp: t): bool =>
```
nmsmith: Because this isn't used anywhere I have to put an underscore to please the linter. If someone eventually needs to use it we can just delete the _.

src/util/HazelUtil.re
```diff
@@ -359,15 +359,15 @@ module ListMinTwo = {
     switch (xs, ys) {
     | (_, []) => xs
     | (Pair(x1, x2), [y]) => Cons(x1, Pair(x2, y))
-    | (Cons(x, xs), [y]) => Cons(x, append_list(xs, ys))
-    | (_, _) => append(xs, to_tuple_list(ys))
+    | (Pair(_, _), _) => append(xs, to_tuple_list(ys))
```
nmsmith: I made this pattern more explicit, but it should behave the same.

src/semantics/Action.re
```diff
@@ -2353,12 +2349,12 @@ let rec perform_syn =
...
       | Tm(_, EmptyHole(_)) =>
         let (rule_p, u_gen) = UHPat.new_EmptyHole(u_gen);
-        let (rule_e, u_gen) = UHExp.new_EmptyHole(u_gen);
+        let (rule_e, ___u_gen) = UHExp.new_EmptyHole(u_gen);
         let rule = UHExp.Rule(rule_p, rule_e);
         ZExp.Deeper(NotInHole, CaseZE(ze1, [rule], Some(Hole)));
       | _ =>
         let (zp, u_gen) = ZPat.new_EmptyHole(u_gen);
-        let (rule_e, u_gen) = UHExp.new_EmptyHole(u_gen);
+        let (rule_e, ___u_gen) = UHExp.new_EmptyHole(u_gen);
```
nmsmith: The fact that these u_gens are *supposed* to shadow each other, despite the last instance not being used, is a tricky thing to handle. I'm assuming if this code ever needs to be touched the reader will notice that they need to remove these underscores.

src/semantics/ZExp.re
cyrus-: ah yeah I believe so

src/semantics/ZExp.re
```diff
@@ -224,7 +224,7 @@ let rec place_before = (e: UHExp.t): t =>
   | Tm(err_status, LineItem(ExpLine(e1), e2)) =>
     let ze1 = place_before(e1);
     Deeper(err_status, LineItemZL(DeeperL(ExpLineZ(ze1)), e2));
-  | Tm(err_status, LineItem(LetLine(_, _, _), _)) =>
+  | Tm(_err_status, LineItem(LetLine(_, _, _), _)) =>
```
cyrus-: can just use _ for unused variables

src/semantics/ZPat.re
cyrus-: it complains even about top-level definitions that are unused? do we need to make .rei files to specify a public interface?

src/semantics/Action.re
cyrus-: leave a comment to clarify what's going on here (maybe @dmoon1221 should do that)

src/semantics/ZPat.re
nmsmith: Yeah a .rei may resolve this warning

src/semantics/Action.re
dmoon1221: I think this [commit](https://github.com/hazelgrove/hazel/commit/92c775eabeb08d832b2946a301299b177be1948e) resolves this.

src/semantics/ZPat.re
nmsmith: Turns out this definition was being shadowed! There was a newer definition at the bottom of the file (I checked the timestamps), so given that this version wasn't used I deleted it.

PR #66, opened 2019-06-06
Cursor on delimiters

dmoon1221:
- Modifies our cursor model to support cursors on syntactic delimiters and infix operators. When the cursor is on a delimiter, the owner node of the delimiter is highlighted as the selected expression; when the cursor is on an operator, the operator subsequence whose root is that operator is highlighted.
- Introduces a `CursorEscape` mechanism by which pressing Backspace at the beginning of a node pops the cursor out to the preceding cursor position (e.g., the end of a preceding delimiter); pressing Delete at end of a node has similar behavior.
- Otherwise, if not `CursorEscape`, deleting a delimiter deletes the owner of the delimiter (including all its children). One exception is when there is only a single child node of the same type as the parent node, in which case deleting a parent delimiter unwraps the parent node and leaves the child node in place.
- Implements `do_transport` cases for delimiters, such that the cursor can only be at the start and end of delimiters.
- Adds cursor on case rule delimiters such that the cursor on `|` or `=>` highlights the rule and applies the deletion rules described above to the rule node.


dmoon1221:
To be reviewed and merged once https://github.com/hazelgrove/hazel/pull/65 is merged. There is a remaining bug where relying on our current `set_cursor` code moves the cursor into some nondisplay characters cushioning delimiters at the moment. Will try to fix today, might not be until tomorrow.

dmoon1221:
Closes https://github.com/hazelgrove/hazel/issues/67

dmoon1221:
Subsumed by https://github.com/hazelgrove/hazel/pull/81

Issue #49, opened 2019-01-09
Refactor view layer into Elm architecture


cyrus-:
There is a tyxml interface to virtual_dom so it should not be too complicated: 

https://github.com/janestreet/virtual_dom/blob/master/src/tyxml.ml

The benefit is that we won't be re-rendering the entire expression view on each change. Haven't really experienced any slowness yet, so this isn't high priority, but I imagine that for thousand-line Hazel programs in the future, this will be handy.

PR #81, opened 2019-09-04
Cursor on delimiters + incr_dom + cursor indicators + node staging + cardstacks

dmoon1221:
This is a massive pull request containing many changes over the past several months. Most notable changes include:
- updating our cursor model so that the caret can be on node delimiters
- restructuring our term datatypes such that each variant has its own err_status field, such that certain variants that don't require it (e.g. EmptyHole, OpSeqs) no longer carry it
- refactor our view layer to use incr_dom
- somewhat simplify pretty printing logic
- enrich our cursor indicators to include current node + current term
- implement first version of node staging
- implements a basic cardstack system for running through exercises in sequence

There are still some fundamental warts to resolve. Some issues are remaining bugs, especially in node staging. The more fundamental issue is that contenteditable, cursor indicators, and incr_dom do not play nicely together in the current implementation. contenteditable requires code to be rendered in dom via inline elements. This means that the shapes we want to overlay with cursor indicators cannot be identified structurally. Instead we must wait until the dom gets rendered on each animation frame by incr_dom, then we take the viewport coordinates of the nodes we want to be overlaid and paint over them. This is also true for nonempty holes. Already we are running into performance issues repainting the cursor indicators and err holes on small programs on each animation frame.

In my next set of changes, I will be addressing these issues by taking an approach similar to VSCode: they create the illusion of text editing by rendering their code in a noneditable structured form (thereby freeing it from the constraints of contenteditable) while using an invisible textarea element to hook into its caret movement+selection logic. Essentially this separate the concerns of editing+movement from the presentation. This will greatly simplify our pretty printing logic and remove the need to paint overlays over our code view on each animation frame.

For now, however, we have been moving forward assuming approximately the current implementation approach, so I'm gonna go ahead and merge these changes so that we can start working off of dev again.


dmoon1221:
Closes
https://github.com/hazelgrove/hazel/issues/49
https://github.com/hazelgrove/hazel/issues/67
https://github.com/hazelgrove/hazel/issues/68
https://github.com/hazelgrove/hazel/issues/71
https://github.com/hazelgrove/hazel/issues/73
https://github.com/hazelgrove/hazel/issues/74
https://github.com/hazelgrove/hazel/issues/75

PR #82, opened 2019-09-10
[WIP] Variable Splitting

wlitwin:
Moving the variable splitting changes to the dev branch. Changed approach, now rebuilds the ZExp more correctly. Seems to work for `+` `-` `*`. Splitting multiple times works, as well as deleting, or adding onto a split variable and then splitting again.


src/semantics/Action.re
```diff
@@ -3408,6 +3420,47 @@ let rec syn_perform_block =
...
     let zblock = ZExp.BlockZE(lines, ze);
     syn_perform_block(~ci, ctx, keyword_action(k), (zblock, Hole, u_gen));
+
+  /* Variable Splitting */
+
+  | (Construct(SOp(op)), BlockZE(lines, zexp))
+      when zexp_is_suitable_for_var_split(zexp) =>
+    let split_var = (pos, err, verr, name) => {
+      let left_var = String.sub(name, 0, pos);
+      let right_var = String.sub(name, pos, String.length(name) - pos);
```
dm0n3y: I think should be `String.sub(name, pos, String.length(name))`?

src/semantics/Action.re
```diff
@@ -3408,6 +3420,47 @@ let rec syn_perform_block =
...
+      let cursor =
+        ZExp.CursorE(
+          OnText(String.length(name) - pos),
+          Var(err, verr, right_var),
+        );
+      let left_var = UHExp.Var(err, verr, left_var);
```
dm0n3y: You don't want to preserve the `err` and `verr` values here. `err: ErrStatus.t` represents the type-checked status of the original var (e.g. `InHole(TypeInconsistent, <hole-num>)` if the original var had type `Num` when there was a `Bool` expected). `verr: VarErrStatus.t` represents the variable-specific error status of the original var (e.g. `InVarHole(Free, <hole-num>)` if the var isn't bound, `InVarHole(Keyword, <hole-num>)` if it's a reserved keyword). These values no longer hold for the new split vars. What I'd do here instead is just use `UHExp.var` to construct the new split vars, then later you'll need to call `Statics.syn_fix_holes_zblock` on the final result zblock to fix the hole statuses.

src/semantics/Action.re
```diff
@@ -3408,6 +3420,47 @@ let rec syn_perform_block =
...
+  | (Construct(SOp(op)), BlockZE(lines, zexp))
+      when zexp_is_suitable_for_var_split(zexp) =>
+    let split_var = (pos, err, verr, name) => {
+      let left_var = String.sub(name, 0, pos);
+      let right_var = String.sub(name, pos, String.length(name) - pos);
+      let cursor =
+        ZExp.CursorE(
+          OnText(String.length(name) - pos),
+          Var(err, verr, right_var),
+        );
```
dm0n3y: Here you're placing the cursor on the right split var. Let's instead place the cursor to the right of the new op. This will be necessary in the long run for when we have multi-character ops (e.g. did the user mean a `+` or a `++`?).

dm0n3y (commented):
Looks good overall modulo comments. Some next steps:
- handle case `BlockZL((_, ExpLineZ(ze), _), _)` in `syn_perform_block`
- look at cases that handle reserved keywords like `let` and `case` and consider what should happen when I split on edit state `let|x` or `case|x`
- whatever cases you add to `syn_perform_block`, you'll also need to handle in `ana_perform_block` (should be roughly the same code but e.g. you'll make calls to `Statics.ana_fix_holes_zexp` instead of `Statics.syn_fix_holes_zexp`

src/semantics/Action.re
dm0n3y: If the op is a space, though, you will want to place the cursor to the left of the right split var

src/semantics/Action.re
wlitwin: I thought so too at first, but it seems it's `pos` and `length` as arguments instead of `pos` `pos`

wlitwin:
Working on handling the `let|x` and `case|x` situations now.

src/semantics/Action.re
```diff
@@ -2747,14 +2747,100 @@ let make_zexp_or_zblock_inconsistent =
...
         (
           ~ci: CursorInfo.t,
           ctx: Contexts.t,
           a: t,
           (zblock, ty, u_gen): (ZExp.zblock, HTyp.t, MetaVarGen.t),
         )
-        : result((ZExp.zblock, HTyp.t, MetaVarGen.t)) =>
+        : result((ZExp.zblock, HTyp.t, MetaVarGen.t)) => {
+  print_endline("=======");
+  let str = Core_kernel.Sexp.to_string_hum(sexp_of_t(a));
```
wlitwin: Will remove this debug stuff for final PR

wlitwin:
Also still need to handle `BlockZL((_, ExpLineZ(ze), _), _)` in syn_perform_block

wlitwin:
OK! I think I've handled everything except parentheses. I've done a bit of testing, but it could use some other eyes on it.

src/hazelcore/semantics/ZExp.re
```diff
@@ -180,6 +180,13 @@ let is_before_rule =
   | CursorR(OnDelim(0, Before), _) => true
   | _ => false;
 
+let rec cursor_on_var =
```
dm0n3y, line 183: No need to return `err` and `verr`

src/hazelcore/GeneralUtil.re
```diff
@@ -538,6 +538,11 @@ module Opt = {
     | None => default
     | Some(a) => f(a)
     };
+  let get_exn =
```
dm0n3y: Let's call this `force_get` to align with existing naming convention for these sorts of things

src/hazelcore/semantics/Action.re
```diff
@@ -2761,14 +2761,133 @@ let make_zexp_or_zblock_inconsistent =
     (B(zblock), u_gen);
   };
 
+let split_variable_name_into_strings = (pos, name) => {
```
dm0n3y: Let's move this into `Var.re` and call it `split`

src/hazelcore/semantics/Action.re
```diff
@@ -3422,6 +3579,64 @@ let rec syn_perform_block =
...
+        String.sub(name, 0, pos)
+        |> Keyword.of_string
+        |> GeneralUtil.Opt.get_exn;
+      let var = String.sub(name, pos, String.length(name) - pos);
+      let (ze, u_gen) = ZExp.new_EmptyHole(u_gen);
+      let action = Construct(SVar(var, OnText(String.length(name) - pos)));
+      switch (zexp) {
+      | CursorE(_) =>
+        let zblock = ZExp.BlockZE(lines, ze);
+        let tuple = Statics.syn_fix_holes_zblock(ctx, u_gen, zblock);
```
dm0n3y: Call this `edit_state`

src/hazelcore/semantics/Action.re
```diff
@@ -3422,6 +3579,64 @@ let rec syn_perform_block =
...
+      when is_zexp_split_on_keyword(zexp) =>
+    switch (ZExp.cursor_on_var(zexp)) {
+    | Some((pos, _, _, name)) =>
+      let keyword =
+        String.sub(name, 0, pos)
+        |> Keyword.of_string
+        |> GeneralUtil.Opt.get_exn;
+      let var = String.sub(name, pos, String.length(name) - pos);
+      let (ze, u_gen) = ZExp.new_EmptyHole(u_gen);
+      let action = Construct(SVar(var, OnText(String.length(name) - pos)));
```
dm0n3y: Let's leave the cursor at `OnText(0)` since that's where the cursor would end up if user actually inserted a text space

src/hazelcore/semantics/Action.re
```diff
@@ -3422,6 +3579,64 @@ let rec syn_perform_block =
...
+  | (Construct(SOp(SSpace)), BlockZE(lines, zexp))
+      when is_zexp_split_on_keyword(zexp) =>
+    switch (ZExp.cursor_on_var(zexp)) {
+    | Some((pos, _, _, name)) =>
+      let keyword =
+        String.sub(name, 0, pos)
+        |> Keyword.of_string
+        |> GeneralUtil.Opt.get_exn;
+      let var = String.sub(name, pos, String.length(name) - pos);
+      let (ze, u_gen) = ZExp.new_EmptyHole(u_gen);
```
dm0n3y: You can move this down into `CursorE` branch

dm0n3y (changes requested):
Looks good overall. Aside from small comments, my main request is organizational. I'd rather that the logic of `handle_variable_split` just be inlined into `syn_perform_block` and `ana_perform_block`. You'll have to duplicate code across those two places but we already do that systematically and I'd rather deal with duplicated code than kinda vague abstraction boundaries.

Your helper functions `handle_variable_split_space_case` and `handle_general_op_case` would still be helpful to keep around, but rather than separating between space and non-space cases, I think it would cleaner to separate across whether you're operating on a single var (`split_single_var(x: Var.t, pos: int, op: UHExp.op): ZExp.t`) vs in the context of an opseq (`split_operand(x: Var.t, pos: int, surround: UHExp.opseq_surround): ZExp.t`) or something like that.

src/hazelcore/GeneralUtil.re
wlitwin: I removed this since I don't need it anymore either.

wlitwin:
> Your helper functions `handle_variable_split_space_case` and `handle_general_op_case` would still be helpful to keep around, but rather than separating between space and non-space cases, I think it would cleaner to separate across whether you're operating on a single var (`split_single_var(x: Var.t, pos: int, op: UHExp.op): ZExp.t`) vs in the context of an opseq (`split_operand(x: Var.t, pos: int, surround: UHExp.opseq_surround): ZExp.t`) or something like that.

I've done everything but this change so far. I'll have to noodle around a bit on how to move things around as I've been gone for a bit.

dm0n3y:
This looks good to me. I think something may have broken while merging but I remember it working and code looks good -- I assume the issue is just one of matching-order-based assumptions being entangled across the merged work and yours. I'm going to go ahead and merge this now because I have a big set of changes that redo some of the foundations of what you're working with. I'll get your code working on these new foundations, then I can get you started again with the new codebase (either continuing to flesh out this feature or exploring something else).

wlitwin:
Thanks for merging!

PR #85, opened 2019-10-28
Variable usage

cyrus-:


src/hazelcore/semantics/CursorInfo.re
```diff
@@ -121,6 +129,10 @@ type t = {
   subblock_start: option(int),
 };
 
+type cursor_t =
+  | CursorNotOnVar(t)
```
cyrus-: CursorNotOnDeferredVarPat / CursorOnDeferredVarPat

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -121,6 +129,10 @@ type t = {
   subblock_start: option(int),
 };
 
+type cursor_t =
```
cyrus-: deferrable_t

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -922,46 +1027,146 @@ let ana_cursor_found_exp =
...
+    shadow_in_pat(x, e1) || shadow_in_pat(x, e2)
+  | OpSeq(skel, SeqOpExp(seq1, _, e1)) =>
+    shadow_in_pat(x, e1) || shadow_in_pat(x, OpSeq(skel, seq1))
+  | Inj(NotInHole, _, e) => shadow_in_pat(x, e)
+  };
+
 let rec _syn_cursor_info_block =
         (~node_steps, ~term_steps, ctx: Contexts.t, zblock: ZExp.zblock)
-        : option(t) =>
+        : option(cursor_t) =>
```
cyrus-: change this back to option(t)

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -985,20 +1190,22 @@ and _syn_cursor_info_line =
       ctx: Contexts.t,
       zli: ZExp.zline,
     )
-    : option(t) =>
+    : option(cursor_t) =>
```
cyrus-: this is the only one that needs to be option(deferrable_t)

VictorYYW:
I may have to calculate node steps along the recursion tree of _syn_cursor_info_block, because to find where a variable is used, the current location(node_steps) is needed. What's your opinion, @cyrus- ?

cyrus-:
Yes I agree, you will have to count node steps while determining cursor info.

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -121,6 +129,10 @@ type t = {
   subblock_start: option(int),
 };
 
+type deferrable_t =
```
cyrus-: Let's (1) move this definition to just before the first function that uses it and (2) include a comment describing what is going on here.

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -79,10 +79,18 @@ type typed =
   | OnRule
   | OnOp;
 
+[@deriving sexp]
+type uses_list = list(CursorPath.steps);
+
+[@deriving sexp]
+type pat_node =
+  | VarPat(UHPat.t, uses_list)
```
cyrus-: The VarPat case should only occur when the UHPat.t is a variable pattern, not an arbitrary pattern. Following the principle of "make invalid states unrepresentable", let's change this to VarPat(ErrStatus.t, VarErrStatus.t, Var.t, uses_list). You can reconstruct the UHPat.t from the first three arguments when necessary.

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -202,7 +216,11 @@ let staging_left_border = ci =>
       Exp(
         Inj(_, _, _) | Parenthesized(_) | Case(_, _, _, _) | Lam(_, _, _, _),
       ) |
-      Pat(Inj(_, _, _) | Parenthesized(_)) |
+      Pat(
+        VarPat(Inj(_, _, _), _) | VarPat(Parenthesized(_), _) |
```
cyrus-: (the change I suggested above makes it so we don't need to consider these invalid cases)

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -310,10 +338,11 @@ let rec cursor_info_typ =
 
 let rec _ana_cursor_found_pat =
         (ctx: Contexts.t, p: UHPat.t, ty: HTyp.t)
-        : option((typed, node, Contexts.t)) =>
+        : option((typed, uses_list => node, Contexts.t, option(Var.t))) =>
```
cyrus-: This should return a deferrable_t, or something similar in structure to it, rather than always returning a deferred node.

You might want to change deferrable_t to be a parameterized type rather than fixing it to return a `t` in the non-deferred case and a `uses_list -> t` in the deferred case, so you can use it in situations like this?

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
     );
   };
 
+let rec find_uses_block =
```
cyrus-: Let's move the find_uses functions and the definition of uses_list into UHExp.re.

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
...
+        (x: Var.t, block: UHExp.block, steps, index): uses_list => {
+  let UHExp.Block(lines, e) = block;
+  let (uses, shadow) = find_uses_lines(x, lines, steps @ [0], index, false);
```
cyrus-: use the variable name `shadowed` instead of `shadow`

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
...
+  shadow
+    ? uses
+    : uses @ find_uses_exp(x, e, steps @ [index + List.length(lines)]);
+}
+and find_uses_lines =
+    (x: Var.t, lines: UHExp.lines, steps, index, shadow): (uses_list, bool) =>
+  switch (lines, shadow) {
```
cyrus-: instead of pattern matching on the pair, just do an `if (shadowed) { ([], shadowed) } else { ... }`

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
...
+  | LetLine(p, _, block) => (
+      find_uses_block(x, block, steps @ [1], 0),
+      shadow_in_pat(x, p),
+    )
+  }
+and find_uses_exp = (x: Var.t, e: UHExp.t, steps): uses_list =>
+  switch (e) {
+  | EmptyHole(_)
+  | Var(_, InVarHole(_, _), _)
+  | Var(InHole(_, _), _, _)
```
cyrus-: A variable in an ErrStatus hole (i.e. `InHole`) is still a use (it just happens to be a use in a type-incorrect position). Only when its `InVarHole` is it necessarily not a use of anything (because that hole means that it is unbound)

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
...
+  | ListNil(_)
+  | Lam(InHole(_, _), _, _, _)
+  | Inj(InHole(_, _), _, _)
+  | Case(InHole(_, _), _, _, _)
+  | ApPalette(_, _, _, _) => []
+  | Var(NotInHole, NotInVarHole, y) =>
+    Printf.printf("x is used\n");
+    x == y ? [steps] : [];
+  | Lam(NotInHole, p, _, block) =>
+    shadow_in_pat(x, p) ? [] : find_uses_block(x, block, steps @ [1], 0)
```
cyrus-: move `shadow_in_pat` to `UHPat.binds_var`

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -946,6 +1049,101 @@ let ana_cursor_found_exp =
...
+  switch (opseq) {
+  | ExpOpExp(e1, _, e2) => (
+      find_uses_exp(x, e1, steps @ [0]) @ find_uses_exp(x, e2, steps @ [1]),
+      2,
+    )
+  | SeqOpExp(seq, _, e) =>
+    let (uses, length) = find_uses_opseq(x, seq, steps);
+    (uses @ find_uses_exp(x, e, steps @ [length]), length + 1);
+  }
+and shadow_in_pat = (x: Var.t, p: UHPat.t): bool =>
```
cyrus-: (this doesn't need to be mutually defined because it doesn't call any of the other functions, see comment above)

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -985,25 +1197,31 @@ and _syn_cursor_info_line =
...
+        ),
       ),
     );
   | ExpLineZ(ze) =>
-    _syn_cursor_info(~node_steps, ~term_steps, ~frame, ctx, ze)
+    switch (_syn_cursor_info(~node_steps, ~term_steps, ~frame, ctx, ze)) {
+    | None => None
+    | Some(ci) => Some(CursorNotOnDeferredVarPat(ci))
+    }
   | LetLineZP(zp, ann, block) =>
```
cyrus-, line 1180: isn't this the case where we need to consider whether it was deferred? am I not seeing the code?

src/hazelweb/Model.re
```diff
@@ -139,7 +139,13 @@ let cursor_info_of_edit_state = ((zblock, _, _): edit_state): CursorInfo.t =>
     )
   ) {
   | None => raise(MissingCursorInfo)
-  | Some(ci) => ci
+  | Some(ci) =>
```
cyrus-, line 142: Make sure you drop this before pushing to dev

src/hazelcore/semantics/CursorInfo.re
VictorYYW: Since VarPat case only occur when UHPat.t is a variable pattern without Hole, can I just have `VarPat(Var.t, uses_list)`? BTW, UHPat.t can still be reconstructed by calling constructor of NotInHole and NotInVarHole when constructing UHPat.Var

yongweiy: Since VarPat case only occur when UHPat.t is a variable pattern without Hole, can I just have `VarPat(Var.t, uses_list)`? Also, UHPat.t can still be reconstructed by calling constructor of NotInHole and NotInVarHole when constructing UHPat.Var

src/hazelcore/semantics/CursorInfo.re
yongweiy: I feel like the only case needs CursordefferedVarPat is when the cursor is on LetLineZP. The only two cases that ZExp.t involves ZPat.t is LamZP and CaseZR. And, outside of ZExp.t, any variable defined in the ZExp.t should have already been out of scope. Please correct me if I miss anything.

src/hazelcore/semantics/CursorInfo.re
yongweiy: That's a great idea! I would change `deferrable_t` to `deferrable('t)`.

src/hazelcore/semantics/CursorInfo.re
cyrus-: ah yeah, good point!

src/hazelcore/semantics/CursorInfo.re
cyrus-: oh nevermind, I see how this works

src/hazelcore/semantics/CursorInfo.re
yongweiy: I realized it may not work because the definition of `uses_list` depends on `CursorPath` while some definition in `CursorPath` depends on `UHExp`. And there would be a dependency cycle. How about changing `uses_list` to `CursorPath.steps_list`?

cyrus-:
Ah I see -- can you put the find uses related code in a separate
UsageAnalysis module?

VictorYYW:
Sure! That makes sense and should resolve the cyclic dependency problem.

PR #89, opened 2020-01-03
new year, new codebase

dm0n3y:
- Refactor core uhnode datatypes so that opseqs may consist of a single element. This removes large volumes of code devoted to handling the prior restriction that opseqs always carry two or more elements. Opseqs are now distinguished types in each uhnode datatype.
- Reorganize modules that traverse uhnode and znode datatype structures into submodules `Typ`, `Pat`, `Exp`. Now, instead of calling `CursorInfo.syn_cursor_info_pat` or `CursorInfo.syn_cursor_info_exp`, you call `CursorInfo.Pat.syn_cursor_info` or `CursorInfo.Exp.syn_cursor_info`.
- We never had proper n-tuples (sometimes there was the illusion of n-tuples working due to other confounding bugs). Modifies `Statics` and `CursorInfo` to support n-tuples.
- Introduce new intermediate `Action` result datatypes that simplify how we handle keywords like `let` and `case`.
- Implement character-level actions and remove all the cruft accumulated in `Cell.re`.
- Rebuild view layer of code on @adamsmd's new pretty printer. See `Doc.re`, `Layout.re`, `LayoutOfDoc.re` for general pretty printing logic. See `TermTag.re`, `TermDoc.re`, and `TermLayout.re` for logic specific to pretty printing Hazel terms. See `Code.re` to see how a `TermLayout.t` gets turned into vdom nodes. Note that `TermLayout` has various decoration functions that get invoked after a `TermDoc.t` gets turned into a `TermLayout.re` (which made it easy to merge @VictorYYW's variable usage decorations).
- Rebuild view layer of code into a separate presentation layer and contenteditable layer. The presentation layer is what the user sees, simulating the blinking caret and showing nice visuals. The actual caret is in the invisible contenteditable element (fixed position outside of viewport) that allows us to hook into existing implementation of text carets. To keep them in sync, we translate caret movements in the contenteditable layer into changes in presentation layer, and we translate clicks on the presentation layer into caret resets in the contenteditable layer. The reason for this approach (which is also taken by VSCode) is that the contenteditable feature limits how we can layout the elements, which makes it hard to style them effectively, so a separate presentation layer frees the implementation of visual styling.
- Introduce distinction between an _open child_ and a _closed child_ of an AST node. An open child is a child that has the same term family (e.g. Typ, Pat, Exp) as the parent -- a closed child is a child with a different term family. A closed child will appear in the UI enclosed in the parent, an open child will have a border but otherwise be unenclosed. This distinction only appears when generating a view of the program, i.e., by the time we have a `TermDoc.t`.
- Introduce new blocks-like UI in which it appears a block representing the current term raises out of the surface to the code text floating above. Use different colors for different term families.
- Remove hacks introduced in the codebase for RC study. In particular, remove all staging functionality and code. For now, deletion immediately removes a node and leaves behind its open children. This will change in a future PR that adds back a proper implementation of staging/restructuring mode.


dm0n3y:
Some remaining TODOs:
- Few bugs here and there in basic editing functionality 
- Take a pass through changes and remove unused code
- Integrate error hole styling with new blocks UI.
- Add back empty hole styling, make sure widths correspond to indices
- Review TODO refactors discussed with @cyrus-, which may be worth doing now or may be deferred to future PR.

cyrus-:
Do you want me to review this weekend or wait until you knock out some of the TODOs that you mentioned? How much time do you think the refactorings we discussed would take?

dm0n3y:
Ready to merge. Recent changes since opening PR consist of:
- fixing bugs and polishing visual warts, can now type out map and quicksort demos
- implementing user newlines by specially handling and displaying the first line of a block if it's an EmptyLine
- merging in @wlitwin's variable splitting changes into new Action code, correcting some issues with numlits/boollits, extending variable splitting support to patterns

Next step is to profile, diagnose, and fix performance issues. Unusably long lags when doing quicksort demo.

dm0n3y:
Addressed all pre-merge tasks in https://github.com/hazelgrove/hazel/projects/6. Can complete quicksort demo. Will go ahead and merge tonight barring objections.

cyrus-:
Go for it as long as you have some cycles to help folks with fixing their branches soon.

PR #93, opened 2020-01-28
Tuple completion

qiranli:


dm0n3y:
Cool. I'll review and help merge today.

dm0n3y:
As part of getting tuple completion merged with new codebase, I removed the `E2`, `E1`, `E0`, etc constructors at the top level of each uhnode datatype.  Now `UHExp.t` is a block and `UHPat.t` and `UHTyp.t` are opseqs. This forces actions always to start at the block/seq level when performing, which makes it easy to implement context-sensitive actions like tuple completion, where we want to invoke it when we know we're at the end of an opseq.

dm0n3y:
Closes https://github.com/hazelgrove/hazel/issues/128

PR #97, opened 2020-02-02
edit action history panel

PPER:
a new version of the panel after merging to the new codebase


adamsmd:
My merge was accidental and premature, so I've opened #208 to replace this one.

PR #210, opened 2020-03-02
action panel reimplementation

wlitwin:
WIP pull request so it's easy to keep tabs on my progress, and so it's easy to provide feedback. So far I've only done the minimum to get a side bar up and enabling/disabling buttons based on edit state. There's a number of styling/formatting issues with the left panel. Notably closing is broken at the moment. Also, some of my descriptions are probably inaccurate, and there's some informational nodes missing.

image


cyrus-:
Let's maybe have it be a little more grouped this time, rather than just a vertical list.

So, for example, maybe have:

Movement:
- move using arrow keys
- click to move directly to a location
- move to next/prev hole: tab/shift+tab

Arithmetic:
- enter number literals directly
- arithmetic operators: + - *

Variables:
- enter variables directly
- type "let<space>" to insert a let expression

(and so on for the other features)

wlitwin:
image

New screenshot with sections

cyrus-:
Looks good!

Small suggestions:
* Change 'type "let"' to 'type "let<space>"' to emphasize that you need to press space
* Make things the user needs to type ("let", "true", "false", etc.) use the mono font that we use for code
* Change "Insert a list" to "Empty list (nil)"
* Add a "functions" section and put lambda and function application (space) there
* "a numeric type" -> "the Num type"
* "a boolean type" -> "the Bool type"
* Put the type ascription thing under both let and functions, and rename the "type ascription" section to "types"
* The "|" operator is part of types, so put that under types too

Also, let's experiment with making better use of horizontal space? For example:
* move to prev/next hole could maybe be on one line, with two hotkeys shown
* the arithmetic operators could all be summarized on one line

Feel free to experiment with other ways to summarize things.

Thanks!

cyrus-:
@wlitwin how's it going?

wlitwin:
Still thinking of different ways to combine things
Selection_001

cyrus-:
Looking good! Delete and Backspace do the same thing in different directions, so they should probably be grouped together. Also type ascription is in there twice. 

I assume you have a section on functions below the fold?

cyrus-:
Also the latest commits to dev by @Yunif3 added floating point numbers and dotted float ops like in OCaml.

PPER:
> Looking good! Delete and Backspace do the same thing in different directions, so they should probably be grouped together. Also type ascription is in there twice.
> 
> I assume you have a section on functions below the fold?

wlitwin:
> I assume you have a section on functions below the fold?

Yes! There's also a miscellaneous section that has some commands that I'm not sure where to place.

Selection_003

> Also the latest commits to dev by @Yunif3 added floating point numbers and dotted float ops like in OCaml.

Cool! I'll add a note about those

cyrus-:
* "Separate terms" should be "apply function"
* "Parenthesize expression" works for patterns and types too, so just call it "Parenthesize"
* the "insert , operator" one should be in a section called "tuples"
* "insert case expression" is now entered with the "case" keyword, similar to "let" (though maybe Alt+C also still works?). it should be in a section called "Pattern Matching". you might also want to write "add new rule" in that section, with "Enter" as the key.

wlitwin:
>though maybe Alt+C also still works?
Seems to still work! I left it in, unless this is planning to be phased out?

Latest changes

Screen Shot 2020-05-10 at 14 53 18

cyrus-:
- The Delete key does not "delete expression", it just does the same thing as backspace but to the right of the cursor. Maybe just have an entry that says "Backspace / Delete" and then has the two keys on the right.
- Everything from Backspace on down in the "Movement" section is not about movement, but rather general editing. Let's create a new section called "General Editing" that includes that + the "Parenthesize" action that is now at the bottom.
- The floating point operator characters can just be in the light brown color on the right, just like the integer operators.
- We don't have a Num type anymore, we have Int and Float ("I" and "F", respectively -- don't need to write "Shift+".
- Let's put the type construction into the corresponding semantic sections, instead of in the "Types" section.
- For variables, maybe we can list the regex for valid variables?
- "injunction" should be "injection". call that section "sum types"
- "type arrow" should be "arrow type"

wlitwin:
Latest changes applied:

Screen Shot 2020-05-25 at 22 07 19

cyrus-:
@wlitwin can you push your code changes? it is probably easier for me to make comments directly on the code for the various messages

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+          text("Backspace / Delete"),
+          Vdom.Node.div(
+            [display_inline_block],
+            [
+              keyboard_button(Backspace) |> add_style(bottom_border_none),
+              keyboard_button(Delete) |> add_style(bottom_border_none),
+            ],
+          ),
+        ]),
+        combo(Ctrl_Alt_I, simple("Swap expression up")),
```
cyrus-: * this should be "swap line up/down"
* can we combine the up/down ones into a dual_line?

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+        combo(Ctrl_Alt_K, simple("Swap expression down")),
+        combo(Ctrl_Alt_J, simple("Swap expression left")),
```
cyrus-: this should be "swap operand left/right"

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+        info_action(
+          [
+            text("Type "),
+            mono_text("\"F\""),
+            text(" to insert a Float type"),
+          ],
+          Action.Construct(SChar("F")),
+        ),
+        spaced_line([
+          text("Floating point operators, type"),
```
cyrus-: drop the ", type"

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+              brown_label([mono_text("-.")])
+              |> add_style(bottom_border_none),
+              brown_label([mono_text("*.")])
+              |> add_style(bottom_border_none),
+            ],
+          ),
+        ]),
+        operator_list(
+          ~on_type=false,
+          "Integer operators",
```
cyrus-: Put integer operators first, to be consistent with order of the type insertion lines

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+          [Plus, Minus, Asterisk],
+        ),
+      ],
+    ),
+    section(
```
cyrus-, line 299: lets move this section above arithmetic

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+      "Sum Types",
+      [
+        combo(Alt_L, simple("Left injection")),
+        combo(Alt_R, simple("Right injection")),
+      ],
+    ),
+    section(
+      "Types",
```
cyrus-: lets get rid of this section and move all of this stuff into the sections for the corresponding expression operations

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+      [
+        combo(Colon, simple("Type ascription")),
```
cyrus-: move to variables section

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+        combo(VBar, simple("Insert | operator")),
```
cyrus-: move to sum types section

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+        combo_and_cursor(
+          ~on_type=true,
+          LeftBracket,
+          simple("Insert type List"),
```
cyrus-: move to lists section

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+        ),
+        combo_and_cursor(~on_type=true, GT, [text("Create an arrow type")]),
```
cyrus-, line 413: move to functions section

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,464 @@
...
+          [
+            text("Type "),
+            mono_text("\"B\""),
+            text(" to insert a Bool type"),
+          ],
+          Action.Construct(SChar("B")),
+        ),
+        operator_list(
+          ~on_type=false,
+          "Operators",
```
cyrus-: these are actually integer comparison operators only. there are also dotted versions of these for float comparison. let's include these in the arithmetic section, and move the boolean section above it so that booleans are described before the comparison operators.

cyrus- (changes requested):

src/hazelweb/gui/ActionPanel.re
wlitwin: I changed it to a the same style as `Backspace / Delete` as it looked a bit weird with the action boxes interspersed with the text. Now the whole UI is more uniform.

wlitwin:
Screen Shot 2020-05-27 at 12 03 16

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+          [keyboard_button(Ctrl_Alt_J), keyboard_button(Ctrl_Alt_L)],
+        ),
+        combo(Enter, simple("Create new line ")),
+        combo(LeftParen, simple("Parenthesize")),
+      ],
+    ),
+    section(
+      "Variables",
+      [
+        info([text("Variable regex: ^[_a-z][_a-zA-Z0-9']*$")]),
```
cyrus-: we can drop the `^` and `$`. also the latest `dev` branch allows initial caps.

cyrus-: also can you use code font for the regex

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+        ),
+        combo(Colon, simple("Type ascription")),
+      ],
+    ),
+    section(
+      "Booleans",
+      [
+        info([
+          text("Enter boolean literals "),
+          mono_text("\"true\", \"false\""),
```
cyrus-: don't need quotation marks around true and false

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+        ]),
+        info_action(
+          [
+            text("Type "),
+            mono_text("\"B\""),
+            text(" to insert a Bool type"),
+          ],
+          Action.Construct(SChar("B")),
+        ),
+        operator_list(~on_type=false, "Logical AND", [Ampersand]),
```
cyrus-: also add OR (`|`) to this entry

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+        ),
+      ],
+    ),
+    section(
+      "Lists",
+      [
+        combo_and_cursor(
+          ~on_type=false,
+          LeftBracket,
+          simple("Insert Empty List (nil)"),
```
cyrus-: replace "Empty List" with `[]` (in code font)

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+      [
+        info([
+          text("Enter number literals directly e.g. "),
+          mono_text("1.0, 2"),
+        ]),
+        info_action(
+          [
+            text("Type "),
+            mono_text("\"I\""),
+            text(" to insert an Integer type"),
```
cyrus-: replace "Integer" with `Int` (in code font)

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+          ],
+          Action.Construct(SChar("I")),
+        ),
+        info_action(
+          [
+            text("Type "),
+            mono_text("\"F\""),
+            text(" to insert a Float type"),
```
cyrus-: put `Float` in code font

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+          ],
+          Action.Construct(SChar("F")),
+        ),
+        operator_list(
+          ~on_type=false,
+          "Integer operators",
```
cyrus-, line 367: latest `dev` has division too `/` and `/.`

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+          simple("Insert type List"),
+        ),
+      ],
+    ),
+    section(
+      "Sum Types",
+      [
+        combo(Alt_L, simple("Left injection")),
+        combo(Alt_R, simple("Right injection")),
+        combo(VBar, simple("Insert | operator")),
```
cyrus-: replace "operator" with "type operator" (this is currently broken on `dev`)

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,437 @@
...
+      ],
+    ),
+    section(
+      "Functions",
+      [
+        combo(Backslash, simple("Insert Lambda expression")),
```
cyrus-: replace "Lambda expression" with "lambda function"

cyrus- (changes requested):

cyrus-:
Looks pretty close to ready! Another thing we should do is add a simple "checkin" function, the purpose of which is to alert anyone adding a new action that they need to modify this sidebar too.

```
type ack_checkin = AddedAction

/* add a new case to this function only if you have already added that action to the sidebar! */
let checkin = (a : Action.t) => 
  switch (a) {
  | Construct(SVar(_)) -> AddedAction
  | etc
```

wlitwin:
Screen Shot 2020-05-28 at 13 43 41

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,501 @@
...
+    ),
+    section(
+      "Functions",
+      [
+        combo(Backslash, simple("Insert lambda function")),
+        combo(Space, simple("Apply function")),
+        combo_and_cursor(~on_type=true, GT, [text("Create an arrow type")]),
+      ],
+    ),
+    section("Tuples", [combo(Comma, simple("Create a tuple"))]),
```
cyrus-: Let's add another entry for create a tuple type, which also has Comma

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,501 @@
...
+  | Construct(SCase)
+  | Construct(SLet)
+  | Construct(SOp(SVBar))
+  | Construct(SChar(_))
+  | SwapUp
+  | SwapDown
+  | SwapLeft
+  | SwapRight
+  | MoveLeft
+  | MoveRight => ()
```
cyrus-: let's define a type with one constructor, `Added`, instead of using `()`, to avoid people just reflexively copying. also let's do one switch rule per action, instead of or-ing them together like this.

cyrus- (changes requested):

cyrus-:
don't forget to click "Ready for review" on the PR when the changes are ready.

wlitwin:
Screen Shot 2020-05-30 at 16 32 20

src/hazelweb/gui/ActionPanel.re
```diff
@@ -0,0 +1,514 @@
...
+  | Construct(SOp(SVBar)) => Added
+  | Construct(SChar(_)) => Added
+  | SwapUp => Added
+  | SwapDown => Added
+  | SwapLeft => Added
+  | SwapRight => Added
+  | MoveLeft => Added
+  | MoveRight => Added
+  /* Not added */
+  | Construct(SApPalette(_)) => Added
```
cyrus-: change the Palette-related ones to `failwith("Unimplemented")`

cyrus- (changes requested):

cyrus- (approved):

cyrus-:
looks good!

PR #211, opened 2020-03-12
automatically generate patterns based on type

### LINKED ISSUE #157: Automatically generate cases from type of scrutinee

cdfa:
Generate patterns for case expressions based on the type of the expression being matched.
Fixes #157


src/hazelcore/semantics/OpSeq.re
```diff
@@ -17,6 +17,19 @@ let mk =
...
+let wrap_operator =
+    (
+      ~err: ErrStatus.t=NotInHole,
+      left: 'operand,
+      op: 'operator,
+      right: 'operand,
+    )
+    : t('operand, 'operator) =>
+  OpSeq(
+    Skel.(BinOp(err, op, Placeholder(0), Placeholder(0))),
```
cyrus-: you shouldn't have two Placeholder(0)s in a BinOp...

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +156,35 @@ and make_inconsistent_operand =
     (Parenthesized(set_p), u_gen);
   };
 
+let patterns_of_type =
+    (u_gen: MetaVarGen.t, ty: HTyp.t): (list(t), MetaVarGen.t) => {
+  let (hole1, u_gen) = u_gen |> new_EmptyHole;
+  let (hole2, u_gen) = u_gen |> new_EmptyHole;
+  let (hole3, u_gen) = u_gen |> new_EmptyHole;
+  let (hole4, u_gen) = u_gen |> new_EmptyHole;
```
cyrus-: why are you generating 4 fresh holes in all situations?

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +156,35 @@ and make_inconsistent_operand =
...
+  let pattern =
+    OpSeq.(
+      switch (ty) {
+      | HTyp.Hole
+      | Unit
+      | Arrow(_, _) => []
+      | Num => [wrap(hole1), wrap(hole2)]
```
cyrus-: what's happening here? for number types, I think the canonical patterns should maybe be "0" and "_"

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +156,35 @@ and make_inconsistent_operand =
...
+      | Bool => List.map(b => wrap(boollit(b)), [true, false])
+      | Prod(_, _) => [
```
cyrus-: for products, you'll want to figure out how many commas to put in -- you can ask Qiran for pointers to his code to do this...

cyrus- (changes requested):

src/hazelcore/semantics/UHPat.re
cdfa: I wasn't sure if it mattered if generated holes aren't used. I can make another `switch` that generates the right number of holes as an array, but this feels more complicated.
Please tell me if you know a nice solution :)

src/hazelcore/semantics/UHPat.re
cdfa: Yeah, I wasn't sure about this either. I feel like matching on ints is rarely done, but the only use-case I could think of is matching on response/error codes. You would probably have a separate type for this in most applications, but with `0` and `_` I think almost everyone would use `== 0` anyway

cdfa: Oh wait, it seems we don't have something like `== 0` yet in Hazel. I'll change it to your suggestion.

src/hazelcore/semantics/UHPat.re
cdfa: Alright, I'll have a look

src/hazelcore/semantics/UHPat.re
cdfa: This happened to fix itself when I was working on generating the right number of holes for a product type, because all cases get their holes differently now

src/hazelcore/semantics/OpSeq.re
cdfa: I don't really understand the meaning of these numbers, but I found a snippet in UHTyp that had the same code, but with a 1 in one of the placeholders, so I assume that's okay.

dm0n3y: Your fix is good 👍  

An `OpSeq.t` consists of a `Skel.t` and a `Seq.t`. The `Seq.t` is a flat sequential representation, which is good for editing. The `Skel.t` is the tree-structured representation determined by operator precedence, which is necessary for type-checking and evaluation. The `Placeholder` numbers in the `Skel.t` are indexes into the accompanying `Seq.t` so that you know which operand is at each skel leaf. For example, notice how `Statics.Exp.syn_skel` (which synthesizes a type for some part of opseq) recurses down the skel structure, then looks up the right operand in the accompanying seq when it hits `Placeholder`.

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +158,51 @@ and make_inconsistent_operand =
     (Parenthesized(set_p), u_gen);
   };
 
+let patterns_of_type =
+    (u_gen: MetaVarGen.t, ty: HTyp.t): (list(t), MetaVarGen.t) =>
+  OpSeq.(
+    switch (ty) {
+    | HTyp.Hole
+    | Unit
+    | Arrow(_, _) => ([], u_gen)
```
cyrus-: in these cases maybe generate a pattern containing just a single hole, instead of the empty list?

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +158,51 @@ and make_inconsistent_operand =
...
+    | Num => ([numlit(0), wild()] |> List.map(wrap), u_gen)
+    | Bool => ([true, false] |> List.map(b => wrap(boollit(b))), u_gen)
+    | Prod(_, _) =>
+      let prod_element_count = ty |> HTyp.get_prod_elements |> List.length;
+      let patternSkel =
+        ListUtil.range(prod_element_count)
+        |> List.map(i => Skel.Placeholder(i))
+        |> make_tuple(NotInHole);
+      let opseq_of_holes = u_gen => {
```
cyrus-: did you reach out to Qiran to ask about his tuple completion code? I imagine it has very similar logic in it, so you might be able to put it in a common helper function instead of duplicating it.

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +158,51 @@ and make_inconsistent_operand =
...
+          OpSeq(
+            patternSkel,
+            Seq.mk(firstHole, holes |> List.map(hole => (Comma, hole))),
+          );
+        (opseq, u_gen);
+      };
+
+      let (opseq1, u_gen) = opseq_of_holes(u_gen);
+      let (opseq2, u_gen) = opseq_of_holes(u_gen);
+      ([opseq1, opseq2], u_gen);
```
cyrus-: why does this return two patterns? there should only be one I think...

src/hazelcore/semantics/UHPat.re
```diff
@@ -153,6 +158,51 @@ and make_inconsistent_operand =
...
+      let (holes, u_gen) = new_EmptyHoles(2, u_gen);
+      let patterns =
+        List.map2(
+          (side, hole) => wrap(injection(side, wrap(hole))),
+          InjSide.[L, R],
+          holes,
+        );
+      (patterns, u_gen);
+    | List(_) =>
+      let (hole1, u_gen) = u_gen |> new_EmptyHole;
```
cyrus-: can use new_EmptyHoles here too

src/hazelcore/util/ListUtil.re
```diff
@@ -226,3 +226,16 @@ let take_while = (p: 'x => bool, xs: list('x)): list('x) =>
        [],
      )
   |> List.rev;
+
+let iterate = (f: 'x => ('y, 'x), times: int, start: 'x): (list('y), 'x) => {
```
cyrus-: you should use an existing List helper function, e.g. fold_left, for this. for loops shouldn't be part of pure functional code.

cyrus- (changes requested):

src/hazelcore/semantics/OpSeq.re
cdfa: Ahaa, thank you very much. So when combining 2 `OpSeq`s (or integrating one created like this in a full program), these placeholders should be renumbered?
Doesn't that take a lot of work when you insert something at the beginning of your program?

src/hazelcore/semantics/UHPat.re
cdfa: Well, in these cases there are just no patterns to generate. I thought returning an empty list would be clearer about this than one with just a single hole pattern.

src/hazelcore/semantics/UHPat.re
cdfa: I thought about this as well for a little bit, but what use is a `case` expression with a single pattern? For safe code, you'll always want at least 2.

src/hazelcore/semantics/UHPat.re
cdfa: I could, but I would have to destructure the list in the assignment (or rather pattern match on a list with 2 elements). The compiler will complain about not having all cases handled, so I would have to disable the warning.
The current approach is less concise, but it's safe by construction, so overall I think it's better.

src/hazelcore/semantics/UHPat.re
cdfa: I found his code (`Action._complete_tuple` I think, but it think it looks a lot more complicated. I'll see if I can unify the two.
It seems construction of tuples with more than 2 elements is a little bit broken at the moment, but I'll ask about this in Slack.
Personally, I think interpreting nested `HTyp.Prod`s as n-tuples could lead to undesired behaviour. For example, if I have a tuple `1, (2,3,4,5,6)`, I might want to just pattern match on the first element (`1, _`), but now I'm forced to have the same number of elements in my pattern (`1, _,_,_,_,_`). To fix this `HTyp.t` has to be changed to contain information to distinguish a right element from being the rest of the tuple (for the n-tuple case) or a nested tuple as last element of the current tuple, but I think this would be out of scope for this PR.

Edit: I meant `HTyp` instead of `UHType`.

src/hazelcore/semantics/UHPat.re
cyrus-: you can technically pattern match on arrow, its just that the only valid patterns are wildcard and variable. but leaving a hole for this is fine. not a big deal but a naive implementation of the next step would have to special case this situation to avoid removing all the rules in the case.

src/hazelcore/semantics/UHPat.re
cyrus-: that's true when you are at the top-level of a case expression, but we can have this mechanism work within sub-patterns too (e.g. if you had a list of tuples, the first invocation would create the nil and cons, but then you might want to invoke it again in head position of the cons pattern)

there is only one pattern form for tuples -- the tuple pattern

src/hazelcore/semantics/UHPat.re
cyrus-: ah fair

src/hazelcore/semantics/UHPat.re
cyrus-: the nested product types are interpreted as n-tuples only if the right-hand side isn't parenthesized, so in your example if the type is (Num, (Num, Num, Num, Num, Num)), parenthesized like that, then it would be treated as a pair.

src/hazelcore/semantics/UHPat.re
cdfa: Okay, I'll keep it in mind.

src/hazelcore/semantics/UHPat.re
cdfa: Okay, got it. I'll change it to one.

src/hazelcore/semantics/UHPat.re
cdfa: But `HTyp` doesn't contain any information about the parentheses anymore, so doesn't that make it indistinguishable at that level? Also both `master` and `dev` currently expect a 6-tuple pattern.

cyrus-: ah yeah that's true... maybe we should explicitly represent n-tuples in HTyp...

cyrus-: yeah I think that makes sense. this is going to affect both your code and Qiran's tuple completion code. do you want to coordinate with him on doing this refactoring? (change Prod to take a list(t) argument + generate that list in the UHTyp.expand function + changing all the code that uses HTyp)

cdfa: Alright, I'll coordinate with him.

src/hazelcore/semantics/OpSeq.re
dm0n3y: For opseqs of length 1 or 2, like in this case, it's fine to manually number. But generally you should only construct the seq, then make a call to `OpSeq.mk` and pass in the right `Associator` function to handle the skel stuff for you. For example, take a look at `Action.Exp.mk_OpSeq` and how it is used. Any opseq construction you wanna do will take place in `Action`, so I would just make use of the existing `mk_OpSeq` functions there.

src/hazelcore/semantics/UHPat.re
dm0n3y: @cdfa Keep me in the loop re: n-tuples. I recently refactored the typechecker (`Statics.re`) to support n-tuples, I can help you propagate the refactor there if you get stuck.

src/hazelcore/semantics/OpSeq.re
cdfa: Yeah, I wanted to use an `Associator` like that first, but that would result in a dependency cycle. The code would probably be simple if with just an `mk_OpSeq` argument or only return the `Seq`.

cyrus-:
@cdfa @dmoon1221 this is a few commits behind master -- can someone merge the latest changes in

cdfa:
> can someone merge the latest changes in

Done! :)

cyrus-:
@cdfa what is the status of this one?

cdfa:
It's finished, but it depends on the tuple refactoring, so I put it on draft until that's merged

cyrus-:
mark it as ready to review once you merge in `dev`

cdfa:
Just merged in `dev`, but realised it was wrong to always have the cursor jump to the second element of a tuple after completion (which I did when I was unifying tuple completion and pattern generation logic). Going to change it to use the move to next hole action, but the rest can be reviewed now

cyrus-:
@cdfa let's have the SplitCases action work when the cursor is on the scrutinee of a case too

cyrus-:
it would also be nice if it worked in this position, replacing the rule with two rules
Screenshot from 2020-05-17 23-20-32

cyrus-:
Otherwise, behaves well! I'll do a detailed code review once you have everything ready to merge.

cdfa:
> @cdfa let's have the SplitCases action work when the cursor is on the scrutinee of a case too

Done!

> it would also be nice if it worked in this position, replacing the rule with two rules

At first, I did not think I would manage to do this in the time I have, but then I got the idea to make a `map_z` function for `ZPat` analogous to the one in `ZList`. What I still don't know is how to get the subtype of the scrutinee corresponding with the pattern operand the cursor is one. Is there any existing function I could use for this?
Another option would be to simultaneously descend the `UHPat` from erasing the cursor and the `HTyp` and finding the operand on which `skel` is rooted. I think something similar is done in `CursorInfo`, but I feel like this approach is less elegant and would take more time.

cyrus-:
@cdfa you should be able to use the data in CursorInfo.t to get the type at the cursor

cyrus-:
something is off with how you're generating hole numbers... you re-use 1, 2, 3 even though they aren't the next hole numbers in the MetaVarGen...
Screenshot from 2020-05-24 13-38-39

cyrus-:
Also if you press Alt+Y here:
Screenshot from 2020-05-24 13-40-24
You get a duplicate nil case, with a duplicate hole 3...
Screenshot from 2020-05-24 13-40-34

cdfa:
Hole number fix waiting for #307, but otherwise ready

cyrus- (changes requested):
Looks good. Only issue is that there is no undo history entry being generated for case split, which breaks undo. Do you have time to fix that @cdfa?

cdfa:
I can look into it. Don't know how difficult it would be

cyrus-:
Should be pretty straightforward -- see UndoHistory and UndoHistoryPanel. You can reach out to @PPER if you need help with that code.

cyrus- (changes requested):
- needs to be merged with latest dev changes, which renamed a bunch of modules
- can you add support for case split when your cursor is in the pattern of a let binding or function and the case split produces a single pattern, e.g. for tuples? basically an alternative to the current tuple completion mechanism.

cdfa:
> can you add support for case split when your cursor is in the pattern of a let binding or function and the case split produces a single pattern, e.g. for tuples? basically an alternative to the current tuple completion mechanism.

I'll have a look. I think this could also be extended to work for arbitrary patterns in let-bindings if you insert a `case` and patterns match on the argument. You would need a name for the argument, but you could leave a hole if it doesn't have a name already.

cdfa:
While working on this I noticed it is possible to assign values to pattern literals, which does not make sense, but it seems to fail silently. Probably a known issue, but thought I'd mention it anyway.
image
With a sensible let-binding the result is just `3`, so it kind of fails silently at the moment

cyrus-:
That is a failed match, but we don't have a nice user interface for that at the moment so it just shows as stuck at the let binding.

cdfa:
> That is a failed match, but we don't have a nice user interface for that at the moment so it just shows as stuck at the let binding.

Okay :)

I now implemented pattern generation for let-bindings and lambda expressions, but no `case` insertion yet. Let me know what you think about that idea.

cyrus-:
Archiving because cannot be merged into Hazel 3, but cross-referenced in #840. Thank you for the work you did on this @cdfa!

PR #212, opened 2020-03-13
Fix dynamics

dm0n3y:
Revive live results and hole closure navigation. Create `DHDoc`/`DHLayout`/`DHAnnot`/`DHCode` for dhexps. Rename `TermDoc`/`TermLayout`/`TermAnnot`/`Code` to `UHDoc`/`UHLayout`/`UHAnnot`/`UHCode`. Add option to show casts.


dm0n3y:
Ready to merge. Can complete `qsort` example. Noticed a few `InvalidInstance` exceptions a few days back, but haven't been able to repro since. For now, I suppress the exception and just log when that happens.

cyrus-:
Looks to be functional, but:
* Can we provide a visual indication of the currently selected closure, like on master (where it was a purple outline)?
* Can you change the mouse pointer shape to be a hand to indicate that it is clickable when hovering over a closure
* Trying to load qsort example from the dropdown on firefox gives me a stack overflow. On Chrome, it works but slow. Is this related to the pretty printer or to the dynamics?

dm0n3y:
Addressed first two issues. Third issue happens on dev too (not on the optimized hosted version, but if you build dev version). My guess is pretty printer.

cyrus-:
Awesome, seems to work well. Is the DHExp pretty printer working for all forms?

Also, can we change the way failed casts are displayed to show the actual and expected type (Right now, its always hole => expected, and the actual type isn't shown)

After that, I'm happy to merge it.

dm0n3y:
Fixed failed cast display. DHExp pretty printing works for everything. Opseqs might look weird if operands are multiline, I plan to revisit (for UHExp too) in a separate pass.

PR #213, opened 2020-03-17
refactor `HTyp.Prod` to contain list of element types

cdfa:
Currently, tuples with a nested tuple as right element and n-tuples are indistinguishable, because there is no information about parentheses in `HTyp.t`. This is solved by adding a type argument of `list(t)` to `Prod`.
The arguments of `Prod` are also abstracted to a `tuple` type, so all element types can easily be uniformly processed with `Tuple.list_of_tuple, while the compiler ensures tuple types always have at least 2 elements.

NOTE: for some reason I have not been able to figure out, JSOO generates some code that is not optimised in Firefox and causes a stack overflow when trying to load the `qsort` example. Chromium and release profiles work fine.


cdfa:
When I started with this, there were several functions in `HTyp` that could benefit of some of the generalizations made by the division (or rather, I felt more like generalizing stuff than fixing multiple functions the same way). However, the only function that wasn't redundant/unused was `consistent`, so the generalization isn't necessary anymore, so I'll revert it. I also hadn't realised how pervasive the type is.
I'll also follow your advice on `Prod` taking a `list(t)`.

cdfa:
@cyrus- I reverted the requested changes. For `eq`, I know just aliased it to `==` and added a comment. I also added a section on the wildcard issue to the code style guide.

cyrus-:
OK sounds good, thanks!

cdfa:
I like the idea, but I'm want to finish `Prod(t, t)` to `Prod(list(t))` refactor first. About that... I just started propagating it to `Dynamics` and I think `DHExp` and `DHPat` have the same ambiguity through `Pair`. I haven't really looked into it yet, but changing this as well might be a little above my experience level with this repo.

cyrus-:
I think it's okay to leave DHExp/DHPat as-is for now -- the issue isn't really important at that level.

cdfa:
> I think it's okay to leave DHExp/DHPat as-is for now -- the issue isn't really important at that level.

Okay, I fear that it will lead to some weird bugs, but we'll see.

All compilation errors have been fixed now (except for one case which now `failswith("not implemented")`, waiting on word from @dmoon1221 for it), but in many cases I'm not sure how to test it. Could somebody help me with that?

cdfa:
> we could replace Unit with Prod([]), since Unit is really a nullary product

Finished!

src/hazelcore/semantics/HTyp.re
```diff
@@ -1,69 +1,55 @@
...
-  | (Prod(ty1, ty2), Prod(ty1', ty2')) => eq(ty1, ty1') && eq(ty2, ty2')
-  | (Prod(_, _), _) => false
-  | (Sum(ty1, ty2), Sum(ty1', ty2')) => eq(ty1, ty1') && eq(ty2, ty2')
-  | (Sum(_, _), _) => false
-  | (List(ty), List(ty')) => eq(ty, ty')
-  | (List(_), _) => false
-  };
+/* eqity
+   At the moment, this coincides with default equality,
+   but this will change when polymorphic types are implemented */
```
dm0n3y, line 39: 👍

src/hazelcore/layout/HTypDoc.re
```diff
@@ -71,12 +53,28 @@ let rec mk = (~parenthesize=false, ~enforce_inline: bool, ty: HTyp.t): t => {
...
+    | Prod([head, ...tail]) =>
+      [
+        mk'(
+          ~parenthesize=HTyp.precedence(head) >= HTyp.precedence_Prod,
+          head,
+        ),
+        ...List.map(
+             ty =>
+               mk'(
+                 ~parenthesize=HTyp.precedence(ty) > HTyp.precedence_Prod,
```
dm0n3y: I think this should be `>=` now. `>` made sense when comma was right-associative, saying there's no need to parenthesize an inner `Prod`. No more associativity now that `Prod` isn't binary.

src/hazelcore/layout/HTypDoc.re
```diff
@@ -44,14 +28,13 @@ let mk_delim = s => Doc.(annot(HTypAnnot.Delim, text(s)));
 let rec mk = (~parenthesize=false, ~enforce_inline: bool, ty: HTyp.t): t => {
   open Doc;
   let mk' = mk(~enforce_inline);
-  let mk_right_associative_operands = (precedence_op, ty1, ty2) => (
-    mk'(~parenthesize=precedence(ty1) >= precedence_op, ty1),
-    mk'(~parenthesize=precedence(ty2) > precedence_op, ty2),
+  let mk_right_associative_operands = (ty, ty1, ty2) => (
```
dm0n3y: nit: I'd prefer this change be reverted. Less clear how to interpret arguments now that there are three types that overlap.

src/hazelcore/semantics/Action.re
```diff
@@ -93,7 +93,7 @@ module Typ = {
 
   let mk_ZOpSeq =
     ZOpSeq.mk(
-      ~associate=Associator.Typ.associate,
+      ~associate=UHTyp.associate,
```
dm0n3y: Nice. Let's move these `mk_ZOpSeq` functions into `ZTyp`/`ZPat`/`ZExp`. Only reason they were here was because I couldn't put `Associator` things in those modules.

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -909,17 +979,30 @@ module Exp = {
...
-        | Expands(d2, ty2, delta) =>
-          let d = DHExp.Pair(d1, d2);
-          let ty = HTyp.Prod(ty1, ty2);
-          Expands(d, ty, delta);
-        }
+    | BinOp(NotInHole, Comma, _, _) =>
+      switch (UHExp.get_tuple_elements(skel)) {
+      | [skel1, skel2, ...tail] =>
+        let%bind (dp1, ty1, delta) = syn_expand_skel(ctx, delta, skel1, seq);
+        let%bind (dp2, ty2, delta) = syn_expand_skel(ctx, delta, skel2, seq);
```
dm0n3y, line 1024: @cyrus- any opinions on let syntax?

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -8,6 +8,29 @@ module Pat = {
     | Expands(DHPat.t, HTyp.t, Contexts.t, Delta.t)
```
dm0n3y: Let's make this type `t` in a new `ExpandResult` module and add a `map` function instead of shuffling back and forth between options.

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -33,16 +56,32 @@ module Pat = {
...
+          syn_expand_skel(ctx, delta, skel1, seq);
+        let%bind (dp2, ty2, ctx, delta) =
+          syn_expand_skel(ctx, delta, skel2, seq);
+        tail
+        |> ListUtil.map_with_accumulator_opt(
+             ((dp_acc, ctx, delta), skel) => {
+               syn_expand_skel(ctx, delta, skel, seq)
+               |> option_of_expand_result
+               |> Option.map(((dp, ty, ctx, delta)) =>
+                    ((DHPat.Pair(dp_acc, dp), ctx, delta), ty)
```
dm0n3y, line 69: @cyrus- Does left vs right associativity for `DHPat`/`DHExp` matter? If yes, what should it be? This is left.

src/hazelcore/semantics/UHTyp.re
```diff
@@ -37,60 +22,62 @@ let rec get_prod_elements: skel => list(skel) =
...
+      | Hole => Seq.wrap(Hole)
+      | Num => Seq.wrap(Num)
+      | Bool => Seq.wrap(Bool)
+      | Arrow(ty1, ty2) => mk_seq_operand(ty, Arrow, ty1, ty2)
+      | Prod([]) => Seq.wrap(Unit)
+      | Prod([head, ...tail]) =>
+        tail
+        |> List.map(ty =>
+             contract_to_seq(
+               ~parenthesize=HTyp.precedence(ty) > HTyp.precedence_Prod,
```
dm0n3y: This should be `>=`

src/hazelcore/semantics/HTyp.re
```diff
@@ -163,18 +125,18 @@ let rec join = (ty1, ty2) =>
...
     switch (join(ty1, ty1'), join(ty2, ty2')) {
     | (Some(ty1), Some(ty2)) => Some(Sum(ty1, ty2))
     | _ => None
     }
   | (Sum(_), _) => None
+  | (Prod(tys1), Prod(tys2)) =>
+    ListUtil.map2_opt(join, tys1, tys2)
+    |> Option.map(OptUtil.sequence)
+    |> Option.join
+    |> Option.map(joinedTypes => Prod(joinedTypes))
```
dm0n3y: nit: use snake_case

src/hazelcore/semantics/Seq.re
```diff
@@ -254,3 +256,34 @@ let t_of_operator_and_surround =
...
+  | S(hd, A(op, seq)) =>
+    let n = counter^;
+    counter := n + 1;
+    Hashtbl.add(ph_map, n, hd);
+    let skel_str = make_skel_str'(string_of_op, seq, counter, ph_map);
+    let op_str = string_of_op(op);
+    string_of_int(n) ++ op_str ++ skel_str;
+  };
+
+let make_skel_str =
```
dm0n3y: Let's move these functions into `Skel.re`.

dm0n3y (commented):
Looks good overall! Some minor requests + questions for @cyrus- in comments. Sorry for the delay, turnaround should get better now that the semester wrapping up.

src/hazelcore/layout/HTypDoc.re
cdfa: Fair. Will revert.

src/hazelcore/layout/HTypDoc.re
cdfa: True. I think parsing would also be changed accordingly though right? I'm not really sure how to do that.

src/hazelcore/semantics/dynamics/Dynamics.re
cdfa: I moved it to its own module. I considered writing a `map` function when I was writing this too, but this seemed more future proof, since we could use any function that works on option this way. Otherwise we keep duplicating `Option` functions in this module.
What do you think about this?

src/hazelcore/Operator.re
```diff
@@ -0,0 +1,103 @@
+module Typ = {
```
cyrus-: let's rename this module to Operators

src/hazelcore/Operator.re
```diff
@@ -0,0 +1,103 @@
...
+  [@deriving sexp]
+  type operator =
```
cyrus-: since we are now in an Operators module, we can just name these types `t` instead of `operator`

src/hazelcore/Operator.re
```diff
@@ -0,0 +1,103 @@
...
+    | Arrow
+    | Prod
+    | Sum;
+
+  let string_of_operator =
```
cyrus-: can just call these "to_string"

src/hazelcore/Operator.re
```diff
@@ -0,0 +1,103 @@
...
+    fun
+    | Arrow => UnicodeConstants.typeArrowSym
+    | Prod => ","
+    | Sum => "|";
+
+  let parse_string_of_operator = op =>
```
cyrus-: `to_parse_string`

src/hazelcore/layout/HTypDoc.re
cyrus-: This is pretty printing not parsing.

src/hazelcore/semantics/HTyp.re
```diff
@@ -1,69 +1,55 @@
...
-  | (Arrow(ty1, ty2), Arrow(ty1', ty2')) => eq(ty1, ty1') && eq(ty2, ty2')
-  | (Arrow(_, _), _) => false
-  | (Prod(ty1, ty2), Prod(ty1', ty2')) => eq(ty1, ty1') && eq(ty2, ty2')
-  | (Prod(_, _), _) => false
-  | (Sum(ty1, ty2), Sum(ty1', ty2')) => eq(ty1, ty1') && eq(ty2, ty2')
-  | (Sum(_, _), _) => false
-  | (List(ty), List(ty')) => eq(ty, ty')
-  | (List(_), _) => false
-  };
+/* eqity
```
cyrus-: this should say "equality"

src/hazelcore/semantics/HTyp.re
```diff
@@ -1,69 +1,55 @@
...
   | (Bool, _) => false
   | (Arrow(ty1, ty2), Arrow(ty1', ty2'))
-  | (Prod(ty1, ty2), Prod(ty1', ty2'))
   | (Sum(ty1, ty2), Sum(ty1', ty2')) =>
     consistent(ty1, ty1') && consistent(ty2, ty2')
   | (Arrow(_, _), _) => false
-  | (Prod(_, _), _) => false
   | (Sum(_, _), _) => false
+  | (Prod(tys1), Prod(tys2)) =>
+    ListUtil.for_all2_opt(consistent, tys1, tys2)
```
cyrus-, line 59: why not define a ListUtil.for_all_zip that returns false if the lists are different lengths, instead of dealing with option here?

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -146,7 +146,7 @@ module Pat = {
          )
```
cyrus-, line 146: when you have an n-tuple inside a hole, e.g. in the following code:
```
let x : Num = 12,| 13
```
it says "expecting an expression of type 1-tuple, got tuple of the wrong length: 2-tuple" but it should say "expecting an expression of type Num got Num, Num. That needs to be corrected in CursorInfo somewhere probably (or maybe in CursorInspector, but probably here).

src/hazelcore/semantics/UHExp.re
```diff
@@ -1,42 +1,6 @@
...
-let is_Space =
-  fun
-  | Space => true
-  | _ => false;
-
-let is_Comma =
-  fun
-  | Comma => true
-  | _ => false;
+include Operator.Exp;
```
cyrus-: includes are not a great idea generally, let's not do this.

src/hazelcore/semantics/UHExp.re
```diff
@@ -319,3 +283,15 @@ let text_operand =
       u_gen,
     );
   };
+
+let parse = s => {
```
cyrus-: this shouldn't really be exposed / called outside of associator -- maybe just fold it into associate

src/hazelcore/semantics/UHPat.re
```diff
@@ -1,28 +1,8 @@
...
-  fun
-  | Comma => ","
-  | Space => " "
-  | Cons => "::";
-
-let is_Space =
-  fun
-  | Space => true
-  | _ => false;
+include Operator.Pat;
```
cyrus-: see above

src/hazelcore/semantics/UHPat.re
```diff
@@ -167,3 +147,13 @@ let text_operand =
       u_gen,
     );
   };
+
+let parse = s => {
```
cyrus-: see above

src/hazelcore/semantics/UHTyp.re
```diff
@@ -1,19 +1,4 @@
...
-  fun
-  | Arrow => UnicodeConstants.typeArrowSym
-  | Prod => ","
-  | Sum => "|";
-
-let is_Prod =
-  fun
-  | Prod => true
-  | _ => false;
+include Operator.Typ;
```
cyrus-: see above

src/hazelcore/semantics/UHTyp.re
```diff
@@ -37,60 +22,62 @@ let rec get_prod_elements: skel => list(skel) =
...
+  let lexbuf = Lexing.from_string(s);
+  SkelTypParser.skel_typ(SkelTypLexer.read, lexbuf);
+};
 
-/* TODO fix this to only parenthesize when necessary */
+let associate = (seq: seq) => {
+  let (skel_str, _) = Skel.make_skel_str(seq, parse_string_of_operator);
+  parse(skel_str);
+};
 let contract = (ty: HTyp.t): t => {
```
cyrus-, line 46: this isn't working anymore -- its fully parenthesizing everything, which is a regression.

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -4,72 +4,113 @@ open Sexplib.Std;
...
+      | DoesNotExpand => None
+      | Expands(pat, ty, ctx, delta) => Some((pat, ty, ctx, delta));
+
+    let from_option =
+      fun
+      | None => DoesNotExpand
+      | Some((pat, ty, ctx, delta)) => Expands(pat, ty, ctx, delta);
+
+    // for use with ppx_let. Don't know why I need to wrap it in a module twice
+    module Let_syntax = {
```
cyrus-: I don't think you're using this. But if you do want to use it, the right way to avoid nesting is to do `module Let_syntax = Pat.ExpandResult.Let_syntax` instead of `open Pat.ExpandResult.Let_syntax`, since ppx_let uses the current binding of `Let_syntax`.

cyrus-: see https://github.com/janestreet/ppx_let

src/hazelcore/semantics/dynamics/Dynamics.re
cyrus-: Pair isn't an infix operator anymore so associativity isn't relevant anymore...

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -4,72 +4,113 @@ open Sexplib.Std;
...
+          syn_expand_skel(ctx, delta, skel2, seq);
+        tail
+        |> ListUtil.map_with_accumulator_opt(
+             ((dp_acc, ctx, delta), skel) => {
+               syn_expand_skel(ctx, delta, skel, seq)
+               |> ExpandResult.to_option
+               |> Option.map(((dp, ty, ctx, delta)) =>
+                    ((DHPat.Pair(dp_acc, dp), ctx, delta), ty)
+                  )
+             },
```
cyrus-, line 71: hard to test the rest of this until the reverse bug in the comment is fixed, so ping me when that's done

cyrus- (changes requested):

cyrus-:
@cdfa there is some sort of bug here, looks like the tuple type is reversed:

Screenshot from 2020-05-03 00-08-52

will finish reviewing expansion / dynamics stuff after that is fixed.

src/hazelcore/layout/HTypDoc.re
cdfa: I meant `,` is still parsed like a right-associative operator right? I think changing this to `>=` wouldn't directly cause any problems, because it would only generate extra parentheses, but I thought it would be nice to keep them in sync.
Anyway, I'll change it to `>=`.

src/hazelcore/semantics/HTyp.re
cdfa: Like this you can stil differentiate between the cases where the lists have different lengths and where the predicate is false. I think if the default value when the lists are of different lengths would always be `false` it would not matter, but I think the default might be `true` as well sometimes.

src/hazelcore/layout/HTypDoc.re
cdfa: I changed the parsing in 0714710de73f3726c8140c4b36de90d8b4461e7f

src/hazelcore/semantics/UHTyp.re
cdfa: Should be fixed now, but I had to change more things around than I expected. Let me know if you see something incorrect/an easier way.

src/hazelcore/semantics/dynamics/Dynamics.re
cdfa: I think I did what you had in mind. Have a look :)

src/hazelcore/layout/HTypDoc.re
cdfa: While working on this I also noticed the parenthesizing logic is duplicated in `HTypDoc` and `UHTyp`. I think it could be unified by using `contract` in `HTypDoc.mk`, but then you are pretty printing a `UHTyp`, for which `UHDoc` might be useful, but I don't really know anything about how all the layout stuff works.
What do you think about this?

cdfa:
@cyrus- Reverse bug is fixed!

src/hazelcore/semantics/CursorInfo.re
cdfa: I started looking in to this and I found that the code [here](https://github.com/hazelgrove/hazel/blob/fix-tuple-type-ambiguity/src/hazelcore/semantics/CursorInfo.re#L847) assumes `ty` is a tuple type, which is not the case (this does not cause an error, because `HTyp.get_prod_elements` (and similar for `UHTyp`, `UHPat` etc.) just creates a singleton list of the given type when the given type is not a product type. It feels like a nasty shortcut and has been a bit of a thorn in my eye since starting this refactoring, but sadly the function is quite widespread and I don't feel like I'm up to the task of removing it everywhere) (sorry for the rant :sweat_smile:).

Anyway, I started searching for how errors are done for arbitrary operators, but I found `ana_cursor_info_zopseq` really only considers tuple `OpSeq`s and when the operator is not `Comma` it just switches to synthetic cursor info, which just says `got as expected` :(. I'm not sure why it does this.

I think this case should generate some kind of `AnaTypeInconsistent` info, but I wanted to pass this by you before I started working on it.

dm0n3y:
Encountering a `SkelExprParser.MenhirBasics.Error` exception when entering comma here
image

src/hazelcore/semantics/dynamics/Dynamics.re
dm0n3y: Thanks. Adding `Option`-like functions as needed sounds fine to me. Either there's a cost of adding those functions, or there's the cost of converting into `Option` types and back whenever we wanna piggyback on the `Option` type functions, and I'd rather pay the former.

cdfa: Okay, now I looked at it more closely, the `Option` type is also required by the use of `map_with_accumulator_opt`. I could make a similar function `map_with_accumulator_expand_result`, but then we're also duplicating `ListUtil` functions, not just `Option`. Should I still go through with it?

cyrus-:
is this ready for review?

cdfa:
Yes, I fixed the exception @dmoon1221 mentioned  and I've been trying to keep it up to date with `dev`.

cyrus-:
@cdfa you're quite behind dev and there are conflicts -- can you resolve those.

cdfa:
Yeah, I am now because many commits were just pushed to `dev` :laughing: . I'll get started on merging.

cyrus-:
tuple completion doesn't work in all analytic positions, e.g. in the following example pressing `(` doesn't do tuple completion
Screenshot from 2020-05-09 13-02-45

src/hazelcore/semantics/CursorInfo.re
cyrus-: once we add labeled tuples, we're definitely going to want to distinguish 1-tuples from non-tuples. this is probably the right time to fix that. can you take a shot at fixing this @cdfa? @dmoon1221 can you assist as necessary?

cyrus-: yes this case should generate AnaTypeInconsistent. the code you linked to above is the right place to fix it.

src/hazelcore/semantics/Skel.re
```diff
@@ -76,3 +76,34 @@ let subskel_rooted_at_op = (op_index: OpIndex.t, skel: t('op)): t('op) => {
...
+    counter := n + 1;
+    Hashtbl.add(ph_map, n, hd);
+    let skel_str = make_skel_str'(string_of_op, seq, counter, ph_map);
+    let op_str = string_of_op(op);
+    string_of_int(n) ++ op_str ++ skel_str;
+  };
+
+let make_skel_str =
+    (seq: Seq.t('operand, 'op), string_of_op: 'op => string)
+    : (string, Hashtbl.t(int, 'operand)) => {
```
cyrus-: this doesn't need to return the hashtable -- I don't think its used anywhere

cyrus- (changes requested):
I can't press comma here
Screenshot from 2020-05-09 13-21-24

dm0n3y:
@cyrus- @cdfa turns out that bug is on dev. I know what the problem is, working on fix.

dm0n3y:
https://github.com/hazelgrove/hazel/issues/251

src/hazelcore/semantics/CursorInfo.re
cdfa: I'm not sure if this is the right time to fix it. This PR is already quite big and it's generating more and more merge conflicts. To be frank, I also just don't have the time to do a full refactor for this.
What I could do is make the functions fail explicitly when its argument is not a tuple, so at least places like this where the assumption of a tuple type is incorrect will become more obvious.

cdfa:
Thank you for fixing the bug @dmoon1221 !

> tuple completion doesn't work in all analytic positions, e.g. in the following example pressing ( doesn't do tuple completion

This is now fixed as well

src/hazelcore/semantics/CursorInfo.re
cdfa: This specific problem of 1-tuples is fixed in acb63046b6c1c02ddfd7ce995ba810b1be3723e3

cyrus-:
the tuple completion bug I mentioned above doesn't seem to be fixed on my end?

dm0n3y:
Are you using the same repro steps as you screenshotted above or are you invoking tuple completion a different way? Current tuple completion via entering comma on an empty hole is currently only triggered when the cursor is at the end of the hole. Shouldn't be limited that way but it's not caused by these changes.

cyrus-:
I'm invoking it by pressing `(`. Anyway, yeah I guess this isn't related to this PR so I'll make a new issue.

cyrus- (approved):

PR #214, opened 2020-03-17
Line swap

Yuning30:
1. **Action.re**: add SwapUp, SwapDown, SwapLeft and SwapRight to Action.t and handle these actions in 
             module Typ, Pat and Exp and the code is placed after the construct action but before the zipper case.

2. **ZExp.re**: add a function called "line_can_be_swapped" which is used in Action.Exp.syn_perform_block and Action.Exp.ana_perform_block to determine if the line can be swapped at the block level.

3. **JSUtil.re**: add new function "ctrl_alt" to indicate the mod key status and add four key combinations which are ctrl_alt_[up, down, left, right]

4. **Cell.re**: use ctrl_alt_[up, down, left, right] to indicate Swap[Up, Down, Left, Right] actions and add  a few lines to the "view" function to let the key combination with arrows are handled

5. **UndoHistory.re**: add four swap actions as undoable actions

### DIFF
--- a/src/hazelcore/semantics/Action.re
+++ b/src/hazelcore/semantics/Action.re
@@ -2531,6 +2531,7 @@ module Exp = {
       /* avoid swap down for the Let line if it is second to last */
       | ([_], LetLineZP(_) | LetLineZA(_) | CursorL(_, LetLine(_))) =>
         Failed
+      /* handle corner case when the second to last line is an EmptyLine */
       | ([last], CursorL(_, EmptyLine)) => {
         let (new_hole, u_gen) = u_gen |> ZExp.new_EmptyHole;
         let new_zblock = (prefix @ [last], ZExp.ExpLineZ(ZOpSeq.wrap(new_hole)), []) |> ZExp.prune_empty_hole_lines;
@@ -3841,11 +3842,12 @@ module Exp = {
       /* avoid swap down for the Let line if it is second to last */
       | ([_], LetLineZP(_) | LetLineZA(_) | CursorL(_, LetLine(_))) =>
         Failed
+      /* handle corner case when the second to last line is an EmptyLine */
       | ([last], CursorL(_, EmptyLine)) => {
         let (new_hole, u_gen) = u_gen |> ZExp.new_EmptyHole;
         let new_zblock = (prefix @ [last], ZExp.ExpLineZ(ZOpSeq.wrap(new_hole)), []) |> ZExp.prune_empty_hole_lines;
         Succeeded(
-          AnaDone(Statics.Exp.ana_fix_holes_z(ctx, u_gen, new_zblock)),
+          AnaDone(Statics.Exp.ana_fix_holes_z(ctx, u_gen, new_zblock, ty)),
         );
       }
       | ([hd, ...tl], _) =>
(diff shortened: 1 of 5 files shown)


src/hazelcore/semantics/ZExp.re
```diff
@@ -37,6 +37,19 @@ type operand_surround = Seq.operand_surround(UHExp.operand, UHExp.operator);
...
+  | CursorL(_)
+  | LetLineZP(_)
+  | LetLineZA(_)
+  | ExpLineZ(ZOpSeq(_, ZOperator(_)))
+  | ExpLineZ(ZOpSeq(_, ZOperand(CursorE(_), _)))
+  | ExpLineZ(ZOpSeq(_, ZOperand(LamZP(_), _)))
+  | ExpLineZ(ZOpSeq(_, ZOperand(LamZA(_), _)))
+  | ExpLineZ(ZOpSeq(_, ZOperand(CaseZA(_), _)))
+  | ExpLineZ(ZOpSeq(_, ZOperand(ApPaletteZ(_), _))) => true
+  | _ => false
```
cyrus-: don't use _ if possible

src/hazelcore/semantics/Action.re
```diff
@@ -3611,6 +3818,28 @@ module Exp = {
...
+      | Some((rest, last)) =>
+        let new_zblock = (rest, zline, [last, ...suffix]);
+        Succeeded(
+          AnaDone(Statics.Exp.ana_fix_holes_z(ctx, u_gen, new_zblock, ty)),
+        );
+      }
+    | (SwapDown, _) when ZExp.line_can_be_swapped(zline) =>
+      switch (suffix, zline) {
+      | ([], _) => Failed
+      /* avoid swap down fot the Let line if it is second to last */
```
cyrus-: fot -> for

src/hazelcore/semantics/Action.re
```diff
@@ -3611,6 +3818,28 @@ module Exp = {
...
+      | ([_], LetLineZP(_) | LetLineZA(_) | CursorL(_, LetLine(_))) =>
+        Failed
+      | ([hd, ...tl], _) =>
+        let new_zblock = (prefix @ [hd], zline, tl);
```
cyrus-: need to do a normalization pass to prevent ExpLines that consist of single holes from being included in the middle of a block (they should be turned into empty lines). There is already code for dealing with this situation somewhere (@dmoon1221 might have an answer for where)

src/hazelweb/gui/Cell.re
```diff
@@ -65,7 +69,22 @@ let view =
               inject(a),
             ]);
           if (JSUtil.is_movement_key(evt)) {
```
cyrus-: I think is_movement_key should probably return false for the ctrl+alt+arrow combos, since thats not a normal movement key

cyrus- (changes requested):
great job -- just a couple of small issues left to resolve and then we should be able to merge

Yuning30:
Hi, I've made changes according to the comments provided by @cyrus-  , but some not user friendly behavior will happen when the SwapUp and SwapDown actions are applied to **EmptyLine** or **LetLines** at the end of program because according to function **Statics.Exp.ana_fix_holes_z** and **Statics.Exp.syn_fix_holes_z**, only **ExpLine** can be the last line of program but **EmptyLine** and **LetLine** cannot

Yuning30:
Hi, up to the last commit, I've finished and tested the code for the corner cases of swapUp and swapDown action. So the swap actions are basically complete.

src/hazelcore/semantics/Action.re
Yuning30: have done the normalization for all new_zblock

cyrus-:
Looks great, nice job!

PR #215, opened 2020-03-29
support case analysis in synthetic position by joining branch types

hannahpotter:
Remove annotation on Case in synthetic position.
Add an error for inconsistent types in branches of case.


cyrus-:
Error message for inconsistent branches looks great. Will do a more thorough code review and merge in the next day or two.

cyrus- (changes requested):
In the following, I can't enter "2" into hole 25 (Program.InvalidInput error on console):

Screenshot from 2020-05-03 15-05-34

Probably something wrong with cast insertion? If you can't figure it out lmk and I can dive in.

cyrus-:
Expected behavior above would be a true with a cast failure on it. This corresponds to using the least upper bound as the join. It is more permissive to use the glb for join, like we did in A6. We should maybe use that instead. That would mean that the case expression's type would be ? rather than Num, because one of the branches has unknown type.

hannahpotter:
> Expected behavior above would be a true with a cast failure on it. This corresponds to using the least upper bound as the join. It is more permissive to use the glb for join, like we did in A6. We should maybe use that instead. That would mean that the case expression's type would be ? rather than Num, because one of the branches has unknown type.

This should be updated now. Still working on figuring out the other bug (although it doesn't show up in that particular case anymore since the case expressions type is now ?).

cyrus-:
@hannahpotter small style issue: can't see the ? very well. maybe instead of alternating bg color just put lines between? or lighten the dark band enough. 
Screenshot from 2020-05-08 11-42-52

src/hazelcore/semantics/HTyp.re
```diff
@@ -153,8 +153,8 @@ let rec complete =
 
 let rec join = (ty1, ty2) =>
```
cyrus-: call this one glb

src/hazelcore/semantics/HTyp.re
```diff
@@ -188,3 +188,34 @@ let rec join = (ty1, ty2) =>
     }
   | (List(_), _) => None
   };
+
+let glb = (types: list(t)): option(t) => {
```
cyrus-: call this one glb_all

src/hazelcore/semantics/HTyp.re
```diff
@@ -188,3 +188,34 @@ let rec join = (ty1, ty2) =>
...
+  switch (types) {
+  | [] => None
+  | [hd] => Some(hd)
+  | [hd, ...tl] =>
+    let rec exist_inconsistencies = l =>
```
cyrus-: move this into a top-level helper function "consistent_all"

src/hazelcore/semantics/Statics.re
```diff
@@ -895,19 +898,40 @@ module Exp = {
...
+    | Case(InHole(InconsistentBranches(rule_types), _), scrut, rules) =>
+      switch (syn(ctx, scrut)) {
+      | None => None
+      | Some(pat_ty) =>
+        List.fold_left2(
+          (ty, rule_type, rule) => {
+            switch (ty) {
+            | None => None
+            | Some(_) =>
+              switch (ana_rule(ctx, rule, pat_ty, rule_type)) {
```
cyrus-: this should actually be checking that the rule synthesizes the rule_type that the hole reason says it does, which isn't quite the same as analysis

cyrus-: (use List.all or similar for this, and then just return HTyp.Hole if that works, rather than folding)

src/hazelcore/semantics/ErrStatus.re
```diff
@@ -1,9 +1,12 @@
+open Sexplib.Std;
+
 module HoleReason = {
   /* Variable: `reason` */
   [@deriving sexp]
   type t =
     | TypeInconsistent
-    | WrongLength;
+    | WrongLength
+    | InconsistentBranches(list(HTyp.t));
```
cyrus-: it's cleaner to have a separate CaseErrStatus.t that only appears on case than to have InconsistentBranches show up in all the pattern matches. TypeInconsistent can appear on any form, and WrongLength can appear on any opseq containing commas, but InconsistentBranches can only appear on one form.

src/hazelweb/www/style.css
```diff
@@ -360,6 +361,19 @@ html, body {
   padding-right: 5px;
 }
 
+.inconsistent-branches-ty-bar {
+  flex-direction: column;
+  align-items: flex-start;
+  flex-basis: auto;
+  /* enable scroll */
+  overflow-y: auto;
```
cyrus-: this should be on .typebar, .infobar since you set a max height

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1049,15 +1053,55 @@ module Exp = {
...
+      switch (syn_expand(ctx, delta, scrut)) {
+      | DoesNotExpand => DoesNotExpand
+      | Expands(d1, pat_ty, delta) =>
+        let expand_rules =
+          List.fold_left2(
+            (b, r_t, r) =>
+              switch (b) {
+              | None => None
+              | Some((drs, delta)) =>
+                switch (ana_expand_rule(ctx, delta, r, pat_ty, r_t)) {
```
cyrus-: this should be syn_expand_rule (see comment in Statics)

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1127,19 +1171,17 @@ module Exp = {
...
       switch (syn_expand(ctx, delta, scrut)) {
       | DoesNotExpand => DoesNotExpand
-      | Expands(d1, ty1, delta) =>
-        switch (ana_expand_rules(ctx, delta, rules, ty1, ty)) {
+      | Expands(d1, ty, delta) =>
+        switch (syn_expand_rules(ctx, delta, rules, ty)) {
         | None => DoesNotExpand
         | Some((drs, delta)) =>
           let d = DHExp.Case(d1, drs, 0);
           Expands(d, ty, delta);
```
cyrus-: the ty here should be the glb of the rule types, not the type of the scrutinee!

src/hazelcore/semantics/CursorInfo.re
```diff
@@ -1091,11 +1093,33 @@ module Exp = {
           ty,
         )
       }
-    | CaseZA(NotInHole, _, rules, zann) =>
-      Typ.cursor_info(~steps=steps @ [1 + List.length(rules)], ctx, zann)
     | ApPaletteZ(NotInHole, _, _, _) =>
       syn_cursor_info_zoperand(~steps, ctx, zoperand)
     }
+  and syn_cursor_info_rule =
```
cyrus-, line 1168: maybe out of scope for this PR, but it would be nice for the cursor inspector to show the glb of all the other branches as the expected type instead of just "any type" in synthetic position

cyrus- (changes requested):
very close -- see comments

hannahpotter:
> small style issue: can't see the ? very well. maybe instead of alternating bg color just put lines between? or lighten the dark band enough.
> Screenshot from 2020-05-08 11-42-52

This is what it looks like with the lightened band
image
This is what it looks like with lines in between (obviously the color/style of the lines could be different)
image
I personally like the banded colors better than the lines (I think that it is easier to keep track of your place as you scroll and such), but I don't feel super passionately about it.

cyrus-:
Lightened band looks good.

Also it would be cool to be able to click a type there and be taken to the corresponding location in the code. I believe we are already tracking path information when we compute cursor info, so that shouldn't be too hard?

src/hazelcore/semantics/CursorInfo.re
hannahpotter: What would it show in the case where the other branches still had type inconsistencies? Would it just show "any type" in that case?

src/hazelcore/semantics/ErrStatus.re
hannahpotter: Would this be similar to what is going on with VarErrStatus? (in the sense that there is an error status that only appears on a particular form)

cyrus-: yeah though I think it would be better to replace the ErrStatus.t on Case with a CaseErrStatus.t that has one constructor for standard errors and then one more for the new one.

hannahpotter: I am not exactly sure how to just replace ErrStatus on Case. It seems like there are places where it is expected that all of the forms will have an ErrStatus on them... Maybe I am just misunderstanding what you mean. 
Why was the VarErrStatus done the way it was if it is better to just replace ErrStatus?

src/hazelcore/semantics/CursorInfo.re
cyrus-: branches only have type inconsistencies on the whole branch expression in analytic position. in any case, you just treat those branches as ?...

src/hazelcore/semantics/ErrStatus.re
cyrus-: Yeah you'd have to look at all the places where there is a get_err_status call and handle the InconsistentBranches situation separately, if it is relevant there. IIRC that function isn't used much and when it is, it is on opseqs, but I might be wrong. If it's too much work to refactor then don't worry about it.

VarErrStatus should probably be done this way too.

hannahpotter: This commit has my attempt at these changes. https://github.com/hazelgrove/hazel/pull/215/commits/f15a75ac87e3dc13f282dde1a1f8520785571066

src/hazelcore/semantics/CursorInfo.re
hannahpotter: image
Are you talking about how the cursor inspector shows "any type" if the cursor is where the red circle is?

hannahpotter:
> Also it would be cool to be able to click a type there and be taken to the corresponding location in the code. I believe we are already tracking path information when we compute cursor info, so that shouldn't be too hard?

Added https://github.com/hazelgrove/hazel/pull/215/commits/dd2344fac51ead22d5befe70339797ae1f7755b2

src/hazelcore/semantics/CursorInfo.re
hannahpotter: https://github.com/hazelgrove/hazel/pull/215/commits/cb5392f6579b509f3a8798801991e3595aae5271

cyrus-:
@hannahpotter `dev` needs to be merged, can you do that and mark as ready for review again

cyrus-:
@hannahpotter can we have the cursor inspector say "expecting an expression consistent with" instead of "expecting an expression of" when the expected type is determined by joining the branches rather than by type analysis?

cyrus-:
also when I click on one of the inconsistent types, focus isn't returned to the cell so I can't actually use the arrow keys. @dmoon1221 what is the right way to programmatically set focus on the cell?

cyrus-:
also let's put the focus on the offending rule rather than the branch expression, so that it highlights both the pattern and the body of interest.

hannahpotter:
> also when I click on one of the inconsistent types, focus isn't returned to the cell so I can't actually use the arrow keys. @dmoon1221 what is the right way to programmatically set focus on the cell?

As far as I can tell, the focus isn't really set when you click on the hole information in the context inspector (which is what I based this on).

cyrus-:
Something is broken with hole ordering (on dev too)
Screenshot from 2020-05-21 00-41-02

If you press tab, it goes from 6 -> 9 -> 7 -> 12 -> 13, but it should go 6 -> 9 -> 10 -> 7 -> 11 -> 12 -> 13.

Can you take a look at that? Should be somewhere in the path code.

cyrus-:
When clicking on types in the inconsistent branches errors, let's put the cursor to the right of the arrow in the rule rather than on the vertical bar. That way it's quick to go either right into the branch or left into the pattern.

cyrus-:
Tab/Shift-Tab doesn't seem to work at all when the cursor is on the arrow in a rule...

cyrus-:
Other than those minor things, looks great. Should be ready to merge after those fixes. (the tab ordering stuff is also an issue in dev, but let's just get case working properly all at once with this PR)

cyrus-:
Actually one more tweak -- the "expecting an expression consistent with type" message could use the lub of the other branches rather than the glb, since its merely informative. that would give you a more specific expectation in situations where some of the remaining branches are unfilled but there is otherwise a consistent type.

cyrus-:
One last refinement: when cursor is on a branch expression for a case where no lub exists for the other branches, instead of saying "expecting an expression of any type" in green, let's say "no consistent expected type" in red and then list the branch types for all of the other branches in the top half of the cursor inspector, with the same UI/clicking support as when you are on the case expression as a whole.

hannahpotter:
> One last refinement: when cursor is on a branch expression for a case where no lub exists for the other branches, instead of saying "expecting an expression of any type" in green, let's say "no consistent expected type" in red and then list the branch types for all of the other branches in the top half of the cursor inspector, with the same UI/clicking support as when you are on the case expression as a whole.

If there is only a single branch, should it say "expecting an expression of any type" in green still?

cyrus-:
Nicely done.

PR #216, opened 2020-04-03
variable usage

### LINKED ISSUE #152: Duplicate variable checking

### LINKED ISSUE #160: Unused variable analysis

### LINKED ISSUE #161: Go to definition action

### LINKED ISSUE #162: Go to next usage action

yongweiy:
- Support GoToDefinition, GoToFirstUsage, GoToNextUsage, GoToPrevUsage
- Add duplicate name error for variables used in a pattern
- Add critically unused variables warning


dm0n3y:
How far are we from merging this? I have a pending bug fix off dev that would benefit from steps being threaded through `CursorInfo`.

yongweiy:
The pull request is simply waiting for review. What do you mean by "thread steps through `CursorInfo`"? The steps is already passed around CursorInfo, right?

src/hazelweb/gui/CursorInspector.re
```diff
@@ -92,25 +96,37 @@ let view = (model: Model.t): Vdom.Node.t => {
...
+  let got_duplicate_indicator =
+    got_indicator("Got a duplicated variable", typebar(HTyp.Hole));
+  let got_var_indicator = (shadow, num_of_uses) =>
+    got_indicator(
+      "Got a variable",
+      special_msg_bar(
+        (shadow ? " shadowing previous variable, " : "")
+        ++ "used "
+        ++ Int.to_string(num_of_uses)
+        ++ " times",
```
cyrus-: You need to check whether the num_of_uses is 0 and write "time" instead of "times" for this to be correct english

cyrus-: also the sidebar shouldn't be green when there is an unused variable or any other kind of warning -- maybe change the background color of the part that says "used 0 times" to be the same color as the crit unused variable outline

cyrus- (changes requested):
The variable y below should not be marked as critically unused, because its scope is not complete (the function body has a hole)

Screenshot from 2020-05-03 15-17-16

y should also not be marked as critically unused in the following:

Screenshot from 2020-05-03 15-18-39

I will await fixes to these before doing a more thorough review of the code.

yongweiy:
For the first screenshot, I did it on purpose. Although `y` could be a recursive function, as long as it is not used after it is defined, I think it is worth a warning. But by our definition of crit_unused, it is true that we shouldn't give a warning here. What do you think?

yongweiy:
For the second issue, is it because the block under lambda involves hole and we shouldn't treat the block as complete?

cyrus-:
ah okay, maybe leave the warning indicator there but change the cursor inspector to distinguish recursive uses from non-recursive uses (as you would if there were a separate fix construct)

src/hazelweb/gui/CursorInspector.re
yongweiy: Screenshot_2020-05-05_20-47-55
Does this fit what in mind? Would the purple be way too highlighting?

src/hazelweb/gui/CursorInspector.re
```diff
@@ -92,25 +96,46 @@ let view = (model: Model.t): Vdom.Node.t => {
...
+  let got_duplicate_indicator =
+    got_indicator("Got a duplicated variable", typebar(HTyp.Hole));
+  let got_var_indicator = (shadow, num_of_non_rec_uses, num_of_rec_uses) =>
+    got_indicator(
+      "Got a variable",
+      special_msg_bar(
+        (shadow ? " shadowing previous variable, " : "")
+        ++ "used "
+        ++ Int.to_string(num_of_non_rec_uses)
+        ++ (num_of_non_rec_uses <= 1 ? " time" : " times")
```
cyrus-: actually, let's change this to say "n uses" instead of "used n time(s)"

src/hazelweb/gui/CursorInspector.re
```diff
@@ -92,25 +96,46 @@ let view = (model: Model.t): Vdom.Node.t => {
...
+        ++ (
+          num_of_rec_uses == 0
+            ? ""
```
cyrus-: change this phrasing to "n recursive uses and m non-recursive uses"

cyrus- (changes requested):
don't like the purple distinction. proposal: use red for both type errors and binding errors (the latter previously used orange), and use orange for all warnings (unused variables, duplicated variables, etc.)

second proposal: change the phrasing to say "at least n uses" when there are holes in the scope of the variable, and "n uses" when there are no holes. add a little warning icon when there are 0 uses to explain why the sidebar is orange.

yongweiy:
I add a warning icon in front of the clause when the clause is part of the warning.

cyrus-:
@VictorYYW let's use the same colors for the dotted boxes around the errors and warnings as we do in the sidebar

cyrus-:
not sure why this isn't showing a warning? it does show a warning if I change those x's to 1's
Screenshot from 2020-05-15 20-17-24

cyrus-:
oh I guess it's because we are counting non-empty holes as holes. that's maybe a little unintuitive, but okay.

yongweiy:
Yeah. In some sense, non-empty hole means that the user still needs to work on that term. But if we change the definition of complete by simply change the parameter we use to call `is_complete`.

cyrus-:
I think that behavior is fine.

cyrus-:
Shadowing shouldn't be considered a warning (sidebar should stay green, no warning icon). It should just be noted, maybe with an "info" icon.

cyrus-:
How hard would it be to have the gray background around other uses when your cursor is on a variable expression? Also what about putting a gray outline around the binding site in that situation as well? I think we have all the data to be able to find the binding site and then from there the other uses...

cyrus-:
also the grammar is still off, it should say "1 use" when it is 1, and "uses" in all other situations.

yongweiy:
As for the other uses when the cursor is on a variable expression, we need to do deferrable stuff in CursorInfo, which is quite like what we did for finding uses for the variable pattern under the cursor except for one difference. We need to further defer the finding uses until we reach the block where the variable is being defined while previous we only need to defer until the innermost block.

yongweiy:
I thought "use" is the right way to go since you commented "x use" last time. I even turned to dictionary for help. It says, "use" could be uncountable :sweat_smile:

yongweiy:
Actually, I would like to indicate both the binding site and other variable uses since that is what I promise at the very beginning (at least for variable usage). :)

cyrus-:
it's "0 uses", "1 use", "2 uses", "3 uses", ..., I promise :)

cyrus-:
when the cursor is on an expression variable, can't we internally go to its definition and then find uses from there to get a list of paths to other uses?

yongweiy:
We don't have steps information in the context when we are in CursorInfo.re. Not sure which way is easier to do.

cyrus-:
I guess yes that would probably need you to use a deferable

cyrus-:
or keep track of steps in the context

cyrus-:
doing it with a deferrable is probably the way to go

yongweiy:
I just took a closer look at how layout is decorated now. Keeping track of steps in the context is also needed since we want to indicate binding site as well. Besides, deferrable helps with finding other uses.

cyrus-:
didn't you need to track steps to implement GoToDefinition already?

yongweiy:
Yeah. You are right. I was confused by myself.

yongweiy:
I will do the deferrable stuff next week then.

cyrus-:
for the definition indicator, maybe we can use a light version of the same color as the indicator for patterns (currently blue) for the outline color. I assume you're still working on highlighting the other uses? (the sidebar always says "0 uses" when you're on a variable expression)

yongweiy:
I didn't notice patterns have a different color from expression! That makes sense. I will change that.
As for the highlighting of other uses, I simply added the invariant and haven't started to implement it. The plan is to get it done later this week or early next week.

yongweiy:
Next step is to fix conflicts.

cyrus-:
merge conflicts need to be merged

cyrus-:
Looking good!

- the numbering is off  -- it says "0 / 2 uses" and "1 / 2 uses"
- let's say "1 of 2 uses" instead of "1 / 2 uses"
- let's make the numbers bold

src/hazelcore/CursorInfo_common.re
```diff
@@ -13,7 +27,7 @@ type typed =
   /* cursor in analytic position */
   | AnaAnnotatedLambda(HTyp.t, HTyp.t)
   // cursor is on a type inconsistent expression
-  | AnaTypeInconsistent(HTyp.t, HTyp.t)
+  | AnaTypeInconsistent(HTyp.t, HTyp.t, varexp)
```
dm0n3y, line 30: What is `varexp` and what are the reasons for adding it to some constructors vs others?

yongweiy: For those cursor position where variable expressions may appear, I add `varexp` to its constructor.

cyrus-:
@VictorYYW this doesn't build for me:

```
dune build @src/fmt --auto-promote || true
File "src/hazelcore/dune", line 3, characters 0-196:
 3 | (library
 4 |  (name hazelcore)
 5 |  (libraries re sexplib camomile pretty)
 6 |  (wrapped false)
 7 |  (js_of_ocaml
 8 |   (flags
 9 |    (:include js-of-ocaml-flags-%{profile})))
10 |  (preprocess
11 |   (pps ppx_let ppx_sexp_conv)))
Error: Some modules don't have an implementation.
You need to add the following field to this stanza:

  (modules_without_implementation elaborator_Exp elaborator_Pat zIntMap)
dune build src --profile dev
File "src/hazelcore/dune", line 3, characters 0-196:
 3 | (library
 4 |  (name hazelcore)
 5 |  (libraries re sexplib camomile pretty)
 6 |  (wrapped false)
 7 |  (js_of_ocaml
 8 |   (flags
 9 |    (:include js-of-ocaml-flags-%{profile})))
10 |  (preprocess
11 |   (pps ppx_let ppx_sexp_conv)))
Error: Some modules don't have an implementation.
You need to add the following field to this stanza:

  (modules_without_implementation elaborator_Exp elaborator_Pat zIntMap)
make: *** [Makefile:13: dev] Error 1
```

yongweiy:
@cyrus- I can't reproduce the error on my side. I guess the error is complaining that it can only find elaborator_Exp.rei but can't find elaborator.re. I am sure that both files exist in the branch. Is there any chance that you are on the wrong branch?

cyrus-:
can you move the action panel entries for your new actions to the "variables" section rather than "movement"

cyrus-:
in synthetic position, we lose the information about what type the variable is in the cursor inspector. can we figure out a way to display both the type and the number of uses? it's okay if it takes more than one line...

cyrus-:
similarly, we don't have uses information for variables in function position. I think we should separate out the type-related message from the variable uses related message. maybe use a light green background for the uses information, like in the history panel.

yongweiy:
Got it. I kind of randomly determine whether we show uses info or the type info based on which one seems to be more interesting. Would adding a variable usage panel help? Are we going to hide such a panel when the cursor is nor on a variable expression or a variable pattern?

src/hazelcore/CursorInfo_common.re
```diff
@@ -221,5 +259,21 @@ let get_ctx = ci => ci.ctx;
...
+        (
+          UsageAnalysis.uses_list,
+          // recursive uses
+          UsageAnalysis.uses_list
+        ) =>
+        't,
+        Var.t,
+      )
+  | CursorOnDeferredVarExp(
+      UsageAnalysis.uses_list => 't,
```
dm0n3y, line 275: Why is this parameter necessary for when the cursor is on an expression variable?

src/hazelcore/dynamics/DHExp.re
```diff
@@ -141,6 +141,7 @@ type t =
   // TODO rename to ExpandingKeyword
   | Keyword(MetaVar.t, MetaVarInst.t, VarMap.t_(t), ExpandingKeyword.t)
   | FreeVar(MetaVar.t, MetaVarInst.t, VarMap.t_(t), Var.t)
+  | Duplicate(MetaVar.t, MetaVarInst.t, VarMap.t_(t), Var.t)
```
dm0n3y: Why does `DHExp` have a duplicate variant? Doesn't that only apply to pattern variables?

src/hazelcore/VarErrStatus.re
```diff
@@ -4,7 +4,8 @@ module HoleReason = {
   [@deriving sexp]
   type t =
     | Free
-    | Keyword(ExpandingKeyword.t);
+    | Keyword(ExpandingKeyword.t)
+    | Duplicate;
```
dm0n3y, line 8: Does this warning also apply for expression variables?

src/hazelcore/VarWarnStatus.re
```diff
@@ -0,0 +1,5 @@
+/* Variable: var_warn */
+[@deriving sexp]
+type t =
+  | NoWarning
+  | CritUnused;
```
dm0n3y, line 5: Remind me what `CritUnused` means again?

yongweiy:
You are right. I just pushed a commit to get rid of `Duplicate` in `DExp`.
The Duplicate error only applies to patterns. I noticed that Free error also only applies to expressions and thought that VarErrStatus.t includes errors about variable no matter it is for var exp or var patterns.
`CritUnused` just means that all the places where the variable could be used doesn't involve holes (empty or non-empty) and the variable is unused. As for variable bound to recursive functions, I didn't take the body of the functions into considerations because a recursive usage is not a real usage in the program.

cyrus-:
My understanding is that @dmoon1221 is going to integrate this into his refactoring, using this branch as a functional reference. Marking it as draft, can be closed once we have the equivalent functionality merged.

cyrus-:
stale, closing

PR #218, opened 2020-04-26
integers and floats

### LINKED ISSUE #191: Floats

Yunif3:
**Num Type, Literal, and Operators have been split into [Int|Float] Type, [Int|Float] Literals, and [Int|Float] Arithmetic Operators.**

I/F will stand for [Int|Float] from now on.

**Type**
- Int type replaced Num type
- Float type was added
- Int operators will treat floats as wrong inputs now, and vice versa
- I/F type ascription can be made by inputting "I" or "F" respectively

**Literals**
- I/F Lit is represented as string now
- IntLit replaced NumLit
- FloatLit can be constructed by using the "." on IntLit
- Temporary backspace/delete cases were added to take care of solo-".", which will be taken care of by InvalidVar hole in the future
- Leading zeros and trailing zeros in I/F Lits are treated as valid inputs

**Arithmetic Operators**
- I/F have separate operators
- Float operators will have F in front of its name (ex. FPlus)
- Float arithmetic operators can be created by adding "." to Int arithmetic operator, and it can be reversed through backspace/delete
- I/F operators have been put into separate modules in DHExp.re


src/hazelcore/Associator.re
```diff
@@ -1,3 +1,4 @@
+// Anytime a new operator is added, the Skel____Lexer.mll and Skel____Parser.mll should be updated too
```
dm0n3y: The parser file extension is `.mly`. Let's also refine this comment so that it's clear what the alternatives are, e.g., "the corresponding `Skel[Typ|Pat|Expr]Lexer.mll` and `Skel[Typ|Pat|Expr].mly` files should be updated too".

src/hazelcore/semantics/dynamics/DHExp.re
```diff
@@ -54,6 +63,7 @@ type t =
   | Ap(t, t)
   | BoolLit(bool)
   | NumLit(int)
+  | FloatLit(float)
```
dm0n3y, line 95: Let's change `NumLit` to `IntLit` here and in `DHPat` too

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -938,14 +958,33 @@ module Exp = {
...
+          | None => DoesNotExpand
+          | Some((op, ty)) =>
+            let d = DHExp.BinNumOp(op, dc1, dc2);
+            Expands(d, ty, delta);
+          };
+        }
+      }
+    | BinOp(NotInHole, FMinus as op, skel1, skel2)
+    | BinOp(NotInHole, FPlus as op, skel1, skel2)
+    | BinOp(NotInHole, FTimes as op, skel1, skel2) =>
```
dm0n3y: You can condense these cases into a single pattern `BinOp(NotInHole, (FMinus | FPlus | FTimes) as op, skel1, skel2)`

src/hazelcore/semantics/dynamics/DHExp.re
```diff
@@ -5,15 +5,21 @@ type bin_num_op =
   | Minus
   | Plus
   | Times
+  | FMinus
+  | FPlus
+  | FTimes
```
dm0n3y: Instead of adding float operators here, I'd rename `bin_num_op` to `bin_int_op` and add a separate `bin_float_op` type, then add a `BinFloatOp` case to `DHExp.t`

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1739,14 +1793,25 @@ module Evaluator = {
     | List(_) => grounded_List
     };
 
-  let eval_bin_num_op = (op: DHExp.bin_num_op, n1: int, n2: int): DHExp.t =>
+  let eval_bin_int_op =
+      (op: DHExp.bin_num_op, n1: int, n2: int): option(DHExp.t) =>
```
dm0n3y: cf comment at top of `DHExp.re`

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1711,6 +1761,9 @@ module Evaluator = {
      5 = bad pattern match
      6 = Cast BV Hole Ground
    */
+  /* TODO
+     need to include the case for float point numbers?
+      */
```
dm0n3y: cf comment at top of `DHExp.re`

src/hazelcore/Examples.re
```diff
@@ -277,6 +277,35 @@ let qsort_example: UHExp.t = {
...
+               ),
+             ],
+           )
+           |> mk_OpSeq_exp,
+         )
+       );
+
+     UHExp.[append_letline, EmptyLine, partition_letline, EmptyLine, qsort_line];
+   };
+   */
```
dm0n3y: remove

src/hazelcore/semantics/UHExp.re
```diff
@@ -58,6 +64,7 @@ and operand =
   | EmptyHole(MetaVar.t)
   | Var(ErrStatus.t, VarErrStatus.t, Var.t)
   | NumLit(ErrStatus.t, int)
+  | FloatLit(ErrStatus.t, float)
```
dm0n3y: Let's also change the name of `NumLit` to `IntLit` here and in `UHPat.re`

src/hazelcore/semantics/Action.re
```diff
@@ -2186,6 +2232,9 @@ module Exp = {
         let new_ze = ZExp.ZBlock.wrap(CursorE(text_cursor, var));
         Succeeded(SynDone((new_ze, Hole, u_gen)));
       };
+    /* | Some(Dot) =>
+       let ze = ZExp.ZBlock.wrap()
+       Succeeded(SynDone((ze, HType.Float, u_gen))) */
```
dm0n3y: remove

src/hazelcore/layout/UHDoc.re
```diff
@@ -182,6 +182,9 @@ let mk_Unit = (~steps: CursorPath.steps, ()): t =>
 let mk_Num = (~steps: CursorPath.steps, ()): t =>
   Delim.mk(~path=(steps, 0), "Num") |> annot_Operand(~sort=Typ);
```
dm0n3y: Change references to `Num` in this module to use `Int` terminology

src/hazelcore/semantics/Action.re
```diff
@@ -2936,6 +2985,26 @@ module Exp = {
...
+    /* Making Float Operators from Int Operators */
+    /* This is for temporary testing. I should find better cursor */
+    /* TODO FIXME */
+    | (Construct(SChar(".")), ZOperator((pos, oper), seq)) =>
+      let new_operator =
+        switch (oper) {
+        | UHExp.Plus => UHExp.FPlus
+        | UHExp.Minus => UHExp.FMinus
+        | UHExp.Times => UHExp.FTimes
+        | _ => oper
```
dm0n3y: Return `Failed` on an operator that can't be converted

src/hazelcore/semantics/Action.re
```diff
@@ -2936,6 +2985,26 @@ module Exp = {
 
     /* Construction */
 
+    /* Making Float Operators from Int Operators */
+    /* This is for temporary testing. I should find better cursor */
+    /* TODO FIXME */
+    | (Construct(SChar(".")), ZOperator((pos, oper), seq)) =>
```
dm0n3y, line 3025: I would restrict this behavior to when the cursor is after the operator in question

src/hazelcore/semantics/Action.re
```diff
@@ -2936,6 +2985,26 @@ module Exp = {
...
+        | UHExp.Plus => UHExp.FPlus
+        | UHExp.Minus => UHExp.FMinus
+        | UHExp.Times => UHExp.FTimes
+        | _ => oper
+        };
+      /* let new_zoperator =
+         switch (new_operator |> ZExp.place_after_operator) {
+         | Some(x) => x
+         | None => zop
+         }; */
```
dm0n3y: Clean up comments

src/hazelcore/semantics/Action.re
dm0n3y: You'll also need to add a similar case to `ana_perform_opseq`. Right now this editing behavior only works in synthetic position, where there's no expected type. You can see this by trying to convert an int operator into a float operator in the defining expression of an annotated let.

dm0n3y: Let's also add the ability to convert a float operator back to an int operator using `Backspace`

dm0n3y (changes requested):
Looks good overall! Requested some minor changes in comments.

There's also an issue where you can only type out floats if what you type matches OCaml's output for `string_of_float`. So you can't type `1.0` left-to-right in Hazel, nor can you can start with `1|0`, type dot, and get `1.0`. We should let the user type whatever they want. I think the full solution here would be to have `UHExp.FloatLit` take a pair of strings or ints (representing left and right sides of decimal point) and convert to float value as needed for evaluation (and maybe add some check in `Action.re` to ensure that the user-entered value can actually be parsed into an OCaml float). cc @cyrus-

Yunif3:
I pulled from the remote branch and fixed all the conflicts.

- Added the ability to convert Float Op to Int Op using Backspace
- Added the ability to remove ".x" float numbers
- Allowed leading zeros in Int and Float by changing IntLit and FloatLit representation as strings

src/hazelcore/Examples.re
dm0n3y: bump

src/hazelcore/semantics/HTyp.re
```diff
@@ -3,21 +3,23 @@
...
   | Bool
   | Arrow(t, t)
   | Prod(t, t)
   | Sum(t, t)
   | List(t);
 
-// Prod(Num, Num)  ==  (Num, Num)
-// (Prod(Num, Num), Bool)  ==  (Num, Num, Bool)
+// Prod(Int, Int)  ==  (Int, Int)
+// (Prod(Int, Int), Bool)  ==  (Int, Int, Bool)
```
dm0n3y: This wasn't you but let's remove

src/hazelcore/semantics/Statics.re
```diff
@@ -828,10 +836,18 @@ module Exp = {
     | BinOp(NotInHole, Minus, skel1, skel2)
     | BinOp(NotInHole, Plus, skel1, skel2)
     | BinOp(NotInHole, Times, skel1, skel2) =>
```
dm0n3y: nit: collapse these into `BinOp(NotInHole, Minus | Plus | Times, skel1, skel2)`

src/hazelcore/semantics/Statics.re
```diff
@@ -828,10 +836,18 @@ module Exp = {
...
-      switch (ana_skel(ctx, skel1, seq, HTyp.Num)) {
+      switch (ana_skel(ctx, skel1, seq, HTyp.Int)) {
       | None => None
       | Some(_) =>
-        ana_skel(ctx, skel2, seq, Num) |> OptUtil.map(_ => HTyp.Num)
+        ana_skel(ctx, skel2, seq, Int) |> OptUtil.map(_ => HTyp.Int)
+      }
+    | BinOp(NotInHole, FMinus, skel1, skel2)
+    | BinOp(NotInHole, FPlus, skel1, skel2)
+    | BinOp(NotInHole, FTimes, skel1, skel2) =>
```
dm0n3y: nit: collapse

src/hazelcore/semantics/ZExp.re
```diff
@@ -80,7 +80,9 @@ let valid_cursors_operand: UHExp.operand => list(CursorPosition.t) =
   | ListNil(_) => CursorPosition.delim_cursors(1)
   /* outer nodes - text */
   | Var(_, _, x) => CursorPosition.text_cursors(Var.length(x))
-  | NumLit(_, n) => CursorPosition.text_cursors(IntUtil.num_digits(n))
+  | IntLit(_, n) => CursorPosition.text_cursors(IntUtil.num_digits_str(n))
+  | FloatLit(_, f) =>
+    CursorPosition.text_cursors(FloatUtil.num_digits_str(f))
```
dm0n3y: Let's just use `String.length` here and remove these utility functions.

src/hazelcore/semantics/dynamics/DHExp.re
```diff
@@ -1,31 +1,58 @@
 open Sexplib.Std;
 
 [@deriving sexp]
-type bin_num_op =
+type bin_int_op =
```
dm0n3y: Sorry to come back to this but let's package up the types `bin_int_op` and `bin_float_op` in modules, e.g., create modules `BinIntOp` and `BinFloatOp`, rename the types as `t`, move `of_int_op`/`of_float_op` inside respective modules and rename as `of_op`, etc. We started this division on @annie-anna's PR too for strings and string ops.

src/hazelcore/semantics/dynamics/DHExp.re
```diff
@@ -53,8 +87,10 @@ type t =
   | Lam(DHPat.t, HTyp.t, t)
   | Ap(t, t)
   | BoolLit(bool)
-  | NumLit(int)
-  | BinNumOp(bin_num_op, t, t)
+  | IntLit(string)
+  | FloatLit(string)
```
dm0n3y: These constructor args should be `int` and `float`. The `DH*` types are the internal expressions used for evaluation. Don't need to maintain the user-edited string.

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1739,15 +1804,28 @@ module Evaluator = {
...
     };
 
-  let eval_bin_num_op = (op: DHExp.bin_num_op, n1: int, n2: int): DHExp.t =>
+  let eval_bin_int_op =
+      (op: DHExp.bin_int_op, s1: string, s2: string): option(DHExp.t) => {
+    let (n1, n2) = (int_of_string(s1), int_of_string(s2));
+    switch (op) {
+    | Minus => Some(IntLit(string_of_int(n1 - n2)))
+    | Plus => Some(IntLit(string_of_int(n1 + n2)))
+    | Times => Some(IntLit(string_of_int(n1 * n2)))
```
dm0n3y: cf comment on constructor args in `DHExp` and `DHPat`. By making that change, you'll need to push the `int_of_string`/`float_of_string` conversions into `expand` functions rather than calling them here during evaluation. There, you should use `int_of_string_opt`/`int_of_float_opt` and return `DoesNotExpand` if the conversions fail.

src/hazelcore/semantics/Action.re
```diff
@@ -6,6 +6,9 @@ type operator_shape =
   | SMinus
   | SPlus
   | STimes
+  | SFMinus
+  | SFPlus
+  | SFTimes
```
dm0n3y: Do we actually need these? Maybe we'd eventually want to directly construct via keyboard command, but even that seems like overkill to me. cc @cyrus-

src/hazelcore/semantics/Action.re
```diff
@@ -3154,19 +3242,29 @@ module Exp = {
...
     | (Backspace, CursorE(OnText(j), Var(_, _, x))) =>
       syn_backspace_text(ctx, u_gen, j, x)
-    | (Backspace, CursorE(OnText(j), NumLit(_, n))) =>
-      syn_backspace_text(ctx, u_gen, j, string_of_int(n))
+    | (Backspace, CursorE(OnText(j), IntLit(_, n))) =>
+      syn_backspace_text(ctx, u_gen, j, n)
+    /* For Float: (.x<|) ==> ( |_ )  */
+    | (Backspace, CursorE(OnText(2), FloatLit(_, f)))
+        when
+          FloatUtil.less_than_one_str(f) && FloatUtil.num_digits_str(f) == 2 =>
```
dm0n3y: This needs to be a little more general. For example, right now we don't handle backspacing on `1|.` or deleting on `.|1`.

src/hazelcore/util/FloatUtil.re
```diff
@@ -0,0 +1,36 @@
...
+  if (dot_index + 1 < String.length(f)) {
+    let decimal_str = String.sub(f, dot_index, String.length(f) - dot_index);
+    /* 1 is added to length of trimmed float because Ocaml only allows for float with 0.xx, so an extra char is introduced */
+    String.length(decimal_str)
+    - String.length(string_of_float(float_of_string(decimal_str)))
+    + 1;
+  } else {
+    0;
+  };
+};
```
dm0n3y: Let's move `leading_zeros` and `trailing_zeros` into `StringUtil` and simplify implementation to be `int`/`float`-agnostic, e.g., recursively increment some char index down/up the string until you encounter the first nonzero.

src/hazelcore/util/IntUtil.re
```diff
@@ -1 +1,9 @@
 let num_digits = (n: int): int => String.length(string_of_int(n));
+let num_digits_str = (n: string): int => String.length(n);
+let leading_zeros = (n: string): int => {
```
dm0n3y: cf comment in `FloatUtil`

src/hazelcore/util/FloatUtil.re
dm0n3y: Also rename to `num_leading_zeros`/`num_trailing_zeros`

dm0n3y (changes requested):
Nice job, looks and feels good. A couple gaps w.r.t. deleting floats + some minor refactor requests.

src/hazelcore/semantics/Action.re
Yunif3: means delete the whole thing when there is only a dot left

Yunif3:
Fixed the pointed brought up by the 2nd review.

- Removed unnecessary comments and simplified the code with better formatting/modularity
- Removed unnecessary Op shapes from Action.re
- Allowed the ".x" to become an empty hole when there is an action that leaves only "."
- moved the zero helper functions to StringUtil
- fixed the IntLit/FloatLit implementation in DH* files

src/hazelcore/util/FloatUtil.re
```diff
@@ -0,0 +1,8 @@
+/* returns the number of digits in a float */
+let num_digits = (f: float): int => {
+  String.length(string_of_float(f));
+};
+
+let less_than_one_str = (f: string): bool => {
```
dm0n3y: unused, remove

src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -1031,7 +1084,9 @@ module Exp = {
         | Keyword(k) => DHExp.Keyword(u, 0, sigma, k)
         };
       Expands(d, Hole, delta);
-    | NumLit(NotInHole, n) => Expands(NumLit(n), Num, delta)
+    | IntLit(NotInHole, n) => Expands(IntLit(int_of_string(n)), Int, delta)
+    | FloatLit(NotInHole, f) =>
+      Expands(FloatLit(float_of_string(f)), Float, delta)
```
dm0n3y: Gotta do the `opt` version here too

src/hazelcore/semantics/Action.re
```diff
@@ -2152,12 +2183,18 @@ module Exp = {
       if (text |> StringUtil.is_empty) {
         let (zhole, u_gen) = u_gen |> ZExp.new_EmptyHole;
         Succeeded(SynDone((ZExp.ZBlock.wrap(zhole), HTyp.Hole, u_gen)));
+      } else if (text == ".") {
+        let (zhole, u_gen) = u_gen |> ZExp.new_EmptyHole;
+        Succeeded(SynDone((ZExp.ZBlock.wrap(zhole), HTyp.Hole, u_gen)));
```
dm0n3y, line 2206: Add a TODO comment here indicating that this behavior should change once we have `InvalidVar` holes.

dm0n3y: Also you need to add this case to `Pat.mk_[syn|ana]_text` as well

src/hazelcore/semantics/Action.re
```diff
@@ -2204,6 +2241,9 @@ module Exp = {
       if (text |> StringUtil.is_empty) {
         let (zhole, u_gen) = u_gen |> ZExp.new_EmptyHole;
         Succeeded(AnaDone((ZExp.ZBlock.wrap(zhole), u_gen)));
+      } else if (text == ".") {
+        let (zhole, u_gen) = u_gen |> ZExp.new_EmptyHole;
```
dm0n3y, line 2265: Add TODO here as well (and in `Pat` module)

src/hazelcore/semantics/Action.re
```diff
@@ -3273,12 +3363,17 @@ module Exp = {
       Succeeded(SynDone((new_ze, ty, u_gen)));
     | (Construct(SAsc), CursorE(_)) => Failed
 
+    /* It is a temporary fix so that a new hole isn't created when a dot is inputted by itself.
+       It will later be replaced by invalid var and handled by make syntax */
```
dm0n3y: Temporary fix so that a new hole isn't created when a dot is inputted by itself (cf mk_[syn|ana]_text)
TODO remove once we have InvalidVar holes

dm0n3y: Think you'll also need to add this case to `Pat.[syn|ana]_perform_operand`

dm0n3y (changes requested):

src/hazelcore/util/FloatUtil.re
Yunif3: do you mean to remove both functions or just less_than_one_str?

dm0n3y: whatever is unused

dm0n3y (approved):

PR #223, opened 2020-05-01
strings

### LINKED ISSUE #189: Strings

dm0n3y:


cyrus-:
just played with your branch briefly:
* hotkeys don't work correctly when inside a string, e.g. "Tab" doesn't take you to next hole and Ctrl+R to refresh the page doesn't work. you should only be capturing the keys that can be inserted into a string, not all keys.
* multiple spaces in a row do not display correctly -- you should use the "white-space: pre" CSS property to make sure nothing like that happens

cyrus-:
Also, when the line gets too long, it breaks. This might relate to a more general problem with long lines + pretty printing @dmoon1221 @adamsmd ?

cyrus-:
Also in the result, strings don't have quotation marks around them.

adamsmd:
@cyrus- @dmoon1221: I've added support for long lines to `dev`.  Merge `dev` into this branch, and let me know if you still see the issue.

Merge into `dev` to support long lines: 05a309106d8ea21c182d8f3f3ffe8f31547a24f5

Commit that actually implements long lines: 5b46637729861aedb18e4f72da8b8feaf56c3b67

cyrus-:
@annie-anna please merge the latest dev in and see if it fixes the issues with long string literals

annie-anna:
Yes, it works now.

cyrus-:
@adamsmd @annie-anna @dmoon1221 it should overflow with a scrollbar in the cell, not like this
Screenshot from 2020-05-11 23-07-52

adamsmd:
@annie-anna: FYI, Cyrus told me that the way to fix that overflow by setting the CSS `overflow` behavior to `scroll`.  (I am not experienced in these things, so I am just passing this along.)

annie-anna:
Okay, I have fixed that.

src/hazelcore/Operators.re
```diff
@@ -87,7 +88,8 @@ module Exp = {
     | Comma => ","
     | Cons => "::"
     | And => "&&"
-    | Or => "||";
+    | Or => "||"
+    | PlusPlus => "^";
```
cyrus-: Call this `Caret` rather than `PlusPlus`

src/hazelcore/SkelExprLexer.mll
```diff
@@ -17,6 +17,7 @@ rule read =
   | "-." { FMINUS }
   | "*." { FTIMES }
   | "/." { FDIVIDE }
+  | "^" { PLUSPLUS }
```
cyrus-: `CARET`

src/hazelcore/SkelExprParser.mly
```diff
@@ -46,6 +47,7 @@
 %left FMINUS
 %left FTIMES
 %left FDIVIDE
+%left PLUSPLUS
```
cyrus-: let's follow OCaml: https://caml.inria.fr/pub/docs/manual-ocaml/expr.html#ss:precedence-and-associativity
`^` is right associative and binds less tightly than `::`

cyrus-: actually can you check all of our precedences here to make sure they are in the same order as OCaml

src/hazelcore/layout/DHDoc.re
```diff
@@ -7,11 +7,13 @@ type formattable_child = (~enforce_inline: bool) => t;
 
 let precedence_const = 0;
 let precedence_Ap = 1;
+let precedence_Subscript = 1;
 let precedence_Times = 2;
 let precedence_Divide = 2;
 let precedence_Plus = 3;
 let precedence_Minus = 3;
 let precedence_Cons = 4;
+let precedence_PlusPlus = 4;
```
cyrus-: see above regarding precedence -- review all of these to make sure they are consistent with OCaml

src/hazelcore/layout/DHDoc.re
```diff
@@ -229,9 +248,11 @@ module Exp = {
     | Let(_)
     | FixF(_)
     | ConsistentCase(_)
-    | InconsistentBranches(_) => precedence_max /* TODO: is this right */
+    | InconsistentBranches(_) => precedence_max
+    | Subscript(_) => precedence_Subscript /* TODO: is this right */
```
cyrus-: see above -- remove comment once this is done

src/hazelcore/layout/DHDoc.re
```diff
@@ -439,6 +482,24 @@ module Exp = {
           failwith("unexpected FailedCast without inner cast")
         | InvalidOperation(operation) =>
           switch (operation) {
+          | FailedSubscript(Subscript(d1, d2, d3)) =>
```
cyrus-: `InvalidOperation` should take an operation and an error message or error class, so it will replace `FailedSubscript`, not wrap around it.

cyrus- (changes requested):
partial review -- will continue later but some action items in there now. you may want to confer with @Yunif3 about InvalidOperation, I think it's not quite right yet.

src/hazelcore/layout/DHDoc.re
annie-anna: Yunsoo and I have updated `InvalidOperation`. I generalize the runtime error decoration as `InvalidOpDecoration`. Other bugs are fixed.

cyrus-:
@annie-anna we should support newlines in string literals. can you add support for that? we can just allow the enter key to work, we don't have to use `\n` like in other languages.

for other special characters, though, we might need to support escape sequences. for example, `\t` can be used to insert the tab character. if we tried to support that without an escape sequence, then it would disable our ability to use the Tab key to move to the next hole...

cyrus-:
also we should support all characters, currently some don't work like `:` and others (probably ones that have bindings?)

annie-anna:
> also we should support all characters, currently some don't work like `:` and others (probably ones that have bindings?)

Do you mean typing characters inside quotation marks? I have added all characters on keyboard to `Cell.view`

cyrus-:
oh sorry don't know why I thought `:` didn't work. is this approach going to support non-US keyboards and other ways that people insert characters, e.g. on screen keyboards?

annie-anna:
> @annie-anna we should support newlines in string literals. can you add support for that? we can just allow the enter key to work, we don't have to use `\n` like in other languages.
> 
> for other special characters, though, we might need to support escape sequences. for example, `\t` can be used to insert the tab character. if we tried to support that without an escape sequence, then it would disable our ability to use the Tab key to move to the next hole...

Currently I have some problem with cursor. If I enter "Enter", `Cell.view` will call `Action.Construct(SLine)`. However, since string is an operand, I deal with `Construct(SLine)` in `syn_perform_operand`, and I cannot return `CursorL`. It looks like this in ui

cyrus-:
we should make Enter emit an SChar("\n") when the cursor is in a string literal

annie-anna:
> screen keyboards

I think it just support characters I listed in `Cell.view`

annie-anna:
When we insert single slash or double slash, should the resulted string behaves like in python--both show single slash or should it just show what is typed? I suppose we can directly type `'`  to represent `'` , but can we directly type `"` ? Or we must use `\"` ?

annie-anna:
And just want to clarify again, when we type `Enter` , in the block where we type expression, the string should be split into 2 lines rather than be inserted the escaped sequence `"\n"` ?

cyrus-:
- We can use `"` in strings, since we aren't actually parsing anything.
- Pressing enter at `"abc|def"` should cause the string to be visually split into two lines, but internally that should be represented as `StringLit("abc\ndef")`. It should also be possible to put in a `\n` manually, in case you don't want to see a new line visually (e.g. when doing `"some message" ^ x ^ "\n"`). That should internally be represented as `StringLit("abc\\ndef")`, so you have the escape sequence in there.

annie-anna:
So, for split string, is the `UHExp.block` structure a list of 1 `UHExp.line` or 2 `UHExp.line`? Since you said previously it only contains a StringLit operand, I think it still has 1 `UHExp.line`? However, in this way, the cursor will always be on the first line as shown on the ui.

cyrus-:
A string literal is an operand, so I don't think this has anything to do with lines. Pressing enter should just cause a newline to be inserted into the string lit. It will be displayed on multiple lines, but that logic is strictly in the doc generation.

annie-anna:
I have enabled escaped sequence `\t`, `\r`,  and `\b`.

cyrus-:
I believe we discussed that a string literal like "abc\a" should have the red letters in the source code, but I don't think it should have the letters in red in the output (DHExp). the output just displays characters, no escape sequences anymore.

for the same reason, we don't need to say that this is unterminated, since the slash there is really a slash character and the quotation mark is just a visual thing. you may want to show a trailing slash as a red character in the original program though.
Screenshot from 2020-06-08 02-41-47

cyrus-:
ultimately I think we want to be able to interact with a string literal in the output and decide whether it should display special characters directly or as escape sequences, for the purposes of copy-and-paste, but that can wait until we have a more sophisticated DHExp interaction system + copy-and-paste support

cyrus-:
also now that we have the action panel enabled again, can you add the string related items to it in a new section? you can ask @wlitwin for help but I think the code is fairly straightforward.

cyrus- (changes requested):
* in the undo history panel, can you include quotation marks around the string literal. also set a maximum length that is displayed there and then use a ... in gray to elide the rest, so we don't get giant undo history entries
* the undo history panel should group "insert new string" with any subsequent character insertions, rather than showing them as two separate entries. reach out to @PPER if you're unsure of how to do this.
* the cursor inspector should display an error message when the cursor is on or adjacent to an invalid escape sequence. look at CursorInfo and CursorInspector
* can we use a dotted box around invalid characters, similar to type errors? the red text on green background is hard to distinguish, and especially so for people who may have red/green color-blindness.

annie-anna:
If we have a string of tailing backslash, should "got invalid escape sequence" be shown when cursor is on the slash?

I remember that if there is "got invalid escape sequence", the block will be red. Would it be better if it is shown yellow, indicating it is a warning (as in ocaml) rather than error and differentiating it from type error?

When the string is "\o777", I think ocaml views it as a valid escape sequence, but the number is out of range. So, this string will not cause "got invalid escape sequence"?

annie-anna:
If I have ``"abc\v|\n"`` with cursor position as indicated, should the cursor follow ``\v`` and show "got invalid escape sequence" or follow ``\n`` and show nothing?

cyrus-:
doesn't compile right now -- I think you have to modify the .rei files now that they are included

annie-anna:
I have made modifications to ``*.rei``, and it can compile now.

cyrus-:
- when the cursor is on an invalid escape sequence, let's make the sidebar red instead of green
- the cursor inspector should be left biased, so in the example in your comment above it should show an error (so that you see the error as soon as you type the escape sequence, going left to right)

cyrus-:
the subscript operator should treat valid escape sequences as single characters, so this is incorrect
Screenshot from 2020-06-15 00-41-46

cyrus-:
the "edit String literal" entry in undo history needs quotation marks around the resulting string

cyrus-:
also it should show \n for all new lines in the undo history, even ones that have been entered with `Enter` rather than `\n`

annie-anna:
I have fixed these issues.

cyrus-:
- more conflicts to resolve
- when the cursor is on an invalid escape sequence, lets have the sidebar title be "GOT" and then in the body of the got indicator, just write "invalid escape sequence \a" (or whatever the invalid escape sequence is). We don't need to say "got type String" in that situation
- for subscript out of bounds errors, the error message should say whether it was the starting or ending subscript that was out of bounds.

cyrus-:
also the "insert new list / substring" entry in the history panel is not precise -- it should be possible to determine whether it is a list or substring being constructed. you may want to ask @PPER about how to best approach that.

annie-anna:
If there is type inconsistency, like ``_+"ab\|vcd"``, when the cursor is on the indicated position, should we show GOT "invalid escape sequence \v" only and ignore "got type consistency"?

cyrus-:
yes that's fine -- the cursor inspector shows the innermost error related to the cursor

cyrus-:
it might be more clear why that is happening if the erroneous escape sequence has a highlighted red background when the cursor is on it (without eliminating the green background on the string literal as a whole). let's do that.

annie-anna:
> it might be more clear why that is happening if the erroneous escape sequence has a highlighted red background when the cursor is on it (without eliminating the green background on the string literal as a whole). let's do that.

Is the corresponding code in ``UHCode.view``? If so, how to get the cursor position in ``UHCode.re``? And since the color of cursor is already red, would the red background be mixed up with the cursor?

cyrus-:
Screenshot from 2020-06-20 21-17-44

the font size for the invalid escape sequence message is different from the "any type" message -- make sure you are using the same infobar code for both. also can you make the "\a" show in code font, since it is a code construct and not part of the natural language part of the message?

cyrus-:
Let's make the text of string literals + the quotation mark show up in a different color (no strong preference about what it is, but something dark that shows up fine when the cursor is on it)

cyrus-:
subscripting silently returning "" when the start is greater than the end is odd. let's make that an error.

annie-anna:
> Let's make the text of string literals + the quotation mark show up in a different color

Should all delim shapes have different color from text?

cyrus-:
no, just change the color of string literals for this PR

cyrus-:
I'm only seeing color on the delimiters. the string contents should be the same color as the quotation marks

annie-anna:
> I'm only seeing color on the delimiters. the string contents should be the same color as the quotation marks

For invalid escape sequence, which has red brackets around it, should it be brown also?

cyrus-:
no we can use red text for that

annie-anna:
Do I need to keep the red brackets? Would the red text cause problems for people who have color blind?

cyrus-:
keep the red border. that way even if someone can't distinguish red/brown, they can see the error

cyrus-:
Screenshot from 2020-06-23 14-57-12

missing space before the \e

cyrus-:
Screenshot from 2020-06-23 15-04-01
I can't press backspace here... it should move the cursor over to the right of `f`, the way it does for other operands in that position

cyrus-:
also can't press backspace here
Screenshot from 2020-06-23 15-05-58

cyrus-:
when I press backspace here:
Screenshot from 2020-06-29 14-55-25

I get placed here:
Screenshot from 2020-06-29 14-55-12

The correct result should be that the `+` operator becomes a space and the cursor is on the left side of the subscript.

It seems like you might be special casing backspace/delete when an operand is a subscript -- I don't think you want to do this. You should modify the existing code to work when your cursor is before an operand, even one that is not bidelimited like subscript.

cyrus-:
also can we get the string literal coloring to work in the result and in the undo history too?

annie-anna:
> also can we get the string literal coloring to work in the result and in the undo history too?

Should the escape sequence be marked as different color in these two places?

cyrus-:
yes, it should look the same in both places -- ideally you'd factor out the display code so it isn't duplicated.

annie-anna:
> yes, it should look the same in both places -- ideally you'd factor out the display code so it isn't duplicated.

So, for example, the invalid escape sequence should be marked red and has red border around it in the result and undo history? since you said previously that it should be displayed as normal text in the result.

cyrus-:
ah you're right that we need to treat characters in an invalid escape sequence as just normal characters in the result (but not in the undo history)

cyrus-:
Screenshot from 2020-06-30 17-35-49

when I press space at `x|[_:_]` it puts the space in the target expression as shown above instead of it becoming part of the parent op seq. doesn't seem to happen for other operators.

cyrus-:
also make sure you catch up with dev soon, lots of conflicts atm

annie-anna:
> Screenshot from 2020-06-30 17-35-49
> 
> when I press space at `x|[_:_]` it puts the space in the target expression as shown above instead of it becoming part of the parent op seq. doesn't seem to happen for other operators.

So, if I have `string_of_int|[1:2]` and I press `space`, I should have `(string_of_int _)[1:2]`?

cyrus-:
No it should be `string_of_int _[1:2]` where the subscript is acting only on the `_` (the green underline in the screenshot shouldn't extend out to the variable)

src/hazelcore/UHExp.re
```diff
@@ -84,6 +89,10 @@ let case =
     : operand =>
   Case(err, scrut, rules);
 
+let subscript =
+    (~err: ErrStatus.t=NotInHole, body1: t, body2: t, body3: t): operand =>
```
cyrus-: instead of body1/2/3, we should give better names to each position: target, start, end

src/hazelcore/UHExp.re
```diff
@@ -194,12 +224,14 @@ and get_err_status_operand =
...
+  | StringLit(err, _)
   | ListNil(err)
   | Lam(err, _, _, _)
   | Inj(err, _, _)
   | Case(StandardErrStatus(err), _, _)
-  | ApPalette(err, _, _, _) => err
-  | Case(InconsistentBranches(_), _, _) => NotInHole
+  | ApPalette(err, _, _, _)
+  | Subscript(err, _, _, _) => err
+  | Case(InconsistentBranches(_), _, _) => NotInHole /* TODO: What to do here...? */
```
cyrus-: remove TODO (it is gone from dev)

src/hazelcore/UHExp.re
```diff
@@ -219,12 +251,14 @@ and set_err_status_operand = (err, operand) =>
...
   | FloatLit(_, f) => FloatLit(err, f)
   | BoolLit(_, b) => BoolLit(err, b)
+  | StringLit(_, s) => StringLit(err, s)
   | ListNil(_) => ListNil(err)
   | Lam(_, p, ann, def) => Lam(err, p, ann, def)
   | Inj(_, inj_side, body) => Inj(err, inj_side, body)
   | Case(_, scrut, rules) => Case(StandardErrStatus(err), scrut, rules)
   | ApPalette(_, name, model, si) => ApPalette(err, name, model, si)
   | Parenthesized(body) => Parenthesized(body |> set_err_status(err))
+  | Subscript(_, body1, body2, body3) => Subscript(err, body1, body2, body3)
```
cyrus-: use more descriptive variable names (see above)

cyrus-: same with other places in the code (won't mark each one)

src/hazelcore/UndoHistoryCore.re
```diff
@@ -37,6 +37,7 @@ type action_group =
   /* SLine in Action_common.shape stands for both empty line and case rule,
      so an extra type CaseRule is added for construction */
   | CaseRule
+  | Subscript
```
cyrus-, line 40: why do you need to special case subscript here?

src/hazelcore/UndoHistoryCore.re
```diff
@@ -37,6 +37,7 @@ type action_group =
   /* SLine in Action_common.shape stands for both empty line and case rule,
```
cyrus-, line 37: pressing "[" creates an "insert new list" history entry sometimes

src/hazelcore/ZExp.re
```diff
@@ -105,7 +166,10 @@ let valid_cursors_operand: UHExp.operand => list(CursorPosition.t) =
   | Inj(_) => CursorPosition.delim_cursors(2)
   | Case(_) => CursorPosition.delim_cursors(2)
   | Parenthesized(_) => CursorPosition.delim_cursors(2)
-  | ApPalette(_) => CursorPosition.delim_cursors(1); /* TODO[livelits] */
+  | ApPalette(_) => CursorPosition.delim_cursors(1)
+  | Subscript(_) => {
+      CursorPosition.delim_cursors(3);
+    }; /* TODO[livelits] */
```
cyrus-: comment in wrong position

src/hazelcore/ZExp.re
```diff
@@ -162,19 +226,24 @@ and is_before_zoperand =
...
   | CursorE(cursor, BoolLit(_)) => cursor == OnText(0)
+  | CursorE(cursor, StringLit(_))
   | CursorE(cursor, Lam(_))
   | CursorE(cursor, Inj(_))
   | CursorE(cursor, Case(_))
   | CursorE(cursor, Parenthesized(_)) => cursor == OnDelim(0, Before)
-  | CursorE(cursor, ApPalette(_)) => cursor == OnDelim(0, Before) /* TODO[livelits] */
+  | CursorE(cursor, ApPalette(_)) => cursor == OnDelim(0, Before)
+  | CursorE(_, Subscript(_)) => false
+  | SubscriptZE1(_, e, _, _) => is_before(e) /* TODO[livelits] */
```
cyrus-: comment in wrong place

src/hazelcore/ZExp.re
```diff
@@ -414,14 +498,22 @@ and erase_zoperand =
   | ParenthesizedZ(zbody) => Parenthesized(erase(zbody))
   | LamZP(err, zp, ann, body) => Lam(err, ZPat.erase(zp), ann, body)
   | LamZA(err, p, zann, body) => Lam(err, p, Some(ZTyp.erase(zann)), body)
-  | LamZE(err, p, ann, zbody) => Lam(err, p, ann, erase(zbody))
+  | LamZE(err, p, ann, zbody) => {
```
cyrus-: shouldn't need to reformat this in this way

src/hazelcore/ZExp.re
```diff
@@ -512,7 +610,10 @@ and mk_inconsistent_zoperand = (u_gen, zoperand) =>
   | InjZ(InHole(TypeInconsistent, _), _, _)
   | CaseZE(StandardErrStatus(InHole(TypeInconsistent, _)), _, _)
   | CaseZR(StandardErrStatus(InHole(TypeInconsistent, _)), _, _)
-  | ApPaletteZ(InHole(TypeInconsistent, _), _, _, _) => (zoperand, u_gen)
+  | ApPaletteZ(InHole(TypeInconsistent, _), _, _, _)
```
cyrus-: can you go back and make ApPalette the last case in all the pattern sequences involving it, instead of putting Subscript at the end? that will make future merges easier.

src/hazelcore/ZExp.re
```diff
@@ -876,31 +1027,43 @@ and move_cursor_right_zoperator =
...
-    // k == 0 || k == 1 || k == 2
-    switch (k == 0, k == 1, ann) {
-    | (true, _, _) => Some(LamZP(err, ZPat.place_before(arg), ann, body))
-    | (_, true, None) =>
-      // invalid cursor position
-      None
-    | (_, true, Some(ann)) =>
-      Some(LamZA(err, arg, ZTyp.place_before(ann), body))
-    | (false, false, _) => Some(LamZE(err, arg, ann, place_before(body)))
+  | CursorE(OnDelim(k, After), Lam(err, arg, ann, body)) => {
```
cyrus-: don't need this diff -- better to avoid making formatting changes not relevant to your feature

cyrus- (changes requested):
reviewed UHExp/UHPat/UHTyp/HTyp/ZExp/ZPat/ZTyp and a few other modules, but didn't have time to get to other modules yet. go ahead and work on these changes for now.

src/hazelcore/UndoHistoryCore.re
annie-anna: so that I can differentiate action_group when I construct `SLeftBracket` and show different keywords for subscript and list.

nickcollins:
Hi Mengjing, sorry if I disrupted your work earlier when i merged this - I blindly assumed this was David's PR, and I acted hastily, trying to get this in quickly to help make a paper deadline. Hope it didn't cause too much trouble

annie-anna:
That's alright. No need to worry about it.

PR #224, opened 2020-05-01
resetting hole numbering when the expression is complete

### LINKED ISSUE #50: Restart hole numbering whenever the program becomes complete

utkmehta:
In Hazel right now, hole numbers increase monotonically but the numbering is never reset, even when you pass through edit states where there are no holes. This pull request checks if expressions are complete and also renumbers holes. 

The completeness of expression is checked (according to POPL 17 paper) by making sure there are no holes (empty/non-empty/type). For this, I have written is_complete functions in UHPat, UHExp, UHTyp, and OpSeq files. For the purpose of renumbering holes, however, type holes are allowed in expressions. So I have passed a check_type_holes flag to the is_complete functions. This flag is set to false when renumbering holes. Every time there is an edit action in the program, the expression is checked. If there are no empty/non-empty holes (note that type holes are allowed since the check_type_holes flag is set to false), the hole number is reset to 1 (by setting u_gen to MetaVarGen.init). Only empty holes are renumbered.

Testing: Created and deleted holes for different types of expressions and made sure the renumbering of holes happened when required.


cyrus-:
Please make some edits to the pull request summary.

> It is more intuitive to programmers if the numbers associated with holes have some sort of uniformity and structure.  It should help them keep track of the information regarding the holes in their code. Currently, there is some structure to increment the hole numbers as new holes are created. 

This is vague and unclear. Instead, get right to the point: in Hazel right now, hole numbers increase monotonically and numbering is never reset, even when you pass through edit states where there are no holes. 

> This pull request builds upon that and starts renumbering the holes from 0 once expressions are complete.

Be terse and avoid unnecessary phrases like "builds upon that" in technical writing.

> To achieve this, every time there is an edit action in the program, the expression is checked. If it is complete, the hole number turns to 0 (by replacing the u_gen to MetaVarGen.init).

"turns" -> "resets" and "replacing the u_gen to" -> "setting u_gen to"

> To check if the expression is complete, there are is_complete functions in UHPat, UEXp, and OPSeq files. Each of the cases in the pattern matching for expressions, patterns, and Opseqs are checked and if required, sub-expressions and sub-types are checked recursively.

* Writing "there are" makes it unclear if there already were these functions or if you added them.
* UHEXp and OPSeq are not spelled correctly.
* You don't mention UHTyp.
* State the criteria by which an expression/pattern/opseq/type are complete. Does it count only empty holes? What about type holes, which are not currently numbered? What about non-empty holes, i.e. the metavars in ErrStatus.t?

src/hazelweb/model/Program.re
```diff
@@ -101,7 +101,15 @@ let perform_edit_action = (a, program) => {
   switch (Action.Exp.syn_perform(Contexts.empty, a, edit_state)) {
   | Failed => raise(FailedAction)
   | CursorEscaped(_) => raise(CursorEscaped)
-  | Succeeded(new_edit_state) => program |> put_edit_state(new_edit_state)
+  | Succeeded(new_edit_state) => 
+    let (ze, _type, u_gen) = new_edit_state;
```
cyrus-: use ty instead of _type

src/hazelcore/semantics/UHPat.re
```diff
@@ -167,3 +167,32 @@ let text_operand =
       u_gen,
     );
   };
+
+let rec is_complete = (pn: t): bool => {
```
cyrus-: why is this variable "pn" rather than "p"?

src/hazelcore/semantics/UHPat.re
```diff
@@ -167,3 +167,32 @@ let text_operand =
...
+  switch (pn) {
+  | OpSeq(Placeholder(n) as _skel, seq) =>
+    is_complete_operand(seq |> Seq.nth_operand(n))
+  | OpSeq(BinOp(InHole(_), _, _, _), _) => false
+  | OpSeq(BinOp(NotInHole, _, skel1, skel2), seq) =>
+    is_complete(OpSeq(skel1, seq)) && is_complete(OpSeq(skel2, seq))
```
cyrus-: you should define an auxiliary function `is_complete_skel` that takes a skel and a seq, rather than constructing intermediate OpSeqs (which are invalid as standalone OpSeqs)

src/hazelcore/semantics/OpSeq.re
```diff
@@ -70,3 +70,19 @@ let make_inconsistent =
...
+          is_complete_operand: 'operand => bool,
+          opseq: t('operand, 'operator),
+        )
+        : bool => {
+  switch (opseq) {
+  | OpSeq(Placeholder(n) as _skel, seq) =>
+    is_complete_operand(seq |> Seq.nth_operand(n))
+  | OpSeq(BinOp(InHole(_), _, _, _), _) => false
+  | OpSeq(BinOp(NotInHole, _, skel1, skel2), seq) =>
+    is_complete(is_complete_operand, OpSeq(skel1, seq))
```
cyrus-: you should define an auxiliary function is_complete_skel that takes a skel and a seq, rather than constructing intermediate OpSeqs (which are invalid as standalone OpSeqs and cause extra allocation)

src/hazelcore/semantics/UHExp.re
```diff
@@ -319,3 +319,41 @@ let text_operand =
...
+  | Lam(InHole(_), _, _, _) => false
+  | Lam(NotInHole, pat, _, body) =>
+    UHPat.is_complete(pat) && is_complete(body)
+  | Inj(InHole(_), _, _) => false
+  | Inj(NotInHole, _, body) => is_complete(body)
+  | Case(InHole(_), _, _, _) => false
+  | Case(NotInHole, body, _, _) => is_complete(body)
+  | Parenthesized(body) => is_complete(body)
+  | ApPalette(InHole(_), _, _, _) => false
+  | ApPalette(NotInHole, _, _, _) => true
```
cyrus-: raise Unimplemented here rather than returning true (will have to fix this once livelits is merged)

cyrus- (changes requested):

src/hazelweb/model/Program.re
utkmehta: fixed

src/hazelcore/semantics/UHPat.re
utkmehta: fixed

src/hazelcore/semantics/UHExp.re
utkmehta: Unimplemented for all four cases?

src/hazelcore/semantics/OpSeq.re
utkmehta: Why are standalone Opseqs invalid?

src/hazelcore/semantics/UHExp.re
cyrus-: no for the line that the comment is on (354) -- GitHub just shows additional lines for context

src/hazelcore/semantics/OpSeq.re
cyrus-: oh in this case its the same seq so I guess its valid. okay, this is fine though it involves more allocations than doing it separately...

src/hazelcore/semantics/UHExp.re
utkmehta: fixed

src/hazelcore/semantics/UHPat.re
utkmehta: fixed

src/hazelcore/semantics/OpSeq.re
utkmehta: Got it. I created a separate is_complete_skel function so that there are fewer allocations

cyrus-:
Pull request message says it resets to 0 in one place and 1 in the other...

cyrus-:
"non-empty holes are not numbered" -> yes they are, it is just not shown in the UI right now. your code looks like it is counting non-empty holes in the completeness check. can you correct the pull request message and double check that you are including non-empty holes in the completeness check correctly?

cyrus-:
"UHEXp, and OPSeq" both still miscapitalized

cyrus-:
"The criterion for checking completeness is making sure there are no holes in the expression (empty/type/any other non-empty holes)." -- this is not clear. it sounds like you are counting type holes for completeness, but I don't think you are?

src/hazelcore/semantics/OpSeq.re
```diff
@@ -70,3 +70,26 @@ let make_inconsistent =
     let set_skel = Skel.BinOp(InHole(TypeInconsistent, u), op, skel1, skel2);
     (OpSeq(set_skel, seq), u_gen);
   };
+
+let rec is_complete_skel =
```
cyrus-, line 74: completeness according to POPL 17 paper actually requires no type holes, but for renumbering we don't want to do that. can you add an optional bool argument to all the is_complete functions called "check_type_holes" that serves as a flag for whether to count type holes or not

cyrus- (changes requested):
see comments on PR / code

cyrus- (approved):

src/hazelcore/semantics/UHExp.re
```diff
@@ -340,3 +340,65 @@ let text_operand =
       u_gen,
     );
   };
+
+let rec is_complete_line = (l: line, check_type_holes: bool): bool => {
+  switch (l) {
+  | EmptyLine => true
+  | LetLine(pat, _, body) =>
```
cyrus-: this should check the type annotation if check_type_holes is true

cyrus- (changes requested):
one last bug

src/hazelcore/semantics/UHExp.re
utkmehta: fixed

cyrus- (approved):

PR #232, opened 2020-05-06
IntLit/FloatLit accepted inputs

### LINKED ISSUE #129: Support "_" in numbers

### LINKED ISSUE #231: Underscore with Int treated as Float
"_11" is currently being treated as float type with the value of "11."

Yunif3:
**Allowed leading/trailing zeros and underscores for IntLit and FloatLit**
**resolves #231 and resolves #129 **
 
- Removed the length checking logic in `TextShape.of_text` function. It is synchronized to the accepted values of Ocaml's [int|float]_of_string_opt now for everything except for number literals with leading underscore, which will be treated as a variable in hazel instead of float.
- 0 can be used before or after Int/Float now (ex. 0004)
- "_" has no effect on the value of IntLit and FloatLit
- "_" has no effect on the maximum number of digits of IntLit and FloatLit
- Scientific notation is used when the IntLit/FloatLit values get too large
- An Int that is too large is automatically casted to a Float


src/hazelcore/semantics/dynamics/Dynamics.re
```diff
@@ -109,15 +109,17 @@ module Pat = {
       let ctx = Contexts.extend_gamma(ctx, (x, Hole));
       Expands(Var(x), Hole, ctx, delta);
     | IntLit(NotInHole, n) =>
-      switch (int_of_string_opt(n)) {
+      let stripped_n = StringUtil.strip_underscores(n);
```
cyrus-: logic for converting the string argument of IntLit/FloatLit to an OCaml int/float should be in refactored into a helper function in TextShape, not repeated any time its needed.

src/hazelcore/util/StringUtil.re
```diff
@@ -54,3 +54,16 @@ let num_trailing_zeros = (num: string): int => {
     };
   helper(String.length(num) - 1);
 };
+
+let strip_underscores = (s: string): string => {
+  let rec helper = (index: int, stripped: string): string =>
```
cyrus-: There is likely a more efficient way using standard library functions to drop a single character.

src/hazelcore/semantics/TextShape.re
```diff
@@ -10,33 +10,35 @@ type t =
   | Var(Var.t);
 
 let of_text = (text: string): option(t) => {
+  let stripped_text = StringUtil.strip_underscores(text);
```
cyrus-: see comment below

cyrus- (changes requested):
few code cleanup things needed

cyrus-:
you could have automatic comma insertion in long number literals be part of the pretty printing logic in Hazel rather than or in addition to adding support for underscore...

Yunif3:
we could replace underscores in Int/FloatLit with something else in pretty printing logic, but this also fixes "_1" being currently being recognized as Float in Dev. 
Do you suggest making a separate fix for this (i.e. not allowing the input of a number after _"? @cyrus-

cyrus-:
let's go ahead and add the _ support for int/float literals and punt on auto-commas for now.

src/hazelcore/semantics/TextShape.re
```diff
@@ -9,34 +9,46 @@ type t =
...
-    float_of_string_opt(text),
+    int_of_string_opt(stripped_text),
+    float_of_string_opt(stripped_text),
     bool_of_string_opt(text),
     ExpandingKeyword.mk(text),
   ) {
   /* the num_digits double count with leading zeros when the int is evaluated to 0  */
   | (Some(n), _, _, _) when n == 0 =>
-    StringUtil.num_leading_zeros(text) == String.length(text)
+    StringUtil.num_leading_zeros(stripped_text)
```
cyrus-: not sure I understand why we have these checks related to the length here... can you clarify in a comment? what is an example of a situation where int_of_string_opt returns Some(0) but this check fails and the function returns None?

src/hazelcore/semantics/TextShape.re
```diff
@@ -9,34 +9,46 @@ type t =
...
   switch (
-    int_of_string_opt(text),
-    float_of_string_opt(text),
+    int_of_string_opt(stripped_text),
+    float_of_string_opt(stripped_text),
     bool_of_string_opt(text),
     ExpandingKeyword.mk(text),
   ) {
   /* the num_digits double count with leading zeros when the int is evaluated to 0  */
   | (Some(n), _, _, _) when n == 0 =>
```
cyrus-: you don't need "when n == 0", you can just match on "Some(0)", though I'm not entirely sure why 0 is special cased like this?

src/hazelcore/semantics/TextShape.re
```diff
@@ -9,34 +9,46 @@ type t =
...
-    // when parsing ints and floats from strings, we don't
-    StringUtil.num_leading_zeros(text)
-    + IntUtil.num_digits(n) == String.length(text)
+    StringUtil.num_leading_zeros(stripped_text)
+    + IntUtil.num_digits(n) == String.length(stripped_text)
       ? Some(IntLit(text)) : None
   /* 1 is subtracted from num_digits because Ocaml introduces extra 0 in front of the decimal when float is < 1 */
   | (_, Some(f), _, _) when Float.abs(f) < 1.0 =>
-    StringUtil.num_leading_zeros(text)
+    StringUtil.num_leading_zeros(stripped_text)
```
cyrus-: same here, why are we special casing these things? can you provide examples of text that goes into each branch in comments?

cyrus- (changes requested):

cyrus-:
@Yunif3 you're behind `dev`, please make sure you're up-to-date with `dev` before requesting review

cyrus-:
@Yunif3 still need you to get this caught up with `dev`, it is currently 107 commits behind. should be a simple `git merge dev` after you pull.

cyrus-:
@Yunif3 wasn't the plan to support all of OCaml's syntax for floats and ints other than prefix "_"?

cyrus-:
Oh I see, it sort of works but you can't actually type "0b1111" and things like that because "0b" is not valid. Same with scientific notation, you can't type "e" because there is nothing after the "e" at that moment...

Yunif3:
Yes, I didn't add support for binary, octal, and hexadecimal. Do you think it's something that we want right now? Also, many of these problems will be probably solved once the InvalidVar is introduced to replace None in `of_shape` function

cyrus-:
it does actually support those notations now in your branch as long as you put them in between existing digits, e.g. you can type `b` when your cursor is at `1|1`. yes, I think we need to get InvalidVar support working, so I'll go ahead and merge this for now. @nwilsonj, what's the status of invalidliterals?

cyrus- (approved):

PR #243, opened 2020-05-08
Float comparison op

### LINKED ISSUE #236: Add floating point comparison operators
`<.` `>.`, `==.` for now (we don't have multi-char ops yet)

Yunif3:
**Added comparison operators for Float types**
The float point comparison operators shapes and behaviors are analogous to Int comparison operators.
The operators could later be changed with overloading, but the following operators are supported for now:
- `<.`
- `>.`
- `==.`


cyrus-:
the commit message says " allowed boolean float ops to be backspaced to boolean int ops " but backspace on float ops just deletes the op entirely...

cyrus-:
Actually it does behave correctly in synthetic position but not in analytic position.

cyrus- (approved):

PR #253, opened 2020-05-12
Swap actions new key binding

### LINKED ISSUE #248: swap actions not working
tried on my laptop and desktop, both running Linux -- something about keyboard handling? I think we might need to consider an alternative key binding if it works on some machines but not others.

Yuning30:
In this pull request, the keybindings of swap actions are changed to Ctrl+Alt+i/k/j/l. And `is_movement_key` function in `JSUtil.re` is modified because we don't use no arrow keys in involved except basic move.


Yuning30:
A more detailed analysis on the reason why the swap actions fail can be found at the linked issue.
https://github.com/hazelgrove/hazel/issues/248#issuecomment-627583318

cyrus- (approved):

PR #260, opened 2020-05-17
extraction to OCaml

sweetsinpackets:
The extraction to OCaml module for hazel.


extraction_history/Extraction_declear.re
```diff
@@ -0,0 +1,38 @@
+/* This is a decleration of the inner types used in the extraction
+   most base but useless code can be directly re-used in the old version
+       - Note that opseq had been rewrite to a top-level expression, need to re-write indeed.
+
+   TODO: things in order need to be done
+       1. declear the inner type used and functions
+       2. declear the vocabulary_set functions
+       3. include the error type and pass rules*/
+
+type pass_t =
```
cyrus-: why can't you use HTyp.t and VarCtx.t directly?

extraction_history/Extraction_declear.re
```diff
@@ -0,0 +1,38 @@
...
+  | HOLE //hole indicates there's an incomplete hole, like conflict, will throw hole error
+  | Bool
+  | Number
+  | Unit
+  | List(pass_t)
+  | ARROW(pass_t, pass_t) //ARROW(a, b) = a -> b
```
cyrus-: Why are some of these all caps and others initial caps?

extraction_history/Extraction_declear.re
cyrus-: what does "pass_t" mean?

extraction_history/Extraction_tool.re
```diff
@@ -0,0 +1,217 @@
...
+    | (Some(s1), None) => Some(s1)
+    | (None, Some(s2)) => Some(s2)
+    | _ => None
+    }
+  };
+
+//insert indent_space just like previous work
+let rec indent_space = (~level: int): string =>
+  if (level > 0) {
+    "  " ++ indent_space(~level=level - 1);
```
cyrus-: probably easier to just run ocaml's autoformatter on the output than try to manage layout/indentation manually

extraction_history/Extraction_declear.re
```diff
@@ -0,0 +1,38 @@
...
+  | SUM(pass_t, pass_t)
+  | PROD(pass_t, pass_t)
+  //| Pair(pass_t, pass_t)
+  | EMPTY //as None, or not a Type
+  // I think EMPTY should never appear since everything have a type
+  | UNK // dependency type like list(a), or can trans to whatever
+  // Strings should be UNK type
+  // Or let expression should have UNK, meaning "not applied"
+  | CONFLICT; //it's an error, only errors like gradual types
```
cyrus-: I don't understand the difference between hole, empty, unk, and conflict?

extraction_history/Extraction_declear.re
```diff
@@ -0,0 +1,38 @@
...
+// the return type of most function
+// (extracted string, the passing type)
+// FIXME: if pass_t is CONFLICT, EMPTY, HOLE... string may not exactly be None
+//        need a further check maybe at last
+type extract_t = (option(string), pass_t);
+
+// the variable environment
+// from top level to down levels, each level insert new ones
+// (name, type), if local bind, override as rules
+type variable_set_t = list((string, pass_t));
```
cyrus-: this should just be VarCtx.t

extraction_history/Extraction_uhexp.re
```diff
@@ -0,0 +1,392 @@
...
+// FIXME: If we encounter "?" holes, for example in Case, if we leave it empty, it isn't None, but Some(Hole) indeed
+// It means when the user see a "? hole" but don't fill it, we receive Some(Hole) in AST. But if the user doesn't see a hole, it should be None
+// Currently, we don't have None cases except for enter ":" in "let"
+// TODO: It's safe and okey to transform Hole to None only for "? holes", i.e. call Extraction_tool.hole_to_none
+
+// The extract_t is prepared for ExpLine
+// In Letline, we can't actually evaluate the type for a let
+//      but we can evaluate the block in let
+//      Letline itself should have UNK, because we can't let forever
+let rec uhexp_trans = (~t: UHExp.t, ~vs: variable_set_t): extract_t =>
```
cyrus-: these are just syntactic translations right? we had talked last semester about doing it in a bidirectional typing style. I think you should start with the code in Dynamics.Exp.syn_expand and modify it to output an OCaml string or OCaml AST instead of a DHExp, rather than writing something from scratch.

cyrus- (changes requested):
This doesn't seem to be the right approach -- your code should look almost the same as Dynamics.Exp.syn_expand, differing only in that the output is an OCaml expression rather than a DHExp. See comments.

npm-debug.log
```diff
@@ -0,0 +1,19 @@
+0 info it worked if it ends with ok
```
cyrus-: don't include logs like this in commits. not sure why npm would be involved in building Hazel, but if you're doing something locally then you can add these kinds of files to .gitignore.

src/hazelweb/gui/OptionsPanel.re
```diff
@@ -165,6 +165,37 @@ let view =
             ~on_change=() => inject(ToggleMemoizeDoc),
             model.memoize_doc,
           ),
+          // Node.button(
```
cyrus-: you should put this button down at the bottom next to the "Serialize to console" button

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
+module Typ = {
+  type ocaml_typ = string;
```
cyrus-: define a separate OCamlTyp module with a type t

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
+module Typ = {
```
cyrus-: let's put all of the extraction modules into a directory `hazelcore/extractions/ocaml`

cyrus-: also you should separate these out into OCamlExtraction_Typ, OCamlExtraction_Exp, etc. following the recent refactoring

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+  type ocaml_typ = string;
+
+  let rec extract = (~t: HTyp.t): ocaml_typ =>
+    switch (t) {
+    | Hole => "'a"
```
cyrus-: this should be a failure case, not 'a

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+    | Int => "int"
+    | Float => "float"
+    | Bool => "bool"
+    // add parenthesis is necessary
+    // (int -> int) -> int != int -> int -> int
+    | Arrow(t1, t2) =>
+      "(" ++ extract(~t=t1) ++ " -> " ++ extract(~t=t2) ++ ")"
+    | Sum(t1, t2) => "(" ++ extract(~t=t1) ++ " | " ++ extract(~t=t2) ++ ")"
```
cyrus-: I don't think this is valid OCaml syntax for anonymous sums. You'll have to use polymorphic variants for this (you can read more about them online)

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+    | Prod(tl) => "(" ++ list_prod(~tl) ++ ")"
+    | List(t) => extract(~t) ++ " list"
+    }
+  and list_prod = (~tl: list(HTyp.t)): ocaml_typ =>
+    switch (tl) {
+    | [] => ""
```
cyrus-: empty product should extract to `unit`, not `()`

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+    switch (dp) {
+    | EmptyHole(_, _) => failwith("Pat: Empty Hole")
+    | NonEmptyHole(_, _, _, _) => failwith("Pat: NonEmptyHole")
+    | Wild => "_"
+    | Keyword(_, _, _) => failwith("Pat: Incomplete Program, Keyword")
+    | Var(s) => s
+    | IntLit(i) => string_of_int(i)
+    | FloatLit(f) => string_of_float(f)
+    | BoolLit(b) => string_of_bool(b)
+    | Inj(_side, t) => extract(~dp=t)
```
cyrus-: see above on how to encode sums

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+    | ListNil => "[]"
+    | Cons(dp1, dp2) => extract(~dp=dp1) ++ "::" ++ extract(~dp=dp2)
```
cyrus-: will need to add parentheses here

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+    | Pair(dp1, dp2) =>
+      "(" ++ extract(~dp=dp1) ++ ", " ++ extract(~dp=dp2) ++ ")"
+    | Triv => "()"
+    | Ap(dp1, dp2) => extract(~dp=dp1) ++ " " ++ extract(~dp=dp2)
```
cyrus-: will need to add parentheses here

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+  //TODO: replace == with HTyp.consistent
+  let rec extract = (~ctx: Contexts.t, ~de: DHExp.t): extract_result =>
+    switch (de) {
+    | EmptyHole(_) => failwith("Exp: Empty Hole")
+    | NonEmptyHole(_) => failwith("Exp: Non-Empty Hole")
+    // used for variables that are keywords in Hazel, like "let" and "case"
+    | Keyword(_) => failwith("Exp: Incomplete Program (Keyword)")
+    | FreeVar(_, _, map, x) =>
+      //lookup the expression represented by x
+      let map_e = VarMap.lookup(map, x);
```
cyrus-: a free var is one that is not in the context, so you don't need to do this lookup

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+      //directly lookup type in environment
+      let typ = VarCtx.lookup(Contexts.gamma(ctx), x);
+      switch (typ) {
+      | None => failwith("Exp: BoundVar " ++ x ++ " Not Found")
+      // if the hole type, we don't print the type
+      | Some(t) =>
+        if (t == Hole) {
+          ("(" ++ x ++ ")", t);
+        } else {
+          ("(" ++ x ++ " : " ++ Typ.extract(~t) ++ ")", t);
```
cyrus-: do you really need to annotate each variable?

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+        let ocaml_dp = Pat.extract(~dp);
+        let ocaml_de1 = extract(~ctx, ~de=de1);
+        let ctx' = Contexts.extend_gamma(ctx, (x, snd(ocaml_de1)));
+        let ocaml_de2 = extract(~ctx=ctx', ~de=de2);
+        // use the "let .. = ..;;" format
+        // if it's a recursive function, we skip to directly evaluating the FixF (de1)
+        // because the recursive function should be declared by "rec" only in ocaml
+        switch (de1) {
+        | FixF(_, _, _) => (
+            fst(ocaml_de1) ++ fst(ocaml_de2),
```
cyrus-: this should turn into a let rec here I think, not in the FixF case below

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+            "(" ++ fst(ocaml_d1) ++ "::" ++ fst(ocaml_d2) ++ ")",
+            snd(ocaml_d2),
+          );
+        } else {
+          failwith("Exp: Cons, type inconsistent");
+        }
+      | _ => failwith("Exp: Cons")
+      };
+    | Inj(t, side, de) =>
+      //ocaml don't have injection
```
cyrus-: see above

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,374 @@
...
+      };
+    }
+  and rule_extract =
+      (~ctx: Contexts.t, ~rule: DHExp.rule, ~pat_t: HTyp.t): extract_result =>
+    switch (rule) {
+    | Rule(dp, de) =>
+      let ocaml_dp = Pat.extract(~dp);
+      //we seems don't allow constructors as pattern
+      //we should add the patterns into the context
+      let rec update_ctx = (dp: DHPat.t, ctx: Contexts.t): Contexts.t =>
```
cyrus-: you should be able to call into Statics for this, don't need to write it yourself here...

cyrus- (changes requested):

src/hazelcore/OcamlExtraction.re
sweetsinpackets: What do you mean by "call into Statics"? I'm a little bit confused on that, does it means I need to refer to something?

src/hazelcore/OcamlExtraction.re
sweetsinpackets: I also don't like that, because it will make the extracted code very messy... May I change to that we only annotate variables when they're declared?

src/hazelcore/OcamlExtraction.re
sweetsinpackets: May I ask how the FixF will appear? Will it appear as Let(..., FixF(...), ...) or directly FixF if a recursive function is declared? Here when the Let encounters an FixF, I directly pass it to FixF handler to extract to "let rec".

src/hazelcore/OcamlExtraction.re
cyrus-: you can use the Statics_Pat module to generate the context for a pattern, since that is what the pattern type synthesis function does.

src/hazelcore/OcamlExtraction.re
cyrus-: Hazel programs will only ever result in internal expressions containing let ... = fix ..., so you can assume that.

src/hazelcore/OcamlExtraction.re
cyrus-: yes if you annotate every variable binding then the variable will necessarily have that type everywhere

src/hazelcore/OcamlExtraction.re
sweetsinpackets: I've read about the polymorphic variants. I'm a little bit worried that the variant needs a constructor, for example, if I encode the sum type as ``` `Left of t | `Right of t ```, then I need to explicitly add the constructor ``` `Left ```or ``` `Right``` for the expression. I'm wondering how the sum type expression is created in Hazel, can we directly declare something of a sum type?

src/hazelcore/OcamlExtraction.re
sweetsinpackets: I see the syn in Statics_Pat requires type UHPat.t, but the patterns I receive should be already of type DHPat.t. I don't know how I can use the module.

src/hazelcore/OcamlExtraction.re
sweetsinpackets: then should I directly assert an error? Or can you raise an example to explain how could this case happen, I can't imagine how to determine the HTyp.t of it without using lookup.

cyrus-: yes this should be an error -- an expression with a free variable cannot be extracted to OCaml because free variables cause type errors in OCaml

src/hazelcore/OcamlExtraction.re
cyrus-: oh right, I'm sorry.

you should write an type_of function in DHPat that generates both the type and ctx for a pattern, and then call that here.

sweetsinpackets: I think the type_of function for DHPat is not realistic, because there's "Wild" which needs to judge by context. I'll first write a function takes an expected type for a pattern and generate the type and update the ctx instead, I'll first store it in Extraction file, and let's see how we can generalize it later.

cyrus-: ah yeah it should take the type as input, not produce it as output.

src/hazelcore/OcamlExtraction.re
cyrus-: this wasn't resolved

src/hazelcore/OcamlExtraction.re
cyrus-: this wasn't resolved

src/hazelcore/OcamlExtraction.re
```diff
@@ -0,0 +1,375 @@
+module Typ = {
```
cyrus-: this file should be deleted

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+//type extract_result = option((ocaml_exp, HTyp.t));
+type t = (ocaml_exp, HTyp.t)
+and ocaml_exp = string;
+
+/* Theorem: If extract(ctx, d) = Some(ce, ty) and ctx |- d : ty then ce :_ocaml Typ.extract(ty) */
+
+//TODO: replace == with HTyp.consistent
+let rec extract = (~ctx: Contexts.t, ~de: DHExp.t): t =>
+  switch (de) {
+  | EmptyHole(_) => failwith("Exp: Empty Hole")
```
cyrus-: define an ExtractionFailed(string) exception instead of using failwith

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+  | NonEmptyHole(_) => failwith("Exp: Non-Empty Hole")
+  // used for variables that are keywords in Hazel, like "let" and "case"
+  | Keyword(_) => failwith("Exp: Incomplete Program (Keyword)")
+  // an expression with a free variable cannot be extracted to OCaml because free variables cause type errors in OCaml
+  | FreeVar(_, _, _, _) => failwith("Exp: FreeVar is not allowed")
+  | BoundVar(x) =>
```
cyrus-: need to add a check that the variable is a valid ocaml variable name

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+          ++ " : "
+          ++ OCamlExtraction_Typ.extract(~t=snd(ocaml_de1))
+          ++ " = "
+          ++ fst(ocaml_de1)
+          ++ ";;\n"
+          ++ fst(ocaml_de2),
+          snd(ocaml_de2),
+        )
+      };
+    | _ => failwith("Exp: Let, not a variable pattern")
```
cyrus-: ocaml supports patterns in let bindings, so this should not be a failure case

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+      //, because the recursive function should be declared by "rec" only in ocaml
+      switch (de1) {
+      | FixF(_, _, _) => (
+          "let rec "
+          ++ ocaml_dp
+          ++ " : "
+          ++ OCamlExtraction_Typ.extract(~t=snd(ocaml_de1))
+          ++ " = "
+          ++ fst(ocaml_de1)
+          ++ ";;\n"
```
cyrus-: let expressions end with "in", not ";;", in general -- the ";;" syntax will only work at the very top level

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+    //Hazel programs will only ever result in internal expressions containing let ... = fix ...
+    //let rec f: some type = fun x -> f ... (is reasonable)
+    // x should be already inserted into context
+
+    //let ocaml_ht = OCamlExtraction_Typ.extract(~t=ht);
+    let ctx' = Contexts.extend_gamma(ctx, (x, ht));
+    let ocaml_de = extract(~ctx=ctx', ~de);
+    (
+      // "let rec " ++ x ++ " : " ++ ocaml_ht ++ " = " ++ fst(ocaml_de) ++ ";;\n",
+      fst(ocaml_de),
```
cyrus-: instead of using fst, use pattern matching in the let above

cyrus-: (happens a few other times too, will just mention it once here)

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+      if (t1 == snd(ocaml_de2) || t1 == Hole) {
+        ("(" ++ fst(ocaml_de1) ++ " " ++ fst(ocaml_de2) ++ ")", t2);
+      } else {
+        failwith("Exp: Apply, type inconsistent");
+      }
+    | _ => failwith("Exp: Apply, not an Arrow type")
+    };
+  | BoolLit(bool) => (string_of_bool(bool), Bool)
+  | IntLit(int) => (string_of_int(int), Int)
+  | FloatLit(float) => (string_of_float(float), Float)
```
cyrus-: will need to special case special floats like nan and inf (ask Yunsoo for help if you need it)

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+  | BinBoolOp(op, d1, d2) =>
+    switch (d1, d2) {
+    | (BoolLit(b1), BoolLit(b2)) =>
```
cyrus-: d1 and d2 don't need to be bool lits, they can be any boolean expression

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+      | Divide => (" / ", Int)
+      | LessThan => (" < ", Bool)
+      | GreaterThan => (" > ", Bool)
+      | Equals => (" == ", Bool)
+    );
+    let ocaml_op = str_of_op(op);
+    let ocaml_d1 = extract(~ctx, ~de=d1);
+    let ocaml_d2 = extract(~ctx, ~de=d2);
+    switch (snd(ocaml_d1), snd(ocaml_d2)) {
+    | (Int, Int) => (
```
cyrus-: you don't need to check the types, you can assume that the input is well-typed

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+      | FDivide => (" /. ", Float)
+      | FLessThan => (" < ", Bool)
+      | FGreaterThan => (" > ", Bool)
+      | FEquals => (" == ", Bool)
+    );
+    let ocaml_op = str_of_op(op);
+    let ocaml_d1 = extract(~ctx, ~de=d1);
+    let ocaml_d2 = extract(~ctx, ~de=d2);
+    switch (snd(ocaml_d1), snd(ocaml_d2)) {
+    | (Float, Float) => (
```
cyrus-: same

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+    | _ => failwith("Exp: BinFloatOp")
+    };
+  | ListNil(t) => ("[]", List(t))
+  | Cons(d1, d2) =>
+    //TODO: add proper parenthesis to avoid priority in case "map"
+    let ocaml_d1 = extract(~ctx, ~de=d1);
+    let ocaml_d2 = extract(~ctx, ~de=d2);
+    switch (snd(ocaml_d2)) {
+    //FIXME: will ever happen? test!
+    | Hole => (
```
cyrus-: won't ever happen (there would always be a cast), so don't need to handle cases like this

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+    | R => ("(`Right " ++ fst(ocaml_exp) ++ ")", Sum(t, snd(ocaml_exp)))
+    };
+  | Pair(d1, d2) =>
+    let ocaml_d1 = extract(~ctx, ~de=d1);
+    let ocaml_d2 = extract(~ctx, ~de=d2);
+    (
+      "(" ++ fst(ocaml_d1) ++ ", " ++ fst(ocaml_d2) ++ ")",
+      Prod([snd(ocaml_d1), snd(ocaml_d2)]),
+    );
+  | Triv => ("()", Hole) //We have no unit Htype...
```
cyrus-: Prod([]) is the unit type

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+    | Case(de, rules, _) =>
+      let ocaml_de = extract(~ctx, ~de);
+      let ocaml_rules = rules_extract(~ctx, ~rules, ~pat_t=snd(ocaml_de));
+      //FIXME: Do we really need to indicate the type?
+      let str =
+        "((match ("
+        ++ fst(ocaml_de)
+        ++ ") with \n"
+        ++ fst(ocaml_rules)
+        ++ ") : "
```
cyrus-: I don't think you need the type here

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+  | InconsistentBranches(_, _, _, _) =>
+    failwith("Exp: Ocaml don't allow inconsistent branches")
+  | Cast(_, _, _) => failwith("Exp: cast not allowed")
+  | FailedCast(_, _, _) => failwith("Exp: cast not allowed")
+  | InvalidOperation(_, _) => failwith("Exp: Invalid Operation")
+  }
+and rules_extract =
+    (~ctx: Contexts.t, ~rules: list(DHExp.rule), ~pat_t: HTyp.t): t =>
+  switch (rules) {
+  | [] => ("", Hole) //no rules, won't happen actually
```
cyrus-: should raise an exception if it is an impossible case

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,270 @@
...
+      (fst(head) ++ "\n" ++ fst(tail), snd(head));
+    } else {
+      failwith("Exp: Case rules with inconsistent results");
+    };
+  }
+and rule_extract = (~ctx: Contexts.t, ~rule: DHExp.rule, ~pat_t: HTyp.t): t =>
+  switch (rule) {
+  | Rule(dp, de) =>
+    let ocaml_dp = OCamlExtraction_Pat.extract(~dp);
+    //we seems don't allow constructors as pattern
```
cyrus-: not sure what you mean by this?

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,98 @@
+type t = string;
```
cyrus-: some of the same concerns raised above with expressions are relevant here, won't repeat

src/hazelcore/extractions/ocaml/OCamlExtraction_Typ.re
```diff
@@ -0,0 +1,32 @@
...
+  | Bool => "bool"
+  // add parenthesis is necessary
+  // (int -> int) -> int != int -> int -> int
+  | Arrow(t1, t2) =>
+    "(" ++ extract(~t=t1) ++ " -> " ++ extract(~t=t2) ++ ")"
+  // use Polymorphic Variants to encode sum, i.e. 'Left of t1, 'Right of t2, and easier for injection
+  // the type is "[< `Left of t1 | `Right of t2]", < means no more than
+  // so the expression should be `Left e1, `Right e2 to match this design
+  | Sum(t1, t2) =>
+    "[< "
```
cyrus-: I don't think you need the "<", that's only for use with subtyping

src/hazelcore/extractions/ocaml/OCamlExtraction_Typ.re
```diff
@@ -0,0 +1,32 @@
...
+    ++ "`Left of "
+    ++ extract(~t=t1)
+    ++ " | "
+    ++ "`Right of "
+    ++ extract(~t=t2)
+    ++ " ]"
+  | Prod(tl) => "(" ++ list_prod(~tl) ++ ")"
```
cyrus-: careful with products, since internally we only use pairs but ocaml supports n-tuples. you might have to parenthesize in list_prod

cyrus- (changes requested):

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Oh, the reason I use `failwith` is it can directly stop the program and the string can be seen in the console. Should I write a new return type and change the program to that if we encounter a `ExtractionFailed(string)`, we just return it?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: May I ask why we don't use fst and snd? I think fst and snd may make the code looks a lot better, because we will call then multiple times in a block...

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Do I need to check whether they're correct here? I means if I can assume they're bool lits or boolean expression, then I don't need the check.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: is that the case similar in the pattern in Rule (in Case)? If so, I can reuse the code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Is the what you mean "the input is well-typed" in the other comment?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: let's add the check to pattern, so whenever it's declared, it should be correct. I'll use Var.is_valid to check, because ocaml variable follows the same regular expression.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Oh I forget to remove that, sorry for that. I just want to remind myself make things easy because Hazel don't support constructor now (like `| Const(x) -> ...`)

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Oh, I just rewrote, it actually looks better in pattern matching.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Sorry I can't use Var.is_valid, because ocaml variable shouldn't start with A-Z

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
+/* Theorem: If extract(ctx, d) = Some(ce, ty) and ctx |- d : ty then ce :_ocaml Typ.extract(ty) */
+
+type t =
+  | OcamlExp(ocaml_exp, HTyp.t)
+  | ExtractionFailed(string)
+and ocaml_exp = string;
```
adamsmd: Since `ocaml_exp` doesn't depend on `t` it should be a separate type declaration (above `t` because `t` does depend on `ocaml_exp`).

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
+/* Theorem: If extract(ctx, d) = Some(ce, ty) and ctx |- d : ty then ce :_ocaml Typ.extract(ty) */
+
+type t =
+  | OcamlExp(ocaml_exp, HTyp.t)
```
adamsmd: What is this `HTyp.t` used for?  Please add a comment explaining it.

If it is only used to cause extraction to fail on improperly typed code, then I would omit the `HTyp.t` from `OcamlExp` and omit those type checks.  This would greatly simplify your implementation, and I think it is fine that improperly typed code might extract to improperly typed code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+            OcamlExp(
+              "let "
+              ++ pat_str
+              ++ " : "
+              ++ typ_str
+              ++ " = "
+              ++ exp1_str
+              ++ " in\n"
+              ++ exp2_str,
+              exp2_typ,
```
adamsmd: Factor the common code between these two cases

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+  // an expression with a free variable cannot be extracted to OCaml because free variables cause type errors in OCaml
+  | FreeVar(_, _, _, _) => ExtractionFailed("Exp: FreeVar is not allowed")
+  | BoundVar(x) =>
+    //directly lookup type in environment
+    let typ = VarCtx.lookup(Contexts.gamma(ctx), x);
+    switch (typ) {
+    | None => ExtractionFailed("Exp: BoundVar " ++ x ++ " Not Found")
+    | Some(t) => OcamlExp(x, t)
+    };
+  | Let(dp, de1, de2) =>
```
adamsmd: I think the convention in the rest of the Hazel code base would be to name these variables `pat`, `exp1` and `exp2`.  The `d` prefix isn't needed since everything in this file is a `DH` type, and `pat` is easier to read than `p`.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+              ++ exp2_str,
+              exp2_typ,
+            )
+          }
+        }
+      }
+    };
+  | FixF(x, ht, de) =>
+    //it's the case of fixpoint/ recursive functions
+    //Hazel programs will only ever result in internal expressions containing let ... = fix ...
```
adamsmd: To ensure this invariant, you should move this case inside the `let` case.  Then put something that just throws an error here.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+        ExtractionFailed("Exp: Case rules with inconsistent results");
+      }
+    };
+  }
+and rule_extract = (~ctx: Contexts.t, ~rule: DHExp.rule, ~pat_t: HTyp.t): t =>
+  switch (rule) {
+  | Rule(dp, de) =>
+    switch (
+      OCamlExtraction_Pat.extract(dp),
+      OCamlExtraction_Pat.update_pattern(dp, pat_t, ctx),
```
adamsmd: `extract` and `update_pattern` should be combined into one function that takes a pattern and a  context and returns the code for that pattern along with a new context

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+
+    // Here the function name should be added by let, do it here to be careful
+    let ctx' = Contexts.extend_gamma(ctx, (x, ht));
+    switch (extract(ctx', de)) {
+    | ExtractionFailed(err) => ExtractionFailed(err)
+    | OcamlExp(exp_str, _exp_typ) => OcamlExp(exp_str, ht)
+    };
+  | Lam(dp, ht, de) =>
+    //Htyp.t is the ground type of pattern
+    switch (dp) {
```
adamsmd: Since Hazel allows `λx, y:(Int, Int).{x + y}` use extract_pat here instead of an explicit match.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+    }
+  | Ap(de1, de2) =>
+    //apply
+    let ocaml_de1 = extract(ctx, de1);
+    let ocaml_de2 = extract(ctx, de2);
+    switch (ocaml_de1, ocaml_de2) {
+    | (ExtractionFailed(err), _)
+    | (_, ExtractionFailed(err)) => ExtractionFailed(err)
+    | (OcamlExp(exp1_str, exp1_typ), OcamlExp(exp2_str, exp2_typ)) =>
+      //FIXME: May can remove the type check
```
adamsmd: I would omit the type check.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+type t =
+  | OcamlExp(ocaml_exp, HTyp.t)
+  | ExtractionFailed(string)
+and ocaml_exp = string;
+
+//TODO: replace == with HTyp.consistent
+let rec extract = (ctx: Contexts.t, de: DHExp.t): t =>
+  switch (de) {
+  | EmptyHole(_) => ExtractionFailed("Exp: Empty Hole")
+  | NonEmptyHole(_) => ExtractionFailed("Exp: Non-Empty Hole")
```
adamsmd: If we get rid of the returned type (as in the comment for line 4), then instead of failing, just continue extracting the contents of the hole.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
+/* Theorem: If extract(ctx, d) = Some(ce, ty) and ctx |- d : ty then ce :_ocaml Typ.extract(ty) */
+
+type t =
+  | OcamlExp(ocaml_exp, HTyp.t)
+  | ExtractionFailed(string)
```
adamsmd: Using `raise` instead of `ExtractionFailed` would simplify a lot of this code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+and ocaml_exp = string;
+
+//TODO: replace == with HTyp.consistent
+let rec extract = (ctx: Contexts.t, de: DHExp.t): t =>
+  switch (de) {
+  | EmptyHole(_) => ExtractionFailed("Exp: Empty Hole")
```
adamsmd: Extract this to an OCaml call to `raise`.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+  | NonEmptyHole(_) => ExtractionFailed("Exp: Non-Empty Hole")
+  // used for variables that are keywords in Hazel, like "let" and "case"
+  | Keyword(_) => ExtractionFailed("Exp: Incomplete Program (Keyword)")
```
adamsmd: This message should provide more information (e.g., which keyword) and be more human readable.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+  // an expression with a free variable cannot be extracted to OCaml because free variables cause type errors in OCaml
+  | FreeVar(_, _, _, _) => ExtractionFailed("Exp: FreeVar is not allowed")
```
adamsmd: If we get rid of the returned type (as in the comment for line 4), then instead of failing, allow free variables.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: I think `OcamlExp` should be `OCamlExp`

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+          OcamlExp(
+            "(match (" ++ exp_str ++ ") with \n" ++ rules_str ++ ")",
+            rules_typ,
+          )
+        };
+      }
+    }
+  // inconsistbranches is a case have inconsistent types
+  | InconsistentBranches(_, _, _, _) =>
+    ExtractionFailed("Exp: Ocaml don't allow inconsistent branches")
```
adamsmd: If we get rid of the returned type (as in the comment for line 4), then instead of failing, just continue extracting.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+  | Cast(_, _, _) => ExtractionFailed("Exp: cast not allowed")
```
adamsmd: I think these should become type ascription (e.g., `(exp : typ)`)

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
...
+  | FailedCast(_, _, _) => ExtractionFailed("Exp: cast not allowed")
+  | InvalidOperation(_, _) => ExtractionFailed("Exp: Invalid Operation")
+  | InvalidText(_, _, _, err) =>
+    ExtractionFailed("Exp: Invalid Text " ++ err)
+  }
+and rules_extract =
```
adamsmd, line 199: If we get rid of the returned type and the type check (as in the comment for line 4), then this is just a `List.fold_left` or `List.fold_right`.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,139 @@
+type t =
+  | OcamlPat(string)
```
adamsmd: OcamlPat -> OCamlPat

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,139 @@
...
+  | ExtractionFailed(string);
+
+// Check whether a variable name is valid in ocaml, use standard library Re.Str
```
adamsmd: Remove ", use standard library Re.Str".  For one, "standard library" means something on this list "https://caml.inria.fr/pub/docs/manual-ocaml/stdlib.html".  Thus, Re is a library not a standard library.  For another, you don't need to explain that you are using a library.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,139 @@
...
+// should start with _ or lowercase
+let ocaml_name_check = (name: Var.t): bool => {
+  let re = Re.Str.regexp("^[_a-z][_a-zA-Z0-9]*$");
```
adamsmd, line 13: Add a comment citing the exact section number and URL of the standard [1] that defines OCaml names.

[1] https://caml.inria.fr/pub/docs/manual-ocaml/

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,139 @@
...
+      }
+    | ExtractionFailed(s) => ExtractionFailed(s)
+    }
+  };
+
+// take dp as a pattern, pat_t the expected type, and ctx the original context
+// used to update the pattern in case or let
+type ctx_update_t =
+  | UpdateResult(Contexts.t)
+  | UpdateFailed(string);
```
adamsmd: Use `raise`

src/hazelcore/extractions/ocaml/OCamlExtraction_Typ.re
```diff
@@ -0,0 +1,46 @@
+type t =
+  | OcamlTyp(string)
+  | ExtractionFailed(string);
```
adamsmd: Use `raise`

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: You should pass an indentation level as argument so the code you produce is properly indented.

adamsmd (changes requested):

adamsmd:
I would also recommend renaming the `extractions` folder to `extraction`

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Well, actually it's an idea from cyrus, we need it to fully annotate the extracted code (the type will explicitly be declared when a variable is declared). In the implementation, it's necessary because I had to add some patterns into the extraction context, for example, the case rule `| x => x`. It's the simplest way to achieve the full annotation.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: In the last version I was using `failwith`, Cyrus told me to change into a `ExtractionFailed(string)` way. So what's your idea on which way should we go?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: The very first version of the extraction was a direct translation from the hazel expression to OCaml expression, but it seems cyrus want us to fully-annotate and raise errors on holes. For this reason, I think it's not a good idea to drop `Htyp.t` to make it easier.

adamsmd: I will discuss this with Cyrus and get back to you.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: I asked how to handle cast, and Cyrus tells me that I can directly throw an exception.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
cyrus-: We need type information sometimes to be able to put in type annotations. So extraction follows the same structure as elaboration from UHExp to DHexp

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
cyrus-: We use `e` for UHExp, `d` for DHExp, `p` for UHPat, `dp` for DHPat, etc.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
cyrus-: We can discuss in OH

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
cyrus-: We can do this, but it's also fine imo to just run it through an ocaml pretty printer at the end to avoid keeping track of indentation levels.

adamsmd: I would prefer an OCaml pretty printer, but I wasn't able to find a pretty printer that was usable as a library and not an external program that the user would have to run.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: Where does the extraction using the type information?  I can't find it.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: I would prefer using a pretty printer, but I couldn't find one that we could use as a library and not a separate program that users would have to run manually.

(Apologies if this message is a duplicate.  I thought I posted it, but can't find it now.)

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,308 @@
+/* Theorem: If extract(ctx, d) = Some(ce, ty) and ctx |- d : ty then ce :_ocaml Typ.extract(ty) */
+
+type t =
+  | OcamlExp(ocaml_exp, HTyp.t)
+  | ExtractionFailed(string)
+and ocaml_exp = string;
+
+//TODO: replace == with HTyp.consistent
```
cyrus-: does this need to be resolved?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: In most cases, we don't need to update the context. So if we combine them into one function, the context argument will not be used and we have to add extra destruction of the return result for most uses. For this concern, I think separating them is better.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: yes, it's an out-dated comment.

cyrus-: ok go ahead and remove it

src/hazelcore/extractions/ocaml/OCamlExtraction_Typ.re
sweetsinpackets: If we use raise to handle exceptions, then we can directly set `type t = string` here, or shall I still leave type `t = OCamlTyp(string)`?

cyrus-: I think the code is more clear if each string that is meant to be an OCaml type is marked with a constructor like that.

sweetsinpackets: Sounds good, a constructor will tell apart Pat and Typ though they're both string.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+      switch (d1) {
+      | FixF(s, ht, d) =>
+        //fixpoint function name should be added into the context to evaluate itself
+        let ctx_fix = Contexts.extend_gamma(ctx, (s, ht));
+        switch (extract(ctx_fix, d)) {
+        | OCamlExp(exp_str, _exp_typ) => OCamlExp(exp_str, ht)
+        };
+      | _ => extract(ctx, d1)
+      };
+    switch (ocaml_pat, ocaml_exp1) {
```
adamsmd: Since this pattern has only one case, deconstruct in the `let`s where ocaml_pat and ocaml_exp1 are bound.  E.g. `let OCamlPat(pat_str) = OCamlExtration_Pat.extract(db);`

This also applies to other places in the code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+
+/*
+ Naming Conventions:
+   d for DHExp, dp for DHPat, ht for HTyp
+   ocaml_(pat|typ|exp) for return result of extract
+   (typ|pat|exp)(_(str|typ))? for destructed components
+ */
+
+type ocaml_exp = string;
+type t =
```
adamsmd, line 11: Just use a tuple.  E.g., `type t = (ocaml_exp, HTyp.t)`.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+      switch (string_of_float(float)) {
+      | "inf" => "infinity"
+      | "-inf" => "neg_infinity"
+      | "nan" => "nan"
+      | s => s
+      };
+    OCamlExp(str, Float);
+  // Bin(Bool|Int|Float)Op can accept Lit or Op
+  | BinBoolOp(op, d1, d2) =>
+    switch (extract(ctx, d1), extract(ctx, d2)) {
```
adamsmd: I think this would be clearer with for the OCamlExp part, and then a separate switch over just the types.

The same applies to other parts of the code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+        | Minus => (" - ", Int)
+        | Plus => (" + ", Int)
+        | Times => (" * ", Int)
+        | Divide => (" / ", Int)
+        | LessThan => (" < ", Bool)
+        | GreaterThan => (" > ", Bool)
+        | Equals => (" == ", Bool)
+      );
+      // add parenthesis to avoid evaluation priority
+      switch (str_of_op(op)) {
```
adamsmd: Use a `let`

The same applies to other parts of the code.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+    switch (extract(ctx, d1), extract(ctx, d2)) {
+    | (OCamlExp(exp1_str, exp1_typ), OCamlExp(exp2_str, exp2_typ)) =>
+      OCamlExp(
+        "(" ++ exp1_str ++ ", " ++ exp2_str ++ ")",
+        Prod([exp1_typ, exp2_typ]),
+      )
+    }
+  | Triv => OCamlExp("()", Prod([])) //Prod([]) is the base type
+  | ConsistentCase(cases) =>
+    switch (cases) {
```
adamsmd: Use a `let`

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+          )
+        };
+      }
+    }
+  // inconsistbranches is a Case having inconsistent types
+  | InconsistentBranches(_, _, _, _) =>
+    raise(Exp_Invalid(Some("inconsistent Case branches")))
+  | Cast(_, _, _)
+  | FailedCast(_, _, _) => raise(Exp_Invalid(Some("cast")))
+  | InvalidOperation(_, _) => raise(Exp_Invalid(None))
```
adamsmd: Why is this `None`?

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,114 @@
+// mark with a constructor to tell apart with typ, exp
+type t =
+  | OCamlPat(string);
```
adamsmd: Just use a string.  E.g. `type t = string`.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,114 @@
...
+    | (OCamlPat(pat1), OCamlPat(pat2)) =>
+      OCamlPat("(" ++ pat1 ++ "::" ++ pat2 ++ ")")
+    }
+  | Pair(dp1, dp2) =>
+    switch (extract(dp1), extract(dp2)) {
+    | (OCamlPat(pat1), OCamlPat(pat2)) =>
+      OCamlPat("(" ++ pat1 ++ ", " ++ pat2 ++ ")")
+    }
+  | Triv => OCamlPat("()")
+  | Ap(dp1, dp2) =>
```
adamsmd: What cases cause this?  Are there constructors that are applied?  (`Inj` already handles `Left` and `Right'.)

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,114 @@
...
+      let ctx' = update_pattern(dp1, t, ctx);
+      update_pattern(dp2, List(t), ctx');
+    | _ => raise(OCamlExtraction_Typ.Typ_NotMatch("cons", "list"))
+    }
+  | Pair(dp1, dp2) =>
+    switch (pat_t) {
+    | Prod([h, t]) =>
+      let ctx' = update_pattern(dp1, h, ctx);
+      update_pattern(dp2, t, ctx');
+    | Prod([h, m, ...t]) =>
```
adamsmd, line 78: Add a comment explaining what cases cause this

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
```diff
@@ -0,0 +1,114 @@
...
+exception Pat_Keyword(ExpandingKeyword.t); //there's Keyword
+exception Pat_Invalid(string);
+
+// the ocaml naming standard can be found in Chapter 7.3 of ocaml-manual (https://caml.inria.fr/pub/docs/manual-ocaml/names.html)
+let ocaml_name_check = (name: Var.t): bool => {
+  let re = Re.Str.regexp("^[_a-z][_a-zA-Z0-9]*$");
+  Re.Str.string_match(re, name, 0);
+};
+
+let rec extract = (dp: DHPat.t): t =>
```
adamsmd: Can `update_pattern` and `extract` be combined into one function?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+
+//at top level, we use de instead of d because it's used for destruction only
+let rec extract = (ctx: Contexts.t, de: DHExp.t): t =>
+  switch (de) {
+  | EmptyHole(_)
+  | NonEmptyHole(_) => raise(Exp_Hole)
+  // used for variables that are keywords in Hazel, like "let" and "case"
+  | Keyword(_, _, _, keyword) => raise(Exp_Keyword(keyword))
+  // an expression with a free variable cannot be extracted to OCaml because free variables cause type errors in OCaml
+  | FreeVar(_, _, _, _) => raise(Exp_Invalid(Some("free variable")))
```
adamsmd: Include the name of the free variable

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
```diff
@@ -0,0 +1,246 @@
...
+  | BoundVar(s) =>
+    //directly lookup type in environment
+    let typ = VarCtx.lookup(Contexts.gamma(ctx), s);
+    switch (typ) {
+    | None => raise(Exp_Invalid(Some("undeclared variable")))
```
adamsmd: Include the name of the bound variable

adamsmd (changes requested):

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: It can be combined with `extract`, but may cause the `extract` function to return a more complicated result to destruct. However, the `update_pattern` is only used for a limited few times, so I think it's better to separate them.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: Ur, I remembered that Cyrus says adding a constructor is like a "label" to indicate it's a result of pattern extraction. Could you check the idea with Cyrus?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Also a case whether to include a constructor as label, the same as `OCamlPat`

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: Sorry I can't understand that, is that the case of `Ap`?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Have been rewrote.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Have been rewrote.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: Done

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Added the message.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
adamsmd: Yes, what causes an `Ap` pattern?

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: I could go either way on whether to include a constructor as a label. @cyrus-, is there a guideline on when to use tuples vs single constructor types?  E.g., `type t = (a, b)` vs `type t = Ctor(a, b)`?

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
adamsmd: @cyrus-, do you want to chime in on this?

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
adamsmd: Except for internal recursions, `OCamlExtraction_pat.extract` and `update_pattern` are always used together.  Also, the internal recursions will line up once they are the same function.

There are also philosophical reasons that these are the same.  You always have to update the scope when extracting from a pattern.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
sweetsinpackets: Btw as a supplement, I remembered last time we talked about it. The reason that we preserve the constructor is because the `typ` and `pat` will both be extracted to the `string`, and it would actually make them more clear.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: Ur.... Sorry for that, it won't ever exist. My mistake, just fixed.

src/hazelcore/extractions/ocaml/OCamlExtraction_Exp.re
adamsmd: Okay, lets use the constructor then.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
adamsmd: Nvm @cyrus-, we will use the constructor.

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
sweetsinpackets: Oh, it seems a good reason. So may I just append it with `and`, or combine the ctx into the return result of `OCamlExtraction_exp.extract`?

src/hazelcore/extractions/ocaml/OCamlExtraction_Pat.re
cyrus-: it can exist syntactically because we need a space operator for editing reasons (when you delete another operator you need to leave something behind), but it will always be in a hole.

sweetsinpackets: Holes will cause an extraction failure, so I think it's better to remove it because it will never be used.

sweetsinpackets: Oh, I want to preserve it now. The reason is the `extract` function should return a string or raise an exception, but it seems really strange to raise some exception because it's definitely not an exception. Meanwhile, since it can exists syntactically, I don't think it will cause problem.

adamsmd: I'm starting to think it should be an exception because it is a form that extraction does not support, just like extraction doesn't support holes.  Since this case should never occur, maybe a `failwith __LOC__` or some other exception that represents "impossible" cases in the code.  (Note that OCaml `(p1 p2)`, which is what it currently extracts to, does not have the same meaning as Hazel `(p1 p2)`.)

sweetsinpackets: I'm combining the `extract` and `update_pattern`, so here I can throw an exception of "invalid update". Meanwhile, can you explain what's the difference between OCaml `p1 p2` and Hazle `p1 p2`? I thought they're both apply.

sweetsinpackets: Done, now the `OCamlExtraction_Pat.extract` will directly throw an exception on `Ap` case.

cyrus-: Yeah it's fine to throw an exception for this. When we have algebraic datatypes we will need this form.

src/hazelcore/extractions/ocaml/OCamlExtraction.re
```diff
@@ -0,0 +1,65 @@
+type t =
+  | OCamlExp(string)
+  | ExtractionFailed(string);
+
+//TODO: test exception handling (seems good now)
+let extract = (ctx: Contexts.t, exp: DHExp.t): t =>
+  switch (OCamlExtraction_Exp.extract(ctx, exp)) {
+  // normal return value
+  | item =>
```
adamsmd: If you change `item` to `OCamlExp(exp_str, _exp_typ)`, then you don't need the following `switch`.

sweetsinpackets: Fixed.

adamsmd (approved):
Everything looks good.  There is a slight simplification that can be made at https://github.com/hazelgrove/hazel/pull/260/files#r487430068, but I am going to approve this anyway.

adamsmd:
Cyrus has asked for two more features.

First, we should have some tests for this.

Second, we should put the OCaml code in a popup instead of the console.  (The exception handling code does something like that which might be useful to look at.)

cyrus-:
marking as draft pending (1) tests and (2) getting it up to date with dev

cyrus-:
Nice work @sweetsinpackets, but closing because the syntax has changed substantially and not actively being developed.

PR #262, opened 2020-05-18
Invalid var

nj-wilson:
Implement InvalidVars, as a fall-through return of `TextShape.of_text`.
Refactors uses of `TextShape.of_text` as it is no longer an option type.

Marks expressions such as '1a' (or anything that doesn't match any current TextShape criteria) as an "Invalid Variable".

### DIFF
--- a/src/hazelcore/layout/DHDoc.re
+++ b/src/hazelcore/layout/DHDoc.re
@@ -127,6 +127,7 @@ module Pat = {
     | NonEmptyHole(_)
     | Wild
     | Keyword(_)
+    | InvalidVar(_)
     | Var(_)
     | IntLit(_)
     | FloatLit(_)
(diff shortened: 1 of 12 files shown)


cyrus-:
I think we'll need to distinguish invalid variables (starts with `a-zA-Z_`) from invalid number literals (starts with `0-9.`). Probably easiest to do this by creating a new InvalidText form in UHExp, rather than by putting error statuses on Var/IntLit/FloatLit (we don't even know if its going to be a int or float lit actually).

cyrus-:
@nj-wilson did you mean to delete this?

nj-wilson:
I figured that since I had renamed it away from InvalidVar, and as I needed to change emails on commits which would destroy history anyhow, I could just make a different branch. Unless there's a way to change which branch a pull request is associated with.

cyrus-:
gotcha.

PR #263, opened 2020-05-19
assert keyword + keyword refactoring

seanyeon:
Added assert to the code
created 4 new files for PatVar and ExpVar
Still working on adding Var cases for the dynamics


cyrus- (changes requested):
dynamics need work (see Slack messages) + needs `dev` merged

Issue #134, opened 2020-02-07
Use `uutf` instead of `Camomile` for Unicode/Utf-8.


adamsmd:

adamsmd:
When doing "make release" there turns out to be little difference in code side, though there is 1MB difference when running "make dev".  Thus, we are staying with Camomile.

PR #266, opened 2020-05-20
use `uutf` instead of `Camomile`

adamsmd:


adamsmd:
Fixes #134

adamsmd:
Need to measure difference in js_of_ocaml output, and need to measure performance difference on js_of_ocaml.

cyrus-:
I believe we decided to stick with Camomile after doing some measurements that showed a performance benefit and minimal code size impact when doing `make release`

PR #270, opened 2020-05-22
invalid text

### LINKED ISSUE #115: Allow number to be placed before a variable temporarily (hole?)

nj-wilson:
Implement InvalidText, as a fall-through return of TextShape.of_text.
Refactors uses of TextShape.of_text as it is no longer an option type.

Additionally add an InvalidText form to UHExp, UHPat.


nj-wilson:
Currently, have a couple issues:

If I have invalid text in the function position, it currently uses `Got inconsistent type ▶ assumed`. I am not sure if it's intended that it say something akin to `Got invalid text type ▶ assumed`, as well as not being quite sure how to work that into the current BinOp Space handling.

~~Can only type a single character to reach invalid text, and then it results in a failed action afterward. ie, I can type `3a` , but attempting to type past that is a failed action (of which I do not know what specific `Failed;` it is hitting). I am not sure if this is fine for expected behavior, or to allow typing past that.
I can, however, type something like `342234.2341`, and insert a letter midway, making it invalid text `3422a34.2341`. The same issue occurs with the letter added.~~
I figured out where this `Failed` is occurring. Question now is, if it's wanted to allow typing past the first character that causes invalid text (stop at `3a`, or allow `3asdf`).

cyrus-:
It shouldn't say inconsistent type, since InvalidText has hole type. It should say "Got invalid text > matched to" in the same way that free variables do. Check out the code for how free variables are handled.

We should support typing past the first character.

cyrus- (changes requested):
need to disable ligatures: https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures

cyrus-:
looks like unicode `⇒` is available in the font we use -- I tried changing UHDoc/DHDoc to use but it messes up cursor positioning. seems it is still being treated as 2 characters. is this an easy fix? @dmoon1221

cyrus-:
it's not actually treating it as 2 character, but it seems to be rendering wider than a normal character. is this an issue with the font?

nj-wilson:
I'm wondering if it's the font, as if I switch to DejaVu Sans Mono, it looks correct. I'm wondering if Fira Code is just missing that symbol specifically or it's just not rendering correctly.
In Firefox, if I check the computed styling of the span with the arrow, it shows it using "Segoe UI Symbol" instead of Fira Code (Chrome says it's still Fira Code, so I'm not sure).

cyrus-:
yeah that's probably the issue. I think turning ligatures on for delims is working fine though, so let's just go with that for now.

cyrus-:
merge conflicts need to be resolved

cyrus-:
need to implement splitting

cyrus-:
OK, seems to work. 

In `dev`, I can't press `.` to start constructing a float lit because the dot by itself is not valid. This should be possible now that we have invalid text. Can you add support for that to this branch, @nj-wilson?

nj-wilson:
I mistakenly set the commit message wrong, I mean to say that inputting a `.` will indeed work as a float, assuming it's properly formed after the `.` of course. A `.` on its own will be `InvalidText`

cyrus-:
We should treat InvalidText like other non-empty holes, so they should have a hole number and be part of the tab order.

cyrus-:
@nj-wilson looks like there are some merge conflicts with dev

nj-wilson:
I have fixed the conflicts, though it did point on something I'm not sure if I have correct.
In `holes_operand` in say `CursorPath_Exp`, I was pretty much just mimicking what `EmptyHole` was doing (though with the recent dev merge, I do have `is_empty = false`.

I also am not entirely sure on how to test if I have the hole numbering set up correctly. I was mostly just checking that behavior matched what other holes did.

src/hazelcore/CursorPath_Exp.re
```diff
@@ -649,7 +655,8 @@ and holes_zoperand =
     : CursorPath_common.zhole_list =>
   switch (zoperand) {
   | CursorE(OnOp(_), _) => CursorPath_common.no_holes
-  | CursorE(_, EmptyHole(u)) =>
+  | CursorE(_, EmptyHole(u))
+  | CursorE(_, InvalidText(u, _)) =>
     CursorPath_common.mk_zholes(
       ~hole_selected=
         Some({
```
cyrus-: is_empty should be false when its InvalidText (check everywhere)

src/hazelcore/CursorPath_Pat.re
```diff
@@ -207,7 +210,8 @@ and holes_zoperand =
     : CursorPath_common.zhole_list =>
   switch (zoperand) {
   | CursorP(OnOp(_), _) => CursorPath_common.no_holes
-  | CursorP(_, EmptyHole(u)) =>
+  | CursorP(_, EmptyHole(u))
+  | CursorP(_, InvalidText(u, _)) =>
     CursorPath_common.mk_zholes(
       ~hole_selected=
         Some({
```
cyrus-: see above

cyrus- (changes requested):

cyrus-:
hole numbering looks good to me

cyrus- (approved):

cyrus-:
@Yuning30 now that we have invalid text at the level of expressions and patterns, you should add support for the same in your type variable branch

PR #276, opened 2020-05-23
division operator for Int and Float types

### LINKED ISSUE #264: Different Zipper behavior for Plus/Minus Op and Times Op
Plus and Minus operators allow you to right-arrow into the operator while Times operator require you to backspace from the second argument in order to edit the operator.
Also, the spacing is different in the two cases.

image

image

Yunif3:
**Division Operator is introduced for Int and Float Types**

- `Int` type will have a `/` operator that behaves the same way as Ocaml's division operator
- `/` will throw an `divide by zero error` as an `InvalidOperation`, a generalized way of throwing run-time errors that was made with @annie-anna 
- `Float` type will have a `/.` operator that behaves the same way it does in Ocaml.
- `NaN` and `Inf` constants were introduced for `Float` type, and all other variations of it from Ocaml were disabled to be treated as a variable
- `NegInf` constant was introduced for `Float` type temporarily until unary minus is introduced
- Variables can now start with an uppercase letter
- Fixed issue #264

### DIFF
--- a/src/hazelcore/Operators.re
+++ b/src/hazelcore/Operators.re
@@ -55,6 +55,7 @@ module Exp = {
     | FPlus
     | FMinus
     | FTimes
+    | FDivide
     | LessThan
     | GreaterThan
     | Equals
@@ -76,6 +77,7 @@ module Exp = {
     | FPlus => "+."
     | FMinus => "-."
     | FTimes => "*."
+    | FDivide => "/."
     | LessThan => "<"
     | GreaterThan => ">"
     | Equals => "=="
--- a/src/hazelcore/Operators.re
+++ b/src/hazelcore/Operators.re
@@ -51,6 +51,7 @@ module Exp = {
     | Plus
     | Minus
     | Times
+    | Divide
     | FPlus
     | FMinus
     | FTimes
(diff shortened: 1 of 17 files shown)


cyrus-:
let's come up with a unified way of indicating run-time errors on primitive operations. @annie-anna has a nice UI for string subscripts out of bounds in the `strings` branch. why don't you two chat about coming up with a more general mechanism, e.g. a new `InvalidOperation` form in `DHExp` that takes an error message and the erroneous expression as an argument and displays the error message.

cyrus-:
also I think variables starting with an uppercase letter is going to cause problems once we have labeled sum types, because it won't be clear whether a variable in a pattern is a reference to a constructor or a variable to be bound. Let's change the syntax for NaN and Inf to be `nan` and `inf` now to avoid headaches in the future.

cyrus-:
or actually since NaN and Inf aren't variables, we can keep them uppercased and just treat them as "constructors" for the Float type. for now we can allow uppercased variables and just deal with constructors later. so scratch last comment.

cyrus-:
sorry, my comments may have been confusing -- lets go back to using capitalized `Inf` and `NaN` for those float values

cyrus-:
`0.0 -. Inf` evaluates to `Inf`. It should be negative infinity. For now, let's just define NegInf constants.

cyrus-:
I can't type `NegInf` -- it stops at `NegIn`

cyrus- (approved):
Looks good!

PR #287, opened 2020-05-26
assert keyword + keyword refactoring

seanyeon:


cyrus-:
This should be a failed assertion, not false...
Screenshot from 2020-05-27 00-10-08

cyrus-:
we need to think about how to report failed assertions that are on ExpLines -- right now they are just skipped even if they fail.

src/hazelcore/dynamics/Dynamics.re
```diff
@@ -720,6 +722,8 @@ module Exp = {
     | BinBoolOp(_, _, _)
     | BinIntOp(_, _, _)
     | BinFloatOp(_, _, _)
+    | AssertLit => Indet
+    | FailedAssert => Indet
```
dm0n3y, line 726: FYI this is a bug -- notice that you've changed the behavior of all the cases above the new ones. Fixing in my merge.

dm0n3y: I think this should say `AssertLit` is `Indet` but `FailedAssert` is the same as `Triv`, since any non-indet assert evaluates to unit. @cyrus- can you confirm?

seanyeon: Sorry this is not the right version the up to date pull request is Assert numbering ui. This one just have the very simple assert implementation. Assert numbering ui has the feature about assert mapping ad UI implementation.

src/hazelcore/dynamics/Dynamics.re
```diff
@@ -2115,6 +2142,8 @@ module Evaluator = {
     | IntLit(_)
     | FloatLit(_)
     | Triv => BoxedValue(d)
+    | FailedAssert => Indet(FailedAssert)
```
dm0n3y, line 2145: Actually this seems to suggest that failed assertions evaluate to indet rather than unit, which would change my suggestion above. Is this correct?

cyrus-: failed assertions have unit type and are indet. only successful assertions evaluate to triv.

src/hazelcore/dynamics/Dynamics.re
dm0n3y: doh ok

PR #290, opened 2020-05-27
changed invalidOperation to have a operation and an error message

Yunif3:
Worked with @annie-anna to restructure `InvalidOperation` to have an operation and an error message


src/hazelcore/semantics/ErrStatus.re
```diff
@@ -6,6 +6,16 @@ module HoleReason = {
     | WrongLength;
 };
 
+module RunTime = {
```
cyrus-: don't make it a submodule of ErrStatus, since ErrStatus is about an entirely different kind of error. just create a InvalidOperationError module in the dynamics directory

cyrus- (changes requested):
looks good, just one minor change needed.

cyrus- (approved):
Great!

PR #303, opened 2020-06-01
Pull caret out of code html

hannahpotter:
Extract the caret out of being nested within the code elements. 
This is part of moving the cursor inspector to the cursor to avoid issues with z-indexes and stacking contexts.

Also fixes an issue where the font metrics were not being updated correctly.


src/hazelweb/Hazel.re
```diff
@@ -8,23 +8,35 @@
...
-        row_height: rect##.bottom -. rect##.top,
-        col_width: rect##.right -. rect##.left,
-      }),
+  // Necessary to correctly recalculate font metrics if delay in loading font or resize of window
+  let _ =
+    ResizeObserver.observe(
+      ~node=JSUtil.force_get_elem_by_id("font-specimen"),
+      ~f=
+        (entries, _) => {
+          let list = Array.to_list(Js_of_ocaml.Js.to_array(entries));
```
dm0n3y: You can pattern match directly on arrays

src/hazelweb/gui/Cell.re
```diff
@@ -56,6 +56,22 @@ let focus = () => {
...
 
+let caret_from_pos = (x: float, y: float): Vdom.Node.t => {
+  let pos_attr =
+    Vdom.Attr.create(
+      "style",
+      "left: "
+      ++ string_of_float(x)
+      ++ "0px; top: "
+      ++ string_of_float(y)
+      ++ "0px",
```
dm0n3y: Could you see if you can get this working using [`Css_gen`](https://ocaml.janestreet.com/ocaml-core/latest/doc/virtual_dom/Css_gen/index.html)?

src/hazelweb/gui/Cell.re
```diff
@@ -56,6 +56,22 @@ let focus = () => {
   JSUtil.force_get_elem_by_id("cell")##focus;
 };
 
+let caret_from_pos = (x: float, y: float): Vdom.Node.t => {
```
dm0n3y: Let's move this caret positioning logic into `UHCode` (where `UHCode.view` additionally take the `(row, col)` as an argument), reason being we've done that in the `livelits` branch where you have nested splices and cursor maps inside of embedded guis so eventually it will need to move there anyway.

src/hazelweb/gui/UHCode.re
```diff
@@ -86,25 +73,14 @@ let view =
 
         | Annot(Step(_) | EmptyLine | SpaceOp, l) => go(l)
 
-        | Annot(Token({shape, len, has_cursor}), l) => {
```
dm0n3y, line 89: We can now remove these fields from `Token` and the `find_and_decorate_caret` function from `UHLayout`.

src/hazelweb/Hazel.re
```diff
@@ -8,23 +8,35 @@
...
 let on_startup = (~schedule_action, _) => {
-  let update_font_metrics = () => {
-    let rect =
-      JSUtil.force_get_elem_by_id("font-specimen")##getBoundingClientRect;
-    schedule_action(
-      Update.Action.UpdateFontMetrics({
-        row_height: rect##.bottom -. rect##.top,
-        col_width: rect##.right -. rect##.left,
-      }),
+  // Necessary to correctly recalculate font metrics if delay in loading font or resize of window
```
dm0n3y: I think this is the right way to do things so no need to explain why we don't do the wrong way.

dm0n3y (changes requested):
Looks good overall, few minor comments

src/hazelweb/gui/UHCode.re
hannahpotter: CursorMap uses the has_cursor and len fields.

dm0n3y: oh yeah, never mind

hannahpotter:
Should the old Font file and its license be removed since it is not being used anymore?

dm0n3y (approved):
LGTM once old font is removed

PR #304, opened 2020-06-02
Type variable 2

Yuning30:
This is the new pull request for type variable, the old type vraiable branch is not used due to conflict with `dev` branch


Yuning30:
Currently the kind information is displayed via special message bar together with the string "got type", and I'm planning to add a type bar. And you can review the code for other semantics.

cyrus-:
OK, I'll take a look soon.

cyrus- (changes requested):
changes requested verbally

Yuning30:
#### Changes since last review:
- Now the context inspector and cursor inspector is able to display the red dotted box around the free type variable
- The error message is change to "got a free type variable"
- The kind related information is removed from the cursor inspector until we have more useful kinds

cyrus-:
merge conflict needs to be resolved

cyrus-:
valid base types (Int, Float, etc.) should display in bold like they do on dev

src/hazelcore/Contexts.re
```diff
@@ -1,19 +1,27 @@
 [@deriving sexp]
-type t = (VarCtx.t, PaletteCtx.t);
-let empty = (VarCtx.empty, PaletteCtx.empty);
+type t = {
+  vars: VarCtx.t,
+  tyvars: TyVarCtx.t,
+  palettes: PaletteCtx.t,
+};
+let empty = {
```
cyrus-, line 8: leave empty lines between definitions

src/hazelcore/CursorInfo.re
```diff
@@ -95,7 +95,9 @@ type typed =
   | PatSynthesized(HTyp.t)
   | PatSynKeyword(ExpandingKeyword.t)
   /* cursor in type position */
-  | OnType
+  | TypKeyword(ExpandingKeyword.t)
```
cyrus-: what is this for? we don't have any expanding keywords at the type level right?

cyrus- (changes requested):

src/hazelcore/CursorInfo.re
Yuning30: I originally thought `let` and `case` are reserved for both expression position and type position, now I'll delete them.

cyrus-: oh I guess we should probably reserve those everywhere for the sake of extraction. you can leave it.

Yuning30:
The Tracis-ci seems to fail to update correctly. It's always pending state in the pull request page. But when I click into that page, it says the build is successful.

src/hazelcore/UHTyp.re
```diff
@@ -95,35 +99,41 @@ let contract = (ty: HTyp.t): t => {
   ty |> contract_to_seq |> OpSeq.mk(~associate);
 };
 
-let rec expand = (ty: t): HTyp.t => expand_opseq(ty)
-and expand_opseq =
-  fun
-  | OpSeq(skel, seq) => expand_skel(skel, seq)
-and expand_skel = (skel, seq) =>
+let rec expand = (ty: t, ctx: TyVarCtx.t): HTyp.t => expand_opseq(ty, ctx)
```
cyrus-: make the ctx the first argument here, to be consistent with the judgement form

src/hazelcore/UHTyp.re
```diff
@@ -149,3 +161,10 @@ and is_complete = (ty: t) => {
   | OpSeq(sk, sq) => is_complete_skel(sk, sq)
   };
 };
+
+let of_string =
```
cyrus-: this should be called "to_string" not "of_string"

src/hazelcore/UHTyp.rei
```diff
@@ -28,17 +29,19 @@ let mk_OpSeq: OpSeq.seq(operand, operator) => OpSeq.t(operand, operator);
 
 let contract: HTyp.t => t;
 
-let expand: t => HTyp.t;
+let expand: (t, TyVarCtx.t) => HTyp.t;
```
cyrus-: put contexts first in all of these

src/hazelcore/ZTyp.re
```diff
@@ -74,10 +76,11 @@ and is_after_zopseq = zopseq => ZOpSeq.is_after(~is_after_zoperand, zopseq)
 and is_after_zoperand =
   fun
   | CursorT(cursor, Hole)
-  | CursorT(cursor, Unit)
-  | CursorT(cursor, Int)
-  | CursorT(cursor, Float)
-  | CursorT(cursor, Bool) => cursor == OnDelim(0, After)
+  | CursorT(cursor, Unit) => cursor == OnDelim(0, After)
+  | CursorT(cursor, Int) => cursor == OnText(3)
```
cyrus-: instead of magic numbers here, define constants determined by calling length on the string representation of each type (based on to_string)

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -20,13 +20,15 @@ let grounded_List = NotGroundOrHole(List(Hole));
...
+  | Hole
+  | TyVarHole(_, _) => Hole
   | Bool
   | Int
   | Float
   | Arrow(Hole, Hole)
   | Sum(Hole, Hole)
-  | List(Hole) => Ground
+  | List(Hole)
+  | TyVar(_, _) => Ground
```
cyrus-: TyVar shouldn't be ground, it depends on what it stands for. leave it as a failwith("Unimplemented") for now, since we don't have bound type variables yet.

src/hazelcore/layout/UHAnnot.re
```diff
@@ -9,6 +9,7 @@ type term_data = {
 
 [@deriving sexp]
 type token_shape =
+  | BoldText
```
cyrus-: call this Keyword rather than BoldText

src/hazelweb/www/style.css
```diff
@@ -814,6 +815,9 @@ h1 {
 .code-delim {
   font-weight: bold;
 }
+.code-boldtext {
```
cyrus-: rename to keyword everywhere

src/hazelcore/TyId.re
```diff
@@ -0,0 +1,18 @@
+open Sexplib.Std;
```
cyrus-, line 1: include a .rei file for every .re file you add

src/hazelcore/Statics_Typ.re
```diff
@@ -0,0 +1,19 @@
+let rec syn = (ctx: Contexts.t, ty: HTyp.t): Kind.t =>
+  switch (ty) {
+  | TyVar(idx, _) =>
+    let (_, k) = List.nth(ctx.tyvars, idx);
```
cyrus-: Define this as a function TyVarCtx, rather than assuming it will always be a list

src/hazelcore/Statics_Typ.re
```diff
@@ -0,0 +1,19 @@
...
+    k;
+  | TyVarHole(_, _)
+  | Hole => KHole
+  | Int
+  | Float
+  | Bool
+  | Prod([]) => Type
+  | Arrow(ty1, ty2)
+  | Sum(ty1, ty2) =>
+    ana(ctx, ty1, Kind.Type) && ana(ctx, ty2, Kind.Type) ? Type : KHole
```
cyrus-: if analysis fails on either argument, then synthesis should fail, not return KHole. need to change return type to option.

src/hazelcore/Statics_Typ.re
```diff
@@ -0,0 +1,19 @@
...
+  | Prod([x, ...xs]) => ana(ctx, x, Kind.Type) ? syn(ctx, Prod(xs)) : KHole
```
cyrus-: need to check each element against Type, not synthesize. and it shouldn't return KHole, it should return Type if this succeeds

src/hazelcore/Statics_Typ.re
```diff
@@ -0,0 +1,19 @@
...
+  | List(t) => ana(ctx, t, Kind.Type) ? Type : KHole
```
cyrus-: return None if it fails

src/hazelcore/HTyp.re
```diff
@@ -42,13 +49,24 @@ let precedence = (ty: t): int =>
 /* equality
    At the moment, this coincides with default equality,
    but this will change when polymorphic types are implemented */
-let eq = (==);
+let eq = (x, y) =>
+  switch (x, y) {
+  | (TyVarHole(_), TyVarHole(_))
```
cyrus-: we shouldn't call all TyVarHoles equal to each other or to Holes. they should only be equal if they are the same variable.

src/hazelcore/Action_Typ.re
```diff
@@ -24,6 +24,27 @@ let shape_of_operator = (op: UHTyp.operator): Action_common.operator_shape =>
   | Sum => SVBar
```
cyrus-, line 24: haven't reviewed this file yet, will take a look in next review

cyrus- (changes requested):
- see comments
- need to add support for splitting, similar to expression and pattern variables (e.g. if I am at `a|b` and type `>`, it should become `a -> b`)

Yuning30:
Thanks for the feedback. I'll fix these issues ASAP.

cyrus-:
I get this failure when trying to enter most forms into hole 5 (e.g. number literals, lambdas)...
Screenshot from 2020-06-23 15-13-58

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -20,13 +20,15 @@ let grounded_List = NotGroundOrHole(List(Hole));
...
-  | Hole => Hole
+  | Hole
+  | TyVarHole(_, _) => Hole
   | Bool
   | Int
   | Float
   | Arrow(Hole, Hole)
   | Sum(Hole, Hole)
   | List(Hole) => Ground
+  | TyVar(_, _) => failwith("Unimplemented")
```
bkase, line 31: remember to add context for calls to a function

cyrus-:
Superceded by #478

PR #306, opened 2020-06-03
Split Dynamics.re

annie-anna:
Split Dynamics.re into Expansion.re and Evaluation.re


src/hazelcore/dynamics/Evaluation.re
```diff
@@ -0,0 +1,370 @@
+open Sexplib.Std /* hole instance numbers are all 0 after expansion and during evaluation -- * renumbering is done on the final result (see below) */;
+
+module Evaluator = {
```
cyrus-: we don't need a submodule here, just call this file Evaluator.re

cyrus- (changes requested):

dm0n3y:
One is called `Expansion`, the other is called `Evaluator`. Should we standardize on a noun form, i.e., rename one to `Expander` or other to `Evaluation`? I think it would also make sense to split `Statics` into `Typechecking`/`Typechecker` and `HoleFixing`/`HoleFixer` (not a suggestion to add to this PR, just considering guiding theme)

cyrus-:
let's call it "Expander" and "Evaluator".

regarding hole fixing being its own module, I think it makes more sense as part of Statics -- it behaves the same as the type checker, except it outputs a hole fixed version of the program (in other words it marks the errors, the way a "production" type checker would). if you want to propose it still, go ahead and make a separate issue and we can discuss there.

cyrus-:
actually, while we're at it, let's rename "Expander" to "Elaborator" and "expand" to "elab" in prep for eventually merging in livelits. also its called elaboration in the POPL 19 paper anyway.

annie-anna:
Do I need to rename ``ExpandingKeyword.re`` to ``ElaboratingKeyword.re`` and ``expand`` to ``elab`` in ``PaletteDefinition.re``, too?

dm0n3y:
No, those should stay the same

annie-anna:
Should ``ExpandResult``, ``DoesNotExpand``, ``Expands``, ``UHTyp.expand`` still be the same?

dm0n3y:
`UHTyp.expand` should remain the same, the others should be renamed to `ElaborationResult`, `DoesNotElaborate`, `Elaborates`

cyrus- (approved):

PR #314, opened 2020-06-07
Side bars have a lot of content, don't compute it when not shown

wlitwin:
Both side panes do a lot of work to update their GUI state. When the side pane is hidden all this work is for nothing and can slow down the editor unnecessarily. To fix this, the contents of the side pane are generated via a thunk that's only called when the side pane will be visible.


cyrus-:
need `dev` merged into this

PR #315, opened 2020-06-08
Labeled tuples

ecdeuts:


cyrus- (commented):
looks good so far -- discussed verbally

ecdeuts:
Finished up statics and dynamics for labeled tuples types/expressions, will work on patterns and projections later in the project.  As I was looking further at the code, I need to work more on the core code for actions and cursors before moving onto the front end.  I would appreciate feedback on the statics/dynamics code though.

cyrus-:
Nice work @ecdeuts, but closing due to haz3l having a different approach to parsing that will likely allow us to approach the design differently.

PR #317, opened 2020-06-08
Recurse into rules when renumbering holes

utkmehta:
There was a bug in the code for renumberingHoles for complete expressions. This PR fixes that bug by recursing into rules for case expressions to make sure the expression is complete.

Test: case [] 
| _ => _, _, _
Result:
The holes are not automatically renumbered.


cyrus- (approved):
Looks good, thanks!

PR #319, opened 2020-06-09
Fix cursor positioning around empty holes

### LINKED ISSUE #313: cursor positioning on opseqs is incorrect
Screenshot from 2020-06-07 13-19-29

this is on Linux -- seems like something may be off with the latest font metrics changes?

hannahpotter:
Don't round pixels for cursor positioning and padding around empty holes.
fixes #313 

@dmoon1221 Using Css_gen only allows ints for pixels and the rounding is causing issues https://ocaml.janestreet.com/ocaml-core/latest/doc/virtual_dom/Css_gen/Length/index.html. 
Potentially we could still use Css_gen if we convert to pt I think, but I believe you would need to grab the dpi to do this correctly.

### DIFF
--- a/src/hazelweb/gui/UHCode.re
+++ b/src/hazelweb/gui/UHCode.re
@@ -117,20 +117,33 @@ let view =
             let full_space = font_width *. float_of_int(len);
             let shrunk_space = full_space *. font_shrink;
             let per_side_padding = (full_space -. shrunk_space) /. 2.0;
-            let padding =
-              Css_gen.padding(
-                ~left=`Px(int_of_float(per_side_padding)),
-                ~right=`Px(int_of_float(per_side_padding)),
-                (),
-              );
-            let font_size =
-              Css_gen.font_size(
-                `Percent(
-                  Core_kernel.Percent.of_percentage(font_shrink *. 100.0),
-                ),
-              );
+            /*let padding =
+                Css_gen.padding(
+                  ~left=`Px(int_of_float(per_side_padding)),
+                  ~right=`Px(int_of_float(per_side_padding)),
+                  (),
+                );
+              let font_size =
+                Css_gen.font_size(
+                  `Percent(
+                    Core_kernel.Percent.of_percentage(font_shrink *. 100.0),
+                  ),
+                );
+              let styling =
+                Vdom.Attr.style(Css_gen.combine(padding, font_size));*/
             let styling =
-              Vdom.Attr.style(Css_gen.combine(padding, font_size));
+              Vdom.Attr.create(
+                "style",
+                "padding-right: "
+                ++ string_of_float(per_side_padding)
+                ++ "0px; "
+                ++ "padding-left: "
+                ++ string_of_float(per_side_padding)
+                ++ "0px; "
+                ++ "font-size: "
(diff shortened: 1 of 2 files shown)


src/hazelweb/gui/UHCode.re
```diff
@@ -52,11 +52,14 @@ let has_child_clss = (has_child: bool) =>
...
 let caret_from_pos = (x: float, y: float): Vdom.Node.t => {
   let pos_attr =
-    Vdom.Attr.style(
-      Css_gen.combine(
-        Css_gen.left(`Px(int_of_float(x))),
-        Css_gen.top(`Px(int_of_float(y))),
-      ),
+    Vdom.Attr.create(
+      "style",
+      "left: "
```
cyrus-: why aren't you using Css_gen here?

src/hazelweb/gui/UHCode.re
```diff
@@ -117,21 +120,30 @@ let view =
...
+                    Core_kernel.Percent.of_percentage(font_shrink *. 100.0),
+                  ),
                 ),
               );
-            let styling =
-              Vdom.Attr.style(Css_gen.combine(padding, font_size));
-            [Node.span([styling, Attr.classes(["HoleLabel"])], go(l))];
+            let padding =
+              Vdom.Attr.create(
+                "style",
```
cyrus-: same question

src/hazelweb/gui/UHCode.re
hannahpotter: Css_gen only allows px to be ints https://ocaml.janestreet.com/ocaml-core/latest/doc/virtual_dom/Css_gen/Length/index.html, and that rounding is what is causing the issues you were seeing. 
We could maybe convert to pt measurements, but I am pretty sure you would need the dpi.

dm0n3y: What happens if you do proper rounding using `int_of_float(Float.round(x))`? `int_of_float` always truncates

cyrus-: different browsers seem to implement decimal pixel values slightly differently in terms of how they round those to actual pixels. rounding ourselves may at least make behavior consistent between browsers?

cyrus-: can we set the font-size in a way that character width is always an integer?

cyrus-: "Setting the font size in pixel values (px) is a good choice when you need pixel accuracy. A px value is static. This is an OS-independent and cross-browser way of literally telling the browsers to render the letters at exactly the number of pixels in height that you specified."

dm0n3y: The other thing you'd want to tweak is the font size ratio (w.r.t. to the font size) so that both reduced font and surrounding padding can be whole numbers of pixels. If we do this, then let's also fix the issue of potential constant skew of font size ratio between `UHCode.re` and `style.css` (forgot to check for this in original PR).

hannahpotter: The code is just setting the width of the HoleLabel and centering the number inside instead of trying to figure out padding with rounding numbers.

dm0n3y: Good idea 👍

cyrus- (approved):

PR #320, opened 2020-06-10
Use ch instead of px to calculate size of hole labels

hannahpotter:


src/hazelweb/gui/UHCode.re
```diff
@@ -112,11 +112,16 @@ let view =
           ]
 
         | Annot(HoleLabel({len}), l) => {
-            let font_width = font_metrics.col_width;
-            let full_space = font_width *. float_of_int(len);
-            let width =
-              Css_gen.width(`Px(int_of_float(Float.round(full_space))));
-            let styling = Vdom.Attr.style(width);
+            let font_shrink = 0.65;
+            let full_space = 1.0 /. font_shrink *. float_of_int(len);
```
hannahpotter: Need to offset the scaling done to the hole numbers (since ch is based on the size of characters in the element)

cyrus-: can you have a sub-span for the hole numbers that sets the font lower so that the width is calculated based on the parent font size?

hannahpotter: You are correct. I had tried that, but didn't have the set up quite right, but now I have it fixed and updated.

cyrus- (approved):

PR #325, opened 2020-06-10
Optimize hole fixing

dm0n3y:
Ensure pointer stability when hole fixing pass does nothing in order to maximize cache hits in doc memoization table.


dm0n3y:
I recommend hiding whitespace-only changes for this one

src/hazelcore/Statics.re
```diff
@@ -1901,364 +2273,523 @@ module Exp = {
...
+                  opseq |> UHExp.set_err_status_opseq(InHole(WrongLength, u)),
+                  u_gen,
+                );
+              }
+            };
+          }: (
+            UHExp.opseq,
+            MetaVarGen.t,
+          )
+          // handle n-tuples
```
cyrus-: fixme

cyrus- (changes requested):
- add comment about use of Lazy.force

PR #341, opened 2020-06-17
Use standard `Map.Make(Int)` for `MetaVarMap`

### LINKED ISSUE #148: Use OCaml's finite maps instead of hand-rolled?

dm0n3y:


src/hazelcore/CursorPath_Exp.re
```diff
@@ -366,11 +366,10 @@ and of_steps_operand =
       }
     | ApPalette(_, _, _, splice_info) =>
       let splice_map = splice_info.splice_map;
-      switch (NatMap.drop(splice_map, x)) {
+      switch (IntMap.find_opt(x, splice_map)) {
```
cyrus-, line 369: don't need to do this here, but in the `livelits` branch you should define a module `SpliceMap` and use its functions, rather than assuming it will be implemented with `IntMap`

cyrus- (approved):

PR #343, opened 2020-06-17
Move mouse_position state to `Model` rather than as a global ref

### LINKED ISSUE #308: move mouse_position state into State.t rather than as a global ref

PPER:
mouse_position state should be moved to `State` in #308, but `state` is not being provided to view in incr_dom api, so I move it to `Model`.


src/hazelweb/model/Model.rei
```diff
@@ -45,6 +54,7 @@ type t = {
   right_sidebar_open: bool,
   font_metrics: FontMetrics.t,
   is_mac: bool,
+  mouse_position: ref(coordinate),
```
cyrus-: leave a comment here saying why we need to track this

src/hazelweb/util/JSUtil.re
```diff
@@ -235,22 +235,18 @@ let force_get_elem_by_cls = cls =>
...
-    };
+let get_mouse_position = evt => {
+  (
+    Js.Optdef.get(evt##.pageX, () => assert(false)),
+    Js.Optdef.get(evt##.pageY, () => assert(false)),
+  );
 };
 
-let unsafe_get_elt_under_mouse = () => {
+let unsafe_get_elt_by_coordinate = (x: int, y: int) => {
```
cyrus-: you can just name this elementFromPoint, its not unsafe since it will always work

cyrus- (changes requested):

src/hazelweb/model/Model.re
```diff
@@ -18,6 +18,11 @@ type measurements = {
   update_apply_action: bool,
 };
 
+type coordinate = {
```
cyrus-: turn this into `MousePosition.t`

src/hazelweb/util/JSUtil.re
```diff
@@ -235,22 +235,18 @@ let force_get_elem_by_cls = cls =>
...
-    };
+let get_mouse_position = evt => {
+  (
+    Js.Optdef.get(evt##.pageX, () => assert(false)),
+    Js.Optdef.get(evt##.pageY, () => assert(false)),
+  );
 };
 
-let unsafe_get_elt_under_mouse = () => {
+let element_from_point = (x: int, y: int) => {
```
cyrus-: instead of taking x and y separately, have it take an argument of type `MousePosition.t` (see above)

cyrus- (changes requested):

src/hazelweb/gui/UndoHistoryPanel.re
```diff
@@ -1082,8 +1082,10 @@ let view = (~inject: Update.Action.t => Vdom.Event.t, model: Model.t) => {
       )
     );
 
-  let get_elt_id_under_mouse = _: option((int, int)) => {
-    let elt: Js.t(Dom_html.divElement) = JSUtil.unsafe_get_elt_under_mouse();
+  /* return option((group_id, elt_id)) */
```
cyrus-, line 1085: going to approve this one, but for your next clean up task can you define the types group_id and elt_id explicitly and then use them everywhere, instead of int?

cyrus- (approved):

PR #347, opened 2020-06-17
Memoize vdom generation

dm0n3y:
Reduces time of `UHCode.view` on edits to qsort100 from 200-300 ms down to 3-6 ms


dm0n3y:
Still need to deal with cursor map, so interaction with qsort100 is still slow overall. `Hazel.create` takes 50-100ms on edits to qsort100. I also think there may be some other bottlenecks we're not accounting for such as vdom diffing.

dm0n3y:
Closing because I've discovered this approach is not quite enough for decorations

PR #348, opened 2020-06-18
Sidebar help icon

wlitwin:
To make it more obvious what the left side bar contains a standard
question mark with a circle is used as the icon instead of a generic
arrow.

Selection_004
Selection_005


cyrus- (changes requested):
can we center it a little better vertically? on my screen it is sitting almost flush with the bottom of the viewport, but has padding above it, so it looks off-center

wlitwin:
Does this padding look better? It should be more centered now.

cyrus- (approved):
looks good, thanks!

PR #360, opened 2020-06-22
More signatures for hazelcore

Yuning30:


dm0n3y (changes requested):
You should review which functions are actually used outside of a module and remove from signature if not.

Yuning30:
A problem I met is if a function is defined but not used inside the module itself and the function name does not start with an underscore, then the compiler will complain that function as unused value. For example, `shape_of_operator` and `keyword_suffix_to_opseq` in `Action_Exp.re` is not used both inside `Action_Exp` and out side.

What should I do for such functions?

cyrus-:
That might indicate that it's old code that we can remove. In other cases, there are functions that are only there for symmetry with other functions -- e.g. shape_of_operator is just the inverse of operator_of_shape, and helps make sure there is a shape for each operator when the language is extended. In those cases, I think it's fine to add that to the signature.

Yuning30:
There are functions that are polymorphic in some module but we know what type will be used when the function will be used. For example, the signature of `ZTyp.erase_operator` generated by `ocaml-print-intf` is `(('a, 'b)) => 'b`, but we know `'a` should be `CursorPosition.t` and `'b` should be `ZTyp.zoperator`.

Should I leave the signature as it's generated by the `ocaml-print-intf` or apply the type manually?

cyrus-:
@Yuning30 you can apply the more specific type manually in situations like that.

cyrus-:
@Yuning30 what is the status of this PR?

cyrus-:
This is out of date at this point, closing.

PR #361, opened 2020-06-23
Shunting-yard opp

### LINKED ISSUE #153: Simplify "Skel" parsers
- Use shunting yard algorithm?
- Good starter project?
- Move into own folder?

claban3:
Added an operator precedence parser in preparation for adding user defined operators. The current implementation handles only infix binary operators, but allows for creating custom precedence and associativity. 

Note to self - I have some simple unit tests that exist in the commit history that might be useful to fish out when we have a test framework agreed on and set up.


src/hazelcore/Operators_Exp.re
```diff
@@ -55,3 +55,34 @@ let is_Space =
   fun
   | Space => true
   | _ => false;
+
+let precedence =
```
cyrus-, line 59: you should reference the OCaml syntax definition to make sure we are giving these things the right precedence. I don't think Comma, Cons and Space should all be 0 for example...

src/hazelcore/UHExp.re
```diff
@@ -305,10 +305,75 @@ let text_operand =
     );
   };
 
-let associate = (seq: seq) => {
-  let skel_str = Skel.mk_skel_str(seq, Operators_Exp.to_parse_string);
-  let lexbuf = Lexing.from_string(skel_str);
-  SkelExprParser.skel_expr(SkelExprLexer.read, lexbuf);
+let associate = (seq: seq): skel => {
```
cyrus-: can you leave a comment pointing to the reference material that you used for implementing this? hard to understand what is going on here without that...

cyrus- (changes requested):

src/hazelcore/Operators_Exp.re
claban3: Ok. I was confused about these 3 operators, I'll take a look at what OCaml does.

src/hazelcore/UHExp.re
claban3: Yes I'll leave a ref to the wikipedia page, and add some comments explaining what each component does.

src/hazelcore/Operators_Exp.re
cyrus-: OK, you can also look at the current skel expr parser. the order that the operator declarations appears is the precedence order.

claban3: Oh ok that's very helpful.

src/hazelcore/UHExp.re
claban3: @cyrus- Please take a second look.

src/hazelcore/UHExp.re
```diff
@@ -305,10 +305,119 @@ let text_operand =
...
+      : (list(skel), list(Operators_Exp.t)) => {
+    /* Move operators from the operator stack to the output skel list while... */
+    switch (op_stack) {
+    | [] => (skels, op_stack) /* (1) The operator stack is not empty. */
+    | [op, ...op_stack'] =>
+      if (should_mv(op)) {
+        /**  (2) The operator on the top of the stack has greater precedence than the current.
+         *       Note - This impl supports only binary operators.
+         */
+        let rand_2 = List.hd(skels);
```
cyrus-: use pattern matching, not List.hd and List.tl

src/hazelcore/UHExp.re
```diff
@@ -305,10 +305,119 @@ let text_operand =
...
+  }
+  and build_ast_while =
+      (skels: list(skel), op_stack: list(Operators_Exp.t), should_mv)
+      : (list(skel), list(Operators_Exp.t)) => {
+    /* Move operators from the operator stack to the output skel list while... */
+    switch (op_stack) {
+    | [] => (skels, op_stack) /* (1) The operator stack is not empty. */
+    | [op, ...op_stack'] =>
+      if (should_mv(op)) {
+        /**  (2) The operator on the top of the stack has greater precedence than the current.
```
cyrus-: don't you also have to consider the situation where the operator at the top of the operator stack has equal precedence and the token is left associative?

cyrus- (changes requested):
see comments -- need to add tests to really feel confident that it works correctly. also would be good to measure performance relative to what we had before.

src/hazelcore/UHExp.re
claban3: I believe I am doing that inside the [should_mv predicate](https://github.com/hazelgrove/hazel/pull/361/files/f50d0d84ba032b8fda1cd59b875351e781d7aa9a#diff-07032a218ad58ee406b8c5f9f3b5cb41R349-R361) that I pass into this function.

src/hazelcore/UHExp.re
```diff
@@ -305,10 +305,119 @@ let text_operand =
...
+        lex_addr: int,
+      )
+      : skel => {
+    switch (affix) {
+    | A(op, seq) =>
+      /**
+       * If the next token is an operator, pop operators from the operator stack,
+       * then push this operator to the operator stack.
+       */
+      let should_mv = op' =>
```
claban3, line 349: @cyrus- should_mv predicate

src/hazelcore/UHExp.re
```diff
@@ -305,10 +305,119 @@ let text_operand =
...
+        /**
+         * Continue popping operators while the precedence of the top of the operator
+         * stack has greater precedence than the current operator.
+         */
+        {
+          switch (Operators_Exp.associativity(op')) {
+          | Left =>
+            Operators_Exp.precedence(op) <= Operators_Exp.precedence(op')
```
claban3: @cyrus- This logic. Also sounds fair, I'll setup some tests and look into perf.

cyrus-: ah gotcha. maybe mention explicitly what's going on here in comments.

claban3:
I realized that the testing PR was very small, and that those tests were really only relevant to this code, so I merged that branch here and combined the work. 

In the test branch I:
- Created a home for a unit testing suite in src/hazelcore/test composed of ppx_expect/inline_test tests.
- Includes some make commands to run/fix tests.
- Includes small test suite for SkelParser and some example tests.

@cyrus- Please let me know if there are testing scenarios that might also need to be covered.

src/hazelcore/test/SampleTests.re
```diff
@@ -0,0 +1,19 @@
...
+//
+// More at https://github.com/janestreet/ppx_expect
+let%expect_test "expect_test" = {
+  print_endline("hazel");
+  %expect
+  {|hazel|};
+};
+
+// To create an inline test, write a function () -> bool
+// who's result is the assertion you wish to test.
```
dm0n3y: *whose

src/hazelcore/test/SampleTests.re
```diff
@@ -0,0 +1,19 @@
+// To create an expectation test, print to the console,
```
dm0n3y, line 1: 👍

README.md
```diff
@@ -102,3 +102,9 @@ The `make dev` and `make release` commands do three things:
 You can print to the browser console using the standard `print_endline` function. This is probably the easiest method right now.
 
 `js_of_ocaml` does support source maps and has some other flags that might be useful. If you experiment with those and get them to work, please update this README with some notes.
+
+### Testing
+
+You can run all of the unit tests located in `src/hazelcore/test` by running `dune runtest`.
```
dm0n3y: Let's change `dune runtest` to `make test` since we say to use `make` everywhere else.

src/hazelcore/UHExp.re
```diff
@@ -308,10 +308,130 @@ let text_operand =
...
-  SkelExprParser.skel_expr(SkelExprLexer.read, lexbuf);
+let associate = (seq: seq): skel => {
+  /**
+   * Shunting-yard opp algorithm according to the specification here
+   * https://en.wikipedia.org/wiki/Shunting-yard_algorithm#The_algorithm_in_detail
+   *
+   */
+  let rec go_seq =
+          (
+            skels: list(skel), /* List of skels to be combined into single output skel. */
```
dm0n3y: Let's name this something like `skel_stack`, just to make clear it's a stack

src/hazelcore/UHExp.re
```diff
@@ -308,10 +308,130 @@ let text_operand =
...
+      | [final_skel] => final_skel // In this case,
+      | _ => Skel.Placeholder(-1) // This case will never be reached
+      };
+    };
+  }
+  and build_ast_while =
+      (skels: list(skel), op_stack: list(Operators_Exp.t), should_mv)
+      : (list(skel), list(Operators_Exp.t)) => {
+    /* Move operators from the operator stack to the output skel list while... */
+    switch (op_stack) {
```
dm0n3y: You could compress this `switch` expression and the nested one further below into
```
switch (op_stack, skels) {
| ([op, ...op_stack'], [subskel1, subskel2, ...skels']) when should_mv(op) => ...
| _ => ...
}
```

src/hazelcore/UHExp.re
```diff
@@ -308,10 +308,130 @@ let text_operand =
...
+         * skels:
+         * [BinOp(_, Plus, Placeholder(1), Placeholder(0))]
+         *
+         * op_stack:
+         * []
+         *
+         */
+        (
+          switch (skels) {
+          | [rand_2, rand_1, ...skels'] =>
```
dm0n3y: By `rand`, I assume you mean `operand`? I'd call these subskels, we use operand to mean specifically the things you find at `Placeholder` nodes.

src/hazelcore/test/SkelParserTests.re
```diff
@@ -0,0 +1,180 @@
...
+      Skel.BinOp(
+        NotInHole,
+        Operators_Exp.Cons,
+        Skel.Placeholder(4),
+        Skel.Placeholder(5),
+      ),
+    );
+
+  UHExp.associate(precedence_op_seq) == precedence_op_skel;
+};
```
dm0n3y, line 187: waow

src/hazelcore/UHExp.re
```diff
@@ -308,10 +308,130 @@ let text_operand =
   | InvalidTextShape(t) => new_InvalidText(u_gen, t)
   };
 
-let associate = (seq: seq) => {
-  let skel_str = Skel.mk_skel_str(seq, Operators_Exp.to_parse_string);
-  let lexbuf = Lexing.from_string(skel_str);
-  SkelExprParser.skel_expr(SkelExprLexer.read, lexbuf);
+let associate = (seq: seq): skel => {
```
dm0n3y: Let's also migrate `UHPat` and `UHTyp` to use shunting yard too. I'd move the core logic of `associate` to `Skel` (maybe just call it `mk` in that module), add some extra args that allow clients to specify precedence and whatnot, then instantiate for each `UHExp`, `UHPat`, `UHTyp` module.

Once you've done that, you can delete the `Skel[Expr|Pat|Typ][Parser|Lexer]` files.

dm0n3y (changes requested):
Looks good for the most part! Waow tests!! Main request other than nits is to add shunting yard parsing for patterns and types as well.

README.md
claban3: Oh shoot, I already did! This is just a typo.

src/hazelcore/UHExp.re
claban3: Sure thing

src/hazelcore/UHExp.re
claban3: Sure thing.

src/hazelcore/UHExp.re
claban3: Will look into all of this. One comment: I wrote a function to look up precedence/associativity of operators in Operators_Exp. Do you still want arguments like this or is that sufficient?

dm0n3y: There will be a single higher-order `Skel.mk` function that takes a function `precedence: 'operator => int` as an argument, then each `UHExp`/`UHPat`/`UHTyp` module will pass in `Operators_[Exp|Pat|Typ].precedence` function into `Skel.mk`.

claban3: Ok makes sense.

claban3:
I've made all of the requested changes. As for performance, it looks like the old parser could parse 10M seqs in ~12 seconds, while the new one can do the same in ~1 second. Good news!

I did not write tests for typ/pat parsing because it is the same exact logic as in exp, just with different associativity/precedence functions. Could someone verify the precedences/associativities that I added in Operators_[Exp|Typ|Pat]? @cyrus @dmoon1221

src/hazelcore/Operators_Typ.re
```diff
@@ -15,3 +15,14 @@ let to_parse_string = op =>
   | Arrow => "->"
   | _ => to_string(op)
   };
+
+let precedence =
+  fun
+  | Arrow => 0
+  | Sum => 1
+  | Prod => 2;
```
dm0n3y: Let's use the same precedence as specified in `HTyp`. We can also refactor any use of precedence in `HTyp` and `HTypDoc` to refer back to this as single source of truth.

src/hazelcore/Skel.re
```diff
@@ -105,3 +105,123 @@ let mk_skel_str =
   let ph_map = Hashtbl.create(8);
   mk_skel_str'(string_of_op, seq, counter, ph_map);
 };
+
+let mk =
+    (
+      precedence: _ => int,
+      associativity: _ => Associativity.t,
```
dm0n3y: Let's write out `'op` instead of `_`.

src/hazelcore/Operators_Typ.re
dm0n3y: There's also some code in `UHTyp.contract` that should use the `precedence` function defined here.

dm0n3y: One last place where we should redirect precedence lookup: the `DHDoc_*` modules. Those modules should also call `Operators_*.precedence`.

src/hazelcore/Skel.re
```diff
@@ -105,3 +105,123 @@ let mk_skel_str =
...
+      go_seq(skel_stack', [op, ...op_stack'], seq, lex_addr);
+    | E =>
+      /**
+       * Once the input seq is empty, continuously pop
+       * operators in the stack and build up the output skel.
+       */
+      let (skel_stack', _) = build_ast_while(skel_stack, op_stack, _ => true);
+      switch (skel_stack') {
+      | [final_skel] => final_skel // In this case,
+      | _ => Placeholder(-1) // This case will never be reached
```
dm0n3y: Let's replace this with a `failwith("<your err message here>")`

README.md
```diff
@@ -102,3 +102,11 @@ The `make dev` and `make release` commands do three things:
 You can print to the browser console using the standard `print_endline` function. This is probably the easiest method right now.
 
 `js_of_ocaml` does support source maps and has some other flags that might be useful. If you experiment with those and get them to work, please update this README with some notes.
+
+### Testing
+
+You can run all of the unit tests located in `src/hazelcore/test` by running `make test`.
+
+Unit tests are written using the [ppx_expect](https://github.com/janestreet/ppx_expect/tree/master/example) and [ppx_inline_tests](https://github.com/janestreet/ppx_inline_test/tree/master/example). If you would like to adjust your expect tests to assert for the output that was last printed, run `make fix-test-answers`.
```
dm0n3y: Remove "the"

dm0n3y (changes requested):
Looks great, thanks! One last round of changes to make sure the `Operators_*.precedence` functions are the single sources of truth throughout the codebase, then we should be good to merge.

src/hazelcore/Operators_Typ.re
claban3: > Let's use the same precedence as specified in `HTyp`. We can also refactor any use of precedence in `HTyp` and `HTypDoc` to refer back to this as single source of truth.

Should I just rewrite the precedence lookup function in `HTyp` to reference `Operators_Typ`, and leave all the references to `HTyp.precedence*` alone?

Or should I remove all precedence lookup functions and change all references to be to the functions `Operator_*.precedence`?

dm0n3y: The former is fine

cdfa:
Finally had time to take a look at this. Looks good for the most part, but the precedence of `const` seems inconsistent. It is `0` for `Exp`, the maximum precedence for `Typ` and there is no defined precedence for `Pat`, but the one for `Exp` is used in `DHDoc_Pat.precedence`.
I think having the `const` precedence for `Typ` be the maximum precedence was intentional to make `UHTyp.contract` work, so you should definitely test it if you change it.

claban3:
> Finally had time to take a look at this. Looks good for the most part, but the precedence of `const` seems inconsistent. It is `0` for `Exp`, the maximum precedence for `Typ` and there is no defined precedence for `Pat`, but the one for `Exp` is used in `DHDoc_Pat.precedence`.

Hey @cdfa thanks for the feedback. I was confused by this too, but my changes are inline with what was there previously. Could someone explain what the const precedence is used for in the two situations, and maybe suggest how to properly centralize it's definition?

cdfa:
The const precedence is used to give a precedence to clearly delimited patterns/expressions/types, so literals or things in parentheses. In `HTyp.contract` this is used to determine wether some `Seq` needs to parenthesised or not. Clearly delimited `Seq`s should not be parenthesised, so this is why it had the maximum precedence, although you could also reverse the comparison order and the comparison there if that is more convenient.

I don't really know about `DHDoc`, but it seems does have the comparison operators reversed, so I think it would be easiest to reverse the precedence order for `Typ` as I described above to make it consistent, but you should test and make sure :)

claban3:
> I don't really know about `DHDoc`, but it seems does have the comparison operators reversed, so I think it would be easiest to reverse the precedence order for `Typ` as I described above to make it consistent, but you should test and make sure :)

Hey @cdfa, I've been trying to make this work, but I'm running into an issue. I can't reverse the precedence of Arrow, Prod, Sum, Const the way you suggest, because that will break the opp parser that I wrote. The only way to refactor this is to change only one of the const definitions to be either max or min precedence. I don't how to do that nicely, and I am wondering why this inconsistency matters, since the definition of const is used in two separate contexts.

dm0n3y:
@claban3 Could you elaborate on how your parser breaks when making the precedence numbering consistent (i.e. const is always 0)? We should use consistent numbering and adapt everything else. I think all that would be necessary is to flip some comparison operators. Taking a look at those spots now.

claban3:
> @claban3 Could you elaborate on how your parser breaks when making the precedence numbering consistent (i.e. const is always 0)? We should use consistent numbering and adapt everything else. I think all that would be necessary is to flip some comparison operators. Taking a look at those spots now.

@dmoon1221 If the precedence for types is `Arrow < Sum < Prod < const`, and I change the order so that `const < Prod < Sum < Arrow`, then I am able to flip the order of the precedence comparison in `UHTyp.contract` which will work properly, but then the actual skel parser will have the incorrect precedences. I can't flip the comparison in that parser, since it is used by `UHExp`, `UHPat`.  

Sorry if I'm missing something.

dm0n3y:
> but then the actual skel parser will have the incorrect precedences

What do you mean by incorrect here? Is it that a test fails? In that case, the test needs to be updated. `const < Prod < Sum < Arrow` is the right ordering based on OCaml's type expression precedence table https://caml.inria.fr/pub/docs/manual-ocaml/types.html (no sum operator in OCaml so we're making a judgment call there).

src/hazelcore/HTyp.re
```diff
@@ -22,10 +22,10 @@ let is_Prod =
   | Prod(_) => true
   | _ => false;
 
-let precedence_Prod = 1;
```
claban3, line 25: @dmoon1221 Sorry I misswrote, but the precedence was originally (in the dev branch) `Prod < Arrow < Sum < const`. I also wrote a test using the original typ parser as the source of truth. That test passes against my parser only when the type precedences are `Prod < Arrow < Sum`. 

Does this mean that the original parser was just broken? If so I will go ahead and set `const < Prod < Sum < Arrow`, flip the comparisons in `contract`, and re write the test accordingly.

src/hazelcore/test/SkelParserTests.re
```diff
@@ -0,0 +1,236 @@
...
+  //   10_000_000,
+  //   "precedence op skel",
+  //   UHExp.associate,
+  //   precedence_op_seq,
+  // );
+
+  UHExp.associate(precedence_op_seq) == precedence_op_skel;
+};
+
+let%test "type precedence test" = {
```
claban3, line 189: @dmoon1221 This is the aforementioned test.

src/hazelcore/HTyp.re
dm0n3y: Yes, I think it was previously broken. @cyrus-, can you confirm (cf thread since Cory's last commit).

claban3: Ah. Thank you so much for bearing with me, I thought I was going crazy.

dm0n3y: Revision to my recommendation after talking to Cyrus: We should still standardize around the `const` precedence having the lowest numeric value, but the previous order where `Arrow` binds more tightly than `Sum` should be preserved (doesn't really matter once we have labeled sums, as we'll likely get rid of the `Sum` operator, but we'll preserve existing behavior for now). So the final order is `const < Prod < Arrow < Sum`. Sorry for the confusion.

claban3: Will do, but how should I adjust comparisons in `UHTyp.contract`? Since the precedence isn't reversed, I can't just flip the comparisons in that function. Should I just fully rewrite contract?

dm0n3y: Ah, taking a closer look at your diffs, I see now how we encountered this surprising skew in precedence values. `HTyp` used the standard ordering convention for precedence (where a higher value means 'binds tighter') whereas we used the reverse ordering everywhere else (for ease of extensibility). When you transferred `HTyp.precedence_const` to `Operators_Typ.precedence_const` in this [commit](https://github.com/hazelgrove/hazel/pull/361/commits/4617760f6d418a4feaac0f3da5e6650cff75fbf2#diff-21dd9ddc05d40c65ec2e5cb4aaea516fR25), the reversal didn't take place while you had already applied the reversal for the rest of the operators. So changing `precedence_const` to `0` simply completes your intended reversal, and now the operators in `UHTyp.contract` (which were originally based on the `HTyp` standard ordering) just need to be flipped.

dm0n3y (approved):

PR #365, opened 2020-06-24
Fixes #357 Add padding to left sidebar to match right side bar body

### LINKED ISSUE #357: action panel needs symmetric margin on left
currently it has none. see the right sidebar for an example.

wlitwin:
The right side bar has some padding on the far right edge. The left
side bar does not. This commit adds the same padding to the left bar.

Screen Shot 2020-06-23 at 21 21 24


cyrus- (approved):
looks good, thanks!

PR #369, opened 2020-06-24
move GUI-independent aspects of Undo History into hazelcore

PPER:
GUI-dependent aspects  are in `UndoHistory.re`  under hazelweb/model; GUI-independent aspects are in `UndoHistoryUtil.re`  under hazecore


cyrus- (changes requested):
rename Util to Core

cyrus- (approved):

PR #378, opened 2020-06-29
Assert testing

seanyeon:
Currently working on adding a red box for the UHExp "assert" and its argument.


cyrus-:
@seanyeon looks like this is a fork off of strings rather than dev? if so, create a pull request against strings. also there are a number of merge conflicts that need to be resolved.

PR #380, opened 2020-06-30
Unary Operators

### LINKED ISSUE #509: Negative exponents in floating point literals
`1e-5` and similar doesn't currently work

anandrav:
UnaryOpZU is for when the cursor is on the [U]nary operator portion of a negated int/float literal.
UnaryOpZN is for when the cursor is on the opera[N]d portion.
The cursor should not be generically on a UnaryOp, only its fields

Doesn't compile yet but it will soon


src/hazelcore/UHExp.re
```diff
@@ -30,6 +30,7 @@ and operand =
   | Inj(ErrStatus.t, InjSide.t, t)
   | Case(CaseErrStatus.t, t, rules)
   | Parenthesized(t)
+  | UnaryOp(ErrStatus.t, operator, operand)
```
cyrus-: you should define a separate datatype for unary operators, rather than using `operator` (which should be renamed `binop` now)

anandrav: Ok got it

cyrus-:
@anandrav pressing Delete at `1 -| 2` doesn't do the same thing as pressing Backspace at `1 - |2`. Make sure you've mirrored the logic for backspace everywhere. Will review further once that is fixed.

src/hazelcore/UHExp.re
```diff
@@ -196,6 +199,7 @@ and get_err_status_operand =
   | Case(StandardErrStatus(err), _, _)
   | ApPalette(err, _, _, _) => err
   | Case(InconsistentBranches(_), _, _) => NotInHole
+  | UnaryOp(err, _, _) => err
```
cyrus-, line 189: style: this can go in with the cases above instead

src/hazelcore/Unops_Exp.re
```diff
@@ -0,0 +1,13 @@
...
+  | FNegate;
+
+let to_string =
+  fun
+  | Negate => "-"
+  | FNegate => "-.";
+
+let to_parse_string = to_string;
+
+// let is_Space = _ => false;
```
cyrus-: style: don't leave stale comments

src/hazelcore/ZExp.re
```diff
@@ -304,6 +317,8 @@ and is_outer_zoperand =
   | CursorE(_, BoolLit(_))
   | CursorE(_, Lam(_))
   | CursorE(_, Inj(_))
+  | CursorE(_, UnaryOp(_))
+  | UnaryOpZ(_)
```
cyrus-: why is this a true case? the cursor is in the child right?

src/hazelcore/ZExp.re
```diff
@@ -675,28 +711,36 @@ and move_cursor_left_zline = (zline: zline): option(zline) =>
 and move_cursor_left_zopseq = zopseq =>
   ZOpSeq.move_cursor_left(
     ~move_cursor_left_zoperand,
-    ~move_cursor_left_zoperator,
+    ~move_cursor_left_zoperator=move_cursor_left_zbinop,
```
cyrus-: style: these argument labels should also be changed

src/hazelcore/dynamics/DHExp.re
```diff
@@ -82,6 +82,36 @@ module BinIntOp = {
...
+module UnIntOp = {
+  [@deriving sexp]
+  type t =
+    | Negate;
+  let of_op = (op: UHExp.unop): option((t, HTyp.t)) =>
+    switch (op) {
+    | Negate => Some((Negate, Int))
+    | _ => None
+    };
+  let to_op = (bio: t): UHExp.unop =>
```
cyrus-: style: `bio` isn't a good variable name here (i think it stood for "bin int op")

cyrus- (dismissed):
reviewed some of the files, so see feedback for those.

then realized we didn't have support for unary minus in patterns -- we want that because we can have number literal patterns. more fancy would be to able to have patterns like -x which only matches negative numbers, with x being bound to the number that is negated.

this is hopefully a little simpler because we _don't_ have binary minus in patterns, so every time you press - it should just insert a unary minus.

anandrav:
Unary operators with int/float literals for operands are converted to negative int/float literals when applicable. In general, whenever an edit action is made to an int/float literal, it is converted to a unary op first (if applicable) before performing the action, then converted back (if applicable) after the action.

I had to make some interesting choices when implementing the conversion functions. The same string of text could potentially be multiple different AST's in Hazel, so there's no perfect solution. 

For instance, the string "-..1" could have the following structures:
- UnaryOp(FNegate, FloatLit(".1"))  <--- awkward because -..1 doesn't look like a proper negative float.
- UnaryOp(Negate, InvalidText("..1"))   <--- makes more sense? But requires more switch statements and hard-coding

At one point, you could arrive at both edit-states using different permutations of edit-actions.
For consistency, it would be less awkward if only one of them occurred (one of them has to).
I ultimately limited it to the first one because it's how the operator behaved during my previous pull request, and giving precedence to float negation operators when parsing made other situations more natural/intuitive and require less code.

Edit: I limited it to the **first** one
Edit: InvalidText not FreeVar

cyrus-:
I think having "-..1" parse as UnaryOp(FNegate, FloatLit(".1")) makes more sense in that it is compositional.

Are there other ambiguous examples to think through?

anandrav:
No other ambiguous situations, as far as I'm aware. Just different versions of the same one.
Captura de pantalla 2021-08-21 a las 6 52 07 p  m
Captura de pantalla 2021-08-21 a las 6 52 23 p  m
again, a little awkward, but more compositional.

cyrus-:
When the cursor is to the left of a hole and we press `-`, I think it should make it a unop rather than a binop. Otherwise, I can't just type "-2" into an empty hole without using backspace.

cyrus-:
Nice work @anandrav, but closing due to haz3l having a different approach to the problem!

PR #381, opened 2020-07-01
Prune OptUtil of redundant functions

### LINKED ISSUE #349: prune OptUtil
now that we are on 4.08 we can use the standard Option module for some of those things (see todo at the top of file)

dm0n3y:
Closes https://github.com/hazelgrove/hazel/issues/349


src/hazelcore/util/OptUtil.re
```diff
@@ -1,33 +1,16 @@
...
   | (_, None) => None
   | (Some(a), Some(b)) => Some(f(a, b))
   };
-let map_default = (~default: 'b, f: 'a => 'b, opt: option('a)): 'b =>
-  switch (opt) {
-  | None => default
-  | Some(a) => f(a)
-  };
+
 let get = (if_absent: unit => 'a, opt: option('a)): 'a =>
```
cyrus-, line 9: this has a name clash with a different function `Option.get`. it behaves most closely to `Option.value`, except the default is thunked. maybe call it `OptUtil.fvalue` or `OptUtil.thvalue` (whatever you like)

cyrus- (changes requested):

src/hazelcore/util/OptUtil.re
dm0n3y: I say we call it `get` since we should always use this instead of `Option.get`

PR #384, opened 2020-07-02
Fix issues with inconsistent branches on a case 

### LINKED ISSUE #310: Resolve TODO re: case err status
https://github.com/hazelgrove/hazel/blob/49c621972371797e80d2777611b972893483726c/src/hazelcore/UHExp.re#L202

hannahpotter:
- Handle inconsistent branches when the case is in the function position of an application (previously, this would stop Hazel)
- Remove some left-over TODOs 
- Fix issue with usage analysis and case with inconsistent branches (wasn't highlighting uses in a case that had inconsistent branches)

### DIFF
--- a/src/hazelcore/CursorInfo_Exp.re
+++ b/src/hazelcore/CursorInfo_Exp.re
@@ -160,7 +160,7 @@ and get_outer_zrules_from_zrule =
 
 let rec cursor_on_outer_expr:
   ZExp.zoperand =>
-  option((ErrStatus.t, VarErrStatus.t, option(CaseErrStatus.t))) =
+  option((ErrStatus.t, VarErrStatus.t, option(list(HTyp.t)))) =
   fun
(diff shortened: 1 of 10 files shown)


src/hazelcore/CursorInfo_Exp.re
```diff
@@ -159,7 +159,8 @@ and get_outer_zrules_from_zrule =
 };
 
 let rec cursor_on_outer_expr:
-  ZExp.zoperand => option((ErrStatus.t, VarErrStatus.t)) =
+  ZExp.zoperand =>
+  option((ErrStatus.t, VarErrStatus.t, option(list(HTyp.t)))) =
```
cyrus-: the three elements of this tuple are exclusive right? should use a sum type here in that case rather than a product.

cyrus- (changes requested):

cyrus- (approved):

PR #393, opened 2020-07-08
Strings

annie-anna:


dm0n3y:
Something is wrong with this new PR. Changes prior to accidental merge into dev aren't appearing in diff. @annie-anna is this the same branch as before?

annie-anna:
> Something is wrong with this new PR. Changes prior to accidental merge into dev aren't appearing in diff. @annie-anna is this the same branch as before?

I think so. Is it because this branch is directly copied from the dev before strings is reverted?

PR #395, opened 2020-07-09
Fix #356 right sidebar goes off-screen when window isn't wide enough

### LINKED ISSUE #356: right sidebar goes off-screen when window isn't wide enough
Screenshot from 2020-06-20 21-53-08

the central area should scroll if it can't fit into the space remaining after laying out the sidebars.

if the open sidebars + the center area with a reasonable min-width don't fit on screen, then the viewport as a whole should have horizontal scrollbars

if the open sidebars take up more horizontal space than is available, then the whole page should have a horizontal scrollbar leaving

wlitwin:
There's a lot going on with this issue. The first was the code editor
was forced to be a fixed 80ch width which caused a lot of layout problems,
this was removed. The body was changed from flex to grid because that
behaves a bit better in this scenario. The code window was given a min-width
(which is arbitrary can be debated what a better size would be) and the body
was given overflow-x. With overflow-x on the body the side panes behave a bit
better now because if they would force the code area to be too small they push
out the page width instead and add a scroll bar. Overall the layout of the
application should behave a bit better, but there might still be some edge
cases. There were a couple scenarios where overflow wasn't specified and so
the page started vibrating (caught in a layout loop).

Tested with Firefox + Chromium on Linux


dm0n3y (commented):
Looks good, thanks for fixing! The one potential issue is that we had set the card width to be 80ch so that line lengths don't get too long. This is more a concern with card captions rather than code, maybe the white card area can extend to full width but any card caption remains restricted to 80ch. Kinda annoying to test right now because our card content is commented out and probably broken at the moment. I'm fine with merging this change and coming back to restricting card caption width later once we need it. I'll let @cyrus- give final approval.

wlitwin:
Do you think a max-width of 80ch would be ok? I think the issue was mostly the fixed width combined with lack of overflow that caused issues. That way if a card caption gets too long it should be limited.

dm0n3y:
Yeah, that sounds good

dm0n3y:
Looks like the max width constraint isn't getting applied

wlitwin:
Hrm, it seems the card captions actually had a `max-width: 9in` already, it was actually the `code-container` that had the 80ch width. It seems like the `max-width` is mostly working for the code container, but it does handle long spans oddly and it seems in some wrapped positions the cursor is in the wrong spot. I'll look into it some more.

What did you do to not see the 80ch constraint? If you do an inspect element on the `.code-container` div and change the max-width to 100ch does it re-layout wider?

wlitwin:
Ah, I see the issue. The issue is the result view and code text spans are not wrapping. It seems when I enable wrapping the cursor becomes de-synchronized with the text, so I'll have to look into that some more.

dm0n3y:
Hey @wlitwin, sorry to have left you hanging here. The issue I'm seeing is that the cell and the surrounding white card area won't remain restricted to 80ch wide. Below is a screenshot of the editor with all side panels collapsed. I'd expect the cell + card area to remain 80ch and for the tan background to fill out the rest of the width.
image
Not sure if I was clear in my original comment a few weeks back. I don't expect the issue to lie with the inner contents of the cell/card area not wrapping. Don't worry about any cursor desynchronization -- there are other deeper issues there that I'm working in another PR (https://github.com/hazelgrove/hazel/pull/414).

wlitwin:
Ah yeah, I interpreted it as something else. It should limit itself now. The inner code area should limit to 80ch and the page width will be somewhat larger because of the padding. I had a bit of issue getting the HTML/CSS to do exactly what I wanted due to other layout structure, so I might try playing with a larger layout change to make things behave a bit better. But for now it should behave as desired.

dm0n3y (approved):
Looks great. Thanks Walter!

PR #398, opened 2020-07-15
Fix #353 Cleanup Program.move_to_* to return the action only

### LINKED ISSUE #353: refactor Program.move_to_* to return the Action only, with performance happening separately

wlitwin:
This commit changes the two functions move_to_hole and
move_to_case_branch in Program.re to return the action to be performed
and instead of the new program. The corresponding callers have been
updated as well (only one each in Model.re).


src/hazelweb/model/Program.rei
```diff
@@ -77,15 +77,14 @@ let move_via_click:
...
+let move_to_hole: (MetaVar.t, t) => Action_common.t;
 
 /**
  * `select_case_branch(steps, n, program)` moves the cursor to the `n`th branch
  * in case expression found at `steps` (when the user clicks on a branch type
  * in the error message for a case expression with inconsistent branches)
  */
-let move_to_case_branch:
-  (CursorPath_common.steps, int, t) => (t, Action_common.t);
+let move_to_case_branch: (CursorPath_common.steps, int) => Action_common.t;
```
dm0n3y, line 88: need to update documentation

dm0n3y (changes requested):
Looks good, thanks! Just need to update some documentation.

src/hazelweb/model/Program.rei
wlitwin: Fixed.

dm0n3y (approved):

PR #402, opened 2020-07-17
Assert numbering ui

seanyeon:
Changed u_gen to id_gen
added numbering for AssertLit
added components for the UI for UHExp


dm0n3y:
Please include a more detailed PR description

src/hazelweb/gui/UHCode.re
```diff
@@ -250,23 +250,24 @@ let key_handlers =
 };
 
 let box_table: WeakMap.t(UHBox.t, list(Vdom.Node.t)) = WeakMap.mk();
-let rec view_of_box = (box: UHBox.t): list(Vdom.Node.t) => {
+let rec view_of_box = (program: Program.t, box: UHBox.t): list(Vdom.Node.t) => {
```
dm0n3y, line 148: Couple issues here:
- In general you should try to minimize what you pass into only what's necessary. In this case, that's just the assert map.
- First issue aside, this doesn't work with the memoization. The memoization table is only indexed on the `UHBox.t`, so if the assert map changes, there will still be a memo cache hit and the assert map change won't be reflected. Rather than embedding the assert decorations in the code text, we should be rendering them as layers on top of the text like we're doing with other decorations. Talk to me about this.

src/hazelcore/layout/UHAnnot.re
```diff
@@ -26,6 +26,12 @@ type open_child_format =
 [@deriving sexp]
 type t =
   | HoleLabel({len: int})
+  //| AssertLabel
+  /*| AssertPass
+    | AssertFail
+    | AssertIndet
+    | AssertComp*/
+  | AssertNum({num: int})
```
dm0n3y: Clean up unused code

src/hazelcore/layout/DHDoc_Exp.re
```diff
@@ -184,6 +188,20 @@ let rec mk =
...
+      | AssertLit(n) =>
+        switch (AssertMap.lookup(n, map)) {
+        | Some(a) =>
+          switch (AssertMap.check(a)) {
+          | Pass => Doc.text("assert") |> Doc.annot(DHAnnot.AssertPass)
+          | Fail => Doc.text("assert") |> Doc.annot(DHAnnot.AssertFail)
+          | Indet => Doc.text("assert") |> Doc.annot(DHAnnot.AssertIndet)
+          | Comp => Doc.text("assert") |> Doc.annot(DHAnnot.AssertComp)
+          }
+        | None => Doc.text("assert") |> Doc.annot(DHAnnot.AssertIndet)
```
dm0n3y: Can you think of how to minimize repetition of `Doc.text` and `Doc.annot` here?

src/hazelweb/gui/Page.re
```diff
@@ -167,8 +167,15 @@ let view = (~inject: ModelAction.t => Vdom.Event.t, model: Model.t) => {
                     ~selected_instance,
                     ~width=80,
                     model.compute_results.show_unevaluated_expansion
-                      ? program |> Program.get_expansion
+                      ? program |> Program.get_expansion  // change expansion
```
dm0n3y: Is this a remaining task?

dm0n3y:
Rendering bug in assert decorations in the result box when the program is long and you can begin to scroll the page
image

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -69,258 +69,341 @@ let eval_bin_float_op =
   };
 };
 
-let rec evaluate = (d: DHExp.t): result =>
+let rec evaluate =
+        (d: DHExp.t, assert_map: AssertMap.t): (result, AssertMap.t) =>
```
dm0n3y, line 541: It seems `assert_map` isn't being read anywhere here. Should that be the case? If so, I'd re-write this so that `evaluate` doesn't take `assert_map` as an argument and just aggregates the submaps of children. This makes it more amenable to memoization in the future by preserving independence of evaluation of children.

dm0n3y:
Bug: expecting the final assertion here to have a green checkmark
image

edit: fixed by https://github.com/hazelgrove/hazel/pull/402/commits/baf2e45c040cff813390289a4083d84dbae2c629

dm0n3y:
Bug: expecting both assertions here to have green checkmarks
image

edit: fixed by https://github.com/hazelgrove/hazel/pull/402/commits/baf2e45c040cff813390289a4083d84dbae2c629

dm0n3y:
Bug: can't type anything in middle empty line, getting `[Program.InvalidInput]` in console
image

edit: fixed by https://github.com/hazelgrove/hazel/pull/402/commits/baf2e45c040cff813390289a4083d84dbae2c629

src/hazelcore/dynamics/Evaluator.re
seanyeon: I will talk to Cyrus about this and try to find other way to get the assert map

dm0n3y: Not sure what you mean by that. My suggestion is for `evaluate` to have the type signature `DHExp.t => (result, AssertMap.t)`.

seanyeon: Oh I see what you mean, I initially thought this caused the problem when evaluate was recursively called, but now that I think about it I think we can get rid of it.

dm0n3y:
Bug: expecting holes in result to have different instance numbers
image

cyrus-:
Marking as draft pending merge with dev and fixes to issues @dmoon1221 identified above.

cyrus-:
superceded by testing branch, closing PR

PR #412, opened 2020-07-22
Turn off user-select on UI elements 

### LINKED ISSUE #323: Turn off user-select on UI elements
Sometimes double clicking on UI elements will cause the text inside to be selected, e.g. the undo buttons and other things. We can use the CSS `user-select` property to disable this for UI elements containing text:

https://stackoverflow.com/questions/826782/how-to-disable-text-selection-highlighting

PPER:
Sometimes double-clicking on UI elements will cause the text inside to be selected. Use the CSS user-select property to disable this for UI elements containing text.

Simply add the existing `noselect` class to elements.


cyrus-:
@PPER automatic merge fails -- can you get this back up to date when you get a chance?

cyrus-:
inactive, closing

PR #414, opened 2020-07-28
New decorations architecture

### LINKED ISSUE #302: Multi-line operands are laid out incorrectly
They are laid out as inline blocks, instead of causing new lines to be inserted.
Screenshot from 2020-05-31 01-22-41

### LINKED ISSUE #374: Refactor Cell/UHCode
Move editing related things to a EditableUHExp (or similar) module, and then Cell is just that with a border around it for now.

dm0n3y:
Refactor `UHCode` rendering such that decorations (e.g. err holes, var use, current term indicator) are separate elements laid on top of code text elements. The motivation for this refactor is twofold:
- First, separating decorations from code text allows `Doc`/`Layout` specification to concern itself solely with the static code text, independent of the cursor location and other transient decorations. This minimizes pretty printing recalculation, the primary bottleneck in our view pipeline. (Note: Although docs/layouts no longer carry decoration information in their annotations, we must still regenerate docs whenever there is a change in an err hole or var err hole status, since these are embedded in `UH[Exp|Pat]` and thus affect pointer equality.)
- Second, specifying decorations as independent layers improves modularity and makes it easier to add new decorations in the future. In particular, decorations no longer need to hack z-indices, which caused issues with ongoing work on Hazel assistant.

New pipeline:
```
Structure -> Doc -> Layout -> Box -> Vdom (code text)
                      |
                       --> MeasuredLayout -> Vdom (code decorations)
```

`Box` decomposes a `Layout` into boxes that can be easily translated to HTML structure. In the process of converting an annotated `Layout` node into a `Box` node, the annotated subject may be split into up to three boxes (head, body, tail) with the annotation distributed over the constituents. (It's possible go directly from `Layout` to `Vdom` by opting to lay out all code text in inline format, but that would require inserting additional indentation and annotations would then span those indentation nodes, which isn't true of the original `Doc`/`Layout` structure. Such an approach is simpler to implement and wouldn't cause any issues in the current implementation, but could perhaps cause issues in future use cases. That said, I suspect the `Box` abstraction crept in due to prior concerns when decorations weren't separated from the code text. Open to dropping it. I may also be forgetting other concerns, maybe @adamsmd could chime in here.)

`MeasuredLayout` is the same as `Layout` but adorns each `Layout` node with measurement information. These are used to determine the coordinates of SVG paths for code decorations, which are placed below / on top of the code text. In more detail: `Program.re` provides paths to each decoration in the current program; `UHCode.re` uses those paths to traverse the program's `MeasuredLayout` and identify the terms that need to be decorated; `UHDecoration.re` takes the dimensions of the identified terms and specifies the shape of each decoration as a collection of rectangles, the contour of the union of which forms the decoration shape. The algorithm used to compute this contour is taken from Computational Geometry: An Introduction by Preparata & Shamos, Sections 1.2.3.1, 8.3, 8.5.

`MeasuredLayout` takes over the role of `CursorMap`. On each click/arrow key movement, the dimensions of the `MeasuredLayout` nodes are used to traverse the document in search of `Step` and `Token` annotations in order to determine the next `CursorPath`.

Removed much of `style.css` that was previously devoted to hacking together current term indicator decorations directly embedded in the code text.


src/pretty/ListUtil.re
```diff
@@ -0,0 +1,16 @@
...
+let split_last_opt = (xs: list('a)): option((list('a), 'a)) =>
+  switch (List.rev(xs)) {
+  | [] => None
+  | [y, ...ys] => Some((List.rev(ys), y))
+  };
+let split_last = (xs: list('a)): (list('a), 'a) =>
+  switch (split_last_opt(xs)) {
+  | None => failwith("empty list")
+  | Some(r) => r
+  };
```
dm0n3y, line 16: Duplicated some code here for now, created an issue to dedup https://github.com/hazelgrove/hazel/issues/424

dm0n3y:
I remembered why I opted to go with the `Box` intermediate abstraction over going directly from `Layout` to `Vdom` for code text: the `Box` approach is more amenable to memoization. The direct approach requires keeping track of indentation levels, meaning that a layout node that gets indented must have its `Vdom` re-computed. In the current approach, both `Layout`-to-`Box` and `Box`-to-`Vdom` can be cleanly memoized.

I've improved documentation and fixed the initial focus bug we encountered during our last review @cyrus- @adamsmd. Ready for final review.

PR #418, opened 2020-08-06
Show exception error msg

### LINKED ISSUE #130: Catch exceptions at top level

hannahpotter:
resolves #130 

This is how the errors display:
image


src/hazelweb/Main.re
```diff
@@ -1,10 +1,42 @@
...
+            Vdom.Node.body(
+              [Vdom.Attr.id("error-message")],
+              [
+                Vdom.Node.h3([], [Vdom.Node.text("ERROR!")]),
+                Vdom.Node.text(message),
+              ],
+            ),
+          );
+        Dom.appendChild(root, error_box);
+      },
```
adamsmd, line 28: IIUC, Hazel completely stops working if this case is triggered.

In that case, I would suggest putting the message at the top of the screen so there is no way to miss it.

I would also suggest stronger wording than "Error" because that sounds like the user made a mistake that he or she can correct.

Maybe something like "Hazel has crashed" with a message telling users to reload the page (and how to reload).  If there is a way to save and reload the user's code after refreshing, tell them how to do that.  If not, warn them that their code will be lost, and they may want to manually copy it before reloading.

src/hazelweb/Update.re
```diff
@@ -117,7 +117,7 @@ let apply_action =
           JSUtil.log("[Program.MissingCursorInfo]");
           model;
         | exception Program.InvalidInput =>
-          JSUtil.log("[Program.InvalidInput");
```
adamsmd, line 120: Should we be showing errors for any of these?  Maybe not a full crash message but something so developers bother to report or track these down when they occur.

- Program.FailedAction
- Program.CursorEscaped
- Program.MissingCursorInfo
- Program.InvalidInput
- Program.DoesNotElaborate

adamsmd (changes requested):
Overall the code looks good to me.  The comments I have are feature extensions that we can handle either now or add them to our list of cleanup items.

src/hazelweb/Main.re
cyrus-: I agree that we should say "Hazel has crashed" instead fo "Error"

src/hazelweb/Update.re
cyrus-: Yes, ideally all of the invariant violation exceptions get reported this way so that we get reports more specific than just "Hazel froze/crashed"

cyrus-:
@hannahpotter marking as draft pending resolution of comments above

cyrus-:
Would like this in haz3l but closing this as out-of-date

PR #430, opened 2020-09-02
Type annotations for operands in patterns

disconcision:
This adds a new annotation form, which can be used to annotate operands in patterns with an explicit type. This serves as a generic replacement for the optional type field attached to letlines, lambdas, and case patterns. As with those previous cases, an annotation can be added using by inserting a `:`.

One complication with entry is that commas are now overloaded in patterns when inserted at the end of a type annotation; does `,` signify a pattern-level tuple or a type-level one? We default to a pattern-level tuple; type-level tuples at the type/pattern syntactic interface must be surrounded by parentheses. More generally, any operator which has both pattern- and type-level instances will be interpreted on the pattern-level, though there are currently no other such cases.


disconcision:
This is probably good for a look... there's some duplication i'll get rid of in UHDoc; had to mess around a bit to get multiline formatting working properly... there are some heterogeneities there preventing simple abstractions. Mostly it seems to work/look okay in manual testing. One thing that's come up during testing is (directly) nested annotations, e.g. `a : Int : Int` or `(a:Int):Int`. Hadn't thought about this case; should I explicitly forbid these?

cyrus-:
(we discussed this during our meeting today -- now that we have type annotations in patterns, we need to remove type annotations on lambdas and let bindings. this will require a little bit of care because whether an expression is analytic or synthetic depends on whether there is an annotation.)

cyrus- (changes requested):
a few small remaining issues / bugs

* pressing Space when your cursor is immediately after `:` doesn't move to the beginning of the type -- it moves it past the type. this is inconsistent with the behavior of other infix ops
* I think the "insert type annotation" entry in the history panel should have sort "Pat" not "Exp"
* In the action panel, it should say "type annotation" not "type ascription" (ascription is used when you annotate an expression directly -- Hazel doesn't support ascription)
* regression: when I type `\x : Bool {x|}`, I can't press `+` -- this works in dev

disconcision:
@cyrus- all issues addressed. Was confused by the Pat/Exp and annotation/ascription bugs as those weren't part of my additions... Looking back though the commit history it looks like the previous type annotation code was derived from now-removed type ascription support; there are still other traces of this spread around, including the name of the SAsc constructor.

src/hazelcore/UHPat.re
```diff
@@ -203,6 +208,7 @@ and is_complete_operand = (operand: 'operand): bool => {
   | ListNil(InHole(_)) => false
   | ListNil(NotInHole) => true
   | Parenthesized(body) => is_complete(body)
+  | TypeAnn(_, op, ann) => is_complete_operand(op) && UHTyp.is_complete(ann)
```
cyrus-, line 211: there was a `check_type_holes: bool` flag on is_complete_lines -- I don't remember what/if that was used for something, but if so, would need to add that to this function (wasn't an argument before because there were no types in patterns)

src/hazelcore/Statics_Exp.rei
```diff
@@ -129,3 +117,27 @@ let fix_and_renumber_holes:
   (Contexts.t, UHExp.t) => (UHExp.t, HTyp.t, MetaVarGen.t);
 
 let fix_and_renumber_holes_z: (Contexts.t, ZExp.t) => Statics.edit_state;
+
+let joint_pattern_type: (Contexts.t, list(UHExp.rule)) => option(HTyp.t);
```
cyrus-: "joined" rather than "joint"

src/hazelcore/Statics_Exp.re
```diff
@@ -43,14 +63,9 @@ and syn_line = (ctx: Contexts.t, line: UHExp.line): option(Contexts.t) =>
...
-  | LetLine(p, None, def) =>
-    let* ty = syn(ctx, def);
-    Statics_Pat.ana(ctx, p, ty);
-  | LetLine(p, Some(uty), def) =>
-    let ty = UHTyp.expand(uty);
-    let (ctx_def, _) = ctx_for_let(ctx, p, ty, def);
-    let* _ = ana(ctx_def, def, ty);
-    Statics_Pat.ana(ctx, p, ty);
+  | LetLine(p, def) =>
+    let* ty_def = syn(ctx, def);
```
cyrus-: this doesn't seem to be handling recursive lets?

src/hazelcore/Statics_Exp.re
```diff
@@ -1334,7 +1293,16 @@ and ana_fix_holes_operand =
         u_gen,
       );
     };
-  };
+  }
+and extend_let_body_ctx =
+    (ctx: Contexts.t, p: UHPat.t, def: UHExp.t): Contexts.t => {
+  /* precondition: (p)attern and (def)inition have consistent types */
```
cyrus-, line 1300: put this in the .rei file too

src/hazelcore/ZExp.re
```diff
@@ -235,12 +208,11 @@ and is_after_zline =
   | CursorL(cursor, CommentLine(comment)) =>
     cursor == OnText(String.length(comment))
   | CursorL(cursor, EmptyLine) => cursor == OnText(0)
-  | CursorL(cursor, LetLine(_, _, _)) => cursor == OnDelim(3, After)
+  | CursorL(cursor, LetLine(_)) => cursor == OnDelim(3, After)
```
cyrus-: there is no longer a delimiter at index 3 right?

"let", "=", "in"

src/hazelcore/ZExp.re
```diff
@@ -634,42 +597,27 @@ and move_cursor_left_zline = (zline: zline): option(zline) =>
...
-  | CursorL(OnDelim(k, Before), LetLine(p, ann, def)) =>
-    // k == 1 || k == 2 || k == 3
-    switch (k == 1, k == 2, ann) {
-    | (true, _, _) => Some(LetLineZP(ZPat.place_after(p), ann, def))
-    | (_, true, None) => Some(LetLineZP(ZPat.place_after(p), ann, def))
-    | (_, true, Some(ann)) =>
-      Some(LetLineZA(p, ZTyp.place_after(ann), def))
-    | (_, _, _) => Some(LetLineZE(p, ann, place_after(def)))
+  | CursorL(OnDelim(k, Before), LetLine(p, def)) =>
+    // k == 1 || k == 2
```
cyrus-: comment inaccurate

src/hazelcore/ZExp.re
```diff
@@ -634,42 +597,27 @@ and move_cursor_left_zline = (zline: zline): option(zline) =>
...
+    switch (k == 1) {
```
cyrus-: this can just be an if, or a switch on k, now

src/hazelcore/ZExp.re
```diff
@@ -850,54 +779,35 @@ and move_cursor_right_zline =
...
+    | Some(zp) => Some(LetLineZP(zp, def))
     | None =>
       Some(
         CursorL(
           OnDelim(
-            switch (ann) {
-            | None => 2
-            | Some(_) => 1
-            },
+            1, // is this right??
```
cyrus-: looks right to me -- comment unnecessary

src/hazelweb/layout/UHDoc_Pat.re
```diff
@@ -29,6 +29,9 @@ let mk_NTuple:
     ~get_tuple_elements=UHPat.get_tuple_elements,
     ~inline_padding_of_operator,
   );
+let mk_PatternAnnotation:
```
cyrus-: call it TypeAnn instead of PatternAnnotation?

src/hazelweb/layout/UHDoc_common.re
```diff
@@ -140,99 +141,139 @@ type formatted_child =
...
-    ])
-  };
+let pad = ((left, right), ~newline=[], child_doc) => {
+  let lpadding = left == empty_ ? [] : [left];
+  let rpadding = right == empty_ ? [] : [right];
+  Doc.hcats(newline @ lpadding @ [child_doc] @ rpadding);
 };
 
-let pad_closed_child =
+let pad_open_inline_child = (inline_padding, with_border, child_doc) => {
```
cyrus-, line 150: looks like there are some refactorings in here that may not be specific to this PR?

src/hazelcore/UsageAnalysis.re
```diff
@@ -21,6 +21,8 @@ and binds_var_operand = (x, operand: UHPat.operand): bool =>
   | Var(NotInHole, NotInVarHole, y) => x == y
   | Parenthesized(body) => binds_var(x, body)
   | Inj(NotInHole, _, body) => binds_var(x, body)
+  | TypeAnn(InHole(_), _, _) => false
```
cyrus-: can add this to the false block above

src/hazelcore/UsageAnalysis.re
```diff
@@ -48,7 +50,7 @@ and find_uses_line = (~steps, x: Var.t, line: UHExp.line): (uses_list, bool) =>
   | CommentLine(_) => ([], false)
   | ExpLine(opseq) => (find_uses_opseq(~steps, x, opseq), false)
   | EmptyLine => ([], false)
-  | LetLine(p, _, def) => (
+  | LetLine(p, def) => (
```
cyrus-, line 53: regression from dev: usage finding isn't working for the recursive variable

src/hazelcore/CursorInfo_Exp.re
```diff
@@ -289,45 +282,26 @@ and syn_cursor_info_line =
...
     | Some(ci) => Some(CursorNotOnDeferredVarPat(ci))
     }
-  | LetLineZP(zp, ann, def) =>
-    let pat_ci = CursorInfo_Pat.ana_cursor_info(~steps=steps @ [0], ctx, zp);
-    switch (ann) {
-    | None =>
-      switch (Statics_Exp.syn(ctx, def)) {
-      | None => None
-      | Some(ty1) => pat_ci(ty1)
+  | LetLineZP(zp, def) =>
```
cyrus-, line 285: this isn't making use of the synthesized type of the definition -- useful information for the user I think

src/hazelcore/CursorInfo_Pat.re
```diff
@@ -201,20 +205,26 @@ and syn_cursor_info_zoperand =
...
              ctx,
              extract_from_zpat_operand(zoperand),
            ),
          )
-       )
+       })
   | InjZ(_, _, zbody)
   | ParenthesizedZ(zbody) => syn_cursor_info(~steps=steps @ [0], ctx, zbody)
+  | TypeAnnZP(_, zop, _) =>
+    syn_cursor_info_zoperand(~steps=steps @ [0], ctx, zop)
```
cyrus-: bug: should be ana

cyrus- (changes requested):
see comments. some small bugs and tweaks to look into, but almost there. didn't look at Action code in detail yet, will do so after next round of fixes.

cyrus-:
in addition to fixes above, please edit the PR description to be more descriptive of the PR :)

src/hazelweb/layout/UHDoc_common.re
disconcision: The preexisting padding functions had some awkwardly duplicated logic, and did not have full combinatorial coverage, including the left-delimited closed-child case, which is necessary here for type annotations. The refactoring was necessary to prevent more copy-paste code here.

src/hazelcore/UHPat.re
disconcision: I did a basic manual analysis. The flag occurs in UHExp and OpSeq. AFAICT the only actual client call site of any is_complete variant is a single call in Program, with the flag set to false. This is the sole occurrence of the pattern UHExp.is_complete*. The only occurrence of OpSeq.is_complete is in a UHExp is_complete* function. Manual inspection of all is_complete* functions in UHExp and OpSeq indicates that they all just thread the flag down from the root is_complete implementations. Shall I just rip out the flag?

src/hazelcore/UsageAnalysis.re
disconcision: Pretty sure this is fixed now but I'm not 100% sure, let's remember to check tmrw

src/hazelcore/UHPat.re
cyrus-: I don't recall why we had it, but easy to add back later if necessary. Go ahead and rip out.

src/hazelcore/Action_Pat.re
```diff
@@ -753,6 +816,14 @@ and syn_perform_operand =
       Succeeded(mk_and_syn_fix_ZOpSeq(ctx, u_gen, zseq));
     }
 
+  | (Construct(SAsc), CursorP(_)) =>
```
cyrus-: let's rename it to SAnn in this PR

src/hazelcore/Action_Pat.re
```diff
@@ -1102,10 +1213,25 @@ and ana_perform_operand =
...
+        ctx,
+        u_gen,
+        op |> ZPat.place_after_operand |> ZOpSeq.wrap,
+        ty,
+      ),
+    )
   /* Construct */
   | (Construct(SOp(SSpace)), CursorP(OnDelim(_, After), _)) =>
     ana_perform_operand(ctx, u_gen, MoveRight, zoperand, ty)
+  | (Construct(SAsc), CursorP(_)) =>
```
cyrus-: on dev, this causes the expected type to be inserted immediately. I think that's a useful default, can we retain that behavior?

src/hazelcore/Action_Pat.re
```diff
@@ -1255,7 +1381,45 @@ and ana_perform_operand =
...
+    | CursorEscaped(side) =>
+      ana_perform_operand(
+        ctx,
+        u_gen,
+        Action_common.escape(side),
+        zoperand,
+        ty,
+      )
+    | Succeeded((ZOpSeq(_, zseq), ctx, u_gen)) =>
+      let newseq = annotate_last_operand(zseq, ann);
```
cyrus-, line 1400: a comment explaining what's going on here might be useful (have to reassociate the type annotation since its not parsed as part of the opseq system)

cyrus- (changes requested):

src/hazelcore/dynamics/Elaborator_Exp.re
```diff
@@ -529,17 +529,16 @@ and syn_elab_line =
     }
   | EmptyLine
   | CommentLine(_) => LinesExpand(d => d, ctx, delta)
-  | LetLine(p, ann, def) =>
-    switch (ann) {
-    | Some(uty1) =>
-      let ty1 = UHTyp.expand(uty1);
-      let (ctx1, is_recursive_fn) =
-        Statics_Exp.ctx_for_let(ctx, p, ty1, def);
+  | LetLine(p, def) =>
```
dm0n3y, line 532: Coming across this code as I merge dev into livelits. Strikes me as an early change that was never evolved with later realizations about desired behavior. I'd expect the pattern to be elaborated by `Elaborator_Pat.syn_elab` rather than `ana_elab` (further below in unchanged lines) based on the syn/ana dance in `Statics_Exp.syn/ana_fix_holes_line`; also weird that the input to this `ana_elab` call is `ty1` which was synthesized from the pattern.

Not a blocker for me so no rush.

PR #432, opened 2020-09-14
Fix #378 *_common module split

### LINKED ISSUE #387: Split `*_common` modules into `*` and `*_common` modules
The intention of `*_common` modules is to contain any internally shared function, but we should still maintain the original `*` modules for public datatypes (e.g. so that we write `CursorPath.t` instead of `CursorPath_common.t`).

wlitwin:
I believe I got all the modules that had _common in their name. I only moved their types (and dependent types) to a new module without the _common prefix and then fixed up all the references.


dm0n3y (approved):
Looks good, thanks @wlitwin ! Sorry again for the long delay. Appreciate the help.

PR #439, opened 2020-10-15
Strings3

annie-anna:


cyrus-:
this branch is still bugged, and strings are being reworked, so closing this one

PR #450, opened 2020-11-17
Create proper unfocused state in Program, clean up settings

### LINKED ISSUE #351: measurement flags should be passed around in records rather than individually

dm0n3y:
Previously Hazel programs didn't have a proper unfocused state. You could set the focused/unfocused flag on a program, but it still retained a zexp. This meant that cursor dependent editor services would display information out of sync with the editor. This PR introduces a proper unfocused state in which programs may contain an unfocused uhexp instead of a zexp. In addition to resolve in the aforementioned out-of-sync issue, this also lays necessary foundations for upcoming livelits changes in which a livelit needs to be able to clear the cursor as it sets splice contents.

At the root of these changes is a new `Program.EditState.t` datatype:
```
type term =
  | Unfocused(UHExp.t)
  | Focused(ZExp.t)
type t = {
  term,
  ty: HTyp.t,
  u_gen: MetaVarGen.t,
}
```
This datatype takes over the role previously played by `Statics.edit_state`.

As part of this change, I also performed the following housekeeping tasks:
- organized all of our settings (e.g. measure performance, configure evaluation) into a `Settings` module; renamed `OptionsPanel` to `SettingsPanel` (closes https://github.com/hazelgrove/hazel/issues/351)
- moved all pretty printing related modules in hazelweb, which we have discussed doing already and was induced by the need for dhdocs to take settings

@PPER please review changes relevant to undo history. Currently undo history assumes that every edit state has a cursor. Now that this is not the case, some care has to be taken such that only cursor-carrying edit states are pushed onto the undo history. Currently this is accomplished by raising exceptions anytime this invariant is broken. I think undo history should be refactored in the future to contain explicit zexps or something along those lines. In the meantime, let me know if you see a better way to handle the new unfocused states.

@hannahpotter this will likely affect your cursor inspector changes


dm0n3y:
This PR is gonna need some more work. I applied the extraneous housekeeping tasks mentioned above directly in `dev` so that we can take advantage of those changes while I iterate on this branch, also simplifies eventual review of this branch

cyrus-:
@dmoon1221 is this stale?

cyrus-:
Seems stale, closing.

PR #454, opened 2020-11-29
Pins deps with opam.export file

bkase:
I ran into some trouble getting a set of dependencies to work together.

We use an `opam.export` file in MinaProtocol/mina to pin the versions of all the opam dependencies that are known to work together. I figured it would make sense here too.

This PR:
* Checks in an `opam.export` file
* Changes `make deps` to use the export file
* Adds a `make change-deps` to update the opam.export file

I tested by wiping my `~/.opam` switch and re-running `make deps` and `make dev` and everything looks good 👍


dm0n3y (approved):
Thanks! Been meaning to look into this, glad to have someone with prior experience. Tested successfully on my machine.

adamsmd:
I have a couple of concerns.

(1) It looks like all installed packages are tracked, not just dependencies.  Is that the case?  Is there a way to have it track only dependencies?  (I'm not sure whether it would be better to include or not include indirect dependencies, so do we have both options or just one?)

(2) If someone has a different version of the depended-upon libraries, what happens?  Do both versions coexist?  Does it uninstall the old version?  (My concern is that it might break someone's setup if done in a pre-existing switch.)

adamsmd:
I also note that there is `opam lock` (https://opam.ocaml.org/blog/opam-20-tips/#Lock-files) which seems similar to what you are doing with the exports.  Could you look into the difference?

PR #459, opened 2020-12-14
Num op prediction

Yunif3:
Currently, the user has to manually input a `Float` type operator if they wish to perform some operation on two `Float` types.\
However, we suggest a type prediction system for operators where the user's input for an operator will be converted match the context.\
This idea can later be expanded to other types as well.
* Current behavior
    ```
    2.| -> 2. +|
    2.| 3. -> 2. +| 3.
    3| -> 3 +|
    2.| 1 -> 2. +| 1
    ```
* New behavior
    ```
    2.| -> 2. +.|
    2.| 3. -> 2. +.| 3.
    3| -> 3 +|
    2.| 1 -> 2. +.| 1
    ```

In predicting the operator type, we will use a four step rule.
"Type of interest" will denote the types of the arguments that is being considered to determine the appropriate operator to construct given an input.
1. If the edited expression is in an analytical position that is either `Int` or `Float`, the operator of the corresponding type will be used. 
2. When synthesizing, the type of the operand which the operator construction action was called from will be used to predict the operator type, if the operand type is a type of interest.
3. If the operand from which the operator construction was called is an empty hole, the type of the opposing operand will be used to predict the operator type, if this latter type is a type of interest.
4. In the case that neither arguments of the operator is of the type of interest, the "default" operator will be used, which will be the `Int` operator in the case of `Int` and `Float` operators.

Additional Examples
```
let i:Int = 2. | 3. in ... -> let i:Int = 2. +| 3. in ...
2.| 3 -> 2. +.| 3
3| 2. -> 3 +| 2.
3.|_ -> 3. +| _
true | true -> true +| true
```

**Current interface**

https://user-images.githubusercontent.com/47135468/124396382-c9009680-dd3b-11eb-96e1-1a96f98ed14b.mov

**Altered interface**

https://user-images.githubusercontent.com/47135468/124396391-d0c03b00-dd3b-11eb-9fe0-0052a90bfe8e.mov


cyrus-:
@Yunif3 what is the status of this PR?

zhixiangteoh:
Current implementation does not yet treat the case where you have an empty hole in between two literals, e.g. `1. |_ 2.` -> `1. +.| _ 2.` (where a hole is also inserted to the right).

cyrus- (changes requested):
Looking very good, but one remaining functionality issue:

When the expected type is unknown (?) or something other than Float or Int, we should use the same logic as in synthetic position. Currently, it doesn't do any prediction in that situation.

cyrus-:
If the expected type is Float or Int, that should supercede the operands, so e.g.

```
let x : Int = 3.5|
```

Pressing `+` here should insert `+` rather than `+.`.

cyrus- (approved):
Looks great, thank you!

PR #460, opened 2021-01-05
Delayed indet

Yuning30:
This PR is used to fix the small bug in Elaborator_Exp.matches function that `Indet` is returned without looking at later patterns.


cyrus- (commented):
Looks good, thanks @Yuning30.

Issue #41, opened 2018-10-17
Renumber holes action


cyrus-:
Sometimes hole numbers get really large even though there aren't a lot of holes in the program. It would be good to be able to explicitly renumber holes. Should be a nice simple starter project for somebody.

PR #461, opened 2021-01-06
Add option to explicitly renumber holes using Ctrl+#

utkmehta:
This PR addresses the following issue: 
https://github.com/hazelgrove/hazel/issues/41

Since hole numbers can become extremely large, users should be allowed to renumber them explicitly. To enable users to do so, I have introduced a new action called "RenumberHoles" and mapped it to the (Ctrl+#) key combination.

Tested by creating expressions with holes, pressing Ctrl+#, and ensuring that the hole numbers start from 1 again.

### DIFF
--- a/src/hazelcore/Action.re
+++ b/src/hazelcore/Action.re
@@ -48,4 +48,5 @@ type t =
   | SwapRight
   | SwapUp
   | SwapDown
+  | RenumberHoles
   | Init;
--- a/src/hazelcore/Action_Exp.re
+++ b/src/hazelcore/Action_Exp.re
@@ -520,6 +520,7 @@ let rec syn_move =
   | SwapRight
   | SwapUp
   | SwapDown
+  | RenumberHoles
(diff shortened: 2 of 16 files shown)


cyrus-:
@utkmehta This doesn't work correctly for me. For example,

Screenshot from 2021-01-12 23-46-09

Performing the renumbering action here only sets the pattern hole the cursor is on to 1, and doesn't change hole 5.

cyrus-:
Also you can put the renumbering action in the "General Editing" part of the action panel, rather than in a new section.

cyrus-:
Also, I think this action should show up in the undo history so that the user can revert it if they didn't mean to renumber the holes.

cyrus- (changes requested):
still needs the fixes mentioned in the comments -- @utkmehta are you interested in finishing this up?

cyrus-:
This branch was buggy and is no longer active, so closing.

PR #462, opened 2021-01-13
Add small step evaluator

coned:
Add the type EvalCtx, and two related functions.


src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | Triv => true
+  | _ => false
+  };
+
+let rec is_boxedval = (d: DHExp.t): bool =>
+  switch (d) {
+  | Cast(d1, Arrow(ty1, ty2), Arrow(ty3, ty4)) =>
+    is_boxedval(d1) && !(HTyp.eq(ty1, ty3) && HTyp.eq(ty2, ty4))
+  | Cast(d1, ty, Hole) => ground_cases_of(ty) == Ground && is_boxedval(d1)
+  | _ => is_val(d)
```
cyrus-: you'll need boxedval cases for other compound types as well (products, sum, lists). you can use this system called the Gradualizer which can generate the necessary rules:

http://www.cs.uml.edu/~mcimini/gradualizerDynamicSemantics/

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+let is_val = (d: DHExp.t): bool =>
+  switch (d) {
+  | BoundVar(_)
+  | Lam(_, _, _)
+  | ListNil(_)
+  | BoolLit(_)
+  | IntLit(_)
+  | FloatLit(_)
+  | Triv => true
+  | _ => false
```
cyrus-: pairs and injections can also be values if their arguments are values

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | Cast(d1, Hole, ty) => is_indet(d1) && ground_cases_of(ty) == Ground
+  | Cast(d, Arrow(ty1, ty2), Arrow(ty3, ty4)) =>
+    is_indet(d) && !(HTyp.eq(ty1, ty3) && HTyp.eq(ty2, ty4))
+  | FailedCast(d1, ty1, ty2) =>
+    is_final(d1)
+    && ground_cases_of(ty1) == Ground
+    && ground_cases_of(ty2) == Ground
+    && !HTyp.eq(ty1, ty2)
+  | InvalidOperation(d1, _) => is_final(d1)
+  | _ => false
```
cyrus-: need to handle products, sums, lists here as well

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+        let (ctx, d0) = decompose(d2);
+        (Pair2(d1, ctx), d0);
+      };
+    } else {
+      let (ctx, d0) = decompose(d1);
+      (Pair1(ctx, d2), d0);
+    }
+  | Let(dp, d1, d2) =>
+    if (is_final(d1)) {
+      if (is_final(d2)) {
```
cyrus-: no we don't evaluate d2 until the substitutions have been performed

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | Inj(ty, side, d1) =>
+    if (is_final(d1)) {
+      (Mark, d);
+    } else {
+      let (ctx, d0) = decompose(d1);
+      (Inj(ty, side, ctx), d0);
+    }
+  | FixF(var, ty, d1) =>
+    if (is_final(d1)) {
+      (Mark, d);
```
cyrus-: similar here, don't evaluate under binders

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | FLessThan => BoolLit(f1 < f2)
+  | FGreaterThan => BoolLit(f1 > f2)
+  | FEquals => BoolLit(f1 == f2)
+  };
+};
+// End of copying from Evaluator.re
+
+[@deriving sexp]
+type step_result =
+  | End
```
cyrus-: you can call this Final

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | FPlus => FloatLit(f1 +. f2)
+  | FMinus => FloatLit(f1 -. f2)
+  | FTimes => FloatLit(f1 *. f2)
+  | FDivide => FloatLit(f1 /. f2)
+  | FLessThan => BoolLit(f1 < f2)
+  | FGreaterThan => BoolLit(f1 > f2)
+  | FEquals => BoolLit(f1 == f2)
+  };
+};
+// End of copying from Evaluator.re
```
cyrus-: you should refactor the common things into a Evaluation_common.re module

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,504 @@
...
+  | Triv
+  | Lam(_, _, _)
+  | InvalidOperation(_, _) => End
+  //| _ => None
+  };
+
+let evaluate_step = (d: DHExp.t): step_result => {
+  let (ctx, d0) = decompose(d);
+  let res = instruction_step(d0);
+  if (res == End) {
```
cyrus-: use switch rather than if so you get help from the exhaustiveness checker

cyrus- (changes requested):
nice job so far! see comments for some things to fix.

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+        VarMap.t_(DHExp.t),
+        t,
+        list(DHExp.rule),
+        int,
+      );
+};
+
+let rec is_boxedval = (d: DHExp.t): bool =>
+  switch (d) {
+  | BoundVar(_)
```
cyrus-: variables are not values -- they stand for values but the evaluator should never encounter them because substitution should have occurred

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+    | (Ground, NotGroundOrHole(_))
+    | (NotGroundOrHole(_), Ground) => is_boxedval(d1)
+    | (NotGroundOrHole(_), NotGroundOrHole(_)) =>
+      !HTyp.eq(ty, ty') && is_boxedval(d1)
+    | (Hole, Ground) =>
+      switch (d1) {
+      | Cast(_, _, Hole) => false
+      | _ => is_boxedval(d1)
+      }
+    | _ => false
```
cyrus-: add the remaining cases explicitly

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+    ) {
+    | (Ground, Hole)
+    | (Ground, NotGroundOrHole(_))
+    | (NotGroundOrHole(_), Ground) => is_boxedval(d1)
+    | (NotGroundOrHole(_), NotGroundOrHole(_)) =>
+      !HTyp.eq(ty, ty') && is_boxedval(d1)
+    | (Hole, Ground) =>
+      switch (d1) {
+      | Cast(_, _, Hole) => false
+      | _ => is_boxedval(d1)
```
cyrus-: there aren't any values of type hole other than a cast, so we should raise an IllTyped exception in situations ruled out by that kind of reasoning

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+  | FreeVar(_)
+  | Keyword(_)
+  | InvalidText(_)
+  | EmptyHole(_, _, _) => true
+  | NonEmptyHole(_, _, _, _, d1) => is_final(d1)
+  | InvalidOperation(d1, _) =>
+    is_final(d1)
+    || (
+      switch (d1) {
+      | BinIntOp(Divide, _, IntLit(0)) => true
```
cyrus-: I think we can just return true for InvalidOperation in all cases, don't need to check d1

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+      }
+    | _ => false
+    }
+  | FailedCast(d1, _, _) => is_final(d1)
+  // && EvaluatorCommon.ground_cases_of(ty1) == Ground
+  // && EvaluatorCommon.ground_cases_of(ty2) == Ground
+  // && !HTyp.eq(ty1, ty2) // why we need this?
+  | Let(dp, d1, _) =>
+    switch (is_final(d1), Elaborator_Exp.matches(dp, d1)) {
+    | (true, Matches(_)) => true
```
cyrus-: if the pattern matches and d1 is final then it is not indet -- substitution can occur

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+          | Indet => true
+          | _ => false
+          }
+        )
+      | Cast(_, Arrow(_, _), Arrow(_, _)) => false
+      | _ => is_indet(d1) && is_final(d2)
+      }
+    )
+  | Inj(_, _, d1) => is_indet(d1)
+  | Pair(d1, d2) => is_indet(d1) && is_final(d2) || is_indet(d2)
```
cyrus-: I think it is `is_indet(d1) && is_final(d2) || is_final(d1) && is_indet(d2)`?

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+  | Cons(d1, d2) => is_indet(d1) && is_final(d2) || is_indet(d2)
```
cyrus-: same here

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+  | BinIntOp(_, d1, d2) => is_indet(d1) && is_final(d2) || is_indet(d2)
```
cyrus-: same here

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,490 @@
...
+  | BinBoolOp(_, d1, d2) => is_indet(d1) && is_final(d2) || is_indet(d2)
+  | BinFloatOp(_, d1, d2) => is_indet(d1) && is_final(d2) || is_indet(d2)
+  | InconsistentBranches(_, _, _, Case(d1, rules, n))
+  | ConsistentCase(Case(d1, rules, n)) =>
+    is_final(d1)
+    && (
+      switch (List.nth_opt(rules, n)) {
+      | None => true
+      | Some(Rule(dp, _)) => Elaborator_Exp.matches(dp, d1) == Indet
```
cyrus-: you have to keep going down the rules and find one that is Indet if the current one does not match

cyrus- (changes requested):
reviewed only is_boxedval and is_indet for now

src/hazelcore/dynamics/EvaluatorStep.re
cyrus-: or if they all do not match, it is indet due to match failure

coned: I regard each matches as a step. So, if it matches, it will be put into evaluator and in the evaluator, the index will be added one.

cyrus-: see comment on Slack -- I think it makes sense to fold is_indet and is_boxedval into the step function to avoid issues like this

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,514 @@
...
+  } else {
+    let (ctx, d0) = decompose(d);
+    switch (instruction_step(d0)) {
+    | InvalidInput(i) => InvalidInput(i)
+    | Final => Final
+    | Step(d1) => Step(compose((ctx, d1)))
+    };
+  };
+
+let evaluate_step = (d: option(DHExp.t)): option(DHExp.t) => {
```
cyrus-: this isn't used

src/hazelcore/dynamics/EvaluatorStep.re
```diff
@@ -0,0 +1,514 @@
...
+        MetaVar.t,
+        MetaVarInst.t,
+        VarMap.t_(DHExp.t),
+        t,
+        list(DHExp.rule),
+        int,
+      );
+};
+
+let rec is_boxedval = (d: DHExp.t): bool =>
```
cyrus-: you don't need this and is_indet to be defined by case analysis anymore since the step function returns whether it is indet, boxedval or steppable.

cyrus- (changes requested):

src/hazelcore/dynamics/EvaluatorStep.re
coned: I wrote a totally new code in EvaluatorStep2.re. It didn't contain any function like is_indet.

cyrus-: oh sorry, missed that!

src/hazelcore/dynamics/EvaluatorStep2.re
```diff
@@ -0,0 +1,331 @@
+open Sexplib.Std;
+
+[@deriving sexp]
+type step_status =
+  | InvalidInput(int)
```
cyrus-: lets use an exception for InvalidInput now, that will clean this up

src/hazelcore/dynamics/EvaluatorStep2.re
```diff
@@ -0,0 +1,331 @@
+open Sexplib.Std;
+
+[@deriving sexp]
+type step_status =
```
cyrus-: let's call this step_result

src/hazelcore/dynamics/EvaluatorStep2.re
```diff
@@ -0,0 +1,331 @@
...
+  | InvalidInput(int)
+  | BoxedValue(DHExp.t)
+  | Indet(DHExp.t)
+  | Step(DHExp.t);
+
+let rec step_evaluate = (d: DHExp.t): step_status =>
```
cyrus-: let's still use an evaluation context like in your other code -- that will be useful in the user interface to be able to highlight (and even allow the user to select) the next steppable expression

cyrus-: also note that "evaluate" means "turn into a value", so I wouldn't call a function like this "step_evaluate" -- just "step" is fine.

src/hazelcore/test/EvaluatorStepTests.re
```diff
@@ -0,0 +1,31 @@
+// Test suit for the step evaluator.
```
cyrus-: suit -> suite

src/hazelcore/test/EvaluatorStepTests.re
```diff
@@ -0,0 +1,31 @@
...
+    BinIntOp(BinIntOp.Divide, IntLit(5), IntLit(0)),
+    BinIntOp(BinIntOp.Times, IntLit(5), IntLit(4)),
+    BinIntOp(BinIntOp.LessThan, IntLit(5), IntLit(4)),
+    BinIntOp(BinIntOp.LessThan, BoolLit(true), IntLit(4)),
+    BinBoolOp(BinBoolOp.And, BoolLit(true), IntLit(4)),
+    BinFloatOp(BinFloatOp.FPlus, FloatLit(5.5), FloatLit(4.0)),
+  ];
+
+  List.map(testing, test) |> List.fold_left((&&), true);
+};
```
cyrus-, line 31: tests are great. let's also have an option in the user interface to run the original evaluator, this new small-step evaluator, or both with an equality check shown in the UI. that way we can keep an eye on it for a bit to make sure they aren't giving different results.

cyrus- (changes requested):

cyrus- (changes requested):
I have some small UI suggestions that we can discuss during the meeting tomorrow (perhaps this can be a starter project for Hilbert)

cyrus-:
Also needs to be brought up to date with `dev`, there are some conflicts.

cyrus-:
Archiving because this cannot be merged without significant changes into Hazel 3, but cross-referenced in #646. Thank you for the work on this!

PR #463, opened 2021-01-14
Ifs

### LINKED ISSUE #611: Booleans with if expressions

kevechen:
This adds if expressions to Hazel.

Does not compile yet.


kevechen:
@cyrus- There's still a UI bug if you type a new expression in the then or else branch, where a new line is inserted right before the expression, like a "let" expression, but I think most of our functionality looks good.

Although we might've found a bug in the let expression, where if you insert a let expression into an expression, the cursor freezes (This happens in the case expression as well as the let expression, i.e. let inside a let body)

cyrus-:
@kevechen I can't reproduce the bug relating to let on dev, can you?

kevechen:
@cyrus- Oh i misspoke, not the body expression, the e1
image

and here it is for case:
image


Not sure if these are bugs though

cyrus-:
That is what it should look like and I don't get any cursor freezing on `dev`...

kevechen:
Oh try backspacing on that let. I totally forgot to mention to backspace oops

cyrus-:
Ah yeah, that is a bug. Can you make a new issue on github?

kevechen:
Yeah I can do that

cyrus- (changes requested):
seems to work well in synthetic position!

* cannot edit guard when you have an if in analytic position
* the cursor highlight should have the sub-expressions shown as open rather than highlighting them
* needs to be brought up to date with dev, there is one seemingly small conflict to resolve

AlienKevin:
Found another UI bug: let expr in the else position gets an extra newline in front
let_under_if

Also, if you try clicking around the subexpressions and then click before the `else` keyword, the cursor breezes at that position.

AlienKevin:
Remaining issue: Cannot evaluate a number at else branch inside let:
else_inside_let
Results in the error `MlBytes {t: 0, c: '[EvaluatorError.Exception((CastBVHoleGround (IntLit 2)))]', l: 57}`

Dr. Omar sugguested to test in a function and apparently it works just fine. The let expr is said to be a little finicky right now, so I assume that the issue is not related ifs.

else_inside_function

@cyrus- Besides this issue, I think the ifs branch is ready to merge.

AlienKevin:
I fixed the last elaboration bug and tested ifs in many senarios:
* top level:
```
if true then
  1
else
  2
```

* nest inside another if's scrutinee:
```
if
  if true then
    false
  else
    true
then
  1
else
  2
```

* syn inside let:
```
let x =
  if true then
    1
  else
    2
in
x
```

* syn inside function:
```
let f =
  λx.{
    if true then
      1
    else
      2
  }
in
f 0
```

* ana inside let:
```
let x : Int =
  if true then
    1
  else
    2
in
x
```

* ana inside function:
```
let f : Int → Int =
  λx.{
    if true then
      1
    else
      2
  }
in
f 0
```

* inconsistent branches:
```
if true then
  2
else
  true
```

* nested if inside inconsistent branches:
```
if true then
  if false then
    3
  else
    4
else
  if false then
    true
  else
    false
```

cyrus-:
OK great, looks good. There is a merge conflict to resolve with `dev`. Also, clicking on the inconsistent branch types in the error message doesn't take you to the corresponding expression like it does with `case`.

AlienKevin:
> OK great, looks good. There is a merge conflict to resolve with `dev`. Also, clicking on the inconsistent branch types in the error message doesn't take you to the corresponding expression like it does with `case`.

In addition to the error message clicking issue, I discovered that there might be **some general issue with inconsistent if branches**. I cannot create **nested inconsistent if at syn position**. Hazel complains with `[Program.MissingCursorInfo]`:
nested inconsistent if at syn position

Nor **inconsistent else branch at ana position**. Hazel complains with `[EvaluatorError.Exception((CastBVHoleGround (IntLit 3)))`:
inconsistent else branch at ana position

cyrus-:
@AlienKevin do you want to try to finish this up? if not, will likely wait until hazel3 to add if expressions

cyrus-:
Nice work everyone. Closing due to subsumed by haz3l code.

PR #469, opened 2021-02-02
pattern annotation letline holes fix

cyrus-:


cyrus-:
@disconcision I can't press ">" to create an arrow type here:

Screenshot from 2021-02-01 23-30-17

It works if I parenthesize `Int` first though...

cyrus-:
actually it works on the left side of the type, but not the right.

disconcision:
Fixed; hadn't updated the pattern ana_perform construct case with the new logic around sort-based action delegation. Only became an issue after the switch to ana_performing for letline patterns

PR #471, opened 2021-02-10
Functorize weakmap

nj-wilson:
Functorize the WeakMap module (probably needs to be renamed) to allow for other map types to be used.

This will allow for code not compiled with js_of_ocaml to make use of the module, and less dependent on hazel-web.


nj-wilson:
Still a WIP, code using this module is still being updated.

cyrus-:
@nj-wilson is this still an active PR?

nj-wilson:
No, a different method was used, which doesn't necessitate this.

PR #473, opened 2021-02-11
Smyth sketch-sensitive

NiekM:
Currently, smyth is not entirely sketch-sensitive: the synthesizer keeps track of some state, which is not entirely recoverable from a sketch. This means that, even though a sketch is an intermediate step towards a solution, the synthesizer might not find that solution when starting from that sketch, because it lacks the relevant state.
A solution to this problem is to use the same state during type checking and store it in the holes so that its available to the synthesizer.

Ideally, the synthesizer should be monotone in the sense that adding more relevant information should not hamper the synthesis. It is not clear whether this is possible or general. More assertions could lead to timeouts and some intermediate states are nonsensical (e.g. `?? ??`).

### DIFF
--- a/src/hazelcore/Shmyth.re
+++ b/src/hazelcore/Shmyth.re
@@ -891,8 +891,8 @@ let solve_all = (e: UHExp.t): option(hole_fillings) => {
     |> List.map(hole_filling =>
          hole_filling
          |> List.map(((u, smexp)) => {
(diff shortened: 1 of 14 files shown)


cyrus-:
don't believe this is being updated any further, so closing it. feel free to reopen if I am incorrect.

PR #474, opened 2021-03-02
Enrich monad implementation for more reuse

bkase:
I converted existing monads and added one for ActionOutcome. This
formation is similar to the one used in Jane Street's Base library.

The Make functor derives everything you'd want (syntax, extra
functions) based on a basic monad definition.

### DIFF
--- a/src/hazelcore/ActionOutcome.re
+++ b/src/hazelcore/ActionOutcome.re
@@ -1,28 +1,5 @@
-module T = {
-  [@deriving sexp]
-  type t('success) =
-    | Succeeded('success)
-    | CursorEscaped(Side.t)
-    | Failed;
-
-  let map =
-    `Custom(
-      (x, f) =>
-        switch (x) {
-        | Succeeded(a) => Succeeded(f(a))
-        | CursorEscaped(s) => CursorEscaped(s)
-        | Failed => Failed
-        },
-    );
-
-  let bind = (x, f) =>
-    switch (x) {
-    | Succeeded(a) => f(a)
-    | CursorEscaped(s) => CursorEscaped(s)
-    | Failed => Failed
-    };
-
-  let return = a => Succeeded(a);
-};
-include T;
-include Monads.Make(T);
+[@deriving sexp]
+type t('success) =
+  | Succeeded('success)
+  | CursorEscaped(Side.t)
+  | Failed;
--- a/src/hazelcore/ActionOutcome.re
+++ b/src/hazelcore/ActionOutcome.re
@@ -1,4 +1,28 @@
-type t('success) =
-  | Succeeded('success)
-  | CursorEscaped(Side.t)
-  | Failed;
+module T = {
+  [@deriving sexp]
+  type t('success) =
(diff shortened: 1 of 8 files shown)


src/hazelcore/Monads.re
```diff
@@ -6,9 +6,39 @@
...
   type t('a);
   let return: 'a => t('a);
   let bind: (t('a), 'a => t('b)) => t('b);
+  let map: [ | `Define_using_bind | `Custom((t('a), 'a => 'b) => t('b))];
+};
+
+module type MONAD = {
+  include MONAD_BASIC;
+  let map: (t('a), 'a => 'b) => t('b);
+  let zip: (t('a), t('b)) => t(('a, 'b));
```
bkase, line 25: I like to call this `zip` but I saw it was called `product` in another file. Is `zip` okay here?

cyrus-: zip makes sense yeah

bkase:
I removed the sexp requirement from MONAD -- this way folks can derive whatever they want on the underlying carrier type (is that the right vocabulary?)

src/hazelcore/ActionOutcome.re
```diff
@@ -1,4 +1,28 @@
...
+    | Succeeded('success)
+    | CursorEscaped(Side.t)
+    | Failed;
+
+  let map =
+    `Custom(
+      (x, f) =>
+        switch (x) {
+        | Succeeded(a) => Succeeded(f(a))
+        | CursorEscaped(s) => CursorEscaped(s)
```
cyrus-: don't think we generally respond to CursorEscaped by passing it along -- each zipper case should decide how to respond to this situation in each sub-expression position carefully. suggests that maybe ActionOutcome shouldn't form a monad directly, but you could refactor it into an option(success) sort of thing

src/hazelcore/Monads.re
```diff
@@ -6,9 +6,38 @@
 
    In any case, that's a good reference. */
 
-module type MONAD = {
-  [@deriving sexp]
+module type MONAD_BASIC = {
   type t('a);
   let return: 'a => t('a);
   let bind: (t('a), 'a => t('b)) => t('b);
+  let map: [ | `Define_using_bind | `Custom((t('a), 'a => 'b) => t('b))];
```
cyrus-: why not use an either type or define a datatype here? hesitant to start using polymorphic variants because it will make it trickier to self-host hazelcore...

cyrus- (changes requested):

src/hazelcore/ActionOutcome.re
bkase: okay I'll remove the monad instance on this for now, and refactor it to an option-like thing with the type variables work

src/hazelcore/Monads.re
bkase: good point about self-hosting, I'll fix it -- but seriously, why don't more languages have anonymous sums if they have anonymous products.

cyrus-:
merged, thanks!

PR #478, opened 2021-03-04
Type Variables (3)

bkase:


cyrus-:
@bkase is the intention still to merge this before the type aliases PR or is this superceded now?

bkase:
> @bkase is the intention still to merge this before the type aliases PR or is this superceded now?

I figure it's good to merge this if it's good as is so there's not more conflicts just in case it takes a bit to resolve the remaining item on type-aliases

cyrus- (changes requested):
Can't use type operators anymore, e.g. pressing `>` to create an arrow type no longer works.

cyrus-:
Also need to get up to date with a few recent dev commits.

src/hazelcore/Elaborator_Typ.re
```diff
@@ -0,0 +1,272 @@
...
+    let (tys, ks, ds) = ListUtil.unzip3(rs);
+    let+ () = ks |> List.for_all(k => k == Kind.Type) |> OptUtil.of_bool;
+    let delta =
+      ds |> List.fold_left((d1, d2) => Delta.union(d1, d2), Delta.empty);
+    (HTyp.Prod(tys), Kind.Type, delta);
+  | BinOp(_, op, skel1, skel2) =>
+    /* TElabSBinOp */
+    let* (ty1, k1, d1) = ana_skel(ctx, delta, Kind.Type, skel1, seq);
+    let* (ty2, k2, d2) = ana_skel(ctx, delta, Kind.Type, skel2, seq);
+    switch (k1, k2) {
```
dedbox, line 37: When I press ">" on a type hole, the type of the operation is consistent (i.e., neither call to ana_skel returned None) but the kinds of k1 and k2 are both KHole. In this case, syn_skel should succeed.

dedbox: The same is true when I press "," on a type hole, which is handled similarly in the preceding case.

cyrus-:
Closing this PR -- we are folding this into the type alias PR.

PR #495, opened 2021-04-07
refactored Monads module to avoid dynamic dispatch for map

cyrus-:


src/hazelcore/Monads.rei
```diff
@@ -1,30 +1,36 @@
...
     let ( let* ): (t('a), 'a => t('b)) => t('b);
     let (let+): (t('a), 'a => 'b) => t('b);
     let (and+): (t('a), t('b)) => t(('a, 'b));
   };
 };
 
-module Make: (M: MONAD_BASIC) => MONAD with type t('a) := M.t('a);
+// See ActionOutcome.re{i} for an example
+module MakeB: (M: MONAD_BASIC) => MONAD with type t('a) = M.t('a);
+module MakeZ: (M: MONAD_ZIP) => MONAD with type t('a) = M.t('a);
```
bkase: Maybe it would be simpler to only put the `Make*` ML-functors in the rei and keep the Functor_of_Basic and Zip_of_Functor private since a consumer of this library only really wants a module conforming to `MONAD` in the end and doesn't ever really care about the intermediate pieces.

Then you can just expose these functors:

```
module MakeOfBasic: (M: MONAD_BASIC) => // ...
module MakeOfFunctor: (M: MONAD_FUNCTOR) => // ...
module MakeOfZip: (M: MONAD_ZIP) => // ...
```

src/hazelcore/Monads.rei
```diff
@@ -1,30 +1,36 @@
...
+
+module type MONAD_ZIP = {
+  include MONAD_FUNCTOR;
   let zip: (t('a), t('b)) => t(('a, 'b));
+};
+
+module Functor_of_Basic:
+  (M: MONAD_BASIC) => MONAD_FUNCTOR with type t('a) = M.t('a);
+module Zip_of_Functor:
+  (M: MONAD_FUNCTOR) => MONAD_ZIP with type t('a) = M.t('a);
```
bkase: I would recommend hiding these (see note below)

src/hazelcore/Monads.re
```diff
@@ -6,44 +6,51 @@
...
-  let map: MapDefinition.t((t('a), 'a => 'b) => t('b));
 };
 
-module type MONAD = {
+module type MONAD_FUNCTOR = {
   include MONAD_BASIC;
   let map: (t('a), 'a => 'b) => t('b);
+};
+
+module Functor_of_Basic = (M: MONAD_BASIC) => {
```
bkase: nit: I prefer the style where you prefix all ML functors with `Make_` -- this way you can instantly know if you're referring to a module or a functor without having to go to the call-site (in the same way you know you're referring to a module when the symbol has a capital letter).

(Jane Street code uses this style)

bkase (changes requested):
I think this change is a great way to force static dispatch but still retain a lot of the power of the Monads module.

But -- are you certain OCaml's optimizer doesn't optimize out the dynamic dispatch? I feel like it's a trivial optimization to go from: `let map = Custom(f)` => `let map = f`. I don't know for sure though, but we could probably figure out by examining compiler artifacts. I'd be surprised that Jane Street didn't make a similar change in their `Base` library if the optimizer misses this opportunity.

A plus of the old approach is that we could also use the `Define_using_bind |  Custom` for zip and then you can do any permutation of custom or generated map,zip, or anything else. A con is it's a little awkward to wrap the definitions in that extra variant, but on the other hand you don't have to think about the functors as much.

The rest of the comments assume OCaml doesn't optimize this and therefore we want to make this sort of a change.

PR #502, opened 2021-05-04
Update function syntax from `λx.{...}` to `fun x {...}`

minchingtonak:
Pretty self-explanatory, all necessary features are implemented.

The only other thing I want to do is move common textual representations for Hazel constructs to their own file so that there aren't duplicated magic string literals at a couple points in the code base. I've done it just for function syntax (see `Doc_common.re`, `DHDoc_common.re`, `UHDoc_common.re` for defs and usage) but I think it would be a good idea to move all of the syntax over to `Doc_common.re`. That's probably best left for a separate PR, but please let me know if you think I should include it in this PR.


src/hazelcore/Action_Exp.re
```diff
@@ -74,6 +74,7 @@ let keyword_action = (kw: ExpandingKeyword.t): Action.t =>
   switch (kw) {
   | Let => Construct(SLet)
   | Case => Construct(SCase)
+  | Lam => Construct(SLam)
```
cyrus-: Let's call it "Fun" instead of "Lam" given the concrete syntax

src/hazelweb/layout/Doc_common.re
```diff
@@ -0,0 +1,6 @@
+module Delim = {
+  let sym_Lam = "fun";
```
cyrus-: Fun here too

cyrus- (changes requested):
looking close, but needs to be brought up to date with dev in a big way

cyrus-:
@minchingtonak do you want to try to wrap this up? looks quite close -- just a few changes above, plus merging in the latest `dev` changes.

minchingtonak:
Yep, thanks for the reminder! I'll start this today.

minchingtonak:
I refactored Lam to Fun in ExpandingKeyword.t and Action.t. I'm not sure whether you want me to change Action.t so I can rool that back if need be. I also changed the string constants from Lam -> Fun in Doc, DHDoc, and UHDoc files.

cyrus- (approved):

cyrus- (changes requested):
Whoops, one last late-breaking change needed now that we have Hannah's work on the strategy guide working -- the entries for creating function literals should say `fun ` in addition to `\`.

cyrus-:
Screenshot from 2021-11-05 13-36-06

cyrus- (changes requested):
Looks great. One last issue: the undo history leaves a "insert var: fun" entry after pressing space, whereas with other keywords like "let" and "case" it overrides the previous entry.

src/hazelcore/Statics_Exp.re
```diff
@@ -768,9 +768,10 @@ and syn_fix_holes_operand =
       | NotInVarHole =>
         let (u, u_gen) = MetaVarGen.next(u_gen);
         let reason: VarErrStatus.HoleReason.t =
-          switch (Var.is_let(x), Var.is_case(x)) {
-          | (true, _) => Keyword(Let)
-          | (_, true) => Keyword(Case)
+          switch (Var.is_let(x), Var.is_case(x), Var.is_fun(x)) {
```
cyrus-: Lets use ExpandingKeyword.mk here so there is only one place we need to modify when there is a new keyword added to the system

src/hazelcore/Var.re
```diff
@@ -25,10 +25,14 @@ let is_false = eq("false");
 
 let is_let = eq("let");
 
+let is_fun = eq("fun");
+
 let is_case = eq("case");
 
 let is_wild = eq("_");
 
+let is_keyword = s => is_let(s) || is_fun(s) || is_case(s);
```
cyrus-: Let's change this to use ExpandingKeyword.mk -- same goal, to have it so we don't need to find multiple places in the codebase to change when a new keyword is added

src/hazelweb/layout/DHDoc_Util.re
```diff
@@ -58,10 +58,10 @@ module Delim = {
   let open_Parenthesized = mk("(");
   let close_Parenthesized = mk(")");
 
-  let sym_Lam = mk(Unicode.lamSym);
-  let colon_Lam = mk(":");
-  let open_Lam = mk(".{");
-  let close_Lam = mk("}");
+  let sym_Fun = mk(Unicode.lamSym);
+  let colon_Fun = mk(":");
+  let open_Fun = mk(".{");
```
cyrus-: this is inconsistent with the UHExp pretty printing -- no dot there

cyrus- (changes requested):
Just a few last code cleanups and this looks good to go!

cyrus- (approved):
Looks great!

PR #505, opened 2021-05-13
Type alias (2)

bkase:


cyrus-:
Well, I can't break the singleton stuff on a quick play through!

Probably a small thing to fix: I don't think we want to allow bindings of built in type names like Int:

`type Int = ? in ...`

It doesn't actually rebind Int, so it should probably show up as an error hole on the TVar.

cyrus-:
Marking as draft pending type-variable-3 updates.

dedbox (changes requested):
Outstanding issues:

- Can't edit type alias name (at the binding site) if it is used in the body (but uses in the bound type are fine).
- When a type alias name is used before it is bound, the free occurrences retain their error decorations after binding.
- Should type alias names with initial minuscule be an error, to prevent confusion with type variables?
- Built-in type names can be rebound, but the type system appears to ignore the rebinding.
  - Should shadowing *any* existing type name be an error, or just the built-in types?
- Can't create injections in positions where the expected type is an alias to a sum type.
- Aliases to product types survive type checking but instances of the type have error decorations.
- Functions whose expected type is an alias to a function type are not recognized as such.

cyrus-:
@dedbox did you move your changes to a new branch? if so can you close this PR and make a new one

dedbox:
Moved to #581

PR #510, opened 2021-06-04
Refactor KeyComboAction.get to use a switch expression instead of a hashtable

yottalogical:
Previously, `KeyComboAction.get` used a static hashtable to emulate the behavior of a switch expression. As far as I can tell, this has no particular advantages, and it means that the exhausiveness checker won't work on it.

There were some constructors of `HazelKeyCombos.t` that weren't in the original hashtable. Since a switch expression requires exhaustiveness, I had these inputs raise the exception `Not_found`, since that's identical behavior to what the hashtable lookup would have done. If anyone has a better idea of what to do in these cases, let me know.

### DIFF
--- a/src/hazelweb/gui/KeyComboAction.re
+++ b/src/hazelweb/gui/KeyComboAction.re
@@ -1,5 +1,5 @@
 let get_model_action =
-    (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t, is_mac: bool)
+    (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t, is_mac: option(bool))
     : option(ModelAction.t) =>
   switch (kc, cursor_info, is_mac) {
   | (Escape, _, _) => None
@@ -35,24 +35,27 @@ let get_model_action =
   | (Alt_R, _, _) => Some(EditAction(Construct(SInj(R))))
   | (Alt_C, _, _) => Some(EditAction(Construct(SCase)))
   | (Pound, _, _) => Some(EditAction(Construct(SCommentLine)))
-  | (Ctrl_Z, _, true) => None
-  | (Ctrl_Z, _, false) => Some(Undo)
-  | (Ctrl_Shift_Z, _, true) => None
-  | (Ctrl_Shift_Z, _, false) => Some(Redo)
+  | (Ctrl_Z, _, Some(true)) => None
+  | (Ctrl_Z, _, Some(false)) => Some(Undo)
+  | (Ctrl_Z, _, None) => None
+  | (Ctrl_Shift_Z, _, Some(true)) => None
+  | (Ctrl_Shift_Z, _, Some(false)) => Some(Redo)
+  | (Ctrl_Shift_Z, _, None) => None
   | (Ctrl_Alt_I, _, _) => Some(EditAction(SwapUp))
   | (Ctrl_Alt_K, _, _) => Some(EditAction(SwapDown))
   | (Ctrl_Alt_J, _, _) => Some(EditAction(SwapLeft))
   | (Ctrl_Alt_L, _, _) => Some(EditAction(SwapRight))
-  | (Meta_Z, _, true) => Some(Undo)
-  | (Meta_Z, _, false) => None
-  | (Meta_Shift_Z, _, true) => Some(Redo)
-  | (Meta_Shift_Z, _, false) => None
+  | (Meta_Z, _, Some(true)) => Some(Undo)
+  | (Meta_Z, _, Some(false)) => None
+  | (Meta_Z, _, None) => None
+  | (Meta_Shift_Z, _, Some(true)) => Some(Redo)
+  | (Meta_Shift_Z, _, Some(false)) => None
+  | (Meta_Shift_Z, _, None) => None
   };
 
 let get_action =
     (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t): option(Action.t) =>
-  // is_mac shouldn't matter here, so we'll just have it be false
-  switch (get_model_action(cursor_info, kc, false)) {
+  switch (get_model_action(cursor_info, kc, None)) {
   | Some(EditAction(action)) => Some(action)
   | _ => None
   };
--- a/src/hazelweb/gui/KeyComboAction.re
+++ b/src/hazelweb/gui/KeyComboAction.re
@@ -50,9 +50,9 @@ let get_model_action =
   };
 
 let get_action =
-    (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t): option(Action.t) =>
-  // is_mac shouldn't matter here, so we'll just have it be false
-  switch (get_model_action(cursor_info, kc, false)) {
+    (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t, is_mac: bool)
+    : option(Action.t) =>
+  switch (get_model_action(cursor_info, kc, is_mac)) {
   | Some(EditAction(action)) => Some(action)
   | _ => None
   };
--- a/src/hazelweb/gui/KeyComboAction.re
+++ b/src/hazelweb/gui/KeyComboAction.re
@@ -1,42 +1,50 @@
-let get = (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t): option(Action.t) =>
-  switch (kc, cursor_info) {
-  | (Escape, _) => None
-  | (Backspace, _) => Some(Backspace)
-  | (Delete, _) => Some(Delete)
-  | (ShiftTab, _) => Some(MoveToPrevHole)
-  | (Tab, _) => Some(MoveToNextHole)
-  | (GT, {CursorInfo.typed: OnType, _}) => Some(Construct(SOp(SArrow)))
-  | (GT, _) => Some(Construct(SOp(SGreaterThan)))
-  | (Ampersand, _) => Some(Construct(SOp(SAnd)))
-  | (VBar, _) => Some(Construct(SOp(SOr)))
-  | (LeftParen, _) => Some(Construct(SParenthesized))
-  | (Colon, _) => Some(Construct(SAnn))
-  | (Equals, _) => Some(Construct(SOp(SEquals)))
-  | (Enter, _) => Some(Construct(SLine))
-  | (Shift_Enter, _) => Some(Construct(SCommentLine))
-  | (Backslash, _) => Some(Construct(SLam))
-  | (Plus, _) => Some(Construct(SOp(SPlus)))
-  | (Minus, _) => Some(Construct(SOp(SMinus)))
-  | (Asterisk, _) => Some(Construct(SOp(STimes)))
-  | (Slash, _) => Some(Construct(SOp(SDivide)))
-  | (LT, _) => Some(Construct(SOp(SLessThan)))
-  | (Space, {CursorInfo.cursor_term: Line(_, CommentLine(_)), _}) =>
-    Some(Construct(SChar(" ")))
-  | (Space, _) => Some(Construct(SOp(SSpace)))
-  | (Comma, _) => Some(Construct(SOp(SComma)))
-  | (LeftBracket, {CursorInfo.typed: OnType, _}) => Some(Construct(SList))
-  | (LeftBracket, _) => Some(Construct(SListNil))
-  | (Semicolon, _) => Some(Construct(SOp(SCons)))
-  | (Alt_L, _) => Some(Construct(SInj(L)))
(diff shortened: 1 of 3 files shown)


yottalogical:
@dmoon1221, looking through the Git history, I saw that you were the original author of this file.

Do you know of any particular reason why we'd be better off using the hashtable approach here? I just thought it would be worth asking first just to be safe.

dm0n3y (changes requested):
@yottalogical thanks for getting this going. You're right to point out the use of a hash table here, it's a relic of ancient times. Instead of raising an exception, though, let's change the type signature and return `option(Action.t)`. Then you'll want to rearrange the `switch` expression [here](https://github.com/hazelgrove/hazel/blob/02a534cbb580b271adf88158090e2d0f5d1a8d6e/src/hazelweb/gui/UHCode.re#L118) so that it first does a case analysis of the result of `KeyComboAction.get`, does [this case](https://github.com/hazelgrove/hazel/blob/02a534cbb580b271adf88158090e2d0f5d1a8d6e/src/hazelweb/gui/UHCode.re#L143) if it gets back an action, otherwise it does the rest of the special cases enumerated there (which could also be cleaned up with some `when` guards).

yottalogical:
> Then you'll want to rearrange the `switch` expression [here](https://github.com/hazelgrove/hazel/blob/02a534cbb580b271adf88158090e2d0f5d1a8d6e/src/hazelweb/gui/UHCode.re#L118) so that it first does a case analysis of the result of `KeyComboAction.get`, does [this case](https://github.com/hazelgrove/hazel/blob/02a534cbb580b271adf88158090e2d0f5d1a8d6e/src/hazelweb/gui/UHCode.re#L143) if it gets back an action, otherwise it does the rest of the special cases enumerated there (which could also be cleaned up with some `when` guards).

I gave this approach a shot, but it ended up with messier code than before. The big problem is that `HazelKeyCombos.of_evt` does have to be computed first, since its result is used as an input to `KeyComboAction.get`. This then means that the `None` case and the Undo/Redo cases of `HazelKeyCombos.of_evt` are considered separately. It also means that all the other cases that aren't Undo/Redo need to be considered if `KeyComboAction.get` returns `None`. You can see how this turned out in commit [`e508e96`](https://github.com/hazelgrove/hazel/pull/510/commits/e508e9618bdbd9481f9eff9789fc11d4da7c40b7).

I think it would be better off if we just left things in the order that they were, and just make the minimal changes necessary to account for `KeyComboAction.get` returning an `option(Action.t)`. I went with this approach in commit [`5742b5c`](https://github.com/hazelgrove/hazel/pull/510/commits/5742b5c307a59c018bfa21a79ad8761e05f938fa).

cyrus-:
It may be cleaner to have the main switch return an UpdateAction rather than an Action, so we don't need to have a separate switch for the undo/redo stuff?

dm0n3y:
The change you made following my suggestion looks reasonable to me, especially after you folded in the if statements into the switch in your last commit.

That said, I think @cyrus-'s suggested approach is the right one here. With a little more work, we could also fold into `KeyComboAction` the movement and character entry cases too, so that `on_keydown` handler in `UHCode.re` is just
```
evt =>
  switch (HazelKeyCombos.of_evt(evt)) {
  | None => Event.Ignore
  | Some(kc) => prevent_stop_inject(KeyComboAction.get(kc, cursor_info))
  }
```
Here I'm assuming that, with all these changes together, `KeyComboAction.get` doesn't return an `option(Action.t)` as I suggested initially but rather `ModelAction.t`. This would require extending the `HazelKeyCombos.t` datatype with variants `SingleKey(JSUtil.single_key)` and `MoveKey(MoveKey.t)` and extending `HazelKeyCombos.of_evt` to emit them.

yottalogical:
So the most recent commit [`3ff1c8d`](https://github.com/hazelgrove/hazel/pull/510/commits/3ff1c8d90eab0ba19227e9b277bf64a53749a8d3), I gave that approach a try. I did have to alter it somewhat to take `is_mac` as an input and return an `option(ModelAction.t)` because there are some key combos that don't have a corresponding `ModelAction.t` (`Ctrl_Z` on a Mac, `Meta_Z` not on a Mac, `Escape` anywhere, etc).

I'm also somewhat concerned about the `ActionPanel` module, since it also uses `KeyComboAction.get` in multiple places, and I'm not sure how well it would work with the function returning an `option(ModelAction.t)`. Perhaps we could have 2 functions in the `KeyComboAction` module, one that returns an `option(ModelAction.t)` and one that returns an `option(Action.t)`. The latter could be fairly simple wrapper around the former.

yottalogical:
> Perhaps we could have 2 functions in the KeyComboAction module, one that returns an option(ModelAction.t) and one that returns an option(Action.t). The latter could be fairly simple wrapper around the former.

I actually went ahead and tried this out in commit [`c238cf5`](https://github.com/hazelgrove/hazel/pull/510/commits/c238cf58986641742f89c3874eb25a981d216b6a). I think it's a fairly reasonable solution, but I'd like to hear your input.

The code is very close to building again. I just need to account for `KeyComboAction.get_action` returning `option(Action.t)` instead of `Action.t`.

cyrus-:
Looks reasonable. Rather than taking is_mac as input, though, you could
just take the entire model as input.

yottalogical:
@cyrus- What do you mean by taking the entire model as input?

yottalogical:
I think I've pretty much achieved what @dmoon1221 was recommending above in commit [`bad4893`](https://github.com/hazelgrove/hazel/pull/510/commits/bad489379b67d73fc905b698dc234e4cab7c8127). The logic to generate an `option(ModelAction.t)` (regardless of whether it's due to a move, combo, or single key) has been encapsulated so that it can be acted upon with a very simple switch expression:

```
switch (model_action) {
| Some(model_action) => prevent_stop_inject(model_action)
| None => Event.Ignore
};
```

The big difference is that I left this logic inside of `UHCode.key_handlers`. Trying to put it into `KeyComboAction.get_model_action` would have made `UHCode.key_handlers` a bit tidier, but on the flip side it would have made both `HazelKeyCombos` and `KeyComboAction.get_model_action` messier. At least when I tried it, things were only made worse.

src/hazelweb/gui/KeyComboAction.re
```diff
@@ -1,56 +1,58 @@
...
+  | (Meta_Shift_Z, _, false) => None
+  };
 
-let get = (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t): Action.t => {
-  let action_of_ci = Hashtbl.find(table, kc);
-  action_of_ci(cursor_info);
-};
+let get_action =
+    (cursor_info: CursorInfo.t, kc: HazelKeyCombos.t): option(Action.t) =>
+  // is_mac shouldn't matter here, so we'll just have it be false
```
cyrus-: this is just an accident of the current set of actions though -- lets set it up to propagate that information correctly.

src/hazelweb/gui/UHCode.re
```diff
@@ -111,51 +111,27 @@ let key_handlers =
...
-          }
-        | Some(kc) =>
-          prevent_stop_inject(
-            ModelAction.EditAction(KeyComboAction.get(cursor_info, kc)),
+      let model_action: option(ModelAction.t) = {
+        let move_key = MoveKey.of_key(Key.get_key(evt));
+        let key_combo = HazelKeyCombos.of_evt(evt);
+        let single_key = JSUtil.is_single_key(evt);
+        switch (move_key, key_combo, single_key) {
+        | (Some(move_key), _, _) => Some(MoveAction(Key(move_key)))
```
cyrus-, line 119: can we fold this move_key stuff into the same place too?

src/hazelweb/gui/UHCode.re
```diff
@@ -111,51 +111,27 @@ let key_handlers =
...
+        | (_, Some(key_combo), _) =>
+          KeyComboAction.get_model_action(cursor_info, key_combo, is_mac)
+        | (_, _, Some(single_key)) =>
```
cyrus-, line 122: this too -- might as well have all the key handling code in one place

src/hazelweb/gui/KeyComboAction.re
yottalogical: In commit [`78dd45c`](https://github.com/hazelgrove/hazel/pull/510/commits/78dd45c37c61c4cd68205121e8167e821e8e910a) I made it so that instead of `is_mac` being a `bool`, it's an `option(bool)` so that it can account for applications of the function where it's not known whether the user is on a Mac, such as in `ActionPanel.re`.

Is this what you were expecting?

cyrus-: Well I was suggesting that some edit actions might actually need that information in the future, so lets set it up so that is being passed in correctly (which might require changing something upstream of this function to propagate that correctly.)

src/hazelweb/gui/UHCode.re
disconcision: I feel like there are a bunch of modules which do key handling which could be folded together... The changes I'm making in uhcode add a bunch of cases, so for simplicity I made an ad-hoc wrapper around these various modules which presents a unified key data structure. I feel most of /keyboard could be merged into a unified structure, presenting a single outward interface which maps an evt to either a unified key data structure, or even maybe directly to a modelAction.

src/hazelweb/gui/KeyComboAction.re
yottalogical: Okay, I think that makes sense. I made a new commit ([`fb7261`](https://github.com/hazelgrove/hazel/commit/fb7261a5242b0232a73e7316a04558e1df81fec2)) that does this.

yottalogical:
Another thing that I've been trying to do is to get rid of the build errors in `ActionPanel.re` that exist because `KeyComboAction.get_action` now returns `option(Action.t)` instead of `Action.t`.

Something I've noticed is that `None` should never be returned if the function is used correctly, since previously, the `None` case was just an exception that crashed the whole program. Even when I try to handle it safely, there's really not anything that can be done besides showing the user an error message that the input wasn't valid.

Because returning `None` just means the function was used incorrectly, I don't really think it makes sense to return an `option(Action.t)` at all. That way, correct usage of the function is enforced by the compiler.

But since some constructors of `HazelKeyCombos.t` can't evaluate to an `Action.t` in all situations, we'd need a different input type that represents key combos that can always can evaluate to an `Action.t`. I don't think I'm familiar enough with the codebase to come up with the best specific way to actually do this, so I'm looking from any advice and/or wisdom before I give it a try.

disconcision:
Hey! There are a couple things you could do to make your big switch a little cleaner. One is to define and use some quick helpers to remove redundancies, such as:

`let construct = action => Some(EditAction(Construct(action)))`
(for example, you may or may not want to include the Some, etc...)

Also, I'd suggest not switching on is_mac. Instead, use when clauses:
`| (Ctrl_Z, _, true) => None`
becomes
`| (Ctrl_Z, _) when is_mac => None`

You could do similar for cursorinfo as well... this is generally good practice when certain values are used only in a few cases. Here you would need to define a couple new predicates:
`let is_on_type = cursor_info => ...
let is_on_comment_line = cursor_info => ...`

dm0n3y (commented):
This looks good overall, thanks! See @disconcision's comments about reorganizing that switch expression. As for the `KeyComboAction.get_action` thing, I would just move that function into `ActionPanel` and throw an exception if the input doesn't get you an `Action.t`, the perspective being that it's a utility function to be handled responsibly within `ActionPanel` rather than a general utility function for all clients of `KeyComboAction`.

yottalogical:
I implemented the changes recommended by @disconcision in commit [`fb7261`](https://github.com/hazelgrove/hazel/commit/fb7261a5242b0232a73e7316a04558e1df81fec2).

yottalogical:
I implemented the changes recommended by @disconcision in commit [`03a8088`](https://github.com/hazelgrove/hazel/pull/510/commits/03a8088f2b003afbafc8f3df75a11d8665c7baaa).

yottalogical:
I implemented @dmoon1221's most recent recommendations in commit [`b0397e4`](https://github.com/hazelgrove/hazel/pull/510/commits/b0397e4f81789fa54120268bb172f75525d893d7). I also made sure that when it fails, it does so with an error message mentioning what input caused the error. This should help anyone who uses it incorrectly quickly understand what the problem they are experiencing is, and potentially how to fix it. You can try it out by adding something invalid (like `Ctrl_Z`) to the action panel.

Also, as of this commit, the code builds again! The inconsistent types (`option(Action.t)` and `Action.t`) in `ActionPanel.re` were preventing the code from building for quite a while, but that has been addressed now.

dm0n3y (approved):
Looks great, thanks!

PR #514, opened 2021-06-24
Partial Application of Non-Curried Functions

yottalogical:
*Note: This is a just a preliminary plan for how this feature will be implemented. Things may change if anyone comes up with better ideas. See the discussion on this pull request (once there is some).*

If a function takes multiple arguments by currying, then it can be partially applied, but the arguments must be given in order. If a function takes multiple arguments by taking in a tuple, then it cannot be partially applied natively. The programmer *could* construct a new function literal that takes in the unapplied arguments, then inserts them where they belong in the application of the original function, but this rather cumbersome. It could be replaced with some sugar to simplify it.

The purpose of this pull request is to allow for non-curried functions to be partially applied. This will be (subjectively) more elegant than partial application of curried functions, since it will allow for arguments to be partially applied in any order.

It will work by introducing a deferral expression: `~`, which is an external expression. If a function that takes an n-tuple as input is applied to an n-tuple, some of the expressions in the tuple can be replaced with deferrals. This indicates that that argument is being skipped and will be applied later. The partial application will evaluate to a new function that takes in the skipped arguments and outputs the result of the original function applied to *all* arguments.

The deferral cannot synthesize to or analyze against any type. This is fine, as the only place it can exist (without being wrapped in an expression hole) is in an n-tuple literal that is being applied to a function that takes an n-tuple as input.

During elaboration, the partial application is replaced with a new function literal that takes in the unapplied arguments, then inserts them where they belong in the application of the original function. This process is the expansion of the sugar that was described in the first paragraph. Since this is all just sugar, the internal expression language doesn't have to be modified at all.

### Examples

```
val f: ('a, 'b, 'c) → 'd
val x: 'a
val y: 'b
val z: 'c
```
|External Expression|Internal Expression|Synthesized Type|
|:--|:--|:--|
|`f (x, y, z)`|`f (x, y, z)`|`'d`|
|`f (x, ~, z)`|`λb.{f (x, b, z)}`|`'b → 'd`|
|`f (~, y, ~)`|`λ(a, c).{f (a, y, c)}`|`('a, 'c) → 'd`|
|`f (~, ~, ~)`|`λ(a, b, c).{f (a, b, c)}`|`('a, 'b, 'c) → 'd`|


cyrus-:
@yottalogical this looks reasonable to me. I was imagining using underscore rather than tilde for the placeholders -- any particular reason to prefer tildes?

yottalogical:
The reason I'm using tildes for now is to avoid confusion with variable names. This won't be a problem when it's finished, but I wanted to avoid any problems that I might experience during development.

I intend to change it to underscores before merging to `dev`.

cyrus-:
Ah I see. OK, that's fine for now. Underscores shouldn't be too bad, we already parse read them out specially because of how they are used in patterns.

cyrus-:
Archiving because this cannot be merged as-is into Hazel 3, but the design is sound. See Hazel 3 issue #795. Thank you for the work on this @yottalogical !

Issue #280, opened 2020-05-24
cannot backspace/delete over space


cyrus-:
![Screenshot from 2020-05-24 00-02-55](https://user-images.githubusercontent.com/280638/82745415-f960bd80-9d51-11ea-9599-b78821ecefc8.png)

pressing backspace here should delete the space and hole 1, but it currently only moves you over to hole 1. it works correctly on `master`.

Same thing in the other direction with Delete.

PR #516, opened 2021-06-28
Fix for cannot backspace over space

disconcision:
This (partially) addressed #280. This is a bit of a hack; I feel there may be no great way to address this in the current implementation.

The issue is the line-handling code in the block-level perform functions captures Deletes/Backspaces actions made at the end/beginning of the line. This seems reasonable to me; the issue feels more like it arises due to the fact that we are relying on deleting holes from the 'wrong side'; i.e., the way hole removal with backspace works right now, the cursor has to be on the left of the hole. This feels off to me?

Nonetheless, this PR fixes the issue I think while maintaining previous hole removal behavior. I only made the fix for the backspace expr(syn+ana) cases. The delete case and pattern cases would be analogous but I want someone else to test this.

I'm putting up another PR which takes the other tack: changing the way hole deletion works. This is more aggressive change., though I think it feels better for the most part.

### DIFF
--- a/src/hazelcore/Action_Exp.re
+++ b/src/hazelcore/Action_Exp.re
@@ -620,7 +620,7 @@ and syn_perform_block =
         u_gen: MetaVarGen.t,
       ),
     )
-    : ActionOutcome.t(syn_success) =>
+    : ActionOutcome.t(syn_success) => {
   switch (a) {
   /* Movement */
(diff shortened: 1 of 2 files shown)


cyrus-:
Is this still relevant?

disconcision:
superseded by #517

PR #517, opened 2021-06-28
Space op backspace/deletion behavior rework

### LINKED ISSUE #280: cannot backspace/delete over space

disconcision:
This represents a more aggressive approach to #516. It replace all backspace and deletion logic when the cursor is adjacent to a space operator. It also handles operand merging.

This is only implemented for expressions at the moment; will extend when otherwise approved


cyrus-:
Screenshot from 2021-07-07 17-36-32

When I backspace a "3 4" into "34", its reading it as a FloatLit (see type and result). Is it not going through the same text parsing logic as text entry?

cyrus-:
Not clear that pressing backspace at `f ? +| 3`  should lead to `f |3` -- why not `f ? |3`?

disconcision:
First comment: Whoops, typo in the float regexp. Yeah, first time around I missed the existing text logic, feel it's a bit hidden. I can switch over to using it... it looks to me like the relevant function mk_syn_text is doing almost entirely redundant logic? It looks like it should just be calling TextShape.of_text, turning the returned cases into the relevant operands, and then calling mk_and_syn_fix, instead of handling everything individually? Is this correct? I ask mostly because I need to factor out that aspect of it to use it in my case, so if so, I may as well just shorten mk_syn_text while I'm at it

(by way of illustration I just made this change; lmk if it doesnt make sense)

disconcision:
Second comment: I agree, that's not clear to me either (though I think it's debatable). It's not addressed by this PR though, which handles precisely the cases where the caret is (optically) adjacent to a space operator... the case you point out is dev behavior.

src/hazelcore/Action_Exp.re
```diff
@@ -367,55 +302,101 @@ let zcase_of_scrut_and_suffix =
   };
 };
 
+let mk_text_operand_exp = (text: string): UHExp.operand => {
```
cyrus-: don't need the _exp suffix since this is Action_Exp

src/hazelcore/Action_Exp.re
```diff
@@ -367,55 +302,101 @@ let zcase_of_scrut_and_suffix =
...
+  switch (TextShape.of_text(text)) {
+  | InvalidTextShape("") => EmptyHole(0)
+  | InvalidTextShape(s) => InvalidText(0, s)
```
cyrus-: why are these metavars set to 0?

src/hazelcore/Action_Exp.re
cyrus-: (but do need similar refactoring in Action_Pat and Action_Typ probably)

cyrus- (changes requested):
see individual comments

src/hazelcore/Action_Exp.re
disconcision: I do the renumbering and hole fixing centrally. I moved thing around a little to clarify the process. I'm using a pattern I started using in the assistant branch; it's less relevant here but I think conceptually it makes sense. There, to avoid threading u_gen and ctx through a bunch of tiny functions I use to build syntax compositionally, I just fix/(re)number once after construction is complete. There is gets annoying because otherwise you have to monadically thread u_gen through multiple levels of construction when you could just number after-the-fact in a single call. There are pros and cons... for example, I forgot to actually renumber here... but I do think the separation is cleaner.

cyrus-: OK, but leave a comment on functions that generate mis-numbered holes so that other clients know that the caller is expected to renumber

cyrus-:
Behavior seems to work well now. Need to port it over to Pat and Typ. Are there other things needing to be done?

disconcision:
Nah I think that's it.... i'm hoping the core logic is straightforward to make polymorphic and i'll just have to pass in the things that actually touch the operands

disconcision:
Supports patterns now. Core logic is polymorphic. I don't think this can meaningfully support types? There are no space operators rn

cyrus-:
Inserting wildcard patterns isn't working... `let _ = ?`

disconcision:
did you pull/make after that last commit? the one that's called "wildcard!!"? unless i'm not following what you mean... I can make a "let _ = "

cyrus-:
yeah I have that commit. I'm getting a "Unhandled exception (Failure "ana_fix_holes did not preserve path ((0)(OnText 1))")" when I try to do `let _`?

disconcision:
Whoops, was on the wrong branch. Fixed.

I'm wondering if there might be room to homogenize the caret position scheme? It would be nice to judge beforeness and afterness of operands just from patterns without using a predicate. Not immediately obvious to me how to do it though without introducing other issues.

I was also noticing that (on dev) "_" in an expression reads as a free var and not invalidtext, even though there's no way to bind it. Is that intentional?

cyrus-:
Not sure how'd you do that when you have non-bidelimited operands.

Making _ invalid in expression position makes sense to me. Want to include that in this PR?

disconcision:
underscores in expression position are now invalid

cyrus-:
OK. Can you get this branch caught up with `dev`? Anything else you're planning to do here?

Regarding types, this is something that the type aliases PR will need to handle once we merge into `dev`. @bkase

cyrus-:
(Once we have kind errors, we can add a space op to types and make it an error, like we have it for patterns)

cyrus- (changes requested):
Pressing `_` in pattern position leaves the cursor before the underscore rather than after.

cyrus-:
Also pressing backspace on `x |_` doesn't join them into `x_`

disconcision:
Oh yeah I forgot... this is going to have the same undo issues as the parentheses deletion fix. I'll think about it some more, but I suspect there's substantial enough work necessary there that i'm inclined to close this for now.

cyrus-:
This one might be a little simpler if you always just emit "deleted space" regardless of what the merge behavior is?

disconcision:
The problem is, we don't know we're on a space; the cursor is on some rando adjacent operand. To the extent there is a simple solution, it's probably the same as the parens one... add a generic "delete" fallthrough case, and make sure the delete-becomes-movement cases don't get caught in it. If it actually ends up being that straightforward, I can do it.

cyrus-:
Well current behavior already recognizes adjacent operators and works for space deletion when there is a hole, so shouldn't be a stretch to extend that to support the new situations?

cyrus-:
Pressing Delete on `x| _` doesn't turn it into `x_` in pattern position.

disconcision:
Okay that one's fixed... I re copy/pasted the code for expression+pattern analytic/synthetic so they should all be consistent now... all of these last bugs have due to those getting out of sync

cyrus-:
Type annotated patterns have odd behavior still, e.g. pressing Backspace at `let ? |y : Int` or `let x |y : Int` doesn't do what I expect...

disconcision:
hmm yeah i didn't really consider typeanns... looks like they'll need a non-trivial refactor as they are mergable from the left but not from the right...

disconcision:
okay, (indefinitely nested) type annotations are now supported. gonna warn you though this PR is starting to get dangerously cheesy

cyrus-:
- error holes for space ops in patterns are showing up on fun position
- not getting an undo history entry when pressing backspace at `? |(e)` and similar in both expressions and patterns

disconcision:
point one: hmm, looks like I was a bit too aggressive in shortening mk_text_ana; replaced previous version

point two: this is an awkward one to detect... basically, without doing a full action_exp style recursion, i have access to the cursor-term before and after the action. i've been using somewhat complicated logic to use this to infer what happened, but it seems to fall down on precisely the case where you have a hole and a non-hole, and the cursor is on the non-hole. then the cursor will be on the same thing before and after, even though the syntax has changed. the only thing I can think of is what I suggested earlier; comparing the whole syntax tree before and after, and no-oping these in undo. This is what I've done; it appears to fix the issue, but I'm not sure if it has unintended consequences.

cyrus-:
1. Still seeing error holes in the wrong place when using space in pattern position.
2. Hm, I suppose the comparison-based entries are better than random undo bugs...

cyrus- (changes requested):
1. Still seeing error holes in the wrong place when using space in pattern position.
2. Hm, I suppose the comparison-based entries are better than random undo bugs...

disconcision:
Can you clarify on incorrect error hole location? for` let a a = ? `and `let (a a):Float = ?` error holes are around ap

cyrus-:
Screenshot from 2021-07-11 15-20-07

Not seeing that...

disconcision:
ah, noticed that i had merged in dev and not pushed... maybe the way I altered things exposed patterns to the same placement bug fixed in that previous PR. in any case that's built.. hazel.org/build/space-op-space now looks to be behaving correctly...

src/hazelcore/Action_common.re
```diff
@@ -323,3 +333,98 @@ let complete_tuple_ =
     mk_ZOpSeq(ZOperator(zcomma, (seq |> Seq.rev, new_suffix)));
   (new_zopseq, u_gen);
 };
+
+let delete_action = (a: Action.t): delete_action =>
+  switch (a) {
+  | Backspace => Backspace
+  | _ => Delete
```
cyrus-: don't do a catchall, raise an exception for bad cases

src/hazelcore/UHExp.re
```diff
@@ -273,8 +273,10 @@ and mk_inconsistent_operand = (u_gen, operand) =>
 
 let text_operand =
     (u_gen: MetaVarGen.t, shape: TextShape.t): (operand, MetaVarGen.t) =>
+  // TODO: this function and its pattern equivalent are morally
```
cyrus-, line 276: this seems like a good time to do this refactor no?

src/hazelcore/UHPat.re
```diff
@@ -213,3 +213,32 @@ and is_complete_operand = (operand: 'operand): bool => {
...
 };
+
+let string_of_operand = (o: operand): string =>
+  switch (o) {
+  | Wild(_) => "_"
+  | InvalidText(_, s)
+  | Var(_, _, s)
+  | IntLit(_, s)
+  | FloatLit(_, s) => s
+  | ListNil(_) => "[]"
```
cyrus-: [] doesn't count as a text-like operand, probably don't want to include this

cyrus- (changes requested):

disconcision:
Addressed the above issues, but found a bug with the actual space deletion logic that I can't resolve without some more work, so leaving this in draft for now

cyrus-:
going to punt on this until after David's changes land

PR #518, opened 2021-06-28
Fix for issue with creating newline on last line of program

disconcision:
A simple one this time, at least I hope... this guy has been driving me crazy.


cyrus- (changes requested):
Seems to work but the hole renumbering is a bit surprising -- only happens when you do it the first time for some reason. May be worth recycling the existing hole in the case where it is a hole already to avoid the appearance of odd behavior.

disconcision:
can you clarify about the hole renumbering behavior?

also how is this an enhancement, isnt it a bug fix

cyrus-:
if you press Enter on the right of the empty hole when Hazel loads, it becomes hole 2 instead of saying hole 1. subsequent enters leave it as hole 2.

I'm not classifying it as a bug because it is just unexpected behavior but doesn't violate any invariants.

disconcision:
ah i see. i'll take a look.

not being to add a line to the bottom of the program certainly feels like a bug to me... perhaps it needs an associated invariant

disconcision:
fixed

cyrus-:
This isn't fixed for blocks in analytic position, e.g. 

```
let x : Float = 
  let y = 3 in 
  ?|
in 
?
```

disconcision:
I feel this is a separate bug. it doesn't implicate the codepath where I applied the fix, and the outlined situation doesn't exhibit the bug this fix is addressing.  I count 11 non-trivial Construct(SLine) cases in Action_Exp; from a cursory look, it looks like most or all of them will need to be changed.

cyrus-:
Hm, I don't think it's a good idea to special-case top-level or synthetic positions and not fix the corresponding bug in analytic positions or let the two drift apart...

disconcision:
I agree, but feel the right approach here is to revert the last commit, having to do with the hole renumbering, since that's not the bug this is addressing; the hole renumbering seems more pervasive and requires a larger change. I don't mind tacking on two bugs in one if it's a minor addition, but that doesn't seem to be the case here.

cyrus-:
Well reverting the renumbering fix won't make the behavior consistent either -- in analytic position, the hole numbers keep increasing, but in synthetic position it happens only once.

disconcision:
okay i fixed the two obvious instances of analogous renumbering issues, which covers the specific case you mentioned. in general, there's odd behavior including but not limited to renumberings when creating a mid-line linebreak.... reluctant to wade into all that atm

cyrus-:
Pressing enter here in synthetic position:
```
let x = ? in 
?|
```
leaves you at the end of the final line, whereas in analytic position, it puts you at the beginning of the line (which is probably the behavior we want)

disconcision:
yeah I noticed that.. not obvious to me which option is right? when pressing enter before the line, it makes sense that cursor placement stays before; we're pushing the line itself down. pressing enter after the line, we're generally creating a new line, so we should be at the start, but now we're special-casing the situation where the line is an empty hole, effectively pushing it down, so it's not obvious to me which is the right behavior. i feel there's an inherant tension here with empty lines vs holes that creates the awkwardness. but i'm fine special-casing as you suggest (enter after an empty hole line pushes the line down and then moves the cursor to the front)

cyrus-:
Yeah, no strong opinion either way as long as they are consistent.

cyrus-:
OK, ready for review?

cyrus- (approved):

Issue #458, opened 2020-12-14
Fix err hole rendering bug on applications


dm0n3y:

dm0n3y:
<img width="415" alt="image" src="https://user-images.githubusercontent.com/8718124/111076542-132ae700-84c3-11eb-9b5a-1f262f651793.png">

PR #519, opened 2021-06-28
Fix error holes on aps being drawn on function position

### LINKED ISSUE #434: Nonempty hole around pattern ap is only drawn around first argument

disconcision:
Closes #434, #458

the issue here was that the hole accumulating functions were returning steps to function position for aps because that's what we currently have to use for move-to-hole, but not what we want for decorations


cyrus- (approved):
This fixes the problem, so we can merge. However, there is a related bug that may be worth fixing now: errors in parentheses are handled strangely.

Screenshot from 2021-07-08 23-05-16

On trunk, we move error holes inside parens out all the way (even for nested parens) for display, and cursor info reports the error information on this basis as well.

Screenshot from 2021-07-08 23-06-21

PR #520, opened 2021-07-01
Deploy hazel to web migration from Travis to Github Actions

disconcision:
When any branch, (say) `branchy-mc-branchface` ,is pushed, this script will build it and make it available at `http://hazel.org/build/branchy-mc-branchface`.

This is ready for someone else to look over and test; the one outstanding issue is that opam doesn't seem to be caching properly, so runs do a full install and make deps, lasting about 22 minutes. Still looking into it.


.github/workflows/deploy_branches.yml
```diff
@@ -0,0 +1,53 @@
...
+        working-directory: ./server
+      - name: Copy in the newly built source # STEP 6
+        run: mkdir "./server/${BRANCH_NAME}" && cp -r "./source/_build/default/src/hazelweb/www"/* "./server/${BRANCH_NAME}"
+      - name : Commit to the website aka deploy # STEP 7
+        run: |
+          git config user.name github-deploy-action
+          git config user.email hazel-deploy@hazel.org
+          git add -A
+          git status
+          git commit -m "github-deploy-action"
```
cyrus-: can we have the commit message be more descriptive here, e.g. say which branch was modified?

disconcision:
Just to follow up on your question cyrus: any failure during build will result in no changes made to the site

cyrus- (changes requested):

.github/workflows/deploy_branches.yml
disconcision: done. well, we'll see if it works in 22 minutes

disconcision: yep, branch name is added to commits now

cyrus-:
Cool, thanks for fixing this!

I don't think its going to trigger for all the current branches until their next commit. Is there a way to manually run this for each existing branch?

cyrus-:
Oh also it seems to show up as a failed action if there isn't a change to the built artifact, e.g. https://github.com/hazelgrove/hazel/runs/2965811922?check_suite_focus=true. Might be worth fixing.

PR #521, opened 2021-07-02
deploy action fixes

cyrus-:


.github/workflows/deploy_branches.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         id: opam-cache
         uses: actions/cache@v2
         with:
-          path: './source/.opam/'
+          path: './.opam/'
```
cyrus-: this isn't right either... what is the correct path to the .opam directory? /home/runner/.opam?

disconcision: i'll have to ssh in again, i forget the directory structure... definitely need a /source/ in there though, that's where i clone hazelweb and do the install/build. if the default wd is  /home/runner/ then it should be /home/runner/source/.opam... but that's what it was before i think

cyrus-: the .opam directory wouldn't be in the cloned repo though, it should be in the home directory. latest push is trying /home/runner/.opam

disconcision: yes you're right, confirmed that it's /home/runner/.opam

cyrus-: yeah looks like it might have worked with that path, 

```Cache Size: ~679 MB (712230115 B)
Cache saved successfully
Cache saved with key: Linux-modules-b7a9a7e06bc2c821a8c75015671b84d33deba4d44c068f17b39a66dc76242fcf```

(size was previously 22b)

disconcision (approved):

.github/workflows/deploy_branches.yml
cyrus-: yup, decreased build time to <2m. woot.

PR #522, opened 2021-07-02
Parentheses reverse deletion fix

disconcision:
Backspacing an opening parens or deleting closing parens no longer requires two keystrokes.

Not sure if there is a more principled approach here

### DIFF
--- a/src/hazelcore/Action_Exp.re
+++ b/src/hazelcore/Action_Exp.re
@@ -1087,6 +1087,34 @@ and syn_perform_opseq =
 
   /* Deletion */
 
+  /* This case prevents deletion of parentheses from the front with backspace
+     or from the rear with delete from taking two keypresses, the first of
+     which does not result in caret movement. */
+  | (
+      a,
+      ZOperand(
+        ParenthesizedZ((
+          [],
+          ExpLineZ(ZOpSeq(_, ZOperand(zop, (prefix, suffix)))),
+          [],
+        )),
+        (upper_prefix, upper_suffix),
+      ),
+    )
+      when
+        ZExp.is_before_zoperand(zop)
+        && a == Backspace
+        || ZExp.is_after_zoperand(zop)
+        && a == Delete =>
+    let new_zseq =
+      ZSeq.ZOperand(
+        zop,
+        (
+          Seq.affix_affix(upper_prefix, prefix),
+          Seq.affix_affix(suffix, upper_suffix),
+        ),
+      );
+    Succeeded(SynDone(mk_and_syn_fix_ZOpSeq(ctx, u_gen, new_zseq)));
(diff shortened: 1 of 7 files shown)


src/hazelcore/Action_Exp.re
```diff
@@ -1087,6 +1087,34 @@ and syn_perform_opseq =
...
+        ParenthesizedZ((
+          [],
+          ExpLineZ(ZOpSeq(_, ZOperand(zop, (prefix, suffix)))),
+          [],
+        )),
+        (upper_prefix, upper_suffix),
+      ),
+    )
+      when
+        ZExp.is_before_zoperand(zop)
```
cyrus-: this isn't correct -- zop might not be the leftmost position if prefix isn't empty. you should check is_before on the entire opseq. manifests as wrong behavior when you do, e.g. 1 + (2 + |3 + 4)

cyrus- (changes requested):

src/hazelcore/Action_Exp.re
disconcision: Ah whoops. Though couldn't I just change the prefix pattern to E and keep the check the same?

disconcision: or I guess that would require separate logic for delete... i'll do it your way

disconcision: fixed

cyrus- (changes requested):
the "delete parentheses" entry in the history panel isn't showing up when deletion occurs via this pathway -- will need to recognize it

cyrus-:
(this is kind of a brittle aspect of the undo history mechanism -- might be worth a bigger rethink in the future)

disconcision:
Pushed a fix. Ugly though. I'd argue that the issue is with perform_action, not the undo system as-such. Or maybe rather: Action should internalize most of the undo system. As it stands, perform_action takes a high-level user action, like pressing Delete, and implicitly transforms it into (one of several) actual low-level edit actions. If we want to log/report the low-level action rather than the high-level one, then the low-level actions should be reified; for example, perform_action should emit the action actually performed which undo_history_panel could use for reporting. As it stands, undo_history tries to guess what was actually done. This is only possible by partial reimplemention of the logic of perform_action, which isn't sustainable.

cyrus-:
Yes, I agree that we should be getting this info out of perform_action rather than trying to infer it backwards. Let's slate that for post-David refactoring for now.

cyrus-:
fix wasn't applied to patterns or types

disconcision:
Closing this... the fix doesn't really work for expressions either, and I don't think this can reasonably be made to work without basically reproducing the entire recursion structure from Action_* .

PR #523, opened 2021-07-06
Automatically treat Ctrl (on non-Macs) and Cmd (on Macs) the same way

### LINKED ISSUE #480: remove `is_mac` from model
No need to keep `is_mac` in the model. It should just be a global ref set in `Hazel.on_startup`.

### LINKED ISSUE #482: alt keycombos do not work on mac
Dug into this issue with @rntz while trying to set up Alt+H for jumping to hole. Turns out macs change the key value when the alt modifier is held (e.g. Alt+R produces an event with key value "®"). Somehow there does not seem to exist a perfect solution for this. Either we can rely on deprecated keyboard event properties (as suggested in these [StackOverflow](https://stackoverflow.com/questions/52799263/how-to-capture-altn-on-keypress-in-javascript) [posts](https://stackoverflow.com/questions/14109639/how-to-get-alt-cor-any-other-key-to-work-in-javascript-on-chrome-for-mac-and)) or we can use the [`code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) property to figure out which physical key was pressed, which creates inconsistent behavior for users with non-QWERTY key mappings.

yottalogical:
Previously, every part of the code that dealt with the Ctrl/Cmd key had to be written twice, once for non-Macs and once for Macs. This was redundant, since (at least in this project) the behavior of the two should always be the same. At best, this meant duplicate code. At worst, it meant certain cases that weren't accounted for.

This pull request makes it so that this is all handled automatically in `ModKeys.re` and `ModKeys.re`. Except for those two places, all references to the Cmd (Meta) key have been removed. Instead, just use Ctrl in all situations, and it will automatically be treated as Cmd on Macs.


yottalogical:
The vast majority of code changes in this pull request are just there to pass `is_mac` down to where it's needed. Besides that, there are a relatively small number of changes involved.

cyrus-:
There may be situations on Mac in the future where we do want to use ctrl itself though -- how would we express that?

yottalogical:
(Unless I'm forgetting something), the only reason I can think that we'd want to use Ctrl specifically is if we were using it in tandem with the Meta (Cmd) key. Otherwise, Cmd by itself would work just fine.

If this were the case, it would have to be a Mac exclusive feature, since looking through the Mozilla documentation, [the Windows on keyboards key no longer registers as the Meta key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey).

Is this something worth keeping open as an option? If so, I can make a smaller pull request that only fixes the inconsistencies that are currently in the `dev` branch.

yottalogical:
The bug fixed in commit [`67d5bf8`](https://github.com/hazelgrove/hazel/pull/523/commits/67d5bf854b6bfe34b54e3244d6a66a2b2c333d12) is one that currently lives in the `dev` branch too. I originally thought it was specific to this branch, since the problem was made more prevalent by my changes, but after fixing it I noticed that it was the same bug that was causing issue #482.

The basic problem was that the option key changes what key is typed when you press a keyboard button. Kind of like how holding shift give you access to an uppercase version of the keyboard, holding option gives you access to a special character version of the keyboard. `Key.recognize` was looking for `l` and `r`, but instead it was actually getting `¬` and `®`. The fix was to use [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead of [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), because it corresponds to the physical key being pressed, without considering modifier keys.

cyrus-:
I'm thinking about things like using Ctrl+arrow keys in the editor to support word-based movement -- we'd want that to use the actual Ctrl key, not Cmd, on Mac. There may be other situations where we just use Ctrl as well. We could support this by distinguishing between Ctrl and a Ctrl_or_Cmd when specifying the shortcut, pushing the handling of that into the keyboard-related helpers?

yottalogical:
Commit [`2e83adc`](https://github.com/hazelgrove/hazel/pull/523/commits/2e83adc29cee312169c5a937ef60483948d81171) implements the change requested in issue #480, although not as a ref, since `is_mac` should never change while running. I incorporated it into this pull request since otherwise there would be an overwhelming number of functions that would need to take it in as an argument.

yottalogical:
Commit [`ec055ad`](https://github.com/hazelgrove/hazel/pull/523/commits/ec055ad95ee8be883a93a9fb150151ecfaa7c2d6) allows for Ctrl to be directly specified. `CtrlOrCmd` will automatically be Ctrl or Cmd depending on the user's platform. `Ctrl` will always be Ctrl.

yottalogical:
Commit [`15d3922`](https://github.com/hazelgrove/hazel/pull/523/commits/15d3922cc830a01c351b29fff430368cd1e2e28d) refixes issue #482 in a way that doesn't use `Key.get_code`, since it does not respect non-QWERTY keyboard layouts. Instead, it goes back to using `Key.get_key`, but first looks for option key characters and converts them to the key that would have been typed if the user hadn't've been holding option.

src/hazelweb/keyboard/Key.re
```diff
@@ -1,6 +1,110 @@
 module Js = Js_of_ocaml.Js;
 module Dom_html = Js_of_ocaml.Dom_html;
 
+/**
+ * Holding down the option key can result in special characters being typed instead of the intended character
+ * This function tries to return the physical key that was typed in a way that respects alternative keyboard layouts
+ */
+let without_option_key =
```
cyrus-: we probably want this to happen only on Mac -- check IsMac?

src/hazelweb/keyboard/Key.re
```diff
@@ -1,6 +1,110 @@
...
+  // https://en.wikipedia.org/wiki/Dead_key
+  | "Dead" =>
+    failwith(__LOC__ ++ ": Key cannot be recognized because it is a dead key")
+  | other => other;
+
+/**
+ * Avoid Using!
+ * Does not respect non-QWERTY keyboard layouts
+ */
 let get_code = (evt: Js.t(Dom_html.keyboardEvent)) =>
```
cyrus-: looks like we don't actually use code-based recognition for anything anymore? if so, we can remove it entirely since it does indeed not respect alt keyboard layouts...

cyrus- (changes requested):

src/hazelweb/keyboard/Key.re
```diff
@@ -31,7 +135,7 @@ let recognize = (evt: Js.t(Dom_html.keyboardEvent), r) =>
     let code = get_code(evt);
     String.equal(code, c);
   | Key(k) =>
-    let key = get_key(evt);
+    let key = without_option_key(get_key(evt));
```
cyrus-: also probably only want to do this in situations where we are recognizing a key combination involving alt -- don't necessarily want to preclude allowing unicode character entry into the code directly

src/hazelweb/keyboard/Key.re
yottalogical: Implemented this request in commit [`12c0cac`](https://github.com/hazelgrove/hazel/pull/523/commits/12c0cac62cda0a3510b21b329223208c3af96cce).

src/hazelweb/keyboard/Key.re
yottalogical: Implemented this request in commit [`916730d`](https://github.com/hazelgrove/hazel/pull/523/commits/916730deefe0de97ee714cd363235a2bdba1c7c0).

src/hazelweb/keyboard/Key.re
yottalogical: Implemented this request in commit [`916730d`](https://github.com/hazelgrove/hazel/pull/523/commits/916730deefe0de97ee714cd363235a2bdba1c7c0).

cyrus- (approved):
Looks good to me, but I don't have a Mac -- @disconcision can you give it a quick try and make sure everything looks good on a Mac?

disconcision (changes requested):
Mostly looks and feels good! But we've got a couple issues with collisions now:

1: I used Ctrl-S for serialize, which was intended to be sort of cutesy as it's a little save-like. Might want to change this now though since it's the same as save page in Chrome+FF. The current behavior seems to be, if the code has focus, it serializes to console, and if it doesn't , it tries to save the page. So debatably the issue is with the focus model... 'saving the page' is a pretty useless action for webapps, so I don't mind overriding it, but we'd have to make the override more robust, ie a set of keyhandlers which is always active when the page root is active. Probably we should just put a pin on the larger issue for now and just change the shortcut to something else (doesn't matter what).

2: Cmd-Option-I is the Chrome+FF console shortcut, so I can't swap up. On the plus side, I can now swap down, left, and right, so this is still a strict improvement on the current state of affairs! I feel all 4 of these should all just be changed to Alt-Arrow shortcuts for now, unless @cyrus is intend on keeping that real-estate free. Otherwise, how about just Alt-(i/k/j/l)? we have so few keyboard shortcuts that I don't feel double modifiers are strictly necessary atm?

cyrus-:
Yeah, all the keyboard shortcuts only work if the code has focus, which we probably want to change at some point. Let's not do that for now though.

Regarding swapping, yeah lets not override the console shortcut. 

I think Alt+Arrow Keys is going to interfere with Mac window manager functionality?

Alt-I/J/K/L overlaps with Alt-L for left injection, which will also go away at some point. 

Can we make it Ctrl+Alt+I/J/K/L on Mac too (so not Cmd)?

disconcision:
I don't think option-arrows do window management. they default generally to advance/retreat to word boundary. i feel like moving stuff around with them is valuable enough to unseat that though, especially since this VSCode does exactly this

cyrus-:
Alt-Left/Right on Win/Linux map to browser back/forward...

disconcision:
i'm inclined to argue that's a good thing, as i feel it's justifiable to override back for an editor, especially one that currently doesn't retain its state. but the focus issue does make it problematic for now...

cyrus-:
Yeah, I suppose I'm fine with that.

yottalogical:
So just to confirm before I do anything:
1. For now, serializing to console should be temporarily changed to something else. How about just Ctrl+S again?
2. Swapping should be changed to Alt-ArrowKeys.

Is this correct?

cyrus-:
1. We can leave serialization to Ctrl + S.
2. Yes

cyrus- (approved):

cyrus-:
@disconcision can you verify that the changes work on Mac?

disconcision (approved):
lgtm!

cyrus- (approved):

PR #524, opened 2021-07-06
Refactor Elaborator_Exp.ml to contain only functions relevant to expression elaboration.

dedbox:
A simple patch to take the `matches` function out of Elaborator_Exp.ml and put it into Evaluator.ml, the only module that uses it.


cyrus- (changes requested):
Looks good -- can you add subst to this as well (and check for other pieces with a similar problem)? I don't believe that is used in elaboration either.

dedbox:
Ok, I've moved subst, subst_var, and id_env to Environment.ml. They're all used by the same modules: Elaborator_Exp and Evaluator.

src/hazelweb/model/Program.re
```diff
@@ -91,6 +91,249 @@ let get_decoration_paths = (program: t): UHDecorationPaths.t => {
...
+
+let renumber =
+    (path: InstancePath.t, hii: HoleInstanceInfo.t, d: DHExp.t)
+    : (DHExp.t, HoleInstanceInfo.t) => {
+  let (d, hii) = renumber_result_only(path, hii, d);
+  renumber_sigmas_only(path, hii, d);
+};
+
 exception DoesNotElaborate;
 let expand =
```
cyrus-: this should be renamed to elaborate (similarly for other uses of "expand" / "expansion")

cyrus- (changes requested):
looks good, just one more cleanup thing

cyrus-:
doesn't compile

src/hazelcore/dynamics/Elaborator_Exp.re
```diff
@@ -1,452 +1,3 @@
...
-  | ConsistentCase(_)
-  | InconsistentBranches(_) => Indet
-  | EmptyHole(_, _, _) => Indet
-  | NonEmptyHole(_, _, _, _, _) => Indet
-  | FailedCast(_, _, _) => Indet
-  | InvalidOperation(_) => Indet
-  };
-
 type elab_result_lines =
   | LinesExpand(DHExp.t => DHExp.t, Contexts.t, Delta.t)
```
cyrus-: rename to LinesElaborate / DoNotElaborate

there may be some other uses of Expand in hazelcore/dynamics that need the same treatment

cyrus- (changes requested):
need a few more renames from expand to elab in hazelcore/dynamics and any part of hazelweb that interfaces with it, but otherwise looks good!

dedbox:
Ok, I did some more renaming. I did a project-wide search for "xpan", then renamed everything related to elaboration, i.e., not keyword expansion or the inverse of type contraction.

src/hazelcore/dynamics/Environment.re
```diff
@@ -1,3 +1,136 @@
...
+       }
+     )
+and subst_var_env = (d1: DHExp.t, x: Var.t, sigma: t): t =>
+  sigma
+  |> List.map(xd => {
+       let (y, d) = xd;
+       (y, subst_var(d1, x, d));
+     });
+
+let subst = (env: t, d: DHExp.t): DHExp.t =>
```
cyrus-: substitution stuff should be in Evaluator or in a separate Subst module, not part of Environment

cyrus- (changes requested):
looks good -- see one comment

cyrus-:
@dedbox can you fix the issue above so we can get this merged?

cyrus- (approved):

Issue #397, opened 2020-07-14
Crashes on latter "Just a Hole" example choice


nj-wilson:
`Async_kernel: Unhandled exception (Invalid_argument "option is None")`

Multiple example ways to reproduce:

1) From a fresh start, type anything into the empty hole, such as `1`.
    Choose any example from the drop-down besides "Just a Hole"
    Then choose "Just a Hole" from the drop-down.

2) From a fresh start, choose an example with a hole, like map
    Type anything into the hole at the bottom, such as `1`.
    Then choose "Just a Hole" from the drop-down.

disconcision:
Closed by #513

disconcision:
The crash is coming from the Option.get in get_selected_hole_instance. I don't exactly understand what selected_instances are, but it looks like they should probably be scoped to programs/cards/whatever, not global?

cyrus-:
I believe the selected instance is the hole instance from the result being shown in the sidebar -- yes, that should not be global and can be invalidated by edits, so should probably be saved in the undo history too

PR #525, opened 2021-07-07
Undo cardstack fix

disconcision:
Undo/Redo now sets the card/cardstack to the card where the action was made.

Also, the card/cardstack selectors now properly display the current card/cardstack.


cyrus-:
It crashes when I load an example, edit it, and switch back to the "hole" example?

disconcision:
Ah. that was a pre-existing issue: https://github.com/hazelgrove/hazel/issues/397. thought this had fixed it, but i guess it just made triggering it slightly different

src/hazelweb/model/UndoHistory.rei
```diff
@@ -9,6 +9,8 @@ type undo_history_entry = {
   previous_action: Action.t,
   action_group: UndoHistoryCore.action_group,
   timestamp: UndoHistoryCore.timestamp,
+  current_cardstack: int,
```
cyrus-: not sure I'm understanding how this connects to the `cardstacks_after_action` and `cardstacks_after_move`? actually I don't recall what the latter was for and it seems to be doing something buggy, e.g. type `abc`, then move to `ab|` and then do an undo and then a redo, and you end up at `ab|c` rather than `abc|`

cyrus-: oh I think the intended behavior is that the redoing back to the top doesn't lose your cursor position. that isn't a thing that vs code does though, so we might want to drop that complexity.

in any case, doesn't this ZCardstacks.t contain the index of the current cardstack and card?

cyrus- (changes requested):
see comment - I think the current_cardstack and current_card entries are redundant with the existing model?

disconcision:
made the change; this seems to make redo behave in a way that feels weird with respect to card changes in a way i can't quite follow. unsure if it is 'correct' behavior. undo perhaps had a symmetric issue in the previous iteration?

cyrus-:
what is the potentially weird behavior?

disconcision:
i'm not sure how to properly describe it... it feels like after redo i should (sometimes) be on a different card. making a couple edits each on a couple different cards is enough to trigger it

cyrus-:
So the dropdown isn't always showing the current card when an undo causes the selected card to change.

I can't get other weird behavior on quick glance -- can you reproduce what is going on? Might be related to this system where we keep track of movements at the top of the stack.

cyrus-:
@disconcision do you want to try to get this fixed in dev or wait until David's refactorings re:undo history coming out of actions land?

cyrus-:
seems stale, closing -- will fix after David's undo history changes land

PR #526, opened 2021-07-07
(possibly bad) fix of image livelit impossible syn crash

disconcision:
This fixes the crash but I'm not sure the fix is well-considered. The issue is that the livelit is loosing it's InsufficientParams label in the below logic. But I'm not clear on why these checks are happening in the first place, instead of just calling ana_fix_holes_operand and being done with it

### DIFF
--- a/src/hazelcore/Statics_Exp.re
+++ b/src/hazelcore/Statics_Exp.re
@@ -2205,13 +2205,11 @@ module rec M: Statics_Exp_Sig.S = {
         syn_fix_holes_operand(ctx, u_gen, ~renumber_empty_holes, e);
       let is_insufficient_params =
         switch (UHExp.get_err_status_operand(e)) {
-        | InHole(TypeInconsistent(Some(InsufficientParams)), _) => true
+        | InHole(TypeInconsistent(Some(InsufficientParams)), _)
         | _ => false
         };
-      if (is_insufficient_params) {
+      if (is_insufficient_params || HTyp.consistent(ty, ty')) {
         (e, u_gen);
-      } else if (HTyp.consistent(ty, ty')) {
-        (UHExp.set_err_status_operand(NotInHole, e), u_gen);
       } else {
         let (u, u_gen) = MetaVarGen.next_hole(u_gen);
         (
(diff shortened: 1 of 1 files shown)


cyrus-:
yeah the problem here is that we've rolled InsufficientParams into the TypeInconsistency error reason, which otherwise can only happen when there is a subsumption failure. a better solution would be to define a separate LivelitErrStatus (similar to CaseErrStatus) and have subsumption stuff only fix actual type inconsistencies

src/hazelcore/Statics_Exp.re
```diff
@@ -2204,7 +2203,14 @@ module rec M: Statics_Exp_Sig.S = {
     | FreeLivelit(_) =>
       let (e, ty', u_gen) =
         syn_fix_holes_operand(ctx, u_gen, ~renumber_empty_holes, e);
-      if (HTyp.consistent(ty, ty')) {
+      let is_insufficient_params =
+        switch (UHExp.get_err_status_operand(e)) {
+        | InHole(TypeInconsistent(Some(InsufficientParams)), _) => false
```
cyrus-: shouldn't this be true?

src/hazelcore/Statics_Exp.re
```diff
@@ -2204,7 +2203,14 @@ module rec M: Statics_Exp_Sig.S = {
...
+        | _ => true
+        };
+      if (is_insufficient_params) {
+        (e, u_gen);
+      } else if (HTyp.consistent(ty, ty')) {
         (UHExp.set_err_status_operand(NotInHole, e), u_gen);
```
cyrus-: I don't think we actually need this branch (in dev too), because syn_fix_holes should never produce a term in a type inconsistency hole

cyrus-:
@disconcision should we get this merged into livelits?

disconcision:
@cyrus I thought i remembered you making a comment on something to change here... But it's not on the PR...

cyrus-:
The two comments from the PR review above I think is it?

cyrus-:
@disconcision bump

src/hazelcore/Statics_Exp.re
disconcision: Which branch? The middle one? If I remove it, the image example immediately throws an error on-load:

"Async_kernel: Unhandled exception (Failure
  "Unexpected evaluation failure during livelit args fill and resume: 1")"

PR #531, opened 2021-07-26
Make string CommentLine reflect keyboard input

zhixiangteoh:
Currently, CommentLine only allows `SChar` construction.

Goal of this PR is to make string CommentLine reflect exactly keyboard 
input, to achieve the following (as example):
- ``# Abl3 to +ype any symb0! 100.89*% ( [ ) ] \ = -_ '~"hello" ; let := >< ?? // $ ^& @hazel \` ``

Modify `JSUtil.is_single_key` to take in a variable `regex`, and then
within `CommentLine` match a wider `char_regex` (`^[^#]$` everything
except `#`) than `alpha_regex` (`^[a-zA-Z']$` all alphabets and `'`) on
regular lines.

Screen Shot 2021-08-09 at 10 57 41 AM

Currently still doesn't support alt/opt-modified symbols.

PR #535, opened 2021-08-04
Add Serialize result to console button.

coned:
Add serialize result to console button. Change the action SerializeToConsole.


src/hazelweb/gui/Page.re
```diff
@@ -56,6 +56,22 @@ let cell_status_panel = (~settings: Settings.t, ~model: Model.t, ~inject) => {
...
+        [
+          {
+            Node.button(
+              [
+                Attr.id("serialize-DHExp"),
+                Attr.on_click(_ =>
+                  inject(SerializeToConsole(DHExp(result)))
+                ),
+              ],
+              [Node.text("Serialize result to console")],
```
cyrus-: why not put this in the dropdown menu alongside the other serialize option?

cyrus- (changes requested):

src/hazelweb/gui/Page.re
zhixiangteoh: https://github.com/hazelgrove/hazel/pull/531/commits/84756c308186ed12e9cb08399a62ec82db89c476

@coned I sort of did something similar for Zexp (just for my visualization purposes), but I think this is similar to what @cyrus- means (i.e., adding as ActionMenu item)?

disconcision: One reason I added this menu was that I was hoping it would provide an unobtrusive place for people to add stuff they develop for their own debugging purposes, so if anyone has something they think might be more generally useful, please do suggest it or make a PR!
2021-08-09 13_08_56-Hazel

cyrus-: Yup, let's put debugging things like this up there.

coned:
Thanks! I put this button into the dropdwon menu now.

cyrus- (approved):

cyrus-:
Thanks @coned!

PR #537, opened 2021-08-19
Type Hole Inference

RaefM:
Draft PR for type hole inference


src/hazelcore/HTyp.re
```diff
@@ -3,7 +3,14 @@ open Sexplib.Std;
 /* types with holes */
 [@deriving sexp]
 type t =
-  | Hole
+  | Hole(MetaVar.t, (list(MetaVar.t)), MetaVarGen.t) // UHHole [1] -> HHole [1.1.1.1.2.1], HHole [1.1.1.1.2.2]
```
cyrus-: - why is there a MetaVarGen.t here?
- we should define a single type "UHTypHoleId.t" and "HTypHoleId.t" to capture these things, instead of putting them inline

anandrav: the MetaVarGen.t is there for when additional holes are created from that one.
So if two holes are created from hole 3, they will be numbered 3.1 and 3.2 using the MetaVarGen.t
I'll explain this more in OH, maybe we had the wrong idea.

Yeah creating a type for HoleId would be more elegant

anandrav: (discussion from OH)

cyrus: instead of using (MetaVar.t, list(MetaVar.t), MetaVarGen.t) for hole ids, which requires updating MetaVarGen.t in expressions, maybe use a finite sum type like (MetaVar.t, list(HoleReason.t)). That way, we don't have to update a MetaVarGen.t in expressions, but the hole ids are still unique.

hole reason could be something like this:
type HoleReason.t =
| matched_arrow(direction)
| matched_sum(direction)
| matched_prod(index)
| matched_list;

cyrus-: Sounds good. I'd call it HoleProvenance.t.

cyrus-:
@RaefM guessing this is an out-of-date branch?

RaefM:
> @RaefM guessing this is an out-of-date branch?

Yup lol- will make a new PR for inference stuff tonight. It is built on the content of the provenances PR linked here:
https://github.com/hazelgrove/hazel/pull/948

PR #538, opened 2021-08-21
Move caret over closing paren

### LINKED ISSUE #131: Skip over closing parens when you type )

hellyab:
Move caret over the closing paren when ")" is inputted right before to the closing paren
closes #131


src/hazelcore/Action_Exp.re
```diff
@@ -1912,9 +1979,8 @@ and syn_perform_operand =
         };
       }
     }
-  | (Init, _) => failwith("Init action should not be performed.")
-  };
-}
+  | (Construct(SCloseBraces), _) => Failed
```
cyrus-: why does SCloseBraces but not SCloseParens have a fallthrough case like this?

src/hazelcore/Action_Exp.re
```diff
@@ -3233,6 +3361,8 @@ and ana_perform_operand =
       }
     }
 
+  | (Construct(SCloseBraces), _) => Failed
```
cyrus-: same as above comment

src/hazelcore/ZTyp.re
```diff
@@ -168,7 +168,8 @@ let move_cursor_right_zoperator: zoperator => option(zoperator) =
 
 let rec move_cursor_right = (zty: t): option(t) =>
   zty |> move_cursor_right_zopseq
-and move_cursor_right_zopseq = zopseq =>
+and move_cursor_right_zopseq = zopseq => {
+  print_endline("ZTyp move_cursor_right_zopseq");
```
cyrus-: don't leave debug prints when you mark as ready for review

src/hazelweb/model/UndoHistory.re
```diff
@@ -464,7 +464,13 @@ let get_new_action_group =
       ~action: Action.t,
     )
     : option(UndoHistoryCore.action_group) =>
-  if (UndoHistoryCore.is_move_action(new_cursor_term_info)) {
+  /* they areparens and bsraces basically results with movements but
+    * it is an edit action, therefore there is a need to make sure that
+    * the action is not Construct(SCloseParens) or Construct(SCloseBraces)
+   */
+  if (UndoHistoryCore.is_move_action(new_cursor_term_info)
```
cyrus-: can we make this a predicate on the action instead of having it compare the previous and new states?

cyrus- (changes requested):
Looks pretty good, just a few small things to respond to / revise. Also, can you add support for `]` when in a list type like `[Int]`? (We'll have list literals in a future pull too, so this can set up for that as well.)

src/hazelcore/ZTyp.re
hellyab: I thought I removed these. My bad.

src/hazelcore/Action_Exp.re
hellyab: The one for SCloseParens was explicitly put as (Construct(SCloseParens), CursorE(_)) => Failed, doing the same for this one too

src/hazelcore/Action_Exp.re
hellyab: > Looks pretty good, just a few small things to respond to / revise. Also, can you add support for `]` when in a list type like `[Int]`? (We'll have list literals in a future pull too, so this can set up for that as well.)

Done!

src/hazelweb/model/UndoHistory.re
hellyab: The entire if expression looks like it was redundant because the else branch was handling all possible cases. I just removed the if branch. Not sure why it was added, but I have removed it at the moment. It was added in dab69adceb8634f31e334ef922287795a66c0c03

hellyab:
> Looks pretty good, just a few small things to respond to / revise. Also, can you add support for `]` when in a list type like `[Int]`? (We'll have list literals in a future pull too, so this can set up for that as well.)

Thanks! I have addressed the issues and also added support for `]`. It is working fine on the list type.

src/hazelweb/model/UndoHistory.re
```diff
@@ -464,195 +464,195 @@ let get_new_action_group =
...
+    switch (shape) {
+    | SLine =>
+      switch (
+        CursorInfo_Exp.get_outer_zrules(new_cursor_term_info.zexp_before)
+      ) {
+      | None => Some(ConstructEdit(shape))
+      | Some(zrules_before) =>
         switch (
-          CursorInfo_Exp.get_outer_zrules(new_cursor_term_info.zexp_before)
+          CursorInfo_Exp.get_outer_zrules(new_cursor_term_info.zexp_after)
```
cyrus-, line 481: @hellyab was there a bug here (unrelated to the PR)?

hellyab: @cyrus- it has always been `zexp_after`, I think git is comparing it to another line.

cyrus- (approved):

PR #548, opened 2021-10-24
update ocaml to 4.12.1

coned:


cyrus-:
@coned There are instructions for when we update to a new version of OCaml here: https://github.com/hazelgrove/hazel/blob/dev/UPDATING.md. Can you go through those and make sure we are doing that / update the instructions if anything is out of date (as part of this PR)? I think at least the instructions about travis need to be updated to talk about github actions.

cyrus- (changes requested):
see comment

coned:
I decided to update ocaml to 4.12.1 because we have a dependency rtop which only supports up to 4.12.1. I try to make least changes. So, only necessary packages are updated. The only big change is merlin (3.6.1 -> 4.4-412).

cyrus- (changes requested):
Can you also update the UPDATING.md file with up-to-date instructions for future upgrades? Some things there are no longer relevant, e.g. the vscode extension stuff.

PR #554, opened 2021-11-03
Add support for built-in functions

mirryi:
This ports over the mechanisms for supporting builtin functions from the `strings3` branch. Builtin functions are declared in `hazelcore/Builtins.re` and implemented (in Reason/OCaml, not Hazel) in `hazelcore/dynamics/Evaluator.re` (see `evaluate_ap_builtin`).

To start, `int_of_float` and `float_of_int` have been added.


src/hazelcore/Contexts.re
```diff
@@ -1,6 +1,6 @@
 [@deriving sexp]
 type t = (VarCtx.t, PaletteCtx.t);
-let empty = (VarCtx.empty, PaletteCtx.empty);
+let empty = (Builtins.ctx, PaletteCtx.empty);
```
cyrus-: let's call this `initial` since its not actually empty

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -816,4 +829,31 @@ and evaluate_case =
         evaluate_case(inconsistent_info, scrut, rules, current_rule_index + 1)
       }
     }
+  }
+and evaluate_ap_builtin = (ident: string, args: list(DHExp.t)): result => {
+  switch (ident) {
+  | "int_of_float" =>
```
cyrus-: let's decentralize this logic into a lookup table in the Builtins module rather than with a switch here, so new builtins only require updating one file ideally

src/hazelweb/layout/DHDoc_Exp.re
```diff
@@ -188,6 +189,19 @@ let rec mk =
         let (doc1, doc2) =
           mk_left_associative_operands(DHDoc_common.precedence_Ap, d1, d2);
         DHDoc_common.mk_Ap(mk_cast(doc1), mk_cast(doc2));
+      | ApBuiltin(x, args) =>
+        switch (args) {
+        // TODO: How to handle tl?
+        | [hd, ..._] =>
```
cyrus-: arguments are missing here -- best way to handle it is probably to call into the Ap case

cyrus- (changes requested):

src/hazelcore/dynamics/Evaluator.re
mirryi: I tried doing this initially, but I got circular module dependency errors (the table in `Builtins` would rely on `Evaluator.result`, and `Evaluator` has to look up in `Builtins`. One solution would be to move `Evaluator.result` into its own module?

cyrus-: Yeah moving Evaluator.result into EvaluatorResult.t makes sense.

cyrus-: actually calling it EvaluationResult.t is better so that we can reuse it even when we have a compiler

mirryi: Cool, I'll do that

src/hazelcore/Builtins.re
```diff
@@ -1,6 +1,46 @@
+module Impls = {
```
cyrus-, line 10: Let's create a single mapping from the builtin name to both its type and its evaluation behavior, so that there is one localized place where each builtin is implemented.

src/hazelcore/Builtins.re
```diff
@@ -1,6 +1,46 @@
...
+  open EvaluatorResult;
+
+  let int_of_float = (ident, args, evaluate) =>
```
cyrus-: It might be easier to define a functor around the entire Builtins module rather than taking in the evaluator in each function.

cyrus- (changes requested):

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -528,9 +526,15 @@ let eval_bin_float_op =
...
 };
 
-let rec evaluate = (d: DHExp.t): result =>
+let rec evaluate = (d: DHExp.t): EvaluatorResult.t =>
   switch (d) {
-  | BoundVar(_) => InvalidInput(1)
+  | BoundVar(x) =>
+    switch (Builtins.lookup_type(x)) {
+    | Some(ty) =>
+      // TODO: Assumes all builtins are functions
```
cyrus-: this approach also means we can't shadow built-in functions, which I think would be fine to allow. 

I think a better approach is to insert all of the built-ins as let bindings wrapping ApBuiltin at the top of the program after elaboration is complete.

src/hazelcore/Builtins.re
mirryi: How can we create a functor for a file, as opposed to a nested module inside the file module?

mirryi: `evaluate` isn't really relevant to other places that use `Builtins` (such as `Contexts.re`, which uses `Builtins.ctx`), so I think that a functor is perhaps not the best approach.

cyrus-: You can define an FBuiltins.F functor that takes the evaluator as input, then define Builtins.re as `include FBuiltins(Evaluator)` and use Builtins normally from there. That breaks up the circularity.

mirryi: Hmm, I tried some different things but couldn't figure out how to do what you mean and avoid a circular dependency. If Builtins has `include FBuiltins(Evaluator)`, and then Evaluator calls Builtins, there would still be a circular dependency?

src/hazelcore/dynamics/Evaluator.re
mirryi: I added a new `elab` function to `Elaborator_Exp.re` that we call from the outside instead of `syn_elab` directly. In there I insert the builtins at the top. Is that what you were thinking?

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,55 @@
+open EvaluatorResult;
+
+module Impl = {
+  [@deriving sexp]
+  type t = (list(DHExp.t), DHExp.t => EvaluatorResult.t) => EvaluatorResult.t;
+
+  let mk_one_arg = (f, ident, args, evaluate) => {
+    let e = DHExp.ApBuiltin(ident, args);
+    switch (args) {
+    | [] => Indet(e)
```
cyrus-: should raise an exception if supplied an incorrect number of arguments, not just return Indet

src/hazelcore/dynamics/Elaborator_Exp.re
```diff
@@ -833,3 +833,22 @@ and ana_elab_rule =
...
+
+let elab = (ctx: Contexts.t, delta: Delta.t, e: UHExp.t): ElaborationResult.t => {
+  switch (syn_elab(ctx, delta, e)) {
+  | Elaborates(d, ty, delta) =>
+    let d' =
+      List.fold_left(
+        (d', (x, _)) =>
+          DHExp.Let(
+            Var(x),
+            Lam(Var("x"), ty, ApBuiltin(x, [BoundVar("x")])),
```
cyrus-: how would this handle builtins with more than one argument, if curried?

cyrus- (changes requested):

src/hazelcore/Builtins.re
mirryi: Fixed.

mirryi:
I reworked much of it so that after elaboration, the let bindings inserted around the expression have a structure that mirrors how it would be if the functions were actually written in Hazel; currying works now (see the `mod` example I added). Defining new built-ins in the `Builtins` module is still pretty much identical to before.

We can also define built-in non-function constants now. I added a `PI` constant for reference.

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,183 @@
...
+[@deriving sexp]
+type t = (Var.t, HTyp.t, form);
+
+module Impl = {
+  [@deriving sexp]
+  type f =
+    (
+      /* args: */ args,
+      /* evaluate: */ DHExp.t => EvaluatorResult.t,
+      /* exp: */ DHExp.t
```
cyrus-: why are we passing in the initial unevaluated expression? to support Indet? if so, it isn't quite right -- the indet expression should have fully evaluated args even if it can't be further reduced, not left entirely unevaluated...

cyrus- (changes requested):

src/hazelcore/Builtins.re
mirryi: Not sure why I did it like this originally... should be fixed now, I think.

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,181 @@
...
+type args = list(DHExp.t);
+
+[@deriving sexp]
+type form = (
+  /* eval: */ (args, DHExp.t => EvaluatorResult.t) => EvaluatorResult.t,
+  /* elab: */ DHExp.t,
+);
+
+[@deriving sexp]
+type t = (Var.t, HTyp.t, form);
```
cyrus-: let's just put the identifier, type, eval function and elab all in one record with field names instead of broken out like this

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,181 @@
...
+
+module Impl = {
+  [@deriving sexp]
+  type f =
+    (/* args: */ args, /* evaluate: */ DHExp.t => EvaluatorResult.t) =>
```
cyrus-: probably good to define an evaluator type alias for readability

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,181 @@
...
+    mk(ident, ty, fn);
+  };
+};
+
+module Impls = {
+  open EvaluatorResult;
+
+  /* int_of_float implementation. */
+  let int_of_float = (ident, d1') =>
+    switch (d1') {
```
cyrus-: Let's use `r` variables for evaluator results and keep `d` only for DHExps so that its easier to read

cyrus- (changes requested):
OK looks much better now. A few last tweaks!

mirryi:
Implemented the suggestions. Everything look good now?

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,190 @@
...
+type args = list(DHExp.t);
+
+[@deriving sexp]
+type eval = (args, DHExp.t => EvaluatorResult.t) => EvaluatorResult.t;
+
+[@deriving sexp]
+type elab = DHExp.t;
+
+[@deriving sexp]
+type t = {
```
cyrus-: Naming is maybe still a bit off. This type should probably be called Builtin.t. Then you can define a Builtins.t module which actually defines the specific builtins that we are defining for Hazel right now.

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,190 @@
...
+  ty: HTyp.t,
+  eval,
+  elab,
+};
+
+module Impl = {
+  type evaluate = DHExp.t => EvaluatorResult.t;
+
+  [@deriving sexp]
+  type f = (/* args: */ args, /* evaluate: */ evaluate) => EvaluatorResult.t;
```
cyrus-: these comments can be type annotations now I think?

cyrus- (changes requested):

src/hazelcore/Builtins.re
mirryi: Alright, that sounds good. `mod` and `PI` I added for making sure currying and constants work. Do we want to keep those? And what other built-ins might be wanted for now?

src/hazelcore/Builtin.re
```diff
@@ -0,0 +1,115 @@
...
+    };
+
+    mk_elab_inner(ty, 0, []);
+  };
+
+  /*
+     Create a built-in function.
+   */
+  let mk = (ident: Var.t, ty: HTyp.t, fn: f): t => {
+    let eval = (args, evaluate) => fn(args, evaluate);
```
cyrus-: isn't this the same as saying let eval = f?

src/hazelcore/Builtins.re
```diff
@@ -0,0 +1,66 @@
...
+module Impls = {
+  open EvaluatorResult;
+
+  /* int_of_float implementation. */
+  let int_of_float = (ident, r1) =>
+    switch (r1) {
+    | BoxedValue(FloatLit(f)) =>
+      let i = int_of_float(f);
+      BoxedValue(IntLit(i));
+    | BoxedValue(d1)
```
cyrus-: if its any boxed value other than a float lit it should raise an exception not return Indet (also in other cases in this file)

cyrus- (changes requested):

src/hazelcore/Builtin.re
mirryi: Oops, yeah. Left over from when I was errneously passing the unevaluated expression into fn.

src/hazelcore/Builtin.re
```diff
@@ -0,0 +1,113 @@
...
+  ident: Var.t,
+  ty: HTyp.t,
+  eval,
+  elab,
+};
+
+module Impl = {
+  /* Evaluator alias. */
+  [@deriving sexp]
+  type evaluate = DHExp.t => EvaluatorResult.t;
```
cyrus-: this is used in the type `eval` above, maybe move this out to the top level / interface.

everything else in Impl you can also move out and just leave out of the rei interface to avoid it being exported

cyrus- (changes requested):
one last cleanup :)

mirryi:
Are these all the built-ins we want for now? Do we want to add/remove any?

cyrus-:
These should be fine for now -- I don't want to flood the context sidebar with builtins right now. I'm going to add "organize context hierarchically" as a starter project, but we don't need to do that to merge this mechanism in. It looks pretty straightforward to add new builtins now so adding more builtins as needed can be handled by other folks as needed.

cyrus- (approved):
Looks great now, thanks Eric!

PR #555, opened 2021-11-03
Replace InvalidInput codes with proper EvaluatorError type

### LINKED ISSUE #549: Replace InvalidInput codes with proper EvalError type
The evaluator currently uses an `InvalidInput(int)` variant of `Evaluator.result` for returning from evaluation branches that shouldn't ever execute. We should replace those codes with a more informative `EvalError` type.

mirryi:
Resolves #549.

### DIFF
--- a/src/hazelcore/dynamics/Evaluator.re
+++ b/src/hazelcore/dynamics/Evaluator.re
@@ -1,6 +1,5 @@
 [@deriving sexp]
 type result =
-  | InvalidInput(EvaluatorError.t)
   | BoxedValue(DHExp.t)
   | Indet(DHExp.t);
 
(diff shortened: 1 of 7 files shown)


disconcision:
definitely a positive change, but i do wonder why we don't just throw exceptions here, given that all these conditions are non-recoverable and indicate an error in either the typechecker, elaborator, or evaluator.

(signed T. looking to reduce the height of the evaluator monad transformer stack)

mirryi:
Perhaps it was initially the idea that the core stuff (typechecking, elaboration, evaluation) should be purely functional, and to move all the user-facing stuff out (in this case, notifying the user/developer of the bug). Just a guess though.

cyrus-:
Yeah that decision was a remnant of doing everything in Coq at one point -- let's refactor it to use an exception instead.

mirryi:
Then I should just move `Program.InvalidInput` exception directly into `Evaluator` and raise immediately?

cyrus-:
Yep that makes sense. You should be able to reuse the error code datatype you defined here.

cyrus- (approved):

cyrus-:
Thanks, Eric!

PR #557, opened 2021-11-03
fix 556 impossible syn

### LINKED ISSUE #556: Failure: extend_let_body_ctx: impossible syn
Something is wrong here: 
Screenshot from 2021-11-03 15-06-01

disconcision:
closes #556.

not totally sure about this one. i still don't understand where hole fixing is actually supposed to happen, canonically speaking


src/hazelcore/Action_Exp.re
```diff
@@ -1068,6 +1068,8 @@ and syn_perform_line =
       | Succeeded((new_zdef, u_gen)) =>
         let new_zline = ZExp.LetLineZE(p, new_zdef);
         let new_def = ZExp.erase(new_zdef);
+        let (new_def, _, u_gen) =
+          Statics_Exp.syn_fix_holes(def_ctx, u_gen, new_def);
```
cyrus-, line 1099: Invariant we need to maintain is that actions preserve typing (sensibility). The Statics_Exp rule for LetLine says that the def must be able to synthesize, and then the pattern must analyze against the synthesized type. However, the LetLineZP case above does an ana_fix_holes on the def, and this case also does an ana_perform indicating that we're treating this as an analytic position. Shouldn't be doing a syn_fix_holes in that case. The correct resolution is probably to change Statics_Exp to be consistent with what we seem to be doing here, which is synthesizing a type for the pattern and then analyzing the def against that type.

cyrus- (changes requested):
See comment. This branch is also several commits behind `dev`.

src/hazelcore/Action_Exp.re
disconcision: I feel like we've been through this one before, or even changed it around multiple times now, probably to do with the possibly-conflicting goal of making sure we have access to the joined type information in both the pattern and the definition.

PR #558, opened 2021-11-04
Parsing

nj-wilson:
Adds ability to import text into hazel as a program, as well as exporting the current program to plain text.

Currently is part of the hamburger menu as separate options (could easily be combined).
Import errors are shown, but currently are quite rudimentary, only showing what character caused a problem.

A couple commits are still local, for cleanup and additional tests.
There is a behavior difference with regard to empty or comment lines, but nothing that impacts correct programs.

ExternalLineBreak was added to Layout.t, but may be better as an Annot (for a future PR).
The module naming for the parser and lexer are somewhat confusing, i.e. Parsing.Parser, and Lexing.Lexer. Could use better names.


src/hazelcore/parse/dune
```diff
@@ -0,0 +1,14 @@
+(include_subdirs no)
```
cyrus-: Since it can be its own library, lets move the parser out to the `src` directory instead of it being inside `hazelcore`. You can call it `hazeltext`.

src/hazelcore/parse/print.re
```diff
@@ -0,0 +1,19 @@
...
+      (s1 ++ s2, c2);
+    | Linebreak => ("\n" ++ String.make(indent, ' '), indent)
+    | ExternalLinebreak => (";\n" ++ String.make(indent, ' '), indent)
+    | Align(l) => go(col, col, l)
+    | Annot(UHAnnot.HoleLabel(_), _) => ("_?", col + 2)
+    | Annot(_, l) => go(indent, col, l)
+    };
+  let (s, _) = go(0, 0, l);
+  s;
+};
```
cyrus-: let's add an end-to-end `print_exp : UHExp.t -> string` as well

src/hazelcore/parse/test/test.ml
```diff
@@ -0,0 +1,75 @@
...
+let test_parse text: bool =
+ (*Get the first AST*)
+  let ast_a = parse text in
+  match ast_a with
+  | Some(ast) ->
+    let doc = Lazy.force UHDoc_Exp.mk ~memoize:true ~enforce_inline:false ast in
+    let lay = Pretty.LayoutOfDoc.layout_of_doc ~width:100 ~pos:0 doc in
+    let l = match lay with
+    | Some l -> l
+    | None -> Pretty.Layout.Text("")
```
cyrus-: this should be a failure case, not ""

src/hazelcore/parse/parse.mly
```diff
@@ -0,0 +1,223 @@
...
+;
+
+%inline typ_op:
+  COMMA { Operators_Typ.Prod }
+  | TARROW { Operators_Typ.Arrow }
+  | BAR { Operators_Typ.Sum }
+;
+
+pat:
+  pat COLONCOLON pat { mk_binop $1 Operators_Pat.Cons $3 }
```
cyrus-: type annotations can now appear on any pattern, not just at the top level of a let or lambda

src/hazelcore/parse/test/test.ml
```diff
@@ -0,0 +1,75 @@
...
+  let map : (Int -> Int) -> [Int] -> [Int] =
+      \\f.{
+        \\xs.{
+          case xs
+          | [] => []
+          | y::ys => (f y)::(map f ys)
+          end
+        }
+      }
+  in map"
```
cyrus-: let's add a few tests of longer opseqs involving all of the various operators

src/pretty/Layout.re
```diff
@@ -6,6 +6,7 @@ type t('annot) =
   | Text(string) // Invariant: contains no newlines. Text("") is identity for `Cat`
   | Cat(t('annot), t('annot)) // Associative
   | Linebreak
+  | ExternalLinebreak
```
cyrus-: Let's use Annot in this PR rather than changing pretty

cyrus- (changes requested):

cyrus- (changes requested):
Given that the popup text editor looks almost exactly like the structure editor UI, what if we actually had it integrated into the main editor as an editing mode toggle -- you can either edit using the structure editor or text editor, and there was a toggle to go back and forth which would trigger parsing/pretty printing as necessary?

In text mode, we could have it attempt to parse on every edit and run the evaluator as usual, but if parsing fails then you get the parse error instead of the result?

nj-wilson:
> 

Is an interesting idea, I shall look into it.
My only questions would be on how to display errors resulting in trying to go back to structured, though maybe that'll be taking care of from running the evaluator during edit. Though I do wonder if that would be problematic to run on every character typed.

I'll also have to see how easy it would be to switch the main text box between modes, or do some other method.

cyrus-:
You can use the same area in the UI that results are displayed to show errors, since no result is available when parsing fails, and the button to toggle back to structure editing can be disabled when there is a parse error. Don't worry about the performance cost of running the parser on each edit. If that becomes an issue down the line we can look into incremental parsing approaches, but nothing at the scale we're using Hazel should cause an issue. Learn OCaml already parses and pretty prints on every character. 

Shouldn't be too difficult GUI wise with the model-view-controller approach. When in text mode, render the text box you already have set up from the pop up in your branch, otherwise render the existing structure editor box.

cyrus-:
no longer relevant with the two other PRs.

PR #559, opened 2021-11-05
Add primitive String type

mirryi:
This PR adds a primitive string type, string literals, and builtin operations on those strings:

- String literals 
  - [x] OCaml escape sequences (`"abc\t\n")
    - [x] Strings with invalid escapes are indet values 
  - [x] Evaluation result syntax highlighting
    - [x] Invalid escape sequences

- String operations
  - [x] Binary concatenation operator (`s1 ^ s2`)
    - [x] Doesn't work on indet values (maybe it should)
  - [x] Subscript operator (`s1[n1:n2]`)
    - [x] Correct error handling and indicators for out-of-bound indices
    - [x] Also doesn't work on indet values (maybe it should)
  - Builtin functions for reading strings
    - [x] `length : String -> Int` returns length of a string
    - [x] `{int,float,bool}_of_string : String -> {Int,Float,Bool}` parses a string as an integer/float/bool (can fail)
    - [x] `string_of_{int,float,bool} : {Int,Float,Bool} -> String` converts an integer/float/bool to a string
  
- String pattern matching
  - [x] String literal patterns

- Parser integration
  - [x] String literal expressions and patterns
  - [x] Binary concatenation operator
  - [x] Mix-fix subscript operation

Leaving off for now:

- Code syntax highlighting


mirryi:
There are still some issues that I observed from the Hazel assistant's popup with regards to type analysis, but otherwise I think everything is pretty much here?

cyrus-:
@Dophin2009 sorry for the delay in reviewing this. can you get it up to date with dev again?

mirryi:
Feeling especially not confident about this code now after a few months...

cyrus-:
Do you want to give it another pass or should I review it soon?

mirryi:
Probably a good idea. I'll give it another pass today/tomorrow.

mirryi:
@cyrus- Should be good for a review now, I think!

cyrus-:
In the previous version of the strings branch, we had syntax highlighting for escape characters, including red for invalid escape characters. That seems to be missing here.
Screenshot from 2022-06-15 16-50-33

mirryi:
I still have next to no real clue how the layout/doc system works, so I'm having a problem with implementing the syntax highlighting:

image

If you were to hit the right arrow key when the cursor is right after `abc` and before `\`, it'd jump back to the beginning of the string, right after the quotation mark, instead of to the right of `\`. I think it's because the `abc` and `\` segments are two different `Doc.t` or something?

I modeled [the doc code](https://github.com/hazelgrove/hazel/blob/f07e4dd1ee1389d899e3a6ff7f969e2fa6afbd60/src/hazelweb/layout/UHDoc_common.re#L340) off of [the code in strings3](https://github.com/hazelgrove/hazel/blob/255516cac90a00bcd7d2189599c8edd2591ecf9c/src/hazelweb/layout/UHDoc_common.re#L176), but didn't want to repeat the parsing logic (which is already in `StringLitLexer`). Following strings3, I modified the [`Text` variant of `UHAnnot.token_shape`](https://github.com/hazelgrove/hazel/blob/f07e4dd1ee1389d899e3a6ff7f969e2fa6afbd60/src/hazelweb/layout/UHAnnot.re#L10) by adding a start index field, but then I'm not really sure what to do with it (e.g. [in `UHMeasuredLayout.re`](https://github.com/hazelgrove/hazel/blob/f07e4dd1ee1389d899e3a6ff7f969e2fa6afbd60/src/hazelweb/layout/UHMeasuredLayout.re#L23)).

mirryi:
Also, I fixed the escape sequence error highlighting for the evaluation code box.

cyrus-:
Hm, not sure what the problem is. can you point me to the relevant code?

cyrus-:
You can look at Mengjing's strings branch, I believe she had it working before...

mirryi:
Yeah, it's the code I linked in the above comment (`UHDoc_common`). I've been going off the strings3 branch, which seemed to duplicate a bunch of escape sequence parsing logic for syntax highlighting purposes? I made it so that lexer already produces the information necessary for that. 

If you were to make `inner` in `UHDoc_common.mk_StringLit` to just `mk_text(s)`, it would have no syntax highlighting but work fine. I followed strings3 by changing `UHAnnot.token_shape` like I said above, but the layout/doc code has changed a lot since then, so I'm not sure how to use it to make movements work correctly.

cyrus-:
Hm, I'm not super familiar with that code either. @disconcision can you take a quick look?

disconcision:
I took a quick look but don't really have an idea of what's going on... I have basic understanding of the layout system but I've never touched measured layout. I don't immediately understand the way strings are constructed here, I'd have to dig into the code in-depth before I could start asking the right questions.

From the nature of the bug (looping on first segment as opposed to getting stuck or skipping out of the string) I would tend to guess that there is something like two indices to represent the position in a string, the segment and the position within the segment, and the latter is being correctly reset at the end of a segment but the former is erroneously not being incremented. But I had trouble finding where in the code the relevant logic is done.

cyrus-:
@mirryi mark as ready for review when its good to go again

cyrus-:
A few things that we should fix when strings are reimplemented via tylr, but that we can pass on for now:
* highlighting of escape sequences and bad escape sequences
* be able to press enter within a string literal
* think about how we want to handle closing quotation marks vs. internal quotation marks (currently, closing quotation mark is automatic so `"` key inserts the quotation mark character
* undo history shows the entire string literal in the message even if it is very long, probably want to elide after a certain length and show only the part that changed
* ~~term highlighting on subscript is incorrect, should show start and end indices as open children~~

However, I'm okay with merging it pending a code review so we have baseline behavior specification for the tylr strings effort in dev.

cyrus-:
thing we should fix now: the `length` built in function doesn't return an integer or error when given a string with an invalid escape sequence in it

Screenshot from 2022-07-14 00-06-47

src/hazelcore/UHExp.rei
```diff
@@ -17,10 +17,12 @@ and operand =
   | IntLit(ErrStatus.t, string)
   | FloatLit(ErrStatus.t, string)
   | BoolLit(ErrStatus.t, bool)
+  | StringLit(ErrStatus.t, string)
```
cyrus-: string argument here should really be of type UnescapedString.t, right?

mirryi: No, since the strings in UHExp are unparsed. They're parsed during elaboration. Maybe "unescaped" isn't the clearest naming, though.

cyrus-: Oh I see, the name makes sense but maybe we should also define EscapedString.t so its clear which is which

src/hazelcore/UHExp.re
```diff
@@ -19,10 +19,12 @@ and operand =
   | IntLit(ErrStatus.t, string)
   | FloatLit(ErrStatus.t, string)
   | BoolLit(ErrStatus.t, bool)
+  | StringLit(ErrStatus.t, string)
   | ListNil(ErrStatus.t)
   | Fun(ErrStatus.t, UHPat.t, t)
   | Inj(ErrStatus.t, InjSide.t, t)
   | Case(CaseErrStatus.t, t, rules)
+  | Subscript(ErrStatus.t, t, t, t)
```
cyrus-, line 27: It might make more sense to have the subscript target be of type operand rather than t, since we don't allow it to be an opseq or any other kind of block expression anyway.

src/hazelcore/ZExp.re
```diff
@@ -626,7 +697,10 @@ and move_cursor_left_zoperand =
   fun
   | z when is_before_zoperand(z) => None
   | CursorE(OnOp(_), _) => None
-  | CursorE(OnText(j), e) => Some(CursorE(OnText(j - 1), e))
+  | CursorE(OnText(j), e) => {
+      print_endline(string_of_int(j));
```
cyrus-: remove

src/hazelcore/ZExp.re
```diff
@@ -626,7 +697,10 @@ and move_cursor_left_zoperand =
...
+      Some(CursorE(OnText(j - 1), e));
```
cyrus-: doesn't this break when j = 0?

src/hazelcore/ZExp.re
```diff
@@ -805,7 +904,10 @@ and move_cursor_right_zoperand =
   fun
   | z when is_after_zoperand(z) => None
   | CursorE(OnOp(_), _) => None
-  | CursorE(OnText(j), e) => Some(CursorE(OnText(j + 1), e))
+  | CursorE(OnText(j), e) => {
+      print_endline(string_of_int(j));
```
cyrus-: remove

src/hazelcore/ZExp.re
```diff
@@ -805,7 +904,10 @@ and move_cursor_right_zoperand =
...
+      Some(CursorE(OnText(j + 1), e));
```
cyrus-: what about if you hit the right side?

src/hazelcore/ZPat.re
```diff
@@ -330,7 +337,8 @@ and move_cursor_right_zoperand =
   | CursorP(
       OnDelim(_, _),
       InvalidText(_, _) | Var(_, _, _) | BoolLit(_, _) | IntLit(_, _) |
-      FloatLit(_, _),
+      FloatLit(_, _) |
+      StringLit(_, _),
```
cyrus-: String literals have delimiters...

src/hazelcore/ZPat.re
```diff
@@ -269,7 +275,8 @@ and move_cursor_left_zoperand =
...
     ) =>
     // invalid cursor position
```
cyrus-: String literals have delimiters...

cyrus- (changes requested):
functionality seems good now, partial review of syntax

src/hazelcore/ZExp.re
mirryi: Hmm, looks like [this has always been here](https://github.com/hazelgrove/hazel/blob/dev/src/hazelcore/ZExp.re#L629=). I guess we should add a guard to do the above when `j > 0` and return `None` otherwise?

cyrus-: Yeah, though analogous issue happens for moving right and there you'd have to actually look at the rightmost valid position (which is returned from the valid cursor positions function)

src/hazelcore/UHExp.re
mirryi: Not sure how to resolve one part of Action_Exp with this change, since `ana_perform` returns a `ZExp.t`, which I'm not sure how to fit back into `ZExp.SubscriptZE1` (the target of which I changed to `ZExp.zoperand`): [this part of the function](https://github.com/hazelgrove/hazel/blob/1d76b18930e2e17026041db617488df91e6975c9/src/hazelcore/Action_Exp.re#L2059-L2098), and a similiar part down below.

cyrus-:
Archiving this Hazel 2 PR, but see #786 -- this PR had many more features than the very basic string literals we have in Hazel 3 right now, so it should serve as a reference for how we eventually want to realize strings. Thank you all for the work on this!

PR #564, opened 2021-11-17
Tests 1

### LINKED ISSUE #618: Parity with Learn OCaml testing system
Can break down into individual tasks as needed

disconcision:
NOTE: when this gets merged, we might want to advise people to do a codebase-wide rename of u_gen to id_gen before they pull this in... might make it less painful.

Supersedes #402 and see also previous discussion on #287

Features:

1. Test Panel on sidebar: collects all instances of all tests, displayed and labelled in lexicographic order. indicates pass or fail status for each test (color, outline) and the joint status over test instances (check). the joint status check is also displayed on syntax. clicking on a row in the test panel moves the cursor to that test on the code. clicking in a hole in an incomplete test puts the caret in that hole on the code.

2. test display results are, if determinate, the boolean result of the predicate, EXCEPT in the case of an equality (==,==.), in which case, both sides are evaluated and the resulting equality with values on both sides is returned instead.

3. Test inspector on cursor: replaces cursor inspector when on a test. displays the result of the test (if any) at cursor.

4. Adds fuel, which is burned

Implementation:

0. Replaces u_gen with id_gen, for threading through multiple id generators. This is part of the big reason the diff is so huge.

1. New operand/data type Keyword, of which Test is the only current inhabitant. Some old 'Keyword' modules renamed to ExpandingKeyword. 

2. Evaluator now has a state map running through it. Useful for a lot of things. The evaluator could be formally monaded.

3. Loosely related cleanup of top-level UI in Page.re and elsewhere

Known issues:

1. Fuel isn't totally working. Trivial nonterminating programs fail successfully, but larger nonterminating programs seem to die on the way back down, ie after having run out of fuel

Addressed issues:

1. All of (Binary ops, unary aps, binary aps) now show the evaluated arguments in the test inspector/bench.
2. Switched to a lighter red/green to lessen xmas effect.
3. Added completion percentage.

Not in scope (separate projects):

1. Concurrent evaluation
2. Polymorphic equality (needs new error code - complicated refactor)
3. Value diff holes (needs polyeq plus DHDeco architecture to gets paths to draw)
4. Test labels/descriptions (needs strings, though I have a different idea to use comments as compositional labels)


cyrus-:
Screenshot from 2022-02-21 00-22-44

Something wrong with test numbering going on here...

(also needs fresh merge with dev)

disconcision:
Above big fixed; merged with dev.

cyrus-:
@dlz01 this shouldn't be closed

dlz01:
I really apologized for that, could you reopen it? I clicked it by mistake

cyrus-:
Subsumed by school mode in Hazel 3. Thank you for laying the foundations here!

PR #566, opened 2021-11-20
Make impossible syn more impossibler

### LINKED ISSUE #615: Let expressions (incl. fixes)

disconcision:
This is a followup to pattern annotations intended to resolve an unanticipated interaction between lets and cases which manifests as the failure of an assertion within statics. It adds two things:

1. A hole type 'provenance' system which adds a flag to type holes which tracks whether or not they originate from an empty type annotation.
2. A new error status for cases which splits InconsistentBranches into InconsistentBranches(Syn) (previously existing case) and a new InconsistentBranches(Ana) case. The latter is not drawn, but behaves as a hole in the limited sense that it synthesizes hole type.


src/hazelcore/HTyp.re
```diff
@@ -79,7 +82,7 @@ let rec consistent_all = (types: list(t)): bool =>
 /* matched arrow types */
 let matched_arrow =
   fun
-  | Hole => Some((Hole, Hole))
+  | Hole(_) => Some((Hole(None), Hole(None)))
```
cyrus-: shouldn't generate None if the original hole was in the program, otherwise function applications like `? (case ...)` will not quite do what you'd expect (analyze the case against hole rather than switch to syn)

src/hazelcore/HTyp.re
```diff
@@ -93,21 +96,21 @@ let get_prod_arity = ty => ty |> get_prod_elements |> List.length;
 /* matched sum types */
 let matched_sum =
   fun
-  | Hole => Some((Hole, Hole))
+  | Hole(_) => Some((Hole(None), Hole(None)))
```
cyrus-: same here and below

src/hazelcore/HTyp.re
```diff
@@ -118,14 +121,14 @@ let rec complete =
 
 let rec join = (j, ty1, ty2) =>
   switch (ty1, ty2) {
-  | (_, Hole) =>
+  | (_, Hole(_)) =>
     switch (j) {
-    | GLB => Some(Hole)
+    | GLB => Some(Hole(None))
```
cyrus-: same thing here -- Hole(None) should only arise from an unannotated pattern variable/wildcard, otherwise it should be Some(()) for now (for the type hole inference work, we'll generalize this to be a provenance)

src/hazelcore/HTyp.rei
```diff
@@ -1,7 +1,11 @@
 /* types with holes */
+
+[@deriving sexp]
+type hole_label = option(unit);
```
cyrus-: hole_provenance might be the better name for this (see comment in HTyp.re)

src/hazelcore/Statics_Exp.re
```diff
@@ -64,7 +65,10 @@ and syn_line = (ctx: Contexts.t, line: UHExp.line): option(Contexts.t) =>
     let+ _ = syn_opseq(ctx, opseq);
     ctx;
   | LetLine(p, def) =>
+    let* (ty_p, _) = Statics_Pat.syn(ctx, p);
     let def_ctx = extend_let_def_ctx(ctx, p, def);
+    let* _ = ana(def_ctx, def, ty_p);
+    // TODO(andrew): ???
```
cyrus-: yes this looks right

src/hazelcore/Statics_Exp.re
```diff
@@ -80,7 +84,7 @@ and syn_skel =
   | BinOp(InHole(_), op, skel1, skel2) =>
     let skel_not_in_hole = Skel.BinOp(NotInHole, op, skel1, skel2);
     let+ _ = syn_skel(ctx, skel_not_in_hole, seq);
-    HTyp.Hole;
+    HTyp.Hole(None);
```
cyrus-: see comments above -- None shouldn't arise except from unannotated variables

src/hazelcore/Statics_Exp.re
```diff
@@ -158,23 +162,24 @@ and syn_operand = (ctx: Contexts.t, operand: UHExp.operand): option(HTyp.t) =>
...
     HTyp.Arrow(ty_p, ty_body);
   | Inj(NotInHole, side, body) =>
     let+ ty = syn(ctx, body);
     switch (side) {
-    | L => HTyp.Sum(ty, Hole)
-    | R => Sum(Hole, ty)
+    | L => HTyp.Sum(ty, Hole(None))
+    | R => Sum(Hole(None), ty)
     };
   | Case(StandardErrStatus(NotInHole), scrut, rules) =>
```
cyrus-: I think Case can't actually have a TypeInconsistency error on it so we might want to refactor this to be either NotInHole or an InconsistentBranches.

src/hazelcore/Statics_Exp.re
```diff
@@ -837,7 +848,7 @@ and syn_fix_holes_operand =
       let (u, u_gen) = MetaVarGen.next(u_gen);
       (
         Case(InconsistentBranches(rule_types, u), scrut, rules),
-        HTyp.Hole,
+        HTyp.Hole(Some()), //TODO(andrew)
         u_gen,
```
cyrus-: this isn't a Hole(Some()) case either for the same reasons

src/hazelcore/Statics_Exp.re
```diff
@@ -1270,18 +1281,33 @@ and ana_fix_holes_operand =
...
-    (Case(StandardErrStatus(NotInHole), scrut, rules), u_gen);
+    print_endline("ana_fix_holes: CASE");
+    print_endline(Sexplib.Sexp.to_string_hum(HTyp.sexp_of_t(ty)));
+
+    switch (ty) {
+    | Hole(Some ()) =>
+      print_endline("case ana fix holes Hole(Some()) case");
+      print_endline(Sexplib.Sexp.to_string_hum(UHExp.sexp_of_operand(e)));
+      let (op, _, u_gen) =
+        syn_fix_holes_operand(ctx, u_gen, ~renumber_empty_holes, e);
```
cyrus-: this looks right

src/hazelcore/Statics_Pat.re
```diff
@@ -65,24 +65,27 @@ and syn_operand =
   | Inj(InHole(WrongLength, _), _, _)
   | TypeAnn(InHole(WrongLength, _), _, _) => None
   /* not in hole */
-  | Wild(NotInHole) => Some((Hole, ctx))
+  | Wild(NotInHole) => Some((Hole(Some()), ctx)) // TODO(andrew)
```
cyrus-: this should be None (unannotated pattern)

src/hazelcore/Statics_Pat.re
```diff
@@ -65,24 +65,27 @@ and syn_operand =
...
   | Var(NotInHole, InVarHole(Free, _), _) => raise(UHPat.FreeVarInPat)
-  | Var(NotInHole, InVarHole(Keyword(_), _), _) => Some((Hole, ctx))
+  | Var(NotInHole, InVarHole(Keyword(_), _), _) => Some((Hole(None), ctx)) //TODO(andrew)
   | Var(NotInHole, NotInVarHole, x) =>
     Var.check_valid(
-      x,
-      Some((HTyp.Hole, Contexts.extend_gamma(ctx, (x, Hole)))),
+      x, //TODO(andrew): correct?
+      Some((
+        HTyp.Hole(Some()),
```
cyrus-: None case here as well

cyrus- (changes requested):
The Hole(None) cases should almost all be Hole(Some()) cases -- None (the "unspecified" unknown type) should only be used for the type of a variable pattern or wildcard pattern in synthetic position. Everything else is the "specified" unknown type. See specific comments.

disconcision:
Hmm yeah I think I got mixed up as to which hole case was which about halfway through.. maybe I should just create a data type so they're labelled?

disconcision:
This is good for another look. Unnannoted lets seem to work now, but not sure I have the cursorinfo logic right.

src/hazelcore/Action_Exp.re
```diff
@@ -1085,16 +1087,18 @@ and syn_perform_line =
...
       switch (ana_perform(def_ctx, a, (zdef, u_gen), ty_p)) {
       | Failed => Failed
       | CursorEscaped(side) => escape(u_gen, side)
       | Succeeded((new_zdef, u_gen)) =>
-        let new_zline = ZExp.LetLineZE(p, new_zdef);
-        let new_def = ZExp.erase(new_zdef);
-        let body_ctx = Statics_Exp.extend_let_body_ctx(ctx, p, new_def);
-        Succeeded(LineDone((([], new_zline, []), body_ctx, u_gen)));
+        let (new_zdef, u_gen) =
+          Statics_Exp.ana_fix_holes_z(def_ctx, u_gen, new_zdef, ty_p);
```
cyrus-: ana_perform should inductively have already fixed the holes so this shouldn't be necessary

src/hazelcore/HTyp.re
```diff
@@ -1,9 +1,12 @@
 open Sexplib.Std;
 
+[@deriving sexp]
+type hole_provenance = option(unit);
```
cyrus-: My interpretation was that Some(()) meant that it was a type hole (i.e. from a syntactic location in the code -- in the future, the Some case would contain the provenance data itself) and None was that it was an internally generated unknown type, but that seems to be backwards from what you implemented e.g. in Action_Exp? (and perhaps other places, haven't read through it all yet...)

cyrus- (changes requested):

src/hazelcore/Action_Exp.re
```diff
@@ -289,11 +289,13 @@ let mk_syn_text =
   | InvalidTextShape(t) =>
     if (text |> StringUtil.is_empty) {
       let (zhole, u_gen) = u_gen |> ZExp.new_EmptyHole;
-      Succeeded(SynDone((ZExp.ZBlock.wrap(zhole), HTyp.Hole, u_gen)));
+      Succeeded(
+        SynDone((ZExp.ZBlock.wrap(zhole), HTyp.Hole(Some()), u_gen)),
```
cyrus-: (see comment below -- my impression was that this should be a None case, not a Some() case because there is no syntactic type hole involved here

src/hazelcore/HTyp.re
disconcision: My recollection is that this was the way I did it initially, and then you told me it should be the other way around. I've definitely already done a bulk-replace to get them in the current correspondence.

disconcision: Unless these are inconsistencies you're pointing out versus a systematic issue?

cyrus-: My recollection was that we had settled on the interpretation above, which seems to make sense to me. Apologies if it was unclear though. It's probably worth explicitly documenting what Some() and None mean on this line and then make sure that is how they are used. If it is systematically swapped everywhere then that should be a simple find/replace to fix it. But I can also do a more careful pass through the key syn/ana branches once that is settled...

src/hazelcore/Action_Exp.re
disconcision: If I remove this, I start getting impossible syn again in both the annotated and unannotated cases... not sure what's up

disconcision:
So to review the new nomenclature... on the board you sketched out a two-variant type called unknown with variants 
'TypHole(MetaVar.t)' and 'Internal'. but what is this replacing exactly? like should the Hole case in HTyp become Hole(uknown), or should I add TypHole(..) and Internal as top-level variants in HTyp.t?

disconcision:
Also to make sure of the mapping.. Internal is current Hole(None) right and TypHole(MetaVar.t) is Hole(Some())? at least I thought it should be this way but the board says the other way around...?

src/hazelcore/UHTyp.re
```diff
@@ -56,7 +56,7 @@ let contract = (ty: HTyp.t): t => {
   and contract_to_seq = (~parenthesize=false, ty: HTyp.t) => {
     let seq =
       switch (ty) {
-      | Hole => Seq.wrap(Hole)
+      | Unknown(_) => Seq.wrap(Hole)
```
cyrus-: write out the three cases instead of wildcard here since this wouldn't always just go to Hole if we had unique IDs

src/hazelcore/HTyp.re
```diff
@@ -1,9 +1,15 @@
 open Sexplib.Std;
 
+[@deriving sexp]
+type hole_provenance =
```
cyrus-: lets call it unknown_type_provenance since only some unknown types are holes

src/hazelcore/HTyp.re
```diff
@@ -93,21 +99,21 @@ let get_prod_arity = ty => ty |> get_prod_elements |> List.length;
...
 
 /* matched list types */
 let matched_list =
   fun
-  | Hole => Some(Hole)
+  | Unknown(prov) => Some(Unknown(prov))
   | List(ty) => Some(ty)
   | _ => None;
 
 /* complete (i.e. does not have any holes) */
```
cyrus-: I don't think we use complete anywhere, and its a bit underspecified here since its detecting all unknowns not just holes. Assuming it isn't actually used anywhere lets go ahead and remove this. We need only define it for UHTyp, not HTyp.

src/hazelcore/HTyp.re
```diff
@@ -118,14 +124,14 @@ let rec complete =
 
 let rec join = (j, ty1, ty2) =>
   switch (ty1, ty2) {
-  | (_, Hole) =>
+  | (_, Unknown(x)) =>
     switch (j) {
-    | GLB => Some(Hole)
+    | GLB => Some(Unknown(x))
```
cyrus-: I think the provenance here should always be Internal rather than inheriting from the one type, since there are potentially two Unknowns involved and this would just be picking one of them arbitrarily.

src/hazelcore/HTyp.rei
```diff
@@ -1,7 +1,17 @@
 /* types with holes */
+
+[@deriving sexp]
+type hole_provenance =
+  | SynPatternVar
+  /* comes from a pattern var being asked to synthesize a type either directly or via matched arrow/prod/etc.
+     When analyzing against such a type, expression can be treated as if synthetic when the distinction matters
+     e.g. for inconsistent branches or cast insertion */
+  | Internal /* other internally generated unknown types, e.g. the type synthesized by a hole in synthetic position etc. */
+  | TypHole; /* comes from an actual type hole, will add an `of MetaVar.t`, once we have unique IDs for type holes */
```
cyrus-: for readability lets make this the first constructor of the three and Internal the last

src/hazelcore/dynamics/Elaborator_Exp.re
```diff
@@ -93,7 +93,12 @@ and syn_elab_line =
               ),
             )
           };
-        let d1 = DHExp.cast(d1, ty1', ty1);
+        let d1 =
+          switch (ty1) {
```
cyrus-: this logic might be better to put into DHExp.cast function since I think you'll need it in a few places -- a cast with target type Unknown(SynPatternVar) should be a no op

cyrus- (changes requested):
partial review of recent discussion items

src/hazelcore/HTyp.re
cyrus-: or just provenance

src/hazelcore/Action_Exp.re
cyrus-: check the zipper case in ana_perform for case?

src/hazelcore/Action_Exp.re
```diff
@@ -1511,12 +1517,16 @@ and syn_perform_operand =
   /* \x :<| Int . x + 1   ==>   \x| . x + 1 */
   | (Backspace, CursorE(OnDelim(1, After), Lam(_, p, body))) =>
     let (p, body_ctx, u_gen) =
-      Statics_Pat.ana_fix_holes(ctx, u_gen, p, Hole);
+      Statics_Pat.ana_fix_holes(ctx, u_gen, p, Unknown(Internal));
```
cyrus-: I think unrelated but maybe could cause related problems -- this seems incorrect? we should be synthesizing the argument pattern for a lambda in synthetic position right?

src/hazelcore/Action_Exp.re
```diff
@@ -3458,3 +3476,25 @@ and ana_perform_subsume =
       }
     }
   };
+
+let syn_perform =
```
cyrus-: this is from that conversation we had about adding contracts into the codebase. probably shouldn't be in this PR.

src/hazelcore/CursorInfo_Exp.re
```diff
@@ -868,18 +892,19 @@ and ana_cursor_info_zoperand =
           ),
         )
       };
+    | Case(InconsistentBranches(_, _, Ana), _, _) =>
```
cyrus-, line 895: this is an impossible case I think? Ana should only appear if the case is actually in analytic position?

src/hazelcore/Action_Exp.re
disconcision: resolved (possibly)

src/hazelweb/gui/CursorInspector.re
```diff
@@ -70,7 +70,8 @@ let exp_keyword_msg = (term, keyword, main_msg) =>
 
 let pat_ana_subsumed_msg =
     (expected_ty, got_ty, expecting_msg, consistency_msg) =>
-  if (HTyp.eq(expected_ty, got_ty) || HTyp.eq(got_ty, HTyp.Hole)) {
+  if (HTyp.eq(expected_ty, got_ty)
+      || HTyp.eq(got_ty, HTyp.Unknown(Internal))) {
```
cyrus-: I think this should work for Unknown(_) not just Unknown(Internal) -- at least TypHole should work, and the other case won't actually happen I think

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -4,23 +4,27 @@ type ground_cases =
...
   | Float
-  | Arrow(Hole, Hole)
-  | Sum(Hole, Hole)
-  | List(Hole) => Ground
+  | Arrow(Unknown(_), Unknown(_))
+  | Sum(Unknown(_), Unknown(_))
+  | List(Unknown(_)) => Ground
   | Prod(tys) =>
-    if (List.for_all(HTyp.eq(HTyp.Hole), tys)) {
+    if (List.for_all(HTyp.eq(HTyp.Unknown(Internal)), tys)) {
```
cyrus-: I don't think this file should be restricted to Internal unknowns, TypHole unknowns are semantically the same thing at this point. need to use wildcards / replace the eq calls with pattern matching...

src/hazelcore/Action_Exp.re
disconcision: why not? isn't this a reasonable check to have?

src/hazelcore/CursorInfo_Exp.re
```diff
@@ -304,7 +304,7 @@ and syn_cursor_info_line =
     let ty_def =
       switch (Statics_Exp.syn(ctx, def)) {
       | Some(ty) => ty
-      | None => HTyp.Hole
+      | None => HTyp.Unknown(Internal)
```
cyrus-: pretty sure we shouldn't be getting a None case here at all anymore

src/hazelcore/Action_Exp.re
disconcision: not really sure but i've made that change

src/hazelcore/CursorInfo_Exp.re
```diff
@@ -868,18 +892,19 @@ and ana_cursor_info_zoperand =
           ),
         )
       };
+    | Case(InconsistentBranches(_, _, Ana), _, _) =>
+      //TODO: Do we want something else here?
+      Some(CursorInfo_common.mk(Analyzed(ty), ctx, cursor_term))
+    | Case(InconsistentBranches(_, _, Syn), _, _) =>
+      syn_cursor_info_zoperand(~steps, ctx, zoperand)
```
cyrus-, line 950: sorry this is the impossible case for ana_cursor_info_zoperand

disconcision:
@cyrus- it won't let me respond to your comments above but I don't think that case is impossible? i get it e.g. here:
Screen Shot 2022-02-28 at 12 00 45 AM
which is as expected since the case is morally in synthetic position but actually analytic since it's always analytic

disconcision:
(other changes made)

cyrus-:
Getting an impossible syn error trying to press open parens or lambda here (possibly other construction actions too) 
Screenshot from 2022-02-28 00-38-34

src/hazelcore/Statics_Exp.re
```diff
@@ -285,7 +288,8 @@ and ana_operand =
   | Inj(InHole(WrongLength, _), _, _)
   | Case(StandardErrStatus(InHole(WrongLength, _)), _, _) =>
     ty |> HTyp.get_prod_elements |> List.length > 1 ? Some() : None
-  | Case(InconsistentBranches(_, _), _, _) => None
+
+  | Case(InconsistentBranches(_), _, _) => Some()
```
cyrus-: why is this not checking the branches against the provided type?

src/hazelcore/Statics_Exp.re
```diff
@@ -285,7 +288,8 @@ and ana_operand =
   | Inj(InHole(WrongLength, _), _, _)
   | Case(StandardErrStatus(InHole(WrongLength, _)), _, _) =>
     ty |> HTyp.get_prod_elements |> List.length > 1 ? Some() : None
```
cyrus-, line 310: not sure whats going on here -- case should never have a WrongLength hole on it.

src/hazelcore/Statics_Exp.re
```diff
@@ -1204,18 +1221,39 @@ and ana_fix_holes_operand =
...
-        scrut_ty,
-        ty,
-      );
-    (Case(StandardErrStatus(NotInHole), scrut, rules), u_gen);
+    switch (ty) {
+    | Unknown(SynPatternVar) =>
+      let (op, _, u_gen) =
+        syn_fix_holes_operand(ctx, u_gen, ~renumber_empty_holes, e);
+      (op, u_gen);
+    | Unknown(Internal | TypHole) =>
```
cyrus-: I don't think these cases need to be special cased here, they should go through the catchall case -- need to mark the situation where there are inconsistent branches even when analyzing against a non-hole type.

disconcision:
@cyrus: still won't let me respond directly to your comments... it says that i've 'started a review', which is maybe why, but i can't cancel the review because i can't review because it's my PR... confusing. anyway:

last comment: neither the current Unknown(Internal | TypHole) case or the current catchall case seem sufficient though. we can't just use the latter as that doesn't take into account the no-common-type case, and we can't use the former case as it doesn't ana-fix-rules in the case where the pattern actually provides a type. or are you suggesting combining the logic here somehow?

previous two comments: from looking at the other cases here i don't really understand in general what ana_operand is supposed to do in the case where the operand is in a non-empty hole. for the wronglength case, i haven't added any new logic there... are you saying Case should work differently than the other forms there, or that all those cases are wrong? in the inconsistent branches case; are you saying it should subsume like the TypeInconsistent case? if so, then it's going to fail to synthesize, and it's going to break. the whole point of this new modification to the hole status, i thought, was to prevent this breakage.

cyrus- (changes requested):
this may help?

cyrus-:
> last comment: neither the current Unknown(Internal | TypHole) case or the current catchall case seem sufficient though. we can't just use the latter as that doesn't take into account the no-common-type case, and we can't use the former case as it doesn't ana-fix-rules in the case where the pattern actually provides a type. or are you suggesting combining the logic here somehow?

I think we need to do both a syn_fix_holes_rules (to get the rule_types) and an ana_fix_holes_rules (to get the actual fixed rules) whether or not the expected type is Unknown(Internal | TypHole).

cyrus-:
> previous two comments: from looking at the other cases here i don't really understand in general what ana_operand is supposed to do in the case where the operand is in a non-empty hole.

ana should be actually verifying that the provided type can be used to successfully analyze, so we need to do ana_rules even when there is an inconsistent branches error. we also should do a syn_rules and make sure that the error is indeed correctly indicating an error. this mirrors the comment above.

cyrus-:
I'm not sure that CaseErrStatus actually even needs a StandardErrStatus case -- you can't ever get a TypeInconsistent with case, because that is when you have a synthesized type different from analyzed type but here we have a rule specifically for analysis. you can't get WrongLength either.

src/hazelcore/Statics_Exp.re
cyrus-: I believe this is a minor bug in `dev` unrelated to this PR -- this should always return None because these forms should never have WrongLength holes on them. we really should special case WrongLength errors to be an argument to the Comma operator in a BinOp...

disconcision:
I think I've looped around again to not really understanding what we're doing. I made the above changes (I think?), but I'm back to getting impossible syn, now on explicit annotated type case, e.g. if it do a case with int and float branches and try to annotate it with Int

cyrus-:
I think you need to fix the Action_Exp code too, not just Statics, otherwise they are out of sync...

cyrus-:
Oh wait I think the issue is that if we have ana actually checking that the branches synthesize inconsistent types then we can't do that on the ana_fixed versions of the branches because the branches that had type errors will synthesize unknown type

cyrus-:
I think we only want InconsistentBranches(..., Ana) to be added if analyzed against the SynPatternVar hole. So you can check that that is the expected type in ana before doing the check that there are inconsistent branches. Any other expected type and it should be NotInHole.

disconcision:
@cyrus not sure what you mean, that will just break the annotated-with-type-hole case... that's the case that needs the InconsistentBranches(..., Ana) to avoid impossible-syn

disconcision:
my understanding:
1. if we're analyzing against SynPatternVar, we just switch to synthetic mode; everything just works as it did in the classic bidi pre-pattern annotation world.
2. if we're analyzing against a TypHole, that's the case where we have to incorporate the new pseudo-error status InconsistentBranches Ana (in the event the branches are inconsistent), in order to prevent the Case from failing synthesis.
3. Not sure about Internal but to prevent synthesis from failing we basically have to do one of the above.
4.  everything else just goes to the NotInHole case.

This was reflected in the code prior to my most recent revision... if we're departing from this approach then I no longer understand the big picture and can't effectively make changes.

cyrus-:
Made some fixes that I think addresses the issue, but there seem to be hole display issues when you have InconsistentBranches(..., Ana). The console suggests the holes are being placed correctly but aren't being shown. Don't have time to debug but take a look at the diff...

disconcision:
@cyrus cursorinfo also seems wonky now:
Screen Shot 2022-03-01 at 3 34 22 PM
Screen Shot 2022-03-01 at 3 34 15 PM

cyrus-:
Yeah I think you want to do a pass through the CaseZE/ZR zipper cases, and CursorInfo, and elaboration, and make sure they are all following the same logic from Statics, which I updated. Might also want to look at LetLine stuff again. (I tweaked the logic for that in the last commit to not ana_perform on the pattern, but that might not be necessary -- it does propagate some information in that direction for use by the type annotation completion).

Also not clear what's going on with the display of holes since it does seem to be generating the right ErrStatuses internally based on a print_endline (that I left in there). Something to do with the logic for not display InconsistentBranches(...Ana)?

disconcision:
@cyrus: made the changed we talked, in particular moved the SynPatternVar check to the top of ana. A couple issues:

1. If I put in the check for SynPatternVar in HTyp.consistent, it immediately throws on startup. Not sure how much time to put into chasing this.
2. If I add the rule-checking discussed for Case InconsistentBranches in syn_operand (Statics_Exp line 150), I start getting impossible syn thrown, specifically in the case where I try to create a case with inconsistent branches with is annotated with a concrete type, e.g. let x:Int = case ? of | _ => 1 | _ => 1.0. Been trying to debug but haven't figured it out, any thoughts?

Other than those issues, this is good for a go

cyrus-:
1. Weird. Not sure how a SynPatternVar is even being produced on startup.
Are all the cards being type checked right away maybe?

2. Not sure why that would be throwing since the branches do all synthesize
and consistency fails. This is worth debugging I think...

disconcision:
@cyrus ah i figured out (2), kind of... so this is the problematic expression:

```
((ExpLine
  (OpSeq (Placeholder 0)
   (S
    (Case (InconsistentBranches 22 Ana)
     ((ExpLine (OpSeq (Placeholder 0) (S (EmptyHole 11) E))))
     ((Rule (OpSeq (Placeholder 0) (S (EmptyHole 12) E))
       ((ExpLine (OpSeq (Placeholder 0) (S (IntLit NotInHole 5) E)))))
      (Rule (OpSeq (Placeholder 0) (S (EmptyHole 14) E))
       ((ExpLine
         (OpSeq (Placeholder 0)
          (S (FloatLit (InHole TypeInconsistent 21) 5.) E)))))))
   E))))
```

this is after hole fixing i guess. the branches are consistent since the float one is wrapped in a hole, so the InconsistentBranches case returns None, so syn_operand fails, triggering impossible syn.

so not sure what part of this is wrong exactly... the hole-fixing of the above is correct right? do i need to erasing holes first?

disconcision:
For (1): syn_fix_holes is being called on the examples to generate an edit state from a zexp; this is what's causing the immediate crash. If I bypass that though there are more issues... for example if I try to write "let x = 1" then the consistency check in mk_ana_text fails... would we maybe need to (at least) do similar SynPatternVar switches at the top level of action, elab, etc?

cyrus-:
(1) yeah we'd need to put that check at the top of everything that follows
the statics

(2) ah yeah I see the problem there -- maybe we need to only do the check
when the InconsistentBranches has the Syn flag, or else not put the
InconsistentBranches flag on analytic cases at all, since hole fixing will
make them all have a type consistent with the expected type anyway?

disconcision:
(2) okay changed it to only do the check in the Syn case

disconcision:
(1) 80 printlns later: it's not enough to add checks in the root ana function. for example if the pattern is a pair (x,x) then the type won't be synpatternvar at the top level, and the constitutent synpatternvars will be passed down through the opseq and skel levels eventually getting to operand without passing through the top-level ana again. 

not transparent to me exactly at which level checks will be required (operand at least, but there are HTyp.consistent calls in skel as well for e.g. lists so that will need to be handled to), and whether this will be consistent modulo implementation decisions across the various statics-driven functions.  this is starting to feel like a boondoggle

disconcision:
Added the SynPatternVar check to all ana exp judgements in statics, elaborator, action, and cursor_info. still getting consistent check fails from the Examples, for example from here
Screen Shot 2022-04-24 at 10 13 10 PM
:

disconcision:
Also:
Screen Shot 2022-04-24 at 11 02 22 PM

disconcision:
@cyrus can you take a look at the latest commit when you have the chance. I've gotten confused about what I'm doing I think. In particular what the scope of this moded change needs to be (I think: need to thread moded through statics_pat, elaborator_pat, action_pat, cursorinfo_pat?) and what the switched behavior actually is. Like are we suppressing both adding SynPatternVar to the context and returning it? What I've done now seems to have removed the issue for the Map example, but the quicksort example is still throwing, now for the "ana_elab_skel" case.

disconcision:
@cyrus Made the changes we talked about last meeting. Still throws a consistency check exception in ana_elab_skel; see "ty is (Prod (ModeSwitch ModeSwitch))". Not sure how current approach can address such cases

disconcision:
Here's a minimized example (errors when I put a reference to z in the hole)
Screen Shot 2022-05-09 at 3 49 35 AM

disconcision:
Also I've further isolated that weird BoolLit thing we saw during our last meeting.... still don't understand it though. What I do know is that if you type "let x = 1", (the 1 can be any expression, but not just a hole) and then delete the x, somehow ana_fix_holes_operand gets called on a BoolLit.

Screen Shot 2022-05-09 at 4 11 31 AM

disconcision:
Yeah not sure how to proceed here. when I try to construct the reference to z, this results in ana_perform being called with type (ModeSwitch,ModeSwitch). Since I'm not trying to construct a pair literal, the type isn't going to get destructured; it just recurses down to ana_operand, and eventually to mk_ana_text,  where, in the Var case, HTyp.consistent is called on the type. Not obvious to me at what stage in this process ModeSwitch should have been intercepted.

disconcision:
@cyrus: added flag to pat hole fixing, disabled consistent moded check. good for another look

src/hazelcore/HTyp.rei
```diff
@@ -1,7 +1,17 @@
 /* types with holes */
+
+[@deriving sexp]
+type unknown_type_provenance =
+  | TypHole /* from an actual type hole; will add a MetaVar.t once we have unique IDs for type holes */
+  /* from a pattern var being asked to synthesize a type either directly or via matched arrow/prod/etc.
```
cyrus-: misplaced comment

src/hazelcore/HTyp.re
```diff
@@ -180,3 +175,8 @@ let join_all = (j: join, types: list(t)): option(t) => {
     }
   };
 };
+
+let is_unknown: t => bool =
+  fun
+  | Unknown(_) => true
+  | _ => false;
```
cyrus-, line 182: if we're treating ModeSwitch as an unknown for consistency, then we should maybe treat it as an unknown everywhere (perhaps putting it back into Unknown as a provenance instead of a top-level thing)

src/hazelcore/dynamics/DHExp.re
```diff
@@ -207,7 +207,7 @@ let rec mk_tuple: list(t) => t =
   | [d, ...ds] => Pair(d, mk_tuple(ds));
 
 let cast = (d: t, t1: HTyp.t, t2: HTyp.t): t =>
-  if (HTyp.eq(t1, t2)) {
+  if (HTyp.eq(t1, t2) || t2 == ModeSwitch) {
```
cyrus-: if we're treating ModeSwitch as an unknown for consistency, then we shouldn't need this anymore I think

cyrus- (changes requested):
partial review -- see comments

disconcision:
@cyrus all changes made (modeswitch is back to unknown)

src/hazelweb/cardstacks/Card.re
```diff
@@ -6,6 +6,6 @@ type t = {
 
 let mk = (info: CardInfo.t): t => {
   info,
-  edit_state:
-    Statics_Exp.fix_and_renumber_holes_z(Contexts.initial, info.init_zexp),
+  edit_state: (info.init_zexp, HTyp.Int, 0),
```
cyrus-: what's happening here?

src/hazelweb/layout/HTypDoc.re
```diff
@@ -40,7 +40,8 @@ let rec mk = (~parenthesize=false, ~enforce_inline: bool, ty: HTyp.t): t => {
   );
   let (doc, parenthesize) =
     switch (ty) {
-    | Hole => (
+    | Unknown(ModeSwitch)
+    | Unknown(_) => (
```
cyrus-: don't use _

src/hazelweb/gui/StrategyGuide.re
```diff
@@ -66,7 +66,8 @@ let lit_msg = (ty: HTyp.t) => {
       AssistantView_common.kc_shortcut_node(HazelKeyCombos.LeftBracket),
     ]);
   switch (ty) {
-  | Hole => [
+  | Unknown(ModeSwitch)
+  | Unknown(_) => [
```
cyrus-: don't use _

src/hazelcore/Statics_Pat.rei
```diff
@@ -1,3 +1,8 @@
+[@deriving sexp]
+type pattern_var_mode =
```
cyrus-: could use some brief documentation here

src/hazelcore/Statics_Pat.rei
```diff
@@ -16,6 +21,7 @@ let ana_nth_type_mode:
  * produces a new context with bindings introduced by `p` (if possible)
  */
 let syn: (Contexts.t, UHPat.t) => option((HTyp.t, Contexts.t));
+let syn_moded: (Contexts.t, UHPat.t) => option(HTyp.t);
```
cyrus-, line 33: naming is a bit confusing here -- was expecting this to take a pattern_var_mode input, but "moded" here refers to the ModedVariable constructor of that type, and it isn't obvious that to get UnknownVariable you just use the default. Thinking it might be better for external clients to have to be explicit about which choice they are making when they call top-level syn...

src/hazelcore/Statics_Exp.re
```diff
@@ -121,95 +123,91 @@ and syn_skel =
...
+  | Inj(InHole(WrongLength, _), _, _) => None
+  | Case(InconsistentBranches(_, Syn), scrut, rules) =>
+    let* ty_scrut = syn(ctx, scrut);
+    let* clause_types = syn_rule_types(ctx, ty_scrut, rules);
+    switch (HTyp.join_all(GLB, clause_types)) {
+    | None => Some(HTyp.Unknown(Internal))
+    | Some(_) => None
+    };
+  | Case(InconsistentBranches(_, Ana), scrut, _) =>
+    let* _ty_scrut = syn(ctx, scrut);
```
cyrus-: I think we do need to check that all of the rules synthesize a type, just don't need to do the join here.

cyrus- (changes requested):

src/hazelcore/Statics_Exp.re
```diff
@@ -304,9 +325,9 @@ and ana_operand =
   | Inj(NotInHole, side, body) =>
     let* (ty1, ty2) = HTyp.matched_sum(ty);
     ana(ctx, body, InjSide.pick(side, ty1, ty2));
-  | Case(StandardErrStatus(NotInHole), scrut, rules) =>
-    let* ty1 = syn(ctx, scrut);
-    ana_rules(ctx, rules, ty1, ty);
+  | Case(_, scrut, rules) =>
```
cyrus-: I think we need to check the error status correctness here too

src/hazelcore/Statics_Exp.re
```diff
@@ -1203,19 +1316,27 @@ and ana_fix_holes_operand =
       };
     }
   | Case(_, scrut, rules) =>
-    let (scrut, scrut_ty, u_gen) =
+    let (scrut, ty_scrut, u_gen) =
       syn_fix_holes(ctx, u_gen, ~renumber_empty_holes, scrut);
+    let (_, _, _, common_type) =
+      syn_fix_holes_rules(ctx, u_gen, ~renumber_empty_holes, rules, ty_scrut);
```
cyrus-: I was expecting an ana_fix_holes followed by a call to syn_rule_types to determine whether a common type exists or not -- not sure we need to do a full syn_fix_holes here

disconcision:
@cyrus: made changes except for above comments

src/hazelcore/Statics_Exp.re
```diff
@@ -304,9 +326,17 @@ and ana_operand =
   | Inj(NotInHole, side, body) =>
     let* (ty1, ty2) = HTyp.matched_sum(ty);
     ana(ctx, body, InjSide.pick(side, ty1, ty2));
-  | Case(StandardErrStatus(NotInHole), scrut, rules) =>
-    let* ty1 = syn(ctx, scrut);
-    ana_rules(ctx, rules, ty1, ty);
+  | Case(_ /*CaseNotInHole*/, scrut, rules) =>
+    let* ty_scrut = syn(ctx, scrut);
+    ana_rules(ctx, rules, ty_scrut, ty);
+  // TODO(andrew): this triggers MissingCursorInfo when adding a concrete type
```
cyrus-: I think we do want to check even in the ana case whether InconsistentBranches errors are accurate so:

* If it is a Syn error then treat it as a synthetic case and do a subsumption (don't think this will ever actually happen in the callers to ana, but correctness wise)
* If it is an Ana error, then make sure that the rules synthesize types but don't need to do a join, as in the syn case

src/hazelcore/Statics_Exp.re
```diff
@@ -1203,19 +1325,49 @@ and ana_fix_holes_operand =
...
-        u_gen,
-        ~renumber_empty_holes,
-        rules,
-        scrut_ty,
-        ty,
-      );
-    (Case(StandardErrStatus(NotInHole), scrut, rules), u_gen);
-  }
+    let (scrut, u_gen) =
+      ana_fix_holes(ctx, u_gen, ~renumber_empty_holes, scrut, ty);
```
cyrus-: not sure what this is doing? scrut should only be synthesizing...

src/hazelcore/Statics_Exp.re
```diff
@@ -1203,19 +1325,49 @@ and ana_fix_holes_operand =
...
+      | Some(clause_types) =>
+        switch (HTyp.join_all(GLB, clause_types)) {
+        | None =>
+          let (u, u_gen) = MetaVarGen.next(u_gen);
+          (Case(InconsistentBranches(u, Ana), scrut, rules), u_gen);
+        | Some(_) => (Case(CaseNotInHole, scrut, rules), u_gen)
+        }
+      }
+    };
+  // TODO(andrew): this triggers MissingCursorInfo when adding a concrete type
```
cyrus-: should be fixed with above fixes? remove if so

cyrus- (changes requested):
reviewed new Statics_Exp, see comments

cyrus-:
Screenshot from 2022-06-15 16-47-46

Something is wrong with cast insertion on let-bound recursive functions, likely an issue in elaboration? When I try to enter an argument in hole 1 in the screenshot I get the EvaluatorError shown, and the cast shown there shouldn't have been inserted...

disconcision:
@cyrus so there's this line in elab that unconditionally adds a cast to the definition (see screenshot). not sure when/why this got added. commenting it out resolves the proximal issue
Screen Shot 2022-06-15 at 5 08 54 PM
:

cyrus-:
OK, I'll review what's going on there soon.

disconcision:
yeah the line exists in dev which does not show the relevant behavior so im confused
https://github.com/hazelgrove/hazel/blob/1deb1ab2cc8e3fa03a8e2557680971000e511540/src/hazelcore/dynamics/Elaborator_Exp.re#L96

src/hazelcore/Statics_Exp.re
```diff
@@ -1203,19 +1329,22 @@ and ana_fix_holes_operand =
...
-        scrut_ty,
-        ty,
-      );
-    (Case(StandardErrStatus(NotInHole), scrut, rules), u_gen);
-  }
+      ana_fix_holes_rules(ctx, u_gen, rules, ty_scrut, ty);
+    switch (syn_rule_types(ctx, ty_scrut, rules)) {
+    | None => failwith("ana_fix_holes_operand' Case 2")
+    | Some(clause_types) =>
+      switch (HTyp.join_all(GLB, clause_types)) {
```
cyrus-, line 1345: are we using LUB or GLB? seems inconsistent with joined_pattern_type function at the top of the file?

cyrus-:
That cast doesn't need to be there in elaboration, but that doesn't entirely fix the issue. It seems there is something wonky with how functions are being elaborated in analytic position -- notice in this unevaluated elaboration that the function argument has ? type when it should have Int type...
Screenshot from 2022-06-17 01-02-30

disconcision:
prev comment reflects current behavior in dev... I've rewritten the analytic fun case in elaborator to bring it in line

cyrus-:
When the pattern is a hole or wildcard, you don't get the same behavior as if it were a variable -- I think we need to use the Moded logic for those forms as well?
Screenshot from 2022-06-18 01-18-10

disconcision:
@cyrus- made the discussed changes about bringing wildcards and holes in patterns in line with the mode switch previously only made for unannotated variables. insodoing it raised the question of whether we should also do this for things like invalidtext, or patterns in holes. any thoughts?

cyrus-:
Yeah I think any time we're synthesizing a fresh unknown type, so not type annotation, it should be a mode switch...

disconcision:
@cyrus so that would mean that all Unknown(Internals) in statics pat elab pat and action pat should be Unknown(Moded) except for those directly derived from TypeAnns?

cyrus-:
Yes, that's right.

cyrus-:
Marking as draft again due to that last change + conflicts needing resolved.

disconcision:
Changes changed, conflicts deconflicted

src/hazelcore/Statics_Pat.re
```diff
@@ -65,45 +104,56 @@ and syn_operand =
...
+    Some((Unknown(ModeSwitch), ctx))
   | Var(NotInHole, NotInVarHole, x) =>
-    Var.check_valid(
-      x,
-      Some((HTyp.Hole, Contexts.extend_gamma(ctx, (x, Hole)))),
-    )
+    switch (pattern_syn) {
+    | Moded => Some((Unknown(ModeSwitch), ctx))
+    | Unknown =>
+      let ty = HTyp.Unknown(ModeSwitch);
```
cyrus-: everything just returns ModeSwitch in here now. I believe the correct behavior is that throughout the file, it should depend on the `pattern_syn` flag. When it is Moded then it should do Unknown(ModeSwitch) and if it is `Unknown` then it should do `Unknown(Internal)`?

cyrus- (changes requested):

cyrus-:
Pressing Backspace on the type annotation here doesn't cause the case to have the inconsistent branches error displayed until you make some other edit that causes hole fixing to happen -- looks like there is a bug in a zipper case somewhere.

https://hazel.org/build/make-impossible-syn-more-impossibler/#let%C2%A0x%C2%A0%3A%C2%A0%3F%C2%A0%3D%0A%C2%A0%C2%A0case%C2%A0%3F%0A%20%20%20%20%7C%C2%A0%3F%C2%A0%3D%3E%C2%A01%0A%20%20%20%20%7C%C2%A0%3F%C2%A0%3D%3E%C2%A0true%0A%20%20%20%20%7C%C2%A0%3F%C2%A0%3D%3E%C2%A03%0A%20%20%20%20end%0Ain%0A%3F

cyrus-:
@disconcision can we get this finished up for hazel2 so we have it as reference? think there is just the zipper case bug above left

cyrus-:
Archiving this, should be addressed in Hazel 3 statics now. Will be writing this up in paper form!

PR #568, opened 2021-11-28
Tests: IDGen Fix

ygaitonde:
depends on #564 

- Refactor examples so that we load an edit_state instead of a UHExp.
- Add option to serialize edit state to console


disconcision:
lgtm, ill merge this into feature tomorrow. in retrospect i think a better approach for the examples might be, instead of retaining ids between sessions, to instead derive a reasonable id_gen state from the syntax, i.e. recurse through the uhexp, gathering ids, and then pick an init idgen such that the starting ids are equal to the largest extant ids plus one. probably not worth the work atm though.

disconcision (approved):

src/hazelweb/cardstacks/builtins/TutorialCards.re
```diff
@@ -42,11 +42,11 @@ let intro_init_zexp =
...
  */
-let intro_card: CardInfo.t = {
-  name: "intro",
-  caption: div([], []),
-  init_zexp: intro_init_zexp,
-};
+// let intro_card: CardInfo.t = {
+//   name: "intro",
+//   caption: div([], []),
+//   init_zexp: intro_init_zexp,
```
cyrus-: why are these commented out?

cyrus-:
> lgtm, ill merge this into feature tomorrow. in retrospect i think a better approach for the examples might be, instead of retaining ids between sessions, to instead derive a reasonable id_gen state from the syntax, i.e. recurse through the uhexp, gathering ids, and then pick an init idgen such that the starting ids are equal to the largest extant ids plus one. probably not worth the work atm though.

you can use the holes list that we generate for hole traversal to figure out the largest hole number. you can also just renumber holes with a fresh id generator in most cases.

disconcision:
Wed need to make similar infrastructure for test IDs. Not bad to do but feel spending time on this is diminishing returns though as all of this can/should be replaced by a generic node ID system.

Re: tutorial cards: not sure that module really does anything at this point?

cyrus-:
If the tutorial cards are no longer needing to be maintained, let's remove them instead of commenting out invalid code. Other than that, this looks fine.

ygaitonde:
> If the tutorial cards are no longer needing to be maintained, let's remove them instead of commenting out invalid code. Other than that, this looks fine.

Sounds good - I deleted the invalid code.

@disconcision are we good to merge?

src/hazelweb/cardstacks/builtins/TutorialCards.re
```diff
@@ -106,9 +101,4 @@ let intro_card: CardInfo.t = {
    caption: empty_hole_insertion_caption,
    init_zexp: empty_hole_insertion_init_zblock,
  };
- */
-
-let cardstack: CardstackInfo.t = {
-  title: "Hazel Tutorial",
-  cards: [intro_card /*backspace_card, empty_hole_insertion_card */],
-};
+ */
```
cyrus-: this whole file isn't usable anymore without the cardstack definition. I'm fine with just removing the file entirely -- not much to salvage here.

ygaitonde: sounds good! I removed the old code

PR #569, opened 2021-11-30
Cleanup KeyComboAction.get_model_action_from_kc

mirryi:
Noticed this when working on string literal character insertions. I could be wrong about this...

Cleans up `KeyComboAction.get_model_action_from_kc` a little bit by removing the unnecessary `cursor_on_comment`. The `single_key_in_comment` already intercepts character insertions (including space) into comment lines.

### DIFF
--- a/src/hazelweb/gui/KeyComboAction.re
+++ b/src/hazelweb/gui/KeyComboAction.re
@@ -7,11 +7,11 @@ let get_model_action_from_kc =
   let construct = (shape: Action.shape): option(ModelAction.t) =>
     Some(EditAction(Construct(shape)));
 
-  let (cursor_on_type, cursor_on_comment) =
+  let cursor_on_type =
     switch (cursor_info) {
-    | {typed: OnType, _} => (true, false)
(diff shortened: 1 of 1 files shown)


src/hazelweb/gui/KeyComboAction.re
```diff
@@ -7,11 +7,11 @@ let get_model_action_from_kc =
...
     Some(EditAction(Construct(shape)));
 
-  let (cursor_on_type, cursor_on_comment) =
+  let cursor_on_type =
     switch (cursor_info) {
-    | {typed: OnType, _} => (true, false)
-    | {cursor_term: Line(_, CommentLine(_)), _} => (false, true)
-    | _ => (false, false)
+    | {typed: OnType, _} => true
+    | {cursor_term: Line(_, CommentLine(_)), _} => false
```
cyrus-: this can be folded into the `_` case too now

cyrus- (changes requested):
the switch on line 78 does the comment line detection -- should drop the _cursor_on_type stuff there too.

cyrus- (approved):

PR #574, opened 2022-01-11
Fix Tuple Completion's Cursor Location

### LINKED ISSUE #561: Wrong cursor position in Let-Pair
Default cursor position for Let-Pair is to place it at the middle comma [,]. The expected behavior is to place the cursor at the left cell of the pair.
Here's a minimal example. After I typed the left parenthesis, the cursor shifts to the comma:
```
let (left, right) = (_[,] _) in
```
The expected behavior should be:
```
let (left, right) = ([_], _) in
```

AlienKevin:
Should fix #561


cyrus- (approved):
Looks good, thanks!

PR #577, opened 2022-02-03
UndoHistory: Switch previous_action from Action.t to ModelAction.t

nj-wilson:
Have UndoHistory track ModelActions instead of Actions for the previous_action.
This will be needed to have an Import entry in the Undo History.


cyrus-:
Is this ready to review?

nj-wilson:
Yes.

cyrus- (approved):

PR #578, opened 2022-02-03
Mutual rec

Zesheng16:
Add mutual recursion support.


src/hazelcore/Statics_Exp.re
```diff
@@ -45,29 +45,82 @@ and syn_block = (ctx: Contexts.t, block: UHExp.block): option(HTyp.t) => {
...
+    (ctx:Contexts.t, lines:list(UHExp.line)):option(Contexts.t) => {
+
+      // first pass: group lines
+
+      let groups = List.fold_left(
+        (accu: (list(UHExp.line),list(list(UHExp.line))), line: UHExp.line) => {
+          switch(line){
+            | EmptyLine
+            | CommentLine(_) 
+            | LetLine(And, _, _) => ([...accu.first, line], accu.second)
```
cyrus-: this isn't valid syntax, the ... can only appear at the end. you probably want to use list append.

src/hazelcore/Statics_Exp.re
```diff
@@ -45,29 +45,82 @@ and syn_block = (ctx: Contexts.t, block: UHExp.block): option(HTyp.t) => {
...
+
+// removed .hd .tl
+
+and syn_lines = 
+    (ctx:Contexts.t, lines:list(UHExp.line)):option(Contexts.t) => {
+
+      // first pass: group lines
+
+      let groups = List.fold_left(
+        (accu: (list(UHExp.line),list(list(UHExp.line))), line: UHExp.line) => {
```
cyrus-: instead of calling it `accu` you should name the two parts immediately using a pair pattern, `(cur_block, blocks)` or something like that

src/hazelcore/Statics_Exp.re
```diff
@@ -45,29 +45,82 @@ and syn_block = (ctx: Contexts.t, block: UHExp.block): option(HTyp.t) => {
...
+              | CommentLine(_) => ctx_line
+              | ExpLine(opseq) =>
+                let *ctx = ctx_line;
+                let+ _ = syn_opseq(ctx, opseq);
+                Some(ctx);
+              | LetLine(_, p, def) =>
+                let *ctx = ctx_line;
+                Some(extend_let_def_ctx(ctx, p, def));
+            },
+            ctx.group,
```
cyrus-: what is ctx.group? this will be easier to review once it compiles...

cyrus- (changes requested):

src/hazelcore/Statics_Exp.re
chenyan-zhang: modified

src/hazelcore/Statics_Exp.re
chenyan-zhang: modified

cyrus-:
Archiving this Hazel 2 PR. We now have mutual recursion by defining tuples of functions with type annotations in Hazel 3. Not the nicest syntax for this (names are far from implementations), so an "and line" approach would be nice down the line (or something else, like Agda's "mutual" blocks). Thank you for the work on this @chenyan-zhang and @Zesheng16 !

PR #580, opened 2022-02-18
Simple Zipper Tutorial

disconcision:


docs/ZipperTutorial.ml
```diff
@@ -0,0 +1,78 @@
...
+   recursive type), plus an aditional variant representing the cursor
+   itself. So in our example, we have 0 variants for the 0-ary Atom,
+   2 variants for the binary Bin (representing the cursor being within
+   the left or right subtrees respectively), and the cursor case,
+   giving 3 variants in total *)
+type zast =
+ | Cursor of ast
+ | BinL of zast * ast
+ | BinR of ast * zast
+
```
cyrus-, line 41: Maybe include an example or two here showing the cursor on a leaf but also on an interior node?

docs/ZipperTutorial.ml
```diff
@@ -0,0 +1,78 @@
...
+ let select_root : ast -> zast = failwith "TODO"
+;; assert (select_root (Bin (Atom, Atom)) = Cursor (Bin (Atom, Atom)))
+
+(* 2. Convert from zippered ast to non-zippered ast by forgetting about the cursor.
+      Note that unlike the other direction, each zippered value corresponds to
+      exactly one non-zippered value *)
+let erase : zast -> ast = failwith "TODO"
+;; assert (erase (BinL (Cursor Atom, Atom)) = Bin (Atom, Atom))
+
+(* 3. Simplified view function: Print a zast, using a caret (>) to indicate the cursor term *)
```
cyrus-, line 67: these tests could be the examples, and maybe have this be the first exercise in that case

docs/ZipperTutorial.ml
```diff
@@ -0,0 +1,78 @@
...
+let view : zast -> string = failwith "TODO"
+;;[ assert (view (Cursor (Bin (Atom, Atom))) = ">(A A)"),
+    assert (view (BinL (Cursor Atom, Atom)) = "(>A A)"),
+    assert (view (BinL (BinR (Atom, Cursor Atom), Atom)) = "((A >A) A)")]
+
+(* 4. Simplified update function: Perform the given move action. If the action is
+      inapplicable to the current cursor position, just return the zast unaltered *)
+let update : action -> zast -> zast = failwith "TODO"
```
cyrus-, line 75: this is called perform in the Hazel codebase so maybe thats better than update here too

docs/ZipperTutorial.ml
```diff
@@ -0,0 +1,78 @@
...
+;; [assert (update SelectChildBinRight (BinL (Cursor (Bin (Atom, Atom)), Atom)) = (BinL (BinR (Atom, Cursor Atom), Atom))),
+    assert (update SelectParent (BinL (BinR (Atom, Cursor Atom), Atom)) = (BinL (Cursor (Bin (Atom, Atom)), Atom)))]
+
+(* Congrats! You've pretty much made a structured editor. If you want to close the loop,
```
cyrus-, line 79: Maybe some pointers into the zippers used in hazelcore here would be helpful. Also a pointer to the HZ repo if they want to expose something intermediate in complexity between the tutorial and Hazel itself?

cyrus- (changes requested):
Looks useful thanks! I made some small suggestions.

cyrus-:
@disconcision want to finish this up?

docs/ZipperTutorial.ml
disconcision: Update seems more standard here; why perform?

docs/ZipperTutorial.ml
disconcision: i feel the flow makes more sense starting with selection rather than erasure; added examples earlier tho

disconcision:
made some changes

cyrus- (approved):
Looks great, thanks Andrew!

PR #584, opened 2022-02-28
fix ana_perform behavior for type inconsistencies to not call syn_perform

cyrus-:
ana_perform should proceed normally even if there is currently a type inconsistency -- it was written in a way that would make sense for explicit non-empty holes a la Hazelnut but Hazel does hole fixing so not necessary here


cyrus-:
Going to merge this. If this causes a bug, then the issue is likely elsewhere anyway.

Issue #536, opened 2021-08-16
slowdown with 12+ letlines


disconcision:
I was originally thinking this was a prettyprinting thing but it might not be.

the below program takes 3 second to register a keypress. it feels exponential in # of lines
![2021-08-16 19_00_53-Hazel](https://user-images.githubusercontent.com/22436459/129639785-e80df0a0-293e-490f-9786-92502005f08a.png)

it seems however that the only lines that 'count' are letlines that provide a pattern but no binding expression.

to illustrate, the following programs exhibit no slowdown:
![2021-08-16 19_01_50-Hazel](https://user-images.githubusercontent.com/22436459/129639924-b50b7ccb-83b9-4d4f-aba7-ceb4e685363d.png)
![2021-08-16 19_02_21-Hazel](https://user-images.githubusercontent.com/22436459/129639927-6bc3b1d9-c3b9-4cee-adce-c0be8f910d0a.png)

however the following program is slowing down as i remove the binding expressions:
![2021-08-16 19_03_09-Hazel](https://user-images.githubusercontent.com/22436459/129639981-13e32d0a-84d0-41ec-9e4b-72a32c419a9a.png)

These are the metrics:

== Hazel.create times ==
hazel.js:1641 Program.get_doc:                   0ms
hazel.js:1641 LayoutOfDoc.layout_of_doc:         0ms
hazel.js:1641 Program.get_doc:                   0ms
hazel.js:1641 LayoutOfDoc.layout_of_doc:         0ms
hazel.js:1641 UHCode.view:                       5ms
hazel.js:1641 Cell.view:                         5ms
hazel.js:1641 Page.view:                      16835ms
hazel.js:1641 Hazel.create:                   16835ms
hazel.js:1641 
== Update.apply_action times ==
hazel.js:1641 Program.get_doc:                   1ms
hazel.js:1641 LayoutOfDoc.layout_of_doc:         0ms
hazel.js:1641 Update.apply_action:

cyrus-:
Ah, this is likely due to closures piling up exponentially -- each empty hole has all of the preceding variables in its closure, which themselves have all of their preceding variables, etc.

cyrus-:
Should work faster if you turn off evaluation but that sidebar option seems buggy -- I get an "Invalid_argument "option is None"" exception when I try to insert let bindings when Evaluate is unchecked. :\

cyrus-:
[...]

PR #586, opened 2022-03-09
Evaluation using the environment model rather than substitution

### LINKED ISSUE #642: Environment-Based Evaluator

jlam55555:
Evaluation with environments rather than environments. Evaluation with environments is useful because:
- Usually (?) better performance (i.e., don't substitute in unevaluated branches)
- Can memoize hole instances with the same environment/instantiation (to handle https://github.com/hazelgrove/hazel/issues/536)

Rough summary of changes:
- New data structure: `EvalEnv.t` (perhaps should be renamed? E.g., to `MemoEnvironment`?) which represents an environment tagged with an ID (`EvalEnvId.t`). Note that unique ID's are generated for an environment each time a new one is created by using `EvalEnvIdGen.t` (very similar to `MetaVarGen.t`). Note also that `EvalEnv.t` maps `Var.t` to `EvaluatorResult.t` rather than to `DHExp.t`, but the choice is somewhat arbitrary
- Change interface of evaluation functions (`Evaluator.evaluate`, `Builtin.evaluate`). `evaluate: DHExp.t => EvaluatorResult.t` becomes `evaluate: (EvalEnvIdGen.t, EvalEnv.t, DHExp.t): (EvalEnvIdGen.t, EvaluatorResult.t)`
- "Generalized closures" to represent any expression that can have a delayed evaluation (e.g., used to encapsulate let/case expr with failed match, lambda closure, hole closure), also neatens up some of the hole handling
- Implement a "postprocessing step" (`EvalPostprocess.postprocess`) that performs two actions:
    - Perform substitution inside closures (lambda bodies, bodies of let/case exprs with failed matches) so that the result matches that of substitution (i.e., context inspector should show same result as before)
    - Number hole instances. Notably, in order to solve https://github.com/hazelgrove/hazel/issues/536, holes that share the same environment are given the same number, i.e., each instantiation of a hole is a single "hole closure", potentially with multiple parents/paths, rather than being split into multiple "hole instances" each time it is encountered in the result. (This causes `HoleInstance.t`, `HoleInstancePath.t`, and `HoleClosureInfo.t` to become `HoleClosure.t`, `HoleClosureParents.t`, and `HoleClosureInfo.t`. Also requires slight changes to the context inspector to display all parents to a hole closure, rather than the (singular) path to the current hole instance.). The old hole instance numbering step, `Program.renumber`, is somewhat simplified and grouped into this postprocessing step.
- For performance reasons, change environments to use a BST implementation rather than a linked list. (`VarBstMap.t` -- currently replaces uses of `VarMap.t` only in `EvalEnv.t` but doesn't affect other usages of `VarMap.t`)
- Implement `DHExp.fast_equals` for fast equality checking memoized by environments, which doesn't recurse through environments, and `Result.fast_equals`, which performs `DHExp.fast_equals` on the result expression and on each expression in each environment
- No more need of `Evaluator.subst*` functions during evaluation. (Substitution is currently being used in one place only: during elaboration when introducing the `FixF` form for let expressions with an type-ascribed lambda binding)
- Evaluation of recursive functions still uses `FixF`, although it is possible to use recursive environments to achieve this (achieved in the [rec-closures](https://github.com/hazelgrove/hazel/tree/rec-closures) branch as a proof-of-concept).

This branch is a work in progress, probably needs more testing and maybe some changing of module interfaces.

Planning to branch off this branch to work on fill-and-resume. The implementation shouldn't be too different from evaluation with substitution, but we can benefit from the optimization in https://github.com/hazelgrove/hazel/issues/536.

I'm also working on a report to try to formalize and clarify some of the above changes, as well as future work with fill-and-resume, but it's not very complete yet.

Worked with Yuchen Jiang (@LighghtEeloo) on this.


cyrus-:
@jlam55555 do you have a timeline for getting this PR ready to review/merge?

jlam55555:
@cyrus- 

> @jlam55555 do you have a timeline for getting this PR ready to review/merge?

I took an impromptu month-long break after school ended because I couldn't bring myself to do anything academic (got caught up in graduation, moving, housefinding, vacation, and COVID). Now that I've recovered from COVID and things are sorted out I think I can kick myself into high gear again for a sprint over the next two weeks. I'll get started right away!

Proposed timeline: I will try to get this branch up for review by next Friday. Tasks:
- Redo the hole numbering step/memoization (put this in a separate postprocessing step so that postprocessing can be more simply memoized)
- Clean up the data structures related to environment numbers (improved version in the `fill-and-resume-backend` branch, which was forked from this branch -- I'll recreate it on this branch)
- Make this branch up-to-date with dev
- Add a PHI for the thesis work and create a TODO list of related tasks

cyrus-:
Sounds good, thanks @jlam55555!

jlam55555:
I made the changes mentioned [here][changes].

See [the new PHI][phi].

[changes]: https://github.com/hazelgrove/hazel/pull/586#issuecomment-1157003522
[phi]: https://github.com/hazelgrove/phi/tree/15-environment-based-evaluation

mirryi:
What if we monadified the state in the evaluator, like in [this branch](https://github.com/hazelgrove/hazel/tree/monadic-evaluator) I wrote two weeks ago?

src/hazelcore/VarMap.re
```diff
@@ -1,7 +1,8 @@
 open Sexplib.Std;
 
 [@deriving sexp]
-type t_('a) = list((Var.t, 'a));
+type t_('a) = list(t__('a))
+and t__('a) = (Var.t, 'a);
```
mirryi: What's the use for separating out `t__`?

mirryi:
Some organizational thoughts:

1. Feels like there's a bit of redundancy between some modules and data types, particularly `DHExp.evalenv` also being `EvalEnv.t` and `DHExp.result = EvaluatorResult.t` and `EvalEnv.result_map =` right component of `DHExp.evalenv`. I reckon there could be a way to reorganize these to streamline the module dependencies and make it easier to follow.
2. Maybe this is just me overthinking these modules, but I feel it would be better if the `EvalEnv` functions that take `EvalState.t` didn't, instead taking some environment id generator (e.g. `EvalIdGen`). It seems like `EvalState` should only be used in `Evaluator`.
3. Perhaps the `Eval*` modules should be `Evaluator*` to maintain naming consistency.

src/hazelcore/dynamics/EvalState.re
```diff
@@ -0,0 +1,16 @@
+[@deriving sexp]
+type t = (EvalEnvId.t, EvalStats.t);
```
mirryi: Make this a record so we can use update syntax.

src/hazelcore/VarMap.re
jlam55555: It comes from [this commit](https://github.com/hazelgrove/hazel/commit/a8afd66a36a6070cc8e8d87a212e0ef344f172b8#diff-df6427418ded9550af74f091c04f5ae6d757004913c60d472543cab543e5f922), and it seems that I wanted to avoid rewriting `(Var.t, DHExp.t)` too many times in `EvalEnv.rei`. As you point out it's not used and it's probably not very useful

src/hazelcore/dynamics/EvalState.re
jlam55555: Makes sense. I think I used a tuple because there doesn't seem to be much use of records in the project (or at least in hazelcore), but that's not a very good reason to avoid record syntax.

`Result.t` could also benefit from record syntax

jlam55555:
> Some organizational thoughts:
> 
> 1. Feels like there's a bit of redundancy between some modules and data types, particularly `DHExp.evalenv` also being `EvalEnv.t` and `DHExp.result = EvaluatorResult.t` and `EvalEnv.result_map =` right component of `DHExp.evalenv`. I reckon there could be a way to reorganize these to streamline the module dependencies and make it easier to follow.
> 2. Maybe this is just me overthinking these modules, but I feel it would be better if the `EvalEnv` functions that take `EvalState.t` didn't, instead taking some environment id generator (e.g. `EvalIdGen`). It seems like `EvalState` should only be used in `Evaluator`.
> 3. Perhaps the `Eval*` modules should be `Evaluator*` to maintain naming consistency.

@mirryi Thanks for your comments Eric! These are some very valid points.

1. I think this was the very first major decision that I ran into, and one that persisted throughout until I got used to it. The redundancy you notice is due to the fact that we have a dependency loop (DHExp.Environment -> `DHExp.result` -> `DHExp.t` -> `DHExp.Closure` -> `DHExp.Environment`) and want `DHExp.result` to be in its own module, hence `EvaluatorResult.t === DHExp.result`. (Note that this is also quite annoying because it causes constant boxing/unboxing of the `Indet` and `BoxedValue` judgments for the environment definiens, even when they're usually irrelevant.) The alternative (which gets rid of this dependency loop) is to have environments map variables to `DHExp.t` rather than `EvaluatorResult.t` (essentially, the expression without the "final" judgment) -- but this forces us to recompute the final judgment each time we need it. In hindsight, the latter may not be so bad -- a call to `Evaluator.evaluate` on a final `DHExp.t` just to get the final judgment for the expression shouldn't take too many evaluation steps. I'm not too sure if that will truly be beneficial though.
2. Your suggestion makes sense. Originally, `EvalState` *was* `EvalEnvIdGen`, since this was the only state carried through evaluation. Only in a [recent commit](https://github.com/hazelgrove/hazel/pull/586/commits/676a7c37f4580315d9b41a0c24302aaff9c1a3ce) was it renamed to `EvalState` and additional fields get added to `EvalState` (e.g., gas counter). So yes, it would make sense to make the id generator a member of `EvalState` and only have that passed to the `EvalEnv` methods.
3. I have a bit of naming indecision. Originally, I was afraid of long names like `EvaluatorEnvironment.t` (since we have the `Evaluator` module and the `Environment.t`) or even longer like `EvaluatorEnvironmentIdGenerator.t`, etc. I think `EvaluatorState` would make sense, but perhaps the `EvalEnv*` names should be overhauled completely. `EvalEnv`'s are just numbered environments, and their name should better reflect that.

I totally agree that 1. and 3. are some of the uglier parts of the implementation and can be improved.

Another part that may be of interest is the postprocessing step in `EvalPostprocess.re`: I already cleaned this up from the original implementation from my thesis, but it still feels overcomplicated. For example, there's the memoization of postprocessing by environments, tracking of holes and hole instantiation numbers using `HoleClosureInfo_.t` and `HoleClosureInfo.t`, and a separate stage for hole parent tracking. If anything, I think it's easier to understand (at least from the comments and naming) than the old postprocessing in `Program.renumber`, but it's still fairly abstruse.

mirryi:
> 1. I think this was the very first major decision that I ran into, and one that persisted throughout until I got used to it. The redundancy you notice is due to the fact that we have a dependency loop (DHExp.Environment -> `DHExp.result` -> `DHExp.t` -> `DHExp.Closure` -> `DHExp.Environment`) and want `DHExp.result` to be in its own module, hence `EvaluatorResult.t === DHExp.result`. (Note that this is also quite annoying because it causes constant boxing/unboxing of the `Indet` and `BoxedValue` judgments for the environment definiens, even when they're usually irrelevant.) The alternative (which gets rid of this dependency loop) is to have environments map variables to `DHExp.t` rather than `EvaluatorResult.t` (essentially, the expression without the "final" judgment) -- but this forces us to recompute the final judgment each time we need it. In hindsight, the latter may not be so bad -- a call to `Evaluator.evaluate` on a final `DHExp.t` just to get the final judgment for the expression shouldn't take too many evaluation steps. I'm not too sure if that will truly be beneficial though.

Hmm yeah this is tricky. Along the lines of the above, we could maybe introduce something like `EvaluatorResult.wrap : DHExp.t -> EvaluatorResult.t` which computes the final judgment in constant time. As you said, I'm not sure if that would be worth it just for the sake of cleaning up the modules/types. Alternatively, and probably not worth it, but apparently mutually recursive modules are a thing? If we went that route, we probably wouldn't be able to define them in separate files.

I also noticed that `DHExp.environment` isn't used at all (unless I missed something). `DHExp.evalenv` can probably renamed to that?

src/hazelweb/model/Program.re
```diff
@@ -370,17 +105,34 @@ let get_elaboration = (program: t): DHExp.t =>
...
-    let (d_renumbered, hii) = renumber([], HoleInstanceInfo.empty, d);
-    (d_renumbered, hii, BoxedValue(d_renumbered));
-  | Indet(d) =>
-    let (d_renumbered, hii) = renumber([], HoleInstanceInfo.empty, d);
-    (d_renumbered, hii, Indet(d_renumbered));
+  | (es, BoxedValue(d)) =>
+    let (hci, d) =
+      switch (d |> EvalPostprocess.postprocess) {
+      | d => d
+      | exception (EvalPostprocessError.Exception(reason)) =>
```
mirryi: This doesn't actually catch any exceptions because the ones thrown in `EvalPostprocess` are `EvalPostprocess.Exception`, not `EvalPostprocessError.Exception`. The error/exception types are duplicated in the two modules---which should be kept?

mirryi:
Also, it would be good to have a brief doc to help people understand how the new environment-based evaluator differs from the previous one, maybe specifically discussing differences from the Hazelnut Live paper, without having to read the PHI or the thesis.

cyrus-:
Styling in the context inspector seems to have broken: 
Screenshot from 2022-07-15 15-14-11

compare to dev:

Screenshot from 2022-07-15 15-14-55

src/hazelcore/VarMap.re
cyrus-: doesn't need to be mutually recursive, and you can name `t__` to `entry`

src/hazelcore/VarBstMap.re
```diff
@@ -0,0 +1,13 @@
+open Sexplib.Std;
```
cyrus-, line 1: Don't see VarBstMap being used anywhere

src/hazelcore/Var.re
```diff
@@ -42,3 +42,6 @@ let split = (pos, name) => {
   let right_var = String.sub(name, pos, String.length(name) - pos);
   (left_var, right_var);
 };
+
+/* Used for VarMap */
+let compare = (x: t, y: t) => compare(x, y);
```
cyrus-, line 47: don't see it actually being used?

src/hazelcore/dynamics/DHExp.re
```diff
@@ -130,22 +130,16 @@ module BinFloatOp = {
...
-    )
-  | ExpandingKeyword(
-      MetaVar.t,
-      MetaVarInst.t,
-      VarMap.t_(t),
-      ExpandingKeyword.t,
-    )
-  | FreeVar(MetaVar.t, MetaVarInst.t, VarMap.t_(t), Var.t)
-  | InvalidText(MetaVar.t, MetaVarInst.t, VarMap.t_(t), string)
+  /* Hole types */
```
cyrus-: Hole forms

src/hazelcore/dynamics/DHExp.re
```diff
@@ -164,26 +158,31 @@ type t =
...
-  | InconsistentBranches(MetaVar.t, MetaVarInst.t, VarMap.t_(t), case)
   | Cast(t, HTyp.t, HTyp.t)
   | FailedCast(t, HTyp.t, HTyp.t)
   | InvalidOperation(t, InvalidOperationError.t)
 and case =
   | Case(t, list(rule), int)
 and rule =
-  | Rule(DHPat.t, t);
+  | Rule(DHPat.t, t)
+and environment = VarMap.t_(t)
```
cyrus-: isn't this already defined as Environment.t?

src/hazelcore/dynamics/EvalStats.re
```diff
@@ -0,0 +1,10 @@
+open Sexplib.Std;
+
+[@deriving sexp]
+type t = int;
```
cyrus-: make this a record

src/hazelweb/model/Program.re
jlam55555: Fixed in https://github.com/hazelgrove/hazel/pull/586/commits/5018f013da425d9733a8f7c2f251c1f3c8f2a4e7 (see commit message).

My bad, didn't remove the `EvalPostprocessError` module when doing a refactor, I was meaning to move it to `EvalPostprocess.error`.

jlam55555:
@mirryi 

> > 1. I think this was the very first major decision that I ran into, and one that persisted throughout until I got used to it. The redundancy you notice is due to the fact that we have a dependency loop (DHExp.Environment -> `DHExp.result` -> `DHExp.t` -> `DHExp.Closure` -> `DHExp.Environment`) and want `DHExp.result` to be in its own module, hence `EvaluatorResult.t === DHExp.result`. (Note that this is also quite annoying because it causes constant boxing/unboxing of the `Indet` and `BoxedValue` judgments for the environment definiens, even when they're usually irrelevant.) The alternative (which gets rid of this dependency loop) is to have environments map variables to `DHExp.t` rather than `EvaluatorResult.t` (essentially, the expression without the "final" judgment) -- but this forces us to recompute the final judgment each time we need it. In hindsight, the latter may not be so bad -- a call to `Evaluator.evaluate` on a final `DHExp.t` just to get the final judgment for the expression shouldn't take too many evaluation steps. I'm not too sure if that will truly be beneficial though.
> 
> Hmm yeah this is tricky. Along the lines of the above, we could maybe introduce something like `EvaluatorResult.wrap : DHExp.t -> EvaluatorResult.t` which computes the final judgment in constant time. As you said, I'm not sure if that would be worth it just for the sake of cleaning up the modules/types. Alternatively, and probably not worth it, but apparently mutually recursive modules are a thing? If we went that route, we probably wouldn't be able to define them in separate files.
> 
> I also noticed that `DHExp.environment` isn't used at all (unless I missed something). `DHExp.evalenv` can probably renamed to that?

I didn't know about recursive modules, I wonder if that has any adverse side effects (i.e., performance issues or workarounds needed). Something to consider in the future if it significantly improves readability.

I'll leave the naming of `DHExp.evalenv` to your discretion. You're right, `DHExp.environment` isn't used and probably comes from a very early iteration of environments for evaluation, so it can be removed. But it may be a little confusing to rename `DHExp.evalenv` to `DHExp.environment` since there's also an `Environment.t` which is different. Again, I don't really like the names but I haven't thought of a better one. I already wrote about how `EvalEnv*` is probably a bad name. Another bad name is `HoleClosure*`, since I think `HoleInstantiation*` is a better name for the concept but haven't gotten around to renaming all the related symbols.

> Also, it would be good to have a brief doc to help people understand how the new environment-based evaluator differs from the previous one, maybe specifically discussing differences from the Hazelnut Live paper, without having to read the PHI or the thesis.

While the thesis report is long (probably too long for the changes made), I don't think that the PHI is too long or obscure, and its purpose is to document this change. I agree that it may benefit from a section explicitly listing the differences from the `dev` evaluator. Let me know if there's anything else that should be clarified/reworded/elaborated on in the PHI.

@cyrus- 

> Styling in the context inspector seems to have broken: Screenshot from 2022-07-15 15-14-11
> 
> compare to dev:
> 
> Screenshot from 2022-07-15 15-14-55

Thanks for pointing out, fixed in https://github.com/hazelgrove/hazel/pull/586/commits/ce73c18aa56030bf6437e8d28f0120d01544c9a7

This brings up the related point that there's no special styling for the "Hole Closure Parents" pane in the context inspector right under the misstyled one, so it's pretty ugly for now especially when the hole instantiation is directly in the result or has multiple parents. I'll leave it to someone else since my design sense is off the charts (ugly).

cyrus-:
Yeah I was also finding all the things called "environment"s confusing. There seem to be evaluation environments and hole closure environments. Can we unify the two? They both seem to be mappings from variables to DHExps at heart, though eval environments map to a result (is that useful for something, or just easier not to pull out the DHExp?)

mirryi:
Bug: getting an error (`Async_kernel: Unhandled exception (Invalid_argument List.for_all2)`) when pressing backspace at this position:

image

Either from `Statics_Exp.syn_operand` or `ListUtil.for_all2_opt` or either `DHExp.fast_equals` or `Result.fast_equals`, but probably the latter since no such bug exists in dev.

src/hazelcore/Builtin.re
```diff
@@ -1,14 +1,19 @@
 open Sexplib.Std;
 
+open EvaluatorMonad.Syntax;
+
 /* Evaluator alias. */
 [@deriving sexp]
-type evaluate = DHExp.t => EvaluatorResult.t;
+type evaluate =
```
cyrus-, line 7: confusing to have a `evaluate` and `eval` type both, how about `evaluate` and `builtin_evaluate`?

src/hazelcore/dynamics/DHExp.re
```diff
@@ -1,228 +1,3 @@
...
-let apply_casts = (d: t, casts: list((HTyp.t, HTyp.t))): t =>
-  List.fold_left(
-    (d, c: (HTyp.t, HTyp.t)) => {
-      let (ty1, ty2) = c;
-      cast(d, ty1, ty2);
-    },
-    d,
-    casts,
-  );
+/* FIXME: Need interface file, but not sure how to write sharing constraint for
```
cyrus-: you can show the inferred interface in utop/rtop

src/hazelcore/dynamics/DH.re
```diff
@@ -0,0 +1,514 @@
...
+  let map_keep_id = (f, env) =>
+    env |> map_of |> Environment.map(f) |> wrap(env |> id_of);
+
+  let filter = (f, env) =>
+    env |> map_of |> Environment.filter(f) |> of_environment;
+
+  let filter_keep_id = (f, env) =>
+    env |> map_of |> Environment.filter(f) |> wrap(env |> id_of);
+
+  let placeholder = wrap(EnvironmentId.invalid, Environment.empty);
```
cyrus-, line 531: maybe include a comment for why this is necessary / figure out a way to make it unnecessary?

src/hazelcore/dynamics/EnvironmentId.re
```diff
@@ -0,0 +1,12 @@
+open Sexplib.Std;
+
+[@deriving sexp]
+type t = int;
+
+let init = 0;
+let equal = (==);
+
+let next = id => id + 1;
```
cyrus-: why is next in here rather than in gen?

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -33,33 +42,38 @@ let ground_cases_of = (ty: HTyp.t): ground_cases =>
...
-  | (_, ConsistentCase(Case(_, _, _))) => Indet
+  | (_, Ap(_, _)) => IndetMatch
+  | (_, BinBoolOp(_, _, _)) => IndetMatch
+  | (_, BinIntOp(_, _, _)) => IndetMatch
+  | (_, BinFloatOp(_, _, _)) => IndetMatch
+  | (_, ConsistentCase(Case(_, _, _))) => IndetMatch
+
+  /* Closure should match like underlying expression */
+  | (_, Closure(_, Fun(_))) => DoesNotMatch
+  | (_, Closure(_, _)) => IndetMatch
```
cyrus-: underscore here may not be a good idea for code evolution -- can we just recurse on what is under the closure?

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -33,33 +42,38 @@ let ground_cases_of = (ty: HTyp.t): ground_cases =>
 type match_result =
   | Matches(Environment.t)
   | DoesNotMatch
-  | Indet;
+  | IndetMatch;
 
 let rec matches = (dp: DHPat.t, d: DHExp.t): match_result =>
   switch (dp, d) {
   | (_, BoundVar(_)) => DoesNotMatch
```
cyrus-: should be a failwith here I think -- we are assuming `d` is final

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -350,14 +369,16 @@ and matches_cast_Cons =
...
+  | InconsistentBranches(_) => IndetMatch
+  | EmptyHole(_, _) => IndetMatch
+  | NonEmptyHole(_, _, _, _) => IndetMatch
+  | FailedCast(_, _, _) => IndetMatch
+  | InvalidOperation(_) => IndetMatch
   };
 
-/* closed substitution [d1/x]d2*/
+/* closed substitution [d1/x]d2
+   Not needed for evaluation with environments,
```
cyrus-: lets move this to its own Substitution module since its not part of the Evaluator anymore

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -393,6 +414,12 @@ let rec subst_var = (d1: DHExp.t, x: Var.t, d2: DHExp.t): DHExp.t =>
       let d3 = subst_var(d1, x, d3);
       Fun(dp, ty, d3);
     }
+  | Closure(env, d3) =>
+    /* Closure shouldn't appear during substitution (which
```
cyrus-: what does this comment mean?

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -524,287 +553,412 @@ let eval_bin_float_op =
...
-        switch (evaluate(d2)) {
-        | BoxedValue(BoolLit(b2)) =>
-          BoxedValue(eval_bin_bool_op(op, b1, b2))
-        | BoxedValue(d2') =>
-          raise(EvaluatorError.Exception(InvalidBoxedBoolLit(d2')))
-        | Indet(d2') => Indet(BinBoolOp(op, d1', d2'))
+let rec evaluate: (ClosureEnvironment.t, DHExp.t) => m(EvaluatorResult.t) =
+  (env, d) => {
+    /* Increment number of evaluation steps (calls to `evaluate`). */
+    let* () = take_step;
```
cyrus-, line 432: what is the step count being used for? in prep for the gas stuff?

src/hazelcore/dynamics/DH.rei
```diff
@@ -0,0 +1,244 @@
...
+  ClosureEnvironment is an environment (mapping variables to expressions) that
+  are used during evaluation. It is different from Environment.t in two ways:
+
+  1. It maps Var.t to Evaluator.result (rather than to DHExp.t) 2. Each
+  ClosureEnvironment has an ID associated with it (if evaluation reaches it),
+  or is unreachable. (i.e., a placeholder environment)
+
+  Environment.t may be useful in certain cases, namely pattern matching, when
+  an evaluated result is not needed. ClosureEnvironment is used for environents
+  during evaluation, including in closures. ClosureEnvironments are numbered so
```
cyrus-: since its not just a closure environment anymore, lets call this one EvalEnvironment

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -524,287 +553,412 @@ let eval_bin_float_op =
...
+      evaluate(env, d);
+
+    | Let(dp, d1, d2) =>
+      let* r1 = evaluate(env, d1);
+      switch (r1) {
+      | BoxedValue(d1)
+      | Indet(d1) =>
+        switch (matches(dp, d1)) {
+        | IndetMatch
+        | DoesNotMatch => Indet(Closure(env, Let(dp, d1, d2))) |> return
```
cyrus-, line 453: the DoesNotMatch case here should really be handled as an error case, same as a case that has no more rules left to apply...

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -524,287 +553,412 @@ let eval_bin_float_op =
...
+
+    switch (d) {
+    | BoundVar(x) =>
+      let d =
+        x
+        |> ClosureEnvironment.lookup(env)
+        |> OptUtil.get(() =>
+             raise(EvaluatorError.Exception(FreeInvalidVar(x)))
+           );
+      evaluate(env, d);
```
cyrus-, line 444: leave a comment explaining why this evaluate call is necessary, since naive assumption would be that the environment already maps to final d

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -524,287 +553,412 @@ let eval_bin_float_op =
...
-            InvalidOperation(
-              BinIntOp(op, IntLit(n1), IntLit(n2)),
-              DivideByZero,
-            ),
-          )
-        | _ => BoxedValue(eval_bin_int_op(op, n1, n2))
+      };
+
+    | FixF(f, _, _) as d =>
+      let* env' = with_eig(ClosureEnvironment.extend(env, (f, d)));
```
cyrus-: shouldn't this be a call to evaluate_extend_env?

src/hazelcore/dynamics/Evaluator.re
```diff
@@ -524,287 +553,412 @@ let eval_bin_float_op =
...
+      let* r1 = evaluate(env, d1);
+      switch (r1) {
+      | BoxedValue(Closure(closure_env, Fun(dp, _, d3)) as d1) =>
+        let* r2 = evaluate(env, d2);
+        switch (r2) {
+        | BoxedValue(d2)
+        | Indet(d2) =>
+          switch (matches(dp, d2)) {
+          | DoesNotMatch
+          | IndetMatch => Indet(Ap(d1, d2)) |> return
```
cyrus-, line 476: same comment here about it being an error if there is a failed match, not a plain indet

cyrus-:
getting a stack overflow error when I press `:` to add a type annotation here
Screenshot from 2022-08-05 00-01-00

cyrus- (changes requested):
partial review

src/hazelcore/VarBstMap.re
mirryi: It's used in `dynamics/DH.re` for `Environment.t`.

src/hazelcore/Var.re
mirryi: Used for `VarBstMap`; I'll fix the comment to say so

src/hazelcore/dynamics/DH.re
mirryi: A comment exists in the interface file and further comment in the location where it's used. Not sure there's a workaround to avoid it.

src/hazelcore/dynamics/Evaluator.re
mirryi: No real usage at the moment. Could be used for gas stuff, but not for the localised approach we discussed last week.

src/hazelcore/dynamics/Evaluator.re
mirryi: I think it's pointing to the fact that `subst_var` is used only in `Elaborator_Exp`, at which point there should not yet be any closures.

src/hazelcore/dynamics/DH.rei
mirryi: In the context of the DH module, its purpose is for the environments of `Closures`, which is why the name `ClosureEnvironment` seems to fit more for me. The evaluator feels separate and happens to use this type for the evaluation process. Either name works fine, though.

src/hazelcore/dynamics/Evaluator.re
mirryi: I'm confused as to why this would be the case...

mirryi:
@cyrus- Merged new web worker stuff in and played around with it a little. One big problem is that after evaluation and postprocessing, for programs with lots of holes, sending the result back to the main thread from the worker can be incredibly slow, as we serialize the result in an sexp on the worker side and deserialize again on the main thread (because copying raw data didn't work for some reason). 

This is mostly because, I think, we have closure environments in their entirety in (1) each hole closure, (2) the hole instance info, making the serialized result large. To alleviate this, we could store all closure environments in a separate map and keep only the ids in the expression tree and hole info. However, this would make the `DHExp.t` datatype not self-containing. Alternatively, we could figure out how to make data transfer between threads faster some other way, but I'm not sure how to do this.

cyrus-:
qsort example doesn't seem to be working -- local variables (smaller, bigger, etc.) are not being shown in the sidebar

Screenshot from 2022-08-22 15-34-07

cyrus-:
merged into haz3l-tests, thanks!

PR #588, opened 2022-03-30
Fix EANEHole rule

jlam55555:
Fix type assignment for type-inconsistent non-empty holes in `ana_elab_operand` in order to match rule EANEHole. Holes in analytic position should be assigned the analyzed type, not type `Hole`.

This caused unnecessary casts to be inserted but otherwise seems to not cause differences in elaboration/evaluation.


cyrus- (approved):
Good catch, thanks!

Issue #583, opened 2022-02-28
cursor placement on comma insertion when analyzing against product type is incorrect


cyrus-:
pressing comma here leaves the cursor before the 5 rather than after the comma as expected. this may be a regression from #574 that slipped in, or never fixed in that PR. correct behavior is to the first hole in the completed tuple, rather than to the first position in the tuple

![Screenshot from 2022-02-28 00-56-44](https://user-images.githubusercontent.com/280638/155931803-5ff13567-1585-4b47-83a7-8babae126ef6.png)

AlienKevin:
> pressing comma here leaves the cursor before the 5 rather than after the comma as expected. this may be a regression from #574 that slipped in, or never fixed in that PR. **correct behavior is to the first hole in the completed tuple**, rather than to the first position in the tuple
> 
> ![Screenshot from 2022-02-28 00-56-44](https://user-images.githubusercontent.com/280638/155931803-5ff13567-1585-4b47-83a7-8babae126ef6.png)

I can think of 3 possible understanding of the correct behavior and not sure which one is the expected behavior:
1. first empty hole after the cursor
2. first empty hole in the tuple
3. first empty/filled hole in the tuple

Personally, I think option 1 makes the most sense because by pressing comma, the user usually intends to fill out the new empty hole after the cursor, not going back to the first empty/filled hole when there's one. This is also consistent with cases with mismatched tuple length. For example: `let x : (Int, Int, Int) = _, 5| in ...` Pressing "," after 5 inserts an empty hole and I think the user expects the cursor to be placed at the new hole rather than going all the way back to the first hole before 5.
Option 3 is particularly tricky to implement, because a hole before tuple completion can either remain as a hole or become a valid element of the tuple. However, at the time of completion, we don't know which option it will take without comparing the type of the hole with the type declaration.
Example of filled hole: `let x : (Int, Int) = t| in ...` Since t is undefined, if I press "," hazel will autocomplete a comma and insert a new hole after t. Should I place the cursor at the filled hole containing t or at the new empty hole after t?

cyrus-:
I think we should place the cursor at the next position after the comma when tuple completion is triggered by a comma, regardless of where any holes may be inserted. Otherwise you'll have situations like `5| 6` when analyzed against a 3-tuple where the first hole skips past the 6.

If tuple completion is triggered by open parentheses, then I think it makes sense to place the cursor at the first hole in the tuple, which will always either be in the first position if there was already a hole there or else in the second position.

[...]

PR #591, opened 2022-05-05
Enhance tuple completion

AlienKevin:
# Expected Behaivors

https://github.com/hazelgrove/hazel/issues/583#issuecomment-1081008183

> I think we should place the cursor at the next position after the comma when tuple completion is triggered by a comma, regardless of where any holes may be inserted. Otherwise you'll have situations like 5| 6 when analyzed against a 3-tuple where the first hole skips past the 6.

> If tuple completion is triggered by open parentheses, then I think it makes sense to place the cursor at the first hole in the tuple, which will always either be in the first position if there was already a hole there or else in the second position.

### 1. Parenthesize a nonempty hole (✅ Fixed)

**Matching type:**
Before `let x : (Int, Int) = |1 in _`
After '(' `let x : (Int, Int) = (|1) in _`

Fixed to `let x : (Int, Int) = (1, |_) in _`

**Mismatched type:**
Before `let x : (Int, Int) = |true in _`
After '(' `let x : (Int, Int) = (|true) in _`

Fixed to `let x : (Int, Int) = (|true, _) in _`

### 2. Tuple completion on type annotation (Not fixed/waiting for updates to type annotation)

Before `let x| = (1, 2, 3) in _`
After ':' `let x : (Int), (Int), (Int) = (1, 2, 3) in _`

Before `let x : |_ = (1, 2, 3) in _`
After '(' `let x : (|_) = (1, 2, 3) in _`

### 3 . Tuple completion on mismatched tuple type annotation (Not fixed/waiting for updates to type annotation)

Before `let x : (Int|) = (1, 2, 3) in _`
After '(' `let x : (Int, |_) = (1, 2, 3) in _`

### 4. Parenthesizing mismatched tuple (✅ Fixed)

Before `let x : (Int, Int, Int) = |3, 4 in _`

After '(' `let x : (Int, Int, Int) = (3), 4 in _`

Fixed to `let x : (Int, Int, Int) = (3, 4, |_) in _`

### 5. Parenthesizing mismatched wrong typed tuple (✅ Fixed)

Before `let x : (Int, Bool, Int) = |true, 3 in _`

After `Async_kernel: Unhandled exception (Failure "extend_let_body_ctx: impossible ana")`

Fixed to `let x : (Int, Bool, Int) = (|true, 3, _) in _`

### 6. Inserting comma moves cursor NOT to the next hole (✅ Fixed)

Before `(1, 2|)`

After ',' `(1, 2,| _)`

Fixed to `(1, 2, |_)`


src/hazelcore/Action_common.re
```diff
@@ -304,16 +324,68 @@ let complete_tuple_ =
...
         | (rev_holes, u_gen) => (rev_holes |> List.rev, u_gen)
       );
-    (
-      List.fold_right(
-        (new_hole, suffix: Seq.affix(_)) => A(comma, S(new_hole, suffix)),
-        new_holes,
-        Seq.E,
-      ),
-      u_gen,
+    print_endline("first_seq_length = " ++ string_of_int(first_seq_length));
```
cyrus-: don't include debug prints in dev

cyrus- (changes requested):
Looks great, thanks! Please remove any debug prints that are remaining and we can merge into dev.

src/hazelcore/Action_common.re
AlienKevin: @cyrus- Pushed a commit that removed all debug prints in this PR.

cyrus- (approved):

PR #592, opened 2022-05-23
Update Neovim editor recommendations

mirryi:
The recommended setup for language server support in Neovim is to use Neovim's built-in LSP client and the related plugin ecosystem. I updated the README to corroborate that.

Not sure how explicit to be.


mirryi:
Oops, I branched off the wrong branch...

cyrus- (approved):

PR #597, opened 2022-06-16
Update Ocaml version to 4.13.1

dedbox:
This PR updates `dev` to Ocaml version 4.13.1. It also updates the instructions in `docs/Updating-OCaml-Version.md` to opam version 2.1+.


cyrus-:
rtop still doesn't support 4.13?

mirryi:
[opam says](https://opam.ocaml.org/packages/rtop/) the `ocaml` constraint is `< 4.15`.

docs/Updating-OCaml-Version.md
```diff
@@ -2,69 +2,94 @@
...
 reasons (which should be re-confirmed periodically):
 
-- The package [`rtop`](https://opam.ocaml.org/packages/rtop/) does not support 4.13 or higher version.
+- The package [`sexplib`](https://opam.ocaml.org/packages/sexplib/) as of v0.15.0 has a bug preventing
+  compilation of expressions of the form
+
+    `[@deriving sexp] type t = u and ...`
+
+  where `u` is an identifier.
+- The package [`rtop`](https://opam.ocaml.org/packages/rtop/) did not support 4.13 or higher version.
```
cyrus-: is this inaccurate?

dedbox: About `rtop`, maybe I misunderstood the preceding instruction to re-check periodically. That was an old blocker that has been fixed. I changed "does not support" to "did not support". We could just remove it. If so, do we also want to remove the step to check `make repl`?

cyrus-: Ah yeah, this list should only be issues blocking upgrade to the next version of OCaml, if any, not a list of past issues. 

No, we should still check `make repl` whenever we upgrade.

dedbox: What about the `sexplib`  item? We only get OCaml 4.13.1 if we downgrade `sexplib` from 0.15 to 0.14, but it also takes several lower-level packages (e.g., `core`) with it.  Since ocaml's latest version is 4.14, it might not matter.

Ignoring the `sexplib` issue, `incr_dom` is blocking OCaml 4.14.0, so I'll add an item for that.

Also, I see a lot of uncapitalized c's.

disconcision: incr_dom 0.15 supports ocaml 4.14 i think?

dedbox: Yes, but it and a couple other packages depend on `js_of_ocaml` < 4.0.0 which depends on `ocaml` < 4.14.

cyrus-: We may want to swap out for Bonsai at some point soon, I don't think Jane Street is going to continue to maintain plain incr_dom?

docs/Updating-OCaml-Version.md
```diff
@@ -2,69 +2,99 @@
...
-
 - Announce the version change on the `#hazel-dev` channel of the `hazelgrove`
   Slack by sending the following message, with the appropriate VERSION:
 
       @channel
       We have switched `dev` to OCaml version VERSION. You can update to VERSION by doing the following things.
-      
-      - Merge your branch with either `dev` or `update_ocaml_VERSION` if that is tricky.
+
+      - Merge your branch with either `dev` or `update_ocaml_NEW-VERSION` if that is tricky.
```
cyrus-, line 92: We use VERSION elsewhere in the document, not NEW-VERSION

docs/Updating-OCaml-Version.md
dedbox: Bonsai still depends on incr_dom, and I vaguely recall something on the Jane Street blog about it being created specifically as their own high-level API for incr_dom, so I doubt they'd kill it suddenly or without notice.

Naively "upgrading" to Bonsai is straight forward, but we wouldn't benefit from e.g. automatic incrementalization without a major redesign.

cyrus- (approved):

PR #599, opened 2022-06-16
Fix issues related to OnDelim with Parenthesized and List

AlienKevin:
Fixed two issues discussed in Jun 16's Office Hour:
1. Fix `OnDelim` for `Parenthesized` and `List` in `holes_zoperand`
2. Fix `OnDelim(0, Before)` in `move_cursor_left_zoperand`

### DIFF
--- a/src/hazelcore/ZTyp.re
+++ b/src/hazelcore/ZTyp.re
@@ -142,6 +142,7 @@ and move_cursor_left_zoperand =
   | CursorT(OnOp(_) | OnText(_), _) => None
   | CursorT(OnDelim(k, After), ty) =>
     Some(CursorT(OnDelim(k, Before), ty))
+  | CursorT(OnDelim(0, Before), _) => None
   | CursorT(OnDelim(_, Before), Hole | Unit | Int | Float | Bool) => None
   | CursorT(OnDelim(_k, Before), Parenthesized(ty1)) =>
     // _k == 1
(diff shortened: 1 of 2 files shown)


src/hazelcore/ZTyp.re
```diff
@@ -142,6 +142,7 @@ and move_cursor_left_zoperand =
   | CursorT(OnOp(_) | OnText(_), _) => None
   | CursorT(OnDelim(k, After), ty) =>
     Some(CursorT(OnDelim(k, Before), ty))
+  | CursorT(OnDelim(0, Before), _) => None
   | CursorT(OnDelim(_, Before), Hole | Unit | Int | Float | Bool) => None
```
cyrus-, line 146: I think this is redundant now because for these forms there is only delim 0

cyrus-: oh but maybe need it to satisfy exhaustiveness checker?

AlienKevin: @cyrus- The line you highlighted (line 145) is used to handle on-the-left-edge case for everyone. The next line (line 146) is there to satisfy the exhuastiveness checker because even though `Hole`, `Unit` etc can only take a delimitor of `0`, the compiler is unaware of that.

cyrus- (approved):

cyrus-:
Thanks!

PR #600, opened 2022-06-22
Evaluation gas

### LINKED ISSUE #602: Function calling gas
Would be nice to catch stack overflows at top-level (or module-level) function applications and turn them into indets instead of causing the overall evaluation to fail. 

This would improve the situation in exercises mode in particular.

Hazel 2 effort: #600

mirryi:
This PR adds gas to the evaluator to prevent freezes caused by bottomless recursion. A new `EvaluatorState.t` record is passed through evaluation:

- On each call to `evaluate`, the `step` count is incremented
- On each call to `evaluate`, the `gas` amount is decremented
  - If `gas == 0`, then `evaluate` stops with an indeterminate result indicating out of gas

**To-do**:

- [ ] Configurable initial gas


disconcision:
lots of simple tests seem to work... here's one that triggers a hang (infinite loop i think, nothing printed to console): when i press x given the following state:
Screen Shot 2022-06-22 at 4 36 22 PM

based on this i'm guessing that the approach i took does not take into account situations where the recursive call is not in tail position.

mirryi:
@disconcision That does seem to be the case. Entering space (calling `f` with a hole argument) seems to lead to the same situation.

cyrus-:
@dlz01 maybe you can take a look at the issue with non-tail recursive positions?

dlz01:
I've fixed the bug and push some changes to allow better performance. It passed all my tests, but on some of the tests it's a little bit laggy. I will try tuning it for next few day so it won't crash on low RAM computer. Could someone test it out and see if there are any other bugs?

disconcision:
1. I haven't been able to crash it, so that's great!
2. I'm slightly confused as to why it's laggy on caret movement when there's an out-of-gas program running. My (I guess naive) mental model says that, since evaluation is memoized, it shouldn't be re-running when the code is the same (i.e. the expanded dhexp stays the same). Or maybe for memoization we're doing pointer equality instead of structural equality? Sort of feeling like we should be doing the latter, at least when the former fails. Not sure here, haven't fully thought through the consequences. This will be partially mitigated as we're eventually going to put evaluation on a separate thread (looking for volunteers!)
3. Thinking about the value display; I'm wondering if we want to consider ways of putting the out of gas error 'further up' rather than at the 'bottom' of a problematic expansion. As in, I'm not sure how useful it is to see all these 1 + 1 + 1s... maybe we just want the error to occur at the first encounter with a problematic fix. But I'm not sure there's a simple approach to prevent that (I can think of slightly complex exception-based ones), or if it's even desirable... Food for thought.
Screen Shot 2022-07-03 at 2 02 29 PM

cyrus-:
Showing the out of fuel error there seems reasonable, you do want to viscerally experience the recursion imo...

cyrus-:
It should probably say "Recursion Limit Exceeded" or something similar, not "Out of Fuel" which is PL jargon

dlz01:
I feel like I've got some idea on why it's being laggy. It seems like the evaluator are exhausting depth instead of breadth, so when one branch encounters dead loops, the evaluator will stop on this branch but would start on another. I could redesign some flows in our gas system, but one drawback could be that once the evaluator reaches a dead loop, all further evaluation would halt. I'm not sure whether that would be desired.

cyrus-:
Archiving this Hazel 2 PR, but see #602.

PR #651, opened 2022-06-29
Make make open make then open

disconcision:


cyrus-:
Actually maybe this isn't what we want, because there are other release profiles like `make release` that you may want to be testing.

If you want a single command you can run every time, you can always do `make dev open` yourself.

PR #660, opened 2022-07-18
Add an interface file for VarMap.re

mirryi:
#586 introduces `VarBstMap` for the purposes of evaluation environments. Currently, backed by the stdlib `Map` underneath, it has a different interface than `VarMap`. We may want to unify their interfaces and make both types opaque?

### DIFF
--- a/src/hazelcore/VarMap.rei
+++ b/src/hazelcore/VarMap.rei
@@ -1,0 +1,56 @@
+/**
+   The type of a variable-indexed map.
+ */
+[@deriving sexp]
+type t_('a) = list((Var.t, 'a));
+
+/**
+   [empty] is the empty map.
+ */
+let empty: t_('a);
+
+/**
+   [is_empty ctx] is true if [ctx] has no members.
+ */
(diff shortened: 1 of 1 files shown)


cyrus-:
Can we just use the standard library Map.S interface? https://v2.ocaml.org/api/Map.S.html

mirryi:
That would break a lot of existing usages of `VarMap`, but we could if we don't mind that.

mirryi:
Also, `VarMap.union` is different from `Map.S.union`.

cyrus-:
I think our VarMap stuff was always a bit unclean, so probably worth fixing this now with the changes coming to the main clients of the interface (the environment-based evaluator and the contexts changes coming in type-aliases-3). Closing this PR in favor of that effort.

mirryi:
@cyrus- Should I work to unify the `VarMap` and `VarBstMap` interfaces to match `Map.S` in the environments PR then?

cyrus-:
Yes, or just use VarBstMap throughout?

PR #666, opened 2022-07-31
Better remolding

dm0n3y:
Relaxes assumption made for remolding optimization in https://github.com/hazelgrove/tylr/pull/9 that tiles should have same nib sorts as output sort, in exchange for other assumptions that match our grammars so far:
- right nib shapes are never ambiguous for a given label
- filtering molds by shape before sort does not create unnecessary sort inconsistencies (imprecise but enough for handling unary/binary minus and paren/ap)

This change may also have uncovered a bug where remolding should in general take place beyond the focal segment, eg any segment directly contained by or directly containing put down shards, but this was an existing bug introduced by https://github.com/hazelgrove/tylr/pull/9 so not fixing here. Concretely this is observed when picking up an `=` in a let and putting back down with caret ending up in the let definition, which leaves the pattern sort-inconsistent. I think previous optimization happened to get away with this because it happened to preserve the original pattern mold, but the proper fix is to reassign molds beyond focal segment.

### DIFF
--- a/src/core/lang/Form.re
+++ b/src/core/lang/Form.re
@@ -95,6 +95,52 @@ let forms: list((string, t)) = [
       mk_pre(P.let_, Exp, [Pat, Typ, Exp]),
     ),
   ),
+  (
+    "typeann",
+    mk(
+      ss,
+      [":"],
+      {
+        out: Pat,
+        in_: [],
+        nibs: (
+          {shape: Concave(P.ann), sort: Pat},
+          {shape: Concave(P.ann), sort: Typ},
+        ),
+      },
+    ),
+  ),
+  (
+    "case",
+    mk(
+      ds,
+      ["case", "of"],
+      {
+        out: Exp,
+        in_: [Exp],
+        nibs: (
+          {shape: Convex, sort: Exp},
+          {shape: Concave(P.case_), sort: Rul},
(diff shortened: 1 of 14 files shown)


dm0n3y:
Realizing there are still some unresolved issues with the current approach.

One issue is specific to our target grammar and the current remolding optimization assumptions: if we're going to use `->` for both rules and type arrows, then there's a conflict in our precedence assignments for the right nib of `->`. Type arrow precedence should be less than comma, which should be less than rule arrow precedence. A temporary quick fix would be to use a different tokens for type arrow and rule arrow for now.

Another issue is more general: now that we have tiles with nib sorts different than output (eg `case .. of` being a prefix with rule sort on the right), and as we start to use overlapping tokens across heterogeneous forms (eg `->` for function arrow, rules, type arrows), it's unclear what should be expanded in general when typing eg `->`. In this specific case we might be fine by restricting function construction to `fun`, which I think is a reasonable restriction, but I haven't yet ruled out other possible ways this could pop up.

dm0n3y:
Talked to @disconcision offline. We'll use a different token for rule arrows for now. This resolves both first and second issues; the second issue was also not an issue given that @disconcision's previous changes restricted delayed expansions to the first keyword.

PR #671, opened 2022-08-04
add www/* to dune source dependencies

dedbox:
This PR adds the `www` directory to dune's default source dependency tree. This provides two concrete benefits:

1. Its contents will always be copied into the corresponding `_build` area.
2. `make` will become aware of modifications to html and css files (and whatever else is in there).


dedbox:
move to #677

PR #672, opened 2022-08-04
Add floating point less-than operator

dedbox:
This PR adds a less-than operator `<.` for floating point numbers.


dedbox:
moved to #679

PR #673, opened 2022-08-04
hazelc: Runtime and runtime library generator

mirryi:
PR contains:

- Substantial refactor of Grain IR library
- Runtime that uses a `MaybeIndet` sum to represent possibly indeterminate results for now
- Tools to safe-ish-ly generate code that uses runtime files and generate runtime files that would be cumbersome to write by hand


cyrus-:
Closing but linked from https://github.com/hazelgrove/hazel/issues/935

PR #675, opened 2022-08-04
Mouse input for movement and selection

dm0n3y:
Mouse input for movement and selection. Single click focuses cell and moves caret toward clicked position relative to cell. Dragging with mouse down invokes selection toward mouse position.


disconcision:
@dmoon1221  omg it feels amazing.

one thing: it only seems to work in school mode. i see the handlers are at the cell level as opposed to the code container level. can they just be on the code container so they will work in the other modes too?

disconcision:
couple other things: 

1. was noticing that the position logic floors as oppose to rounds: when you click on a character, the caret always appears on the position to the left of that character, as opposed to the nearest position. this results in increased frequency of misclicks.
2. (maybe related) was noticing the selection ui is jumpy in a very specific way. it only seems to apply to tokens that are an even number of characters long. if you are dragging over such tokens, while you are passing over the second half of the token, the selected region will rapidly jump around, selecting/unselecting the token. can reproduce if necessary

neither of theses are dealbreakers but if they're quick fixes would be nice to get them in

dm0n3y:
could use some additional polish in the simple + study modes. there's a delay in the selection highlighting if the mousedown-drag starts to the left of the code text. the code containers for those modes should probably fill the screen so that you can start the mousedown-drag anywhere. merging this in the meantime.

PR #680, opened 2022-08-05
Recursive let bindings

dedbox:
This PR implements recursive let-bindings in the same style as hazel 2.


src/core/statics/Statics.re
```diff
@@ -216,6 +216,30 @@ let rec uexp_to_info_map =
     let (ty_ann, m_typ) = utyp_to_info_map(typ);
     let (ty_pat, _ctx_pat, m_pat) =
       upat_to_info_map(~mode=Ana(ty_ann), pat);
+    let ctx =
```
disconcision, line 221: lets split this off into a helper... something like extend_let_def_ctx : (pat, typ, exp, ctx) => ctx

src/core/dynamics/elaborator.re
```diff
@@ -83,9 +83,46 @@ let rec dhexp_of_uexp =
       | InHole(FreeVariable) => Some(FreeVar(u, 0, sigma, name))
       | _ => wrap(BoundVar(name))
       }
+    | LetAnn(
+        {term: Var(x), _} as p,
+        {term: Arrow(_), _},
+        {term: Fun(_), _} as def,
+        body,
+      ) =>
+      let pat_typ =
```
disconcision, line 93: lets define these as helpers in statics... say typ_after_fix_pat and typ_after_fix_exp, both of which take a upat/uexp + the map and return a typ.

disconcision (changes requested):

disconcision:
nm need this just gonna merge this

PR #683, opened 2022-08-06
comments

dm0n3y:


lxguan1:
This adds comments as an extension of Whitespace.

The behavior of comments is similar to that of strings, with the syntax spec modified from there:

1.     # inserts an empty comment ##, with the caret being positioned between the hash signs
2.     Trying to delete either bounding # results in the deletion of the whole comment
3.     Trying to insert a # anywhere inside or adjacent to a comment results in no insertion and the caret being moved to the end of the comment
4.     Comments can contain any character which is otherwise valid in the syntax, except for # due to the above logic
5.     Linebreaks are currently not supported

src/haz3lcore/tiles/Whitespace.re
```diff
@@ -2,15 +2,45 @@ open Sexplib.Std;
 [@deriving (show({with_path: false}), sexp, yojson)]
```
cyrus-: calling this Whitespace still seems confusing -- in the literature, this is known as secondary notation, so maybe we should rename this Secondary

src/haz3lcore/zipper/Zipper.re
```diff
@@ -246,7 +246,20 @@ let rec construct = (from: Direction.t, label: Label.t, z: t): IdGen.t(t) => {
       /* Special case for constructing string literals.
          See Insert.move_into_if_stringlit for more special-casing. */
       construct(Left, [Form.string_delim ++ Form.string_delim], z)
+    | [content] when Form.is_comment(content) =>
```
cyrus-, line 266: more reason to distinguish proper whitespace from other secondary notational things

src/haz3lweb/view/Code.re
```diff
@@ -35,18 +35,18 @@ let of_grout = [Node.text(Unicode.nbsp)];
 let of_whitespace =
   Core.Memo.general(
     ~cache_size_bound=1000000, ((whitespace_icons, indent, content)) =>
-    if (content == Whitespace.linebreak) {
+    if (Whitespace.get_string(content) == Whitespace.linebreak) {
```
cyrus-: need to use String.equal not == for string comparisons (didn't look thoroughly elsewhere so may not just be here)

src/haz3lcore/lang/Molds.re
```diff
@@ -66,7 +66,8 @@ let instant_completes: completions =
     Form.forms,
   )
   |> List.flatten
-  |> List.sort_uniq(compare);
+  |> List.sort_uniq(compare)
+  |> List.append([("#", (["##"], Direction.Left))]);
```
cyrus-: seems ad hoc to include this in Molds.re?

cyrus- (changes requested):
Functionality looks great, nice work! 

Included some code comments for you to take a look at.

@dmoon1221 could you review this too? I know your regrold changes might affect some of this, but should we merge this and then you can port into regrold as needed?

src/haz3lcore/lang/Form.re
```diff
@@ -112,8 +114,8 @@ let is_string_delim = str => str == string_delim;
    approach to invalid text.*/
 let is_whitelisted_char = regexp("[!@]");
 
-/* A. Whitespace: */
-let whitespace = [Whitespace.space, Whitespace.linebreak];
+/* A. Secondary: */
```
cyrus-: more descriptive: "Secondary Notation (whitespace, comments, etc.)"

src/haz3lcore/tiles/Secondary.re
```diff
@@ -0,0 +1,53 @@
...
+  id: Id.t,
+  content: secondary_content,
+}
+and secondary_content =
+  | Whitespace(string)
+  | Comment(string);
+
+let space = " ";
+let linebreak = "⏎"; //alternative: "¶"
+let comment = Re.Str.regexp("^#[^#]*#?$");
```
cyrus-: why is the closing `#` optional?

cyrus- (changes requested):

src/haz3lcore/tiles/Secondary.re
lxguan1: I believe that it was originally a hack to get the insertion of comments to work, since otherwise I wasn't able to get it to insert the comment. I ended up forgetting about it. Right now the insertion and deletion logic should make it impossible for only a single '#' to appear, but you're right I'll work on making the code not require that.

cyrus-:
@lxguan1 also need to merge with latest haz3l-tests

cyrus- (approved):

cyrus-:
@lxguan1 can you update the text generation so that there is a span with a comment css class wrapping the comment text that we can use to style it? currently it is just emitted raw

image

PR #693, opened 2022-08-08
Text Editor

nj-wilson:
Adds Text Editor to the Hazel page.

When in use, hides the normal editor, as well as hiding/disabling the undo history and context elements.

The toggle editor button will be disabled in Text Editor view when there is a pending error, for which the error is the string editor passed back by Hazeltext parsing.


cyrus-:
Functionality looks great! Marking as reference for the haz3l version.

cyrus-:
Archiving this Hazel 2 PR, thanks for the effort on it @nj-wilson ! See #875 for Hazel 3.

PR #716, opened 2022-08-12
Clipboard (Copy, Paste, Cut)

disconcision:
d give this a quick try before I merge pls


disconcision:
@dmoon1221 okay there's actually still a consistency issue with this; you can still end up with non-shards in the backpack if you backspace on a selection containing an incomplete tile which has a bidelimited tile say "let x =". how straightforward would it be to change destruct to take these apart? probably want to do it via a flag as I can see going back and forth on this

dm0n3y:
should be straightforward. add another step that breaks apart the `to_pick_up` incomplete tiles produced in https://github.com/hazelgrove/hazel/blob/haz3l-tests/src/core/zipper/Zipper.re#L182

disconcision:
@dmoon1221 can you sanity check that change?

dm0n3y:
looks good

disconcision:
Merging this. Destruct now removes all children. Pick-up is bound to Ctrl-P.

PR #720, opened 2022-08-15
Relax mold restrictions + use recursive-descent-style molding policy

### LINKED ISSUE #700: Case expressions
Do we want to replicate the GLB branch joining except for currently indicated branch behavior from hazel2?

### LINKED ISSUE #719: fix floating point / unary minus issue

dm0n3y:
Use a sort-specific recursive-descent-style policy for remolding. Removes overly strict limit that a label cannot have distinct right nibs across its assignable molds. Add unary minus (closes #719) and case expressions (closes #700).


dm0n3y:
Works well for left-to-right construction. Lots of red appears when constructing before existing expressions; this will improve with upcoming history-sensitive changes. Merging for now so that others can get started on semantics of case.

PR #731, opened 2022-08-24
Restore basic err hole decorations

dm0n3y:
Restores err hole decorations. Basic rectangular contour following the shape of the error term's textual content. Needs some polish to integrate with existing tile decorations, both indicated and selected -- coming soon. Merging this for now so that others can work with visual semantic feedback.

image


dm0n3y:
image
image

PR #732, opened 2022-08-25
Haz3l Type Variables

dedbox:
This PR pulls in the work from `type-alias-3` and adapts it to work with the new core architecture.


dedbox:
Typing contexts, binding forms, and variable references appear to be working properly.

But I can't get type variables to display correctly in the cursor inspector yet:

image

Tracing does not reveal any related calls to `Typ.consistent` or `Typ.equivalent` or related checks. Local bindings have similar issues:

image

cyrus-:
@dedbox can you make this a PR into haz3l-tests rather than into dev?

disconcision:
i'm confused as to what the scope of this change is. i was expecting only (or mostly?) haz3l-level changes but it looks like there are (hundreds of files of) hazel2 changes as well?

dedbox:
I had to merge `dev` and then `type-alias-3` in order to guarantee that everything continues working once control is handed over to the evaluator. I tried moving only necessary things over a few times, but subtle bugs emerged quickly.

disconcision:
Ah so does this bring haz3l-tests current with hazel2 dev then? that's great, but I think we should maybe split off that change into a separate PR from type variables, both for the purposes of independently assessing the type variables changes and also facilitating the merges of my existing features branches off haz3l-tests, some of which have evaluator changes too.

dedbox:
Sounds good. I'll take a quick look tonight to see if there's an easy way to do the merge.

PR #736, opened 2022-08-25
Add listlit to Haz3l

LutSa:


cyrus-:
Looking good. This doesn't work as we discussed: 
Screenshot from 2022-08-26 15-08-06

dm0n3y (approved):
Took a pass over the haz3l statics and elaborator and made some fixes, looks good to merge there. n-ary visual decorations will be addressed as part of nary-skel changes.

PR #741, opened 2022-08-28
Update haz3l-tests to OCaml 4.14.0

dedbox:
Lots of minor changes needed for compiling under OCaml 4.14. It makes the LSP server very happy.


dm0n3y:
we'd have to update our release builds too, not sure how annoying that would be. also what's with the `core` to `core3` rename?

dedbox:
> what's with the `core` to `core3` rename?

The latest version of the Jane St. package `core_kernel` renames the module `Core_kernel` to `Core`. This is the only version supported by OCaml 4.14.

dm0n3y:
Instead of renaming our core, would it be possible to make only our `Util` library depend on jane street core and expose it via the `Util` library elsewhere as `Jane` or something?

docs/quick-and-dirty-ocaml414.md
```diff
@@ -0,0 +1,55 @@
+These instructions assume you want a local opam switch.
```
cyrus-: Can you make this into a shell script?

cyrus- (changes requested):

docs/quick-and-dirty-ocaml414.md
```diff
@@ -0,0 +1,55 @@
...
+``` shell
+eval $(opam env --switch=. --set-switch)
+```
+
+3. From somewhere outside the Hazel source tree, clone the following opam package repos:
+
+``` shell
+cd ..
+git clone 'git@github.com:janestreet/async_js.git'
+git clone 'git+https://github.com/janestreet/incr_dom.git'
```
cyrus-: git+https doesn't parse on my machine

dedbox: > git+https doesn't parse on my machine

Not sure what you mean. What did you try? I gave `opam pin` some https URLs and it converted them into what you see there.

cyrus- (approved):

Issue #730, opened 2022-08-23
haz3l mouseover capture in simple/study editors


disconcision:
I can't click on the bottom bar UI affordances in these modes, it gets captured by the overlay. more broadly, I think we should either push the click logic down into code, or create another level of abstraction which incorporates mouse logic but can be used uniformly by all editor modes. the difficulty in the latter case is deciding what exactly should live at this level. like i want to say that that level should be 'cell', and it should incorporate an evaluation display as well, but in school mode we use 'cells' where the evaluation part might be spliced from multiple editors, so this isn't fully clean.

PR #746, opened 2022-09-01
Encapsulate click-and-drag logic into Cell component

dm0n3y:
Encapsulates the click-and-drag logic (implemented via a mousedown whole-page overlay) into a reusable component called `Cell`. May want to rename this in the future if we have other intentions with the cell terminology. Fixes https://github.com/hazelgrove/hazel/issues/730 where mouse actions in cursor inspector would trigger the mousedown overlay and cause the cursor to move.

cc @cyrus- since you've been touching school cell stuff

### DIFF
--- a/src/web/view/Cell.re
+++ b/src/web/view/Cell.re
@@ -1,0 +1,76 @@
+open Virtual_dom.Vdom;
+open Core;
+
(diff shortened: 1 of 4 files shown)


disconcision:
Looks great conceptually, seems to work as well.

disconcision (approved):

PR #756, opened 2022-09-02
Haz3l tests new evaluator

dedbox:
Copies in files for the new evaluator.

Everything is nearly ready to go, but I'm totally burnt and can't quite figure out how to deal with some of the older files that are no longer relevant but prevent compilation, like `Program.re`. If anyone can figure out that part quickly, it's done.

Issue #757, opened 2022-09-02
comma binds more tightly than fun


cyrus-:
![image](https://user-images.githubusercontent.com/280638/188170509-b1f2b772-5074-432e-8219-3cdd2bcb9b23.png)

dm0n3y:
this is based on ocaml's operator precedence but I guess with tight ap we don't want this? should comma be lowest?

cyrus-:
Yeah I think so, or else we need a right delimiter for fun...

disconcision:
I think tuples should require parens

cyrus-:
you want function application to be `f((a, b, c))`?

disconcision:
no, but i'm fine with special-casing application

disconcision:
or another a way of looking at it maybe: disambiguation parens aren't a thing. parens with no contained commas always creates a one-tuple. one-tuples have identical semantics to the contained term. function application syntactically requires a tuple for the argument.

cyrus-:
We can discuss but regardless, it wouldn't solve the problem in this issue so lets make a separate thread about that?

PR #761, opened 2022-09-04
Copy over missing evaluator changes

dm0n3y:
https://github.com/hazelgrove/hazel/pull/756 didn't have all of the evaluator changes. This PR fixes that.


mirryi:
@dmoon1221 `ProgramEvaluator` operates on the old `Program.t`; do we want to reintroduce something like that or just pass a tuple of `Statics.map, Term.UExp.t` directly?

dm0n3y:
I changed the request type to DHExp, which is all that was needed afaict. Interface.re is the new Program-level API.

mirryi:
We're doing elaboration main thread side then?

dm0n3y:
Oh hm I guess never committed that change. I'll let you decide that based on what you did before. Passing in term + map should work too.

mirryi:
The map might be a little big to be passing over to the web worker; we can do elaboration client side I guess. Is `Interface` supposed to be concerned with UI view or evaluation? Maybe we could move `Interface.{evaluate, evaluation_result}` to `ProgramEvaluator`.

dm0n3y:
Interface is supposed to centralize general program-level semantic operations. Personally I'd err on the side of keeping things as is for now with the intention of reconsidering/merging APIs in a subsequent PR, but do what you gotta do to get things working.

mirryi:
Yeah, sure that's fine for now. I'll make the request `DHExp.t` and make `ProgramEvaluator` hook into `Interface`.

mirryi:
Is `DoesNotElaborate` getting raised supposed to be a bug or no? We need to catch those somewhere. More broadly, we need to change the calls to `Interface.evaluation_result` inside the ui views to use the web worker model again like in dev.

dm0n3y:
Yes it getting raised would signal a bug

dm0n3y:
Re: changing calls to `Interface.evaluation_result`. What's involved there? Is it just changing the function call?

mirryi:
The web worker model is asynchronous, so instead of making synchronous call to `evaluation_result`, evaluation needs to occur asynchronously when the program changes and pushes a model update when complete, which will then update a "current result" that is displayed

dm0n3y:
@cyrus- checking in. @mirryi and I talked offline. There's going to be a single asynchronously updated `ModelResult` in the top-level model. This should work with the multiple cells in school mode using the approach we discussed yesterday. School cells will get spliced together using a cell-separator operator that binds more tightly than lets. Semantically the cell-separator operator behaves the exact same as tuple commas; the only difference is precedence (I think they will need to be different given that https://github.com/hazelgrove/hazel/issues/757 suggests tuple commas should be lowest). In school mode, the cells get spliced together and evaluated; then, if you want to show individual results for each cell, you can extract the evaluated tuple components from the ModelResult (are we going to show results for anything other than the student impl?). 

We also talked about not internalizing cell separators and threading evaluator environments through cells, but this would require additional plumbing that we shouldn't try to bite off right now. We can revisit this decision later. For now, the only debt we take on right now is a second internal tuple operator.

cyrus-:
School mode actually needs to make multiple calls to the evaluator (running tests with student solution, then separately with instructor solution, etc.), so exposing a simple way to make multiple calls to the worker with a term is necessary.

cyrus-:
Showing the result of student cell would then also be straightforward, just a separate run without anything following it.

dm0n3y:
I see. Ok then I propose we add a string key (eg `"student_impl"`, `"instructor_solution"`, etc) to the evaluator request datatype and make the `ModelResult` a map from keys to evaluator responses.

cyrus-:
Sure, that works. The values in that map should tell you the state of the request, maybe like Pending/TimedOut/Finished, so that we can put in UI to indicate that for long-running requests / show timeout feedback.

cyrus-:
Unless there is a trick I'm missing, you can't run this PR from a `file:///` URL due to browser security restrictions.

It does seem to work okay if you spawn up a server, e.g. using `python3 -m http.server 8000` in the `_build/default/src/haz3lweb/www/index.html` directory. I guess this is fine, but we should modify the Makefile and README.

cyrus-:
Actually, getting really odd results when running it from localhost. Sometimes it works, sometimes it doesn't. Seems to be some sort of race condition. Works consistently when I add a print_endline and have the console open so I think it may have something to do with timeouts / fast path logic? I can show you on screen sharing (maybe, if that doesn't affect the race...) if you can't reproduce. I've been able to do it on both my Linux desktop and Mac, and in both Chrome and Firefox...

cyrus-:
If this is going to take a long time to debug, can we get a PR that is just the new evaluator but without the web worker stuff? Current haz3l-tests evaluator doesn't work if you have tests in your code. Making it difficult to make progress on school mode stuff now.

cyrus-:
Hm, problem also seems to go away if serving a `make release` version of the code. Not sure if that's just because it is faster. Very odd. But given that I'm blocked on this, I'm going to merge this into my branch for now.

mirryi:
For me it seemed like one evaluation would succeed, and then the rest would just timeout. I was trying to figure out where in the worker pipeline it was failing but still haven't a good idea.

PR #770, opened 2022-09-19
Haz3l string literals

### LINKED ISSUE #610: Strings
String-related language features:

- [ ] Basic string type, literals, concatenation, subscripting (#559)
- [ ] Code syntax highlighting

disconcision:
This adds string literals with basic statics + dynamics.

This is the intended syntax spec which string literals should conform to:

1. Pressing `"` inserts an empty  string literal `""`, with the caret being positioned between the quotes
2. Trying to delete either bounding " results in the deletion of the whole literal
3. Trying to insert a " anywhere inside or adjacent to a string literal results in no insertion and the caret being moved to the end of the literal
4. String literals can contain any character which is otherwise valid in the syntax, except for " due to the above logic
5. linebreaks are currently not supported


disconcision:
@d i think you wanted to be pinged on this;

cyrus-:
String literals themselves seem to work okay. Kind of annoying not to be able to select within the literal (how hard would that be to implement?)

The String type seems to be behaving differently from other built in types like Int -- doesn't turn purple when entered.

disconcision:
Fixed the type annotation color issue.

Not sure its meaningful implementation-wise to do in-string selections as a special case, so we're basically talking about general within-token selections. this touches a few different subsystems so i'm not sure how long it would take; possibly several days

cyrus-:
Seems reasonable for string literals to be distinct from other tokens in certain ways (also like comments), but also I do find myself wanting to be able to select part of a variable too (e.g. turning fold_left into fold_right). I think moving on to ADTs / building in Expr is higher priority though, so okay with merging this without that feature.

cyrus-:
- Can we give string literals their own text color just to visually distinguish them from the "metalanguage", especially since the syntax will be similar in 490?

- Not supporting symbols that aren't already in Hazel's syntax might be problematic, e.g. in class I have a factorial operator "!" but it doesn't work here. Type application syntax in 490 uses @ but that also doesn't work.

disconcision:
agreed it's fine if strings/comments are special wrt selection, it's just that such a subselection system necessarily touches everything in Zipper. i have long been contemplating a refactor which makes subtoken positioning into a sub-zipper as opposed to the current index-based system, which (i conjecture) would make the system more better suited to robustly adding features like subtoken selection, but i'm just not confident i've figured out the right level of abstraction there... another part of me feels like we should bite the bullet, flatten the whole data structure (so that arbitrary selections are just character ranges), and treat the entire tree/zipper structure as derived metadata attached to token ids

disconcision:
we can include whatever symbols you want, i just think it's a good idea from a robustness perspective to whitelist

disconcision:
re: color: sure, makes sense. let me check how straightforward that is

disconcision:
String literals now have a special color. Feel free to change it I just picked something random. I whitelisted ! and @ in strings, feel free to add more to Form.is_whitelisted_char

cyrus-:
if I press backspace here: 
Screen Shot 2022-09-22 at 9 15 31 PM

I get this weird state:
Screen Shot 2022-09-22 at 9 16 05 PM

Seems to combine the two into a string literal with the `"` inside?

cyrus-:
(I would expect it to leave a concave grout in between the two separate literals)

cyrus-:
We also need a string equality operator/builtin function for the next assignment (to implement substitution and context lookup).

cyrus-:
(I'm open to just doing overloaded == for now if you'd rather do it that way)

disconcision:
Lol didn't we have a two-meeting argument last year about budget adhoc polymorphic ==

cyrus-:
I was thinking we'd have the ability to overload == in a local scope at some point, but absent that (which is being worked on but probably won't be done until end-of-semester), I think its fine temporarily...

disconcision:
Fixed the string merge issue. probably will just add another operator or builtin for string equality as overloading is not straightforward in current system. any preferences for syntax/name?

cyrus-:
maybe `$==`?

disconcision:
Added $== operator

PR #772, opened 2022-09-20
Haz3l ctx separation

LighghtEeloo:
The implementation of separating term and type context as discussed.

### DIFF
--- a/src/haz3lcore/statics/Ctx.re
+++ b/src/haz3lcore/statics/Ctx.re
@@ -1,54 +1,60 @@
-open Sexplib.Std;
-
 [@deriving (show({with_path: false}), sexp, yojson)]
-type entry = {
-  id: Id.t,
-  typ: Typ.t,
+type t = {
+  term: TermCtx.t,
+  typ: TypCtx.t,
 };
 
 [@deriving (show({with_path: false}), sexp, yojson)]
-type t = VarMap.t_(entry);
-
-[@deriving (show({with_path: false}), sexp, yojson)]
-type co_item = {
-  id: Id.t,
-  mode: Typ.mode,
+type co = {
+  term: TermCtx.co,
+  typ: TypCtx.co,
 };
 
-[@deriving (show({with_path: false}), sexp, yojson)]
-type co_entry = list(co_item);
+let empty: t = {term: TermCtx.empty, typ: TypCtx.empty};
+let extend_term = (ctx: t, x: Token.t, term_entry): t => {
+  term: TermCtx.extend(ctx.term, (x, term_entry)),
+  typ: ctx.typ,
+};
+let extend_typ = (ctx: t, x: Token.t, typ_entry): t => {
+  term: ctx.term,
+  typ: TypCtx.extend(ctx.typ, (x, typ_entry)),
+};
+let lookup_term = (ctx: t, x) => TermCtx.lookup(ctx.term, x);
+let lookup_typ = (ctx: t, x) => TypCtx.lookup(ctx.typ, x);
+let union = (ctx1: t, ctx2: t): t => {
+  term: TermCtx.union(ctx1.term, ctx2.term),
+  typ: TermCtx.union(ctx1.typ, ctx2.typ),
+};
 
-[@deriving (show({with_path: false}), sexp, yojson)]
-type co = VarMap.t_(co_entry);
+let co_empty: co = {term: TermCtx.empty, typ: TypCtx.empty};
+let co_extend_term = (ctx: co, x: Token.t, term_entry): co => {
+  term: TermCtx.extend(ctx.term, (x, term_entry)),
+  typ: ctx.typ,
+};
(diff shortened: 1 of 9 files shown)


src/haz3lcore/statics/Ctx.re
```diff
@@ -1,54 +1,60 @@
-open Sexplib.Std;
-
 [@deriving (show({with_path: false}), sexp, yojson)]
-type entry = {
-  id: Id.t,
-  typ: Typ.t,
+type t = {
+  term: TermCtx.t,
```
cyrus-, line 3: call it Expr not Term (we use Term as an umbrella for Expr, Pat, Typ, etc.)

src/haz3lcore/statics/TypCtx.re
```diff
@@ -0,0 +1,17 @@
+include CtxBase;
+
+module Kind = {
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type t =
+    | Singleton;
```
cyrus-, line 6: Singleton should take an argument

src/haz3lcore/statics/Ctx.re
```diff
@@ -1,54 +1,60 @@
...
 
-[@deriving (show({with_path: false}), sexp, yojson)]
-type co_entry = list(co_item);
-
-[@deriving (show({with_path: false}), sexp, yojson)]
-type co = VarMap.t_(co_entry);
-
-let empty = VarMap.empty;
+let empty: t = {term: TermCtx.empty, typ: TypCtx.empty};
+let extend_term = (ctx: t, x: Token.t, term_entry): t => {
```
cyrus-, line 14: why Token.t in some places and Id.t in others?

(In either case, don't we need to also have the unique ID in the domain of the map with the approach we have planned?)

disconcision (commented):
Couple questions:
1. why the separation between CtxBase and TermCtx? why not just have TermCtx?
2. why the singleton mode type in Kind? is this a stub for future use?

otherwise lgtm

src/haz3lweb/view/CtxInspector.re
```diff
@@ -3,21 +3,27 @@ open Node;
...
+    : Node.t =>
   div(
     ~attr=clss(["context-entry"]),
     [text(name), text(":"), Type.view(typ)],
   );
 
 let ctxc = "context-entries";
 
-let exp_ctx_view = (ctx: Haz3lcore.Ctx.t): Node.t =>
+let exp_ctx_view = (ctx: Haz3lcore.TermCtx.t): Node.t =>
```
cyrus-, line 20: still concerned about not having a single context if we want to view a context in order, including type variables / aliases that are in scope. remind me again why it's beneficial to maintain them separately?

src/haz3lweb/view/CtxInspector.re
```diff
@@ -3,21 +3,27 @@ open Node;
...
   div(
```
cyrus-, line 21: something is broken with either the context itself or the view code -- entries are showing up doubled in the context inspector

cyrus- (changes requested):

src/haz3lcore/statics/Ctx.re
LighghtEeloo: I see, changing it to Expr

src/haz3lcore/statics/TypCtx.re
LighghtEeloo: It's just a stub, will be moved out as a separate module and actually complete it.

src/haz3lcore/statics/Ctx.re
LighghtEeloo: Our previous `Ctx.t` is a `VarMap.t_`, which takes `Token.t`s as keys. It's a map from `Token.t` to `{ id: Id.t, typ: Typ.t }`. This is why my rewrite tends to remain the same. I don't know if I'm missing anything... Afaik, `Ctx.t` is a map from `Token.t` to `Id.t` and it's `Typ.t` / `Kind.t`. If we want to lookup by id probably we should just look at `Statics.map` instead?

src/haz3lweb/view/CtxInspector.re
LighghtEeloo: 1. It's easier to debug and maintain because one lookup results in only one sort of term.
2. It separates the domain of expression variable names from the type variable names, which makes things easier to track.
3. (newly mentioned) It might not be a bad idea to display two different sorts of contexts separately.

I understand it's concerning when we lose information. If the reasons above don't seem to be convincing, I've drafted a mixed context and we can switch to that too.

src/haz3lweb/view/CtxInspector.re
LighghtEeloo: Looking into it...

AlienKevin:
@cyrus- We created a new mixed ctx PR to replace context separation.

PR #774, opened 2022-09-21
basic exercise editing with export to ML module file

cyrus-:
changed exercise format to use Zipper.t instead of strings, and creat…ed exercise export functionality for instructors

benefits: 
* students can't find the raw strings containing solutions in the .js file anymore
* you can create a fresh exercise using BlankTemplate.ml, create the exercise content in Hazel UI, then export it back to a .ml file (using .ml rather than .re because it is built on ppx_deriving.show to serialize the exercise spec, which does not seem to have a reason mode?)
* if Zipper.t changes there will be a build error

downsides:
* if Zipper.t changes there will be a build error, which might be a tricky refactor. not sure what the best way to deal with this is, but probably we want the ppx that compile-time parses strings ultimately. until then, maybe have to version changes to Zipper.t and create an explicit transformation from previous to new?


cyrus-:
@dmoon1221 you may want to read the PR notes here as it pertains to changes to Zipper.t

PR #775, opened 2022-09-21
Fix listlit element evaluation

cyrus-:


src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -710,20 +710,11 @@ let rec evaluate: (ClosureEnvironment.t, DHExp.t) => m(EvaluatorResult.t) =
...
-          },
-          lst,
-          return([]),
-        );
-      BoxedValue(ListLit(x1, x2, x3, ty, evaluated_lst));
+      let+ lst =
+        lst
+        |> List.map(el => evaluate(env, el) >>| EvaluatorResult.unbox)
+        |> sequence;
+      BoxedValue(ListLit(x1, x2, x3, ty, lst));
```
cyrus-: If any element of the list is Indet, then the whole list should be indet. So not always a boxed value...

mirryi: Oops. Do we want to stop evaluation of elements when we hit an indet one, or just do everything regardless?

cyrus-: we evaluate everything but its only a BoxedValue if all of the arguments are BoxedValues (similar to tuples)

PR #776, opened 2022-09-23
Update opam init command for WSL and opam version

lxguan1:
Update the opam init command for Windows, and update the opam version to install to 4.14.0


cyrus- (approved):

cyrus-:
Thanks!

PR #857, opened 2022-09-23
DHExp -> UExp

deerHDE:


cyrus-:
@deerHDE when you use a new ID, you have to generate it from the ID gen and return the _new_ ID generator state. here, you are using the state of the ID gen as the next ID but not returning an incremented state.

cyrus-:
closing, work on this moved to #931

PR #858, opened 2022-09-23
Haz3l school module separation and serde

LighghtEeloo:
Separating all non-web-related but school-related stuff from web into library `haz3lschool` and provide offline serde utilities that are independent of js.

FYI serde means: https://serde.rs/


cyrus-:
I think the haz3lschooldata library wasn't pushed btw

LighghtEeloo:
Not actually a school data separation but all data under "school" field in the exported json file has gained native access.

Run
```
dune exec ./src/haz3lweb/serde.exe <path-to-school-export-file>
```
to observe the output json.

Can I have a few more clarification on what type of data do we need to hand to the gradescope's side? Maybe just print the student's code in human-readable form?

Btw, I used functor instead of generics to abstract the `Node.t`. Ends up well I guess?

cyrus-:
@LighghtEeloo We need to show:

- The student's tests
- The student's implementation
- An autograder report

The default grade that gradescope gets should be the autograder grade, then the graders can manually apply deductions from there. 

This is similar to how we did Learn OCaml upload in previous semesters, so take a look at that code. There is a Gradescope API for doing this that we use I believe.

LighghtEeloo:
Deprecated. See [haz3l-school-separation](https://github.com/hazelgrove/hazel/pull/892) instead.

PR #863, opened 2022-09-25
Built-in ADTs

disconcision:
This adds hardcoded algebraic data types. Types can be added in BuiltinADTs.re. Type names AND constructor names must be globally unique. This is enforced by startup assertions. Types may be (mutually) recursive.

This introduces basic type variables, intended to be supplanted by type aliases. These are of the form Var(string). These are globally scoped, and type equality is nominative, not structural, although this doesn't currently matter because of the enforced uniqueness constraints.

Constructors (Tags) are essentially a new kind of atomic literal. Constructors may be nullary, in which case their type is simply Var("name"), or unary, in which case their type is Arrow(something, Var("name")), where 'something' may be a tuple, to achieve n-ary constructors.

Given the above type assignment, constructors/tags are treated basically the same was as function literals by the static semantics. Both function and constructor application pass through the same Ap form. Constructors are first-class and may be used any way functions can. The dynamic semantics are essentially the same as for tuples.

This implementation is fully functional AFAICT. I've written a basic LC interpreter, which doesn't quite work due to casting issues, but these should (probably, hopefully) be resolved separately in the forthcoming casting branch


src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -119,6 +119,16 @@ let rec matches = (dp: DHPat.t, d: DHExp.t): match_result =>
...
   | (StringLit(_), Cast(d, Unknown(_), String)) => matches(dp, d)
   | (StringLit(_), _) => DoesNotMatch
+  | (Tag(n1), Tag(n2)) =>
+    if (n1 == n2) {
+      Matches(Environment.empty);
+    } else {
+      DoesNotMatch;
+    }
+  //TODO(andrew): below cases?
+  //| (Tag(_), Cast(d, ty, Unknown(_))) => matches(dp, d)
```
cyrus-: commented out code looks correct -- can strip off casts on tags

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -130,7 +140,8 @@ let rec matches = (dp: DHPat.t, d: DHExp.t): match_result =>
   | (Inj(_, _), Cast(d, Sum(_, _), Unknown(_))) => matches(dp, d)
   | (Inj(_, _), Cast(d, Unknown(_), Sum(_, _))) => matches(dp, d)
   | (Inj(_, _), _) => DoesNotMatch
-  | (Pair(dp1, dp2), Pair(d1, d2)) =>
+  | (Pair(dp1, dp2), Pair(d1, d2))
+  | (Ap(dp1, dp2), Ap(d1, d2)) =>
```
cyrus-, line 142: we probably want to special case Ap with a Tag immediately in the dp1 position as distinct from one with a var or other pattern there, which is typically not allowed (matching on arbitrary tags)

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -564,7 +578,14 @@ let rec evaluate: (ClosureEnvironment.t, DHExp.t) => m(EvaluatorResult.t) =
...
       switch (r1) {
       | BoxedValue(TestLit(id)) => evaluate_test(env, id, d2)
-
+      | BoxedValue(Tag(_)) =>
+        let* r2 = evaluate(env, d2);
+        switch (r2) {
+        | BoxedValue(d2)
+        | Indet(d2) =>
+          //TODO(andrew): finish this. do we need a seperate value form?
+          BoxedValue(Ap(d1, d2)) |> return
```
cyrus-: should be Indet if d2 is Indet

src/haz3lcore/lang/Form.re
```diff
@@ -117,7 +119,8 @@ let whitespace = [Whitespace.space, Whitespace.linebreak];
    priority for forms with overlapping regexps */
 let atomic_forms: list((string, (string => bool, list(Mold.t)))) = [
   ("bad_lit", (is_bad_lit, [mk_op(Any, [])])),
-  ("var", (is_var, [mk_op(Exp, []), mk_op(Pat, [])])),
+  ("var", (is_var, [mk_op(Exp, []), mk_op(Pat, []), mk_op(Typ, [])])),
+  ("ctr", (is_constructor, [mk_op(Exp, []), mk_op(Pat, [])])),
```
cyrus-: is a constructor different from a tag?

src/haz3lcore/statics/BuiltinADTs.re
```diff
@@ -0,0 +1,65 @@
...
+};
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type adt = (string, list(tag));
+
+/* Add Built-In ADTs here. Type names and constructor names
+   must be globally unique (enforced by assertions below) */
+let adts: list(adt) = [
+  (
+    "color",
```
cyrus-: let's use capitalized names for all types (not just the primitives)

src/haz3lcore/statics/Statics.re
```diff
@@ -168,6 +168,15 @@ let typ_after_fix = (mode: Typ.mode, self: Typ.self): Typ.t =>
   | NotInHole(AnaInternalInconsistent(ty_ana, _)) => ty_ana
   };
 
+/* Type of a type annotation; set to unknown if it's
+   a free type variable. */
+let typ_ann_after_fix = (self: Typ.self): Typ.t =>
```
cyrus-: this is kind of a hack? probably want to add UnboundVar to UTyp and have it behave like Unknown semantically (but errors still report the identifier inside)

src/haz3lcore/statics/Typ.re
```diff
@@ -117,6 +118,9 @@ let rec join = (ty1: t, ty2: t): option(t) =>
     | None => None
     }
   | (List(_), _) => None
+  | (Var(n1), Var(n2)) when n1 == n2 => Some(ty1)
+  //TODO: structural comparison?
```
cyrus-: not sure what structural comparison would mean here for built in ADTs

cyrus- (changes requested):
Looks reasonable and straightforward, see a few comments but should be suitable to merge soon.

disconcision:
Leaving the issue of permissible tag positions aside until we do exhaustiveness. I agree that we'll likely want to restrict these in the future, to prohibit weird looking things if nothing else; on reflection it's not obvious to me that leaving it as-is actually complicates exhaustiveness checking; the general implementation i'm imagining would i think be ecumenical here

PR #864, opened 2022-09-26
haz3l: Restore builtins

### LINKED ISSUE #780: Builtin Functions

mirryi:
Closes #780

### DIFF
--- a/src/haz3lcore/dynamics/Builtin.re
+++ b/src/haz3lcore/dynamics/Builtin.re
@@ -17,7 +17,7 @@ type builtin_evaluate =
 
 [@deriving (show({with_path: false}), sexp, yojson)]
 type t = {
-  ty: Typ.t,
+  typ: Typ.t,
   eval: builtin_evaluate,
   elab: DHExp.t,
 };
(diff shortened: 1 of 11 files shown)


mirryi:
@cyrus- I've cleaned up the `Builtins` API, and changed the API of evaluator to thread a map of builtin forms throughout evaluation. But this API change (along with changing `Statics.mk_map` to take an initial typing context) has to propogate throughout the codebase to stuff like `Interface`, then across anywhere that uses `Interface`. Not sure what the correct design decisions are here, especially with the exercise stuff.

Also, when web worker stuff returns, this design means we'll have to pass builtins across to the worker thread every single time. Not sure if there's a better way to do this (functor maybe?)

cyrus-:
For now, let's just have a global set of builtins to avoid some of those issues.

cyrus-:
Can you get that done tonight/tomorrow morning? Need this to get a parser integrated into the assignment...

cyrus-:
For web worker stuff, maybe the thing to do would be to "mask in" + "mask out" builtins that we do / don't want to be available in a given evaluation, chosen from the global list of all possible builtins.

mirryi:
@cyrus- Implemented the global builtins. Just add whatever is needed to `Builtins.Pervasives` for now. I'm not sure if it's working entirely because the page isn't loading for some reason... Not feeling too great this evening so probably won't be able to try to debug any more tonight.

cyrus-:
Looks to be working, thanks!

PR #868, opened 2022-09-29
Haz3l improved casts

disconcision:
still buggy...


disconcision:
@cyrus still a bug here... do you see what's being miscast?
Screen Shot 2022-09-28 at 11 30 10 PM

disconcision:
Also we probably need some more tests for this... any other ideas for what to try?

cyrus-:
On the last line of the elaboration, missing a cast to ? on the "x + 1"
operation, and each element of the list literal should also have a cast to
? because they are all in analytic position against ?. The function
argument as a whole should also be cast to ?.

The function body seems right if I'm reading those casts correctly...

cyrus-:
Also the type in the fixpoint annotation seems wrong, but the casts seem to
correctly cast to ?.

disconcision:
just adding a list of minified erroring examples to check against later:
Screen Shot 2022-10-04 at 1 33 58 PM
Screen Shot 2022-10-04 at 1 33 52 PM
Screen Shot 2022-10-04 at 1 33 48 PM
Screen Shot 2022-10-04 at 1 33 37 PM
Screen Shot 2022-10-04 at 1 33 27 PM

disconcision:
still some problematic cases:
Screen Shot 2022-10-12 at 5 50 11 PM
Screen Shot 2022-10-12 at 5 50 04 PM
Screen Shot 2022-10-12 at 5 49 27 PM

disconcision:
@cyrus still some weird stuff going on here, see above

cyrus-:
First case seems to not be doing synpatternvar correctly anymore -- shouldn't need a cast there since x and y are both synthetic.

cyrus-:
third case... how did there end up being a cast from a list type on the function? that's very odd. can you show the unevaluated elaboration?

cyrus-:
second case seems to be missing casts to ? on the arguments of the function ap at the bottom (and maybe some others in the function, but at least that)

disconcision:
@cyrus- minified version of the remaining casting bug:
Screen Shot 2022-10-26 at 5 16 37 PM

the problem's immediate cause seems to be the highlighted case we added here:
Screen Shot 2022-10-26 at 5 16 49 PM

tonyfettes:
I just discovered that I can bypass the type-checking of if expression with a handcrafted fix point
image
I'm not sure if this is the ideal behaviour, and I'm also not totally clear what goes wrong here, but after some digging, I suspect putting cast after almost everything (in this example) is not a wise decision? See
https://github.com/hazelgrove/hazel/blob/haz3l-improved-casts/src/haz3lcore/dynamics/elaborator.re#L83

The dump of problematic casts.
image

cyrus-:
> @cyrus- minified version of the remaining casting bug: Screen Shot 2022-10-26 at 5 16 37 PM
> 
> the problem's immediate cause seems to be the highlighted case we added here: Screen Shot 2022-10-26 at 5 16 49 PM

I'm not sure why the function inside the fixpoint shows a cast to `? -> ?` since it is already of that type? That might be causing some issues with the ap?

cyrus-:
I don't see how that issue would cause the exception you're seeing though. 

The result of `g 2<Int => ?>` should be `[]<[?] => ?>` but then that should be cast to `[?]` which as indicated there appears in the tail position of the cons correctly. This should be a successful cast leaving just `[]`... might need to investigate which of those steps isn't happening correctly.

cyrus-:
> I just discovered that I can bypass the type-checking of if expression with a handcrafted fix point image I'm not sure if this is the ideal behaviour, and I'm also not totally clear what goes wrong here, but after some digging, I suspect putting cast after almost everything (in this example) is not a wise decision? See https://github.com/hazelgrove/hazel/blob/haz3l-improved-casts/src/haz3lcore/dynamics/elaborator.re#L83
> 
> The dump of problematic casts. image

@tonyfettes (you can make a separate issue for this kind of thing, not sure its related to this PR.)

tonyfettes:
> > I just discovered that I can bypass the type-checking of if expression with a handcrafted fix point image I'm not sure if this is the ideal behaviour, and I'm also not totally clear what goes wrong here, but after some digging, I suspect putting cast after almost everything (in this example) is not a wise decision? See https://github.com/hazelgrove/hazel/blob/haz3l-improved-casts/src/haz3lcore/dynamics/elaborator.re#L83
> > The dump of problematic casts. image
> 
> @tonyfettes (you can make a separate issue for this kind of thing, not sure its related to this PR.)

This behaviour only shows in this PR, not in haz3l-tests, so I thought it's better to post here. I could also open an issue, but I'm a bit worried about confusing others who don't keep track of this PR.

cyrus-:
Ah ok if it's only on this pr then that's fine, sorry about that. @disconcision can you take a look

disconcision:
> I'm not sure why the function inside the fixpoint shows a cast to ? -> ? since it is already of that type?

@cyrus- : isn't the function of type ? -> [?]

cyrus-:
> > I'm not sure why the function inside the fixpoint shows a cast to ? -> ? since it is already of that type?
> 
> @cyrus- : isn't the function of type ? -> [?]

It's in analytic position, and the elaboration itself shows a cast on `[]` to `?`.

disconcision:
@tonyfettes, @cyrus- : statics isn't affected by casting, so that's orthogonal. the difference is likely with the new mode for function position (previously function position was analytic against ?->?). looks like currently that function body is ending up in analytic against ?, hence no error holes being drawn.

here's a minified version with the same essential statics:
2022-10-27 11_41_16-hazel

which i believe is functioning as-expected? the argument is analyzed against ?, producing the observed behavior

disconcision:
> It's in analytic position, and the elaboration itself shows a cast on [] to ?.

but the types which cast insertion is based on are the types from the statics, not the types after casting

tonyfettes:
@disconcision That makes sense, and thank you for the explanation.

cyrus-:
> > It's in analytic position, and the elaboration itself shows a cast on [] to ?.
> 
> but the types which cast insertion is based on are the types from the statics, not the types after casting

Yes, but its in analytic position. I think the commented out `Fun(_), _` case is maybe the culprit here -- we do want that case because Fun already knows how to propagate the expected type into the body by matching?

cyrus-:
Did that fix it?

disconcision:
Screen Shot 2022-10-28 at 1 02 55 PM
@cyrus adding back the fun case prevents the previous suggestion. still a bit foggy on the justification. There are still problematic listlit cases though, such as above. here the actual type of xs is ?, due to the annotation, and the expected type is [?], so it gets a [?]=>? cast, triggering the InvalidBoxedListLit exception. or is this an issue with the cons casting/statics?

cyrus-:
I think the issue there is that the case shouldn't have a cast on it, since its in analytic position. I think we need that special case in the `wrap` function for anything that has a special analytic case?

disconcision:
okay, no casts on case. i think that exhausts our special analytic cases? 
still getting this guy (invalid boxed bool lit from being cast to unknown typehole)
Screen Shot 2022-10-28 at 3 00 46 PM

cyrus-:
The list lit `[true]` shouldn't need a cast to `[?]` on it since the `true`
is cast to `?` -- I guess thats another analytic special case since the
element types propagate down.

disconcision:
@cyrus this is ready for another look. I'm currently not aware of any more examples that throw exceptions. We need to comment the wrap function better, explain which cases are cast. I refactored and clarified the wrap logic somewhat, but the only substantive change is to the listlit case, pursuant to the above

disconcision:
made the changes we talked about... couple things... first, the new tuple rule is causing errors again:

Screen Shot 2022-11-08 at 7 13 50 PM
Screen Shot 2022-11-08 at 7 13 34 PM

second, you mentioned fix as a form that needs a special rule... but what should it be?

cyrus-:
@disconcision There is a small merge conflict that needs to be resolved now in the Let case of the elaborator due to the recently merged PR adding function name metadata

src/haz3lcore/dynamics/DH.re
```diff
@@ -225,6 +225,7 @@ module rec DHExp: {
     | (ListLit(_, _, _, _, []), List(_)) =>
```
cyrus-: this may no longer be necessary?

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -3,6 +3,9 @@ open EvaluatorMonad;
 open EvaluatorMonad.Syntax;
 open EvaluatorResult;
 
+open Sexplib.Std;
+[@deriving sexp]
+type blah = list(list((Typ.t, Typ.t)));
```
cyrus-: pretty blah name for a type. is it used somewhere?

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -51,222 +51,237 @@ let exp_binop_of: Term.UExp.op_bin => (Typ.t, (_, _) => DHExp.t) =
...
 
+/* Wrap: Handles cast insertion and non-empty-hole wrapping
+   for elaborated expressions */
+let wrap = (u, mode, self, d: DHExp.t): option(DHExp.t) =>
+  switch (Statics.error_status(mode, self)) {
+  | NotInHole(_) =>
+    switch (mode) {
+    | Syn => Some(d)
+    | SynFun =>
+      /* Things in function position get cast to an arrow type */
```
cyrus-: nit: get cast to the matched arrow type. maybe make the `_` case an Arrow case and assert in the fall through case that it should never happen in this and the cases below

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -51,222 +51,237 @@ let exp_binop_of: Term.UExp.op_bin => (Typ.t, (_, _) => DHExp.t) =
...
+          Some(DHExp.cast(d, List(Unknown(prov)), ana_ty))
+        | _ => Some(d)
+        }
+      | Fun(_) =>
+        switch (ana_ty) {
+        | Unknown(prov) =>
+          Some(DHExp.cast(d, Arrow(Unknown(prov), Unknown(prov)), ana_ty))
+        | _ => Some(d)
+        }
+      /*| Tuple(_) =>
```
cyrus-: can delete

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -51,222 +51,237 @@ let exp_binop_of: Term.UExp.op_bin => (Typ.t, (_, _) => DHExp.t) =
...
+          Some(DHExp.cast(d, Prod(us), ana_ty));
+        | _ => Some(d)
+        }
+      | Inj(_) =>
+        switch (ana_ty) {
+        | Unknown(prov) =>
+          Some(DHExp.cast(d, Sum(Unknown(prov), Unknown(prov)), ana_ty))
+        | _ => Some(d)
+        }
+      | FixF(_) =>
```
cyrus-: I don't think this case ever arises but in principle, it should behave like Let in that the expected type is passed down in the statics so I'd put it down in the next group of cases

src/haz3lcore/statics/Statics.re
```diff
@@ -106,8 +107,28 @@ type error_status =
...
   switch (mode, self) {
-  | (Syn | Ana(_), Free(free_error)) => InHole(Free(free_error))
-  | (Syn | Ana(_), Multi) => NotInHole(SynConsistent(Unknown(Internal)))
+  | (SynFun, Just(ty)) =>
+    switch (Typ.join(Arrow(Unknown(Internal), Unknown(Internal)), ty)) {
+    | None => InHole(NoFun(ty))
+    | Some(_) => NotInHole(SynConsistent(ty))
+    }
+  | (SynFun, Joined(_wrap, tys_syn)) =>
+    /*| (Ana(Unknown(SynSwitch)), Joined(tys_syn))*/
```
cyrus-: can be removed

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -206,7 +206,12 @@ let rec mk =
         | InconsistentBranches(u, i, Case(dscrut, drs, _)) =>
           go_case(dscrut, drs)
           |> annot(DHAnnot.InconsistentBranches((u, i)))
-        | _ => raise(EvaluatorPost.Exception(PostprocessedNonHoleInClosure))
+        | _ =>
+          //raise(EvaluatorPost.Exception(PostprocessedNonHoleInClosure))
+          print_endline("DHDoc_exp EXCEPTION 1:");
+          print_endline(Sexplib.Sexp.to_string_hum(DHExp.sexp_of_t(d)));
+          DHDoc_common.mk_StringLit("Help:PostprocessedNonHoleInClosure");
+        //TODO(andrew): reinstate
```
cyrus-: not sure what this means

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -217,7 +222,13 @@ let rec mk =
       | FreeVar(_)
       | InvalidText(_)
       | InconsistentBranches(_) =>
-        raise(EvaluatorPost.Exception(PostprocessedHoleOutsideClosure))
+        //TODO(andrew): reinstate
```
cyrus-: is this still a problem?

cyrus- (changes requested):
just a few more cleanup / minor issues

cyrus- (approved):

PR #869, opened 2022-09-29
Haz3l mutual recursion

cyrus-:
Goal: 
```
let (even : Int -> Bool, odd : Int -> Bool).= (
  fun n -> 
    if n == 0 then true else odd(n - 1), 
  fun n -> 
    if n == 0 then false else even(n - 1)) in 
even(4)
```

Approach is to elaborate to a fixpoint over a tuple with projections out of the self reference. To do that, modifying internal pairs to be tuples and adding tuple projection to the internal language.


cyrus-:
Screen Shot 2022-09-29 at 9 27 06 PM

PR #892, opened 2022-10-03
Haz3l school separation

LighghtEeloo:


cyrus-:
There are some conflicts here -- can you do a merge and cleanup pass after you get the A1 grades uploaded?

LighghtEeloo:
Attempted 10 mins during break. Worked out easier than thought.

src/haz3lschool/Gradescope.re
```diff
@@ -0,0 +1,126 @@
+open Haz3lcore;
```
cyrus-, line 1: Can you update the INSTRUCTORS.md file in the root directory with instructions for how Gradescope export works

src/haz3lschool/specs/Ex_OddlyRecursive_grading.ml
```diff
@@ -0,0 +1,2948 @@
+module SchoolExercise = GradePrelude.SchoolExercise
```
cyrus-: remind me why we need a separate grading module? just because of prompt?

cyrus- (changes requested):

src/haz3lschool/specs/Ex_OddlyRecursive_grading.ml
LighghtEeloo: Because of web dependency. We also want a command line executable to run on students' uploaded editor state.

src/haz3lschool/Gradescope.re
LighghtEeloo: Done

src/haz3lschool/specs/Ex_OddlyRecursive_grading.ml
LighghtEeloo: I'm not sure whether there is a scalable way to grade students' editor states in browser, the IO sounds troublesome. But to run the grader in command line, we need the separation to remove web dependency.

cyrus- (approved):

PR #893, opened 2022-10-03
System clipboard copy/paste

disconcision:
Copy/Paste via system clipboard (For Ctrl-C/Ctrl-V, I'm not actually capturing clipboard events)


disconcision:
After banging my head on this all day it seems to have ended up super clean.... on chrome anyway. Paste doesn't work on firefox... trying to figure out a workaround

disconcision:
Worth a run-through tho @cyrus to check if there's other issues i missed

disconcision:
k actually it works in firefox via the following workaround:

1. go to about:config
2. search for dom.events.asyncClipboard.readText
3. set it to true
4. refresh hazel if open

good enough for government-funded work imo

cyrus-:
So looked around online for a bit and it seems like there is no general workaround for the Firefox permission issue if you want arbitrary clipboard access, but when its specifically to copy/paste text on the page, you can create a hidden text area  and use the clipboard events there to access the clipboard contents when the user presses the appropriate keys.

Since we only need it for paste and not cut/copy I think this should actually be pretty easy -- don't need to keep the text area contents / selection in sync with the code itself, just clear it out after every paste...

cyrus-:
I'm getting the following warning when building this branch (after doing a `make deps`):

```
Warning 58 [no-cmx-file]: no cmx file was found in path for module Promise, and its interface was not compiled with -opaque
```

cyrus-:
Also this approach doesn't work at all when I run it via `make serve`, since that isn't an https server. There is some way to generate a local ceritificate and serve localhost via https I think, but that's getting complicated.

cyrus-:
#896 is an alternative implementation using execCommand that seems to work cross-browser. can you take a look @disconcision ?

cyrus-:
closing in favor of #896

PR #894, opened 2022-10-04
Haz3l mixed ctx

AlienKevin:
Added mixed Ctx for storing both term and type variables in context.


cyrus-:
Entries in the context are doubled up: 
Screen Shot 2022-10-10 at 2 08 07 PM

LighghtEeloo:
This happens because we are preserving all bindings in our context. We've just fixed it by adding a `filter_duplicate` function and call it before showing it in our context inspector (d09c683).

LighghtEeloo:
The latest commit (786b219) fixes the duplicate bindings in a recursive let definition.

src/haz3lcore/statics/Ctx.re
```diff
@@ -23,10 +28,25 @@ type co = VarMap.t_(co_entry);
 
 let empty = VarMap.empty;
 
-let subtract = (ctx: t, free: co): co =>
+let lookup_typ = (ctx: t, x) =>
```
cyrus-: rename to lookup_var (and lookup_tvar)

src/haz3lcore/statics/Kind.re
```diff
@@ -0,0 +1,3 @@
+[@deriving (show({with_path: false}), sexp, yojson)]
+type t =
+  | Singleton;
```
LighghtEeloo: Change to Typ

src/haz3lcore/statics/Ctx.re
```diff
@@ -1,9 +1,14 @@
...
+[@deriving (show({with_path: false}), sexp, yojson)]
+type value =
+  | Typ(Typ.t)
+  | Kind(Kind.t);
+
 [@deriving (show({with_path: false}), sexp, yojson)]
 type entry = {
   id: Id.t,
-  typ: Typ.t,
+  value,
```
cyrus-: create separate Var and TVar types and have the context be a list of entries, which can be either Var or TVar entries

src/haz3lcore/statics/Ctx.re
```diff
@@ -1,9 +1,14 @@
 open Sexplib.Std;
 
+[@deriving (show({with_path: false}), sexp, yojson)]
+type value =
+  | Typ(Typ.t)
+  | Kind(Kind.t);
+
 [@deriving (show({with_path: false}), sexp, yojson)]
 type entry = {
```
LighghtEeloo: Use LIst.t(entry)

src/haz3lcore/statics/Kind.re
cyrus-: rename to Type

src/haz3lweb/view/CtxInspector.re
```diff
@@ -12,22 +12,33 @@ let context_entry_view =
           inject(UpdateAction.PerformAction(JumpToId(id)))
         ),
       ]),
-    [text(name), text(":"), Type.view(typ)],
+    [
+      text(name),
```
cyrus-: have it say "type A" for type entries

cyrus- (changes requested):

AlienKevin:
@cyrus- All requested changes are made. We tested the mixed ctx in the Hazel editor and the context looks fine. We think it's ready to merge.

src/haz3lcore/statics/Statics.re
```diff
@@ -540,7 +540,7 @@ and upat_to_info_map =
     let typ = typ_after_fix(mode, self);
     add(
       ~self,
-      ~ctx=VarMap.extend(ctx, (name, {id: Term.UPat.rep_id(upat), typ})),
+      ~ctx=[VarEntry({name, id: Term.UPat.rep_id(upat), typ}), ...ctx],
```
cyrus-: define a Ctx.extend function instead of relying on the implementation type

cyrus- (changes requested):
should be ready to go after the small change suggested

cyrus- (approved):

PR #896, opened 2022-10-06
Sys copy paste - exec command version

### LINKED ISSUE #704: Clipboard Copy/Paste

cyrus-:
Alternative to #893 that uses execCommand, which is a deprecated API that nevertheless works cross-browser and without the security restrictions of the clipboard API.

Had to use a hidden textArea.

Limitations at the moment: 
* Had to disable the Ctrl+K workaround because it wants to invoke Copy/Paste using non-keyboard actions (there is a way to do this using execCommand though, if we want to preserve this -- but also different ways to implement it not using Copy/paste and hopefully we won't need Ctrl+K at all?)
* Copy/paste does not preserve / handle indentation elegantly.

PR also needs some clean-up before merging (some holdovers because I branched off of #893, don't need promises API anymore)


disconcision:
seems to work! not copying indentation was a considered decision, as if indentation is copied then pasting within hazel will have a bunch of added whitespace. could fix this by preprocessing before paste i guess (eliminating all leading whitespace per line)

cyrus-:
Yeah -- I think near-term one of the uses cases for this is for people who find Hazel too slow, they can edit in a text editor and paste back when they are done. So indentation support would be nice for that.

We could preprocess by using `\t` for automatically inserted indentation, so that manually inserted leading spaces on a line are preserved.

disconcision:
copy preserves indentation now and paste strips leading whitespace, resulting in standard indentation

PR #899, opened 2022-10-07
Haz3l fun name metadata

yujiagao-audrey:
Add optional name field for Fun constructor.


src/haz3lcore/dynamics/elaborator.re
```diff
@@ -199,6 +199,11 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
         /* not recursive */
         let* dp = dhpat_of_upat(m, p);
         let* ddef = dhexp_of_uexp(m, def);
+        let ddef =
+          switch (ddef) {
+          | Fun(a, b, c, _) => DHExp.Fun(a, b, c, Term.UPat.get_fun_name(p))
+          | _ => ddef
```
cyrus-, line 202: need to handle the case of a function in Parens here as well

yujiagao-audrey: image

I think it handled the case of a function in Parens now.

src/haz3lcore/statics/Term.re
```diff
@@ -288,6 +288,33 @@ module UPat = {
...
   };
 
+  let rec get_fun_name = (pat: t) => {
+    switch (pat.term) {
+    | Parens(pat) => get_fun_name(pat)
+    | Var(x) => Some(x)
+    | TypeAnn(pat, typ) =>
+      if (UTyp.is_arrow(typ)) {
+        get_fun_name(pat);
+      } else {
```
cyrus-: we don't actually need to check if it is annotated with an arrow type in the non-recursive case, so this can be removed in favor of a get_var function that just ignores any type annotations entirely

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -382,7 +382,10 @@ let rec mk =
             DHDoc_common.Delim.close_Fun,
           ]);
         } else {
-          annot(DHAnnot.Collapsed, text("<fn>"));
+          switch (s) {
+          | None => annot(DHAnnot.Collapsed, text("<fn>"))
```
cyrus-: this is ambiguous if we have a function named `fn`, let's use `<anon fn>` instead

cyrus- (changes requested):
Looks great, just needs a couple of small fixes and then we can merge.

src/haz3lcore/statics/Term.re
yujiagao-audrey: I just deleted the get_fun_name and used get_var instead.

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
yujiagao-audrey: Sure. I have changed to use <anon fn> instead.

cyrus- (approved):

PR #901, opened 2022-10-08
disable postprocessing to fix performance issues

cyrus-:
with large deeply nested indet cases

causing significant problems rn in the 490 release when you end up in situations like this

Screen Shot 2022-10-08 at 5 04 29 PM

Don't quite understand what is causing the issue here so disabling for now. We also don't use hole closures in the current UI.


cyrus-:
issue seems to be in the "track_children" phase

PR #902, opened 2022-10-10
Haz3l small step

tonyfettes:
Implement/Port small step evaluator to Haz3l.

## Progress update

image

## Implementation

Instead of using indices to locate the mark, we store the whole context into annotation, and can be composed directly when the user clicks.

We use substitution to achieve evaluation of let-expression and plan to use it on function applications, since environment would let the user to click on each site of a bound variable for it to get substituted.

## Known Bugs

- [x] Z combinator doesn't work. Works in `dev`, might due to cast issue.
- [x] #949 

## Roadmap

- [x] Clickable subexpression
- [x] Basic arithmetics and let expressions.
- [x] Use a standalone `DHDoc_Step` module to annotate steps.
- [x] "Previous Step" button.
- [x] History (record in legacy code)
- [x] Add description for each step. (We have tooltips for now)
- [x] Populate `EvalCtx` so that more expression can be decomposed.
  - [x] Projection
  - [x] List Literal (`[1, 2, ...]`)
  - [x] BinStringOp
- [x] Automatic evaluation of Builtin variables and functions.

## Future work

- "Next Step" button. Sometimes the history is long and we accidentally step back too far. (Move to future work since this can be more "correctly" handled by incremental step evaluation)
- Bring back evaluator options and add more to it. Blocked by #985
- Show environment alongside the expression.

## Similar work

- [A stepper for untyped lambda calculus](https://lambdacalc.io/)
  - Shows transformation application on the right hand side.
- [Another stepper for untyped lambda calculus](https://www.cl.cam.ac.uk/~rmk35/lambda_calculus/lambda_calculus.html#)
  - Nice highlight on mouse hovering.


tonyfettes:
Move to the top comment.

cyrus-:
closing in favor of #1006

PR #916, opened 2022-10-16
Go to definition

7h3kk1d:
Addresses https://github.com/hazelgrove/hazel/issues/811
---

There's probably a cleaner way to handle the mapping; I wasn't sure if there's a better way to get the statics/term. 

In a separate PR I'd like to add ctrl or alt-click to go to definition.

### DIFF
--- a/src/haz3lcore/zipper/action/Action.re
+++ b/src/haz3lcore/zipper/action/Action.re
@@ -19,7 +19,8 @@ type t =
   | RotateBackpack
   | MoveToBackpackTarget(planar)
   | Pick_up
-  | Put_down;
+  | Put_down
+  | GoToDefinition;
 
 module Failure = {
   [@deriving (show({with_path: false}), sexp, yojson)]
@@ -28,7 +29,8 @@ module Failure = {
     | Cant_insert
     | Cant_destruct
     | Cant_select
-    | Cant_put_down;
+    | Cant_put_down
+    | Cant_go_to_definition;
 };
 
 module Result = {
--- a/src/haz3lcore/zipper/action/Action.re
+++ b/src/haz3lcore/zipper/action/Action.re
@@ -16,7 +16,6 @@ type jump_target =
 [@deriving (show({with_path: false}), sexp, yojson)]
 type t =
   | Move(move)
-  | JumpToId(Id.t)
   | Jump(jump_target)
   | Select(move)
   | Unselect
--- a/src/haz3lcore/zipper/action/Action.re
+++ b/src/haz3lcore/zipper/action/Action.re
@@ -9,9 +9,15 @@ type move =
   | Goal(Measured.Point.t);
 
 [@deriving (show({with_path: false}), sexp, yojson)]
+type jump_target =
+  | TileId(Id.t)
+  | BindingSiteOfIndicatedVar;
+
+[@deriving (show({with_path: false}), sexp, yojson)]
 type t =
   | Move(move)
   | JumpToId(Id.t)
+  | Jump(jump_target)
   | Select(move)
   | Unselect
   | Destruct(Direction.t)
@@ -19,8 +25,7 @@ type t =
   | RotateBackpack
   | MoveToBackpackTarget(planar)
   | Pick_up
-  | Put_down
-  | GoToDefinition;
+  | Put_down;
 
 module Failure = {
   [@deriving (show({with_path: false}), sexp, yojson)]
@@ -29,8 +34,7 @@ module Failure = {
     | Cant_insert
     | Cant_destruct
     | Cant_select
-    | Cant_put_down
-    | Cant_go_to_definition;
+    | Cant_put_down;
 };
 
 module Result = {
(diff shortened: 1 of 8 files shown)


src/haz3lweb/Keyboard.re
```diff
@@ -43,7 +43,7 @@ let handle_key_event = (k: Key.t, ~model: Model.t): list(Update.t) => {
     | "F3" => toggle(Log.debug_update)
     | "F4" => toggle(Log.debug_keystroke)
     | "F5" => toggle(Log.debug_zipper)
-    | "F6" => []
+    | "F6" => now(Jump(BindingSiteOfIndicatedVar)) // Figure out actual keybinding
```
7h3kk1d: Thoughts?

src/haz3lcore/zipper/action/Action.re
```diff
@@ -8,10 +8,15 @@ type move =
...
+[@deriving (show({with_path: false}), sexp, yojson)]
+type jump_target =
+  | TileId(Id.t)
+  | BindingSiteOfIndicatedVar;
+
 [@deriving (show({with_path: false}), sexp, yojson)]
 type t =
   | Move(move)
-  | JumpToId(Id.t)
+  | Jump(jump_target)
```
disconcision, line 19: Now that I've been doing some similar in Select, I'm somewhat tempted to consolidate these actions under Move.... make these Move(Id(id)), Move(ToDefinition), and make existing Move(whatever) to Move(Relative(whatever))... not quite sure how I feel about this change though. @cyrus- thoughts?

src/haz3lcore/zipper/action/Action.re
```diff
@@ -8,10 +8,15 @@ type move =
...
   | Select(move)
```
disconcision: Now that I've been doing some similar in Select, I'm somewhat tempted to consolidate these actions under Move.... make these Move(Id(id)), Move(ToDefinition), and make existing Move(whatever) to Move(Relative(whatever))... not quite sure how I feel about this change though. @cyrus- thoughts?

src/haz3lcore/zipper/action/Move.re
```diff
@@ -178,6 +178,31 @@ module Make = (M: Editor.Meta.S) => {
...
+      };
+
+    let* entry =
+      switch (info_exp.term.term) {
+      | TermBase.UExp.Var(name) => VarMap.lookup(info_exp.ctx, name)
+      | _ => None
+      };
+    jump_to_id(z, entry.id);
+  };
+
```
disconcision: having thought about this a bit more, lets refactor this:
1. I want to avoid statics dependencies in Move. for now let's consolidate these in Perform.
2. Put the maketerm/mk_map calcs at the top of go_z in perform, so the map can be (hypothetically) re-used across cases, as well as the Indicated call. I forgot earlier but don't bother to call Indicated.piece, you can just do Indicated.index to get the id directly.
3. Make a function in Statics called something like get_binding_site which takes an id and a map and returns option Id; some if id points to an infor_exp which is a variable reference which exists in its context.
4. depending on how small this makes jump_to_indicated_var maybe just inline it; your discretion

src/haz3lweb/Keyboard.re
disconcision: we tend to use vscode as our source of defaults, so let's say F12? (plain F12 not shift F12)

disconcision (changes requested):
looks functionally solid, lets make a couple org changes tho

src/haz3lcore/zipper/action/Move.re
7h3kk1d: This all makes sense to me. Just want to call out that I'm not sure how expensive `maketerm` and `mk_map` are, so I'm not sure if this will slow down things like moving up/down.

src/haz3lcore/zipper/action/Action.re
7h3kk1d: In the long term, it would make sense to think of a semantic taxonomy for the actions. Something similar to the context menu breakdown; with variants for `Navigation`, `Refactoring`, `Selection`, etc. I updated the rest of the pr and can collapse Jump/Move for now if we want though.

src/haz3lcore/zipper/action/Move.re
disconcision: great point. they are moderately expensive. kind of getting ahead of myself here, i have a plan to stick these in the model and condition recomputation on action type (and eventually fully incrementalize their calculation) but we're not there yet

so yeah please just move those three lines into the jump case and i'll replace them when i make the aforementioned change.

src/haz3lcore/zipper/action/Action.re
disconcision: sounds reasonable. just leave it for now, a separate reorg pass is probably in order in a little while.

disconcision (approved):
lgtm except for the other comment i made. actually also could you put a type signature on get_binding_site? i know we're not doing this everywhere but I want to try to be a bit more consistent about it

cyrus-:
@7h3kk1d a couple of small conflicts to resolve

cyrus-:
@7h3kk1d for future PRs, go ahead and make a branch in this repo instead of your own fork -- that way our github actions will autobuild and deploy it to hazel.org/build/<branchname> for ease of testing / demos

cyrus- (approved):

PR #917, opened 2022-10-16
Haz3l user defined operators

### LINKED ISSUE #828: User-defined operators
Previous Hazel 2 effort: https://github.com/hazelgrove/hazel/pull/464

- [x] be able to use _op_ as a variable name in patterns
    
    ```
    let _++_ = fun (left, right) -> left + right in 
    ```

- [x] be able to use _op_ as a variable in expressions
   
    ```
    let _++_ = fun (left, right) -> left + right in 
    _++_(2, 2)
    ```

- [x] be able to parse infix uses of the operator
 
    ```
    let _++_ = fun (left, right) -> left + right in 
    2 ++ 2
    ```

    parses into

    `UserBinOp(Var("++"), Int(2), Int(2))`

- [ ] errors for incorrect arity of functions bound to operators (has to be binary, uncurried)
- [ ] elaboration to function application
- [ ] display function applications with a user op function in infix form when printing a DHExp
- [ ] look into unary operators

shawn-stovall:
Enabling user-defined operators in Hazel 3 - haz3l:

Implemented using a ground-up approach that defines a new pattern form which will be used to recognize user-defined operators.

Details:

Operator precedence based on first symbol in the user-defined operator symbol.

User defined operators will have a special place in the context, for now.  Full integration to come later.

### DIFF
--- a/src/haz3lweb/LanguageRefSlide.ml
+++ b/src/haz3lweb/LanguageRefSlide.ml
@@ -1,5 +1,5 @@
 let slide : ScratchSlide.persistent_state =
-  ( 3570,
+  ( 3588,
     {
       zipper =
         "((selection((focus \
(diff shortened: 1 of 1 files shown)


shawn-stovall:
Working on the pattern now.  My first priority is to get the pattern working on tylr (Making the form appear when a user starts entering the operator).  I've added the information to TermBase.UPat as well as completed the pattern matching in Term.re for the pattern.

Next step is to work on adding the correct information to the context, but it looks like I may have to refactor slightly how I'm approaching this to get it to work as a function.

PR #920, opened 2022-10-18
Select term

### LINKED ISSUE #799: Double-click term selection support

disconcision:
There's more that could be done here so I'm trying to set scope. Right now you can either double-click or use Ctrl-D to select the term currently indicated by the caret.

1. There is one limitation in that the current TermRanges implementation considers the square brackets of list literals and the case end delimiters of cases to be distinct terms from their contents. I don't fully understand the implications of updating TermRanges so this should probably be a separate ticket.

2. I've done some thinking about cycling behaviours / different click arities to change between different kinds of selections, but haven't reached any firm conclusions. My biggest question is around semantic versus syntactic selection options. like in vscode double click selects token, then another single click selects line. i could see us doing something similar, but if we try to incorporate both syntactic and semantic options there's too many to cycle through: token, tile, term, line, and it's not obvious to me what should give. maybe line is no longer necessary given that tile will often do the same (or better). in any case, more design decisions to be made there so should be a separate ticket imo

### DIFF
--- a/src/haz3lcore/Measured.re
+++ b/src/haz3lcore/Measured.re
@@ -177,10 +177,11 @@ let find_shards' = (id: Id.t, map) =>
   | Some(ss) => ss
   };
 
-let find_w = (w: Whitespace.t, map) => Id.Map.find(w.id, map.whitespace);
-let find_g = (g: Grout.t, map) => Id.Map.find(g.id, map.grout);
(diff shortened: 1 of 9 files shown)


cyrus-:
Case rules are also awkward here, for fundamental structural reasons. Don't need to handle that with this PR but just noting it.

PR #924, opened 2022-10-26
Add pipeline operator

### LINKED ISSUE #803: Boolean or operator
Currently doesn't work due to `|` triggering case rule construction

alan123piano:
This pull request is only three lines. It adds a pipeline operator. Some things to note:

* Since the current parsing engine doesn't support invalid partial states for operators, I've added an auxillary operator '|' which creates a binary op form (with no evaluation semantics). Perhaps this can be fixed if the parsing engine is reworked, or if bitwise-or is added. If any operators beginning with '|' are added with a different precedence, then this could definitely cause some weirdness (it just so happens that '||' and '|>' share the same precedence level).
* As a side effect, the auxillary '|' operator also makes the binary op for OR ('||') possible to synthesize in the editor.
* The mold for 'Rule' has some ambiguous states with the pipeline operator '|>'. Of particular concern is that if you add a whitespace after an expression like '1 | ', then it instantly becomes a Rule as per the deferred expansion rule (without a chance to change it back into a pipeline operator).
* The evaluation for the pipeline operator is the same as function application, since it is just syntax sugar.

image


cyrus- (changes requested):
Looks good but lets update the LangDocs description to describe what |> does specifically.

cyrus-:
@alan123piano marking as draft -- can you look into the langdocs thing soon and then mark as ready for review when its good to go. feel free to reach out to @hannahpotter about the langdocs modules if you need a pointer.

alan123piano:
This works WITH the haz3l-improved-casts branch - I merged that branch into this one. I've put up some LangDocs messages as well.

cyrus- (approved):

PR #926, opened 2022-11-02
Add undefined keyword

### LINKED ISSUE #831: Undefined

agrsh:
Adds an undefined keyword to the language. The goal is to provide users the ability to express intent not to fill an empty hole. This may be useful when there is no coherent value to return in certain branches. For example, consider the the following operation on lists:
```
let head = fun xs -> case xs
    | hd::tl => hd
    | nil => undefined
end in
(head([2, 3, 4]), head(nil))
```
This evaluates to `(2, undefined)` because the left has a coherent value, but the head of an empty list is undefined.

The one difference between an empty hole and `undefined` is that `undefined` should not be a valid pattern in a case expression. I'm still working on making that behavior work properly.

This could be extended by allowing users to specify an error message along with the undefined value to be more clear as to what error led to the undefined behavior. This would require a more major overhaul which adds the possibility for a message/value associated with any empty hole (which could be a desirable feature in itself)


cyrus-:
@shbhgrwl marking as draft -- click ready for review when it's up to date with haz3l-tests and ready to look at!

agrsh:
Documentation started working properly on my machine after clearing out the browser cache, but prior to clearing out the cache I was receiving an `async_js` error -- I wonder if that has something to do with the failing build

cyrus-:
@shbhgrwl Hm, can you reproduce the issue again from a clean cache?

agrsh:
once I cleaned my cache I no longer get the error on my machine (including after using/reloading the webpage such that the cache isn't clean). @hannahpotter seemed to think it was some sort of bug with serialization of new language documentation

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -86,6 +86,7 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
         // placeholder logic: sequence
         tl |> List.fold_left((acc, d) => DHExp.Sequence(d, acc), hd) |> wrap
       };
+    | Undefined => Some(DHExp.EmptyHole(u, 0))
```
cyrus-: Let's add an undefined to DHExp as well so results are a bit more comprehensible.

src/haz3lweb/LangDocMessages.re
```diff
@@ -144,6 +144,30 @@ let triv_exp: form = {
   };
 };
 
+let undef_exp_group = "undef_exp_group";
+let undef_exp: form = {
+  let explanation = {
+    message: "Undefined expression. The undefined expression denotes the lack of a value. It is similar to an empty hole, but with no intent to define it in the future.",
```
cyrus-: "denotes the lack of a value" -> "can be used to define a partial function"

cyrus- (dismissed):

agrsh:
I've added `undefined` to `DHExp`, so it now evaluates to `undefined` rather than an empty hole. additional operations are indeterminate, so, for example, `undefined + 3` will show a result of `undefined + 3`

cyrus-:
Great. If it is ready for review, you can click the Ready for review button. Make sure it is up-to-date with haz3l-tests.

cyrus-:
@shbhgrwl can you resolve the small merge conflict and then mark as ready for review again

cyrus-:
As discussed in office hours, undefined should not evaluate to what looks like a string literal. It should be its own DHExp form.

image

linsyking:
> As discussed in office hours, undefined should not evaluate to what looks like a string literal. It should be its own DHExp form.
> 
> image

This should be fixed now

cyrus-:
@linsyking can you fix the merge conflicts here

linsyking:
Still need to update the `Transition` module

linsyking:
> @linsyking can you fix the merge conflicts here

It should work now, but it seems that the latest hazel cannot run in chrome (chrome ver: 120.0.6099.109).

cyrus-:
@linsyking do you have time to take another look at the latest merge conflicts and getting this over the line?

cyrus-:
Closing in favor of #1315

PR #928, opened 2022-11-08
Exercise Development Mode: Title Editor

LutSa:
Add Haz3l title edit.


src/haz3lweb/Log.re
```diff
@@ -36,6 +36,9 @@ let is_action_logged: Update.t => bool =
   | ResetSlide
   | ToggleMode
   | SwitchSlide(_)
+  | SwitchTextEditor
```
cyrus-: Should we unify this with SwitchEditor since the focus can only be in one place at a time?

src/haz3lweb/Log.re
```diff
@@ -36,6 +36,9 @@ let is_action_logged: Update.t => bool =
...
+  | UpdateTitle(_)
```
cyrus-, line 39: Let's group all Exercise related actions into an ExerciseAction.t type that is injected into Action.t at once, to avoid having too many cases?

cyrus- (changes requested):
Looks good, but see small comments.

You also need to update INSTRUCTORS.md with updated instructions, esp re:prompt.

cyrus-:
closing in favor of #1316

PR #929, opened 2022-11-12
Haz3l exponentiation

shawn-stovall:
This pull request implements the `**` (integer exponentiation) and `**.` (floating-point exponentiation) operators in Hazel 3.

- `**` uses the efficient exponentiation by squaring algorithm (O(log n) time complexity).  Its precedence is one level tighter than multiplication/division and one level looser than negation.  It has right associativity.
- `**.` reduces to Reason's `**`.  Its precedence and associativity is the same as the integer version of this operator.

Possible source of errors:  `**` overflows at 2 ** 31.


src/haz3lcore/lang/Form.re
```diff
@@ -141,6 +141,8 @@ let forms: list((string, t)) = [
   ("plus", mk_infix("+", Exp, P.plus)),
   ("minus", mk_infix("-", Exp, P.plus)),
   ("times", mk_infix("*", Exp, P.mult)),
+  ("power_", mk_nul_infix("**", P.power)),
```
cyrus-: Why are both `**` and `**.` mapped to the same operation? Seems we should have an integer and float version.

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -286,6 +286,7 @@ and exp_term: unsorted => (UExp.term, list(Id.t)) = {
           | (["-."], []) => BinOp(Float(Minus), l, r)
           | (["*."], []) => BinOp(Float(Times), l, r)
           | (["/."], []) => BinOp(Float(Divide), l, r)
+          | (["**."], []) => BinOp(Float(Power), l, r)
```
cyrus-, line 290: `**` seems to be only partially implemented

cyrus- (changes requested):

src/haz3lcore/lang/Form.re
shawn-stovall: Added integer power operator.

src/haz3lcore/statics/MakeTerm.re
shawn-stovall: Added integer power operator.

cyrus- (approved):

PR #932, opened 2022-11-16
Automatically scroll to caret and rearrange dom to not overlap

7h3kk1d:
Uses a flexbox for the overall page and then places the header/footer inside. Introduced a `main` div to allow for the `editor` div padding to still work.

Auto-scrolling at the end of the page goes a little bit past the caret since the view box for the caret SVG is doubled in height:
Screenshot from 2022-11-17 21-40-04

I think this is due to https://github.com/hazelgrove/hazel/blob/haz3l-tests/src/haz3lweb/view/dec/DecUtil.re#L98-L102 but I'm not positive. I decided to not address it in this PR since VS Code also over-scrolls when going down lines so you can preview the next line.


src/haz3lweb/view/SchoolMode.re
```diff
@@ -340,43 +340,49 @@ let view =
...
+              : []
           ),
-        ]
-        : []
+        ),
+      ],
     ),
-  );
+    div(~attr=Attr.class_("bottom-bar"), ci_view),
+  ]; // TODO lang doc visibility tied to ci visibility (is this desired?)
```
7h3kk1d: I tried to keep the changes to a minimum to avoid merge conflicts. These now return a list of divs that can be added to the flexbox div.

src/haz3lweb/Main.re
```diff
@@ -123,13 +123,27 @@ module App = {
...
       ~apply_action=apply(model),
       model,
       Haz3lweb.Page.view(~inject, ~handlers, model),
+      ~on_display=(_, ~schedule_action as _) => {
+        let old_zipper = Editors.get_editor(model.editors).state.zipper;
+        let new_zipper = Editors.get_editor(old_model.editors).state.zipper;
+
+        if (old_zipper != new_zipper) {
+          JsUtil.scroll_cursor_into_view_if_needed();
+        };
```
7h3kk1d: I decided to just auto-scroll when the zipper for the current editor changes. Not sure if there's a better strategy.

src/haz3lweb/util/JsUtil.re
```diff
@@ -116,6 +116,19 @@ let copy = (str: string) => {
...
+let scroll_cursor_into_view_if_needed = () => {
+  let caret_elem = get_elem_by_id("caret");
+  let main = get_elem_by_id("main");
+  let main_rect = main##getBoundingClientRect;
+  let caret_rect = caret_elem##getBoundingClientRect;
+
+  if (caret_rect##.top < main_rect##.top) {
+    caret_elem##scrollIntoView(Js._true);
+  } else if (caret_rect##.bottom > main_rect##.bottom) {
+    caret_elem##scrollIntoView(Js._false);
```
7h3kk1d, line 134: I'd like to play with the scroll behaviors but the bindings for js_of_ocaml aren't there and it doesn't seem fully necessary.

src/haz3lweb/www/style.css
```diff
@@ -237,15 +237,15 @@ svg {
 /* PAGE */
 
 #page {
-  overflow-x: auto;
-  overflow-y: auto;
   height: 100vh;
```
7h3kk1d, line 240: Setting `#page` height to `100%` instead of `100vh` seems to fix the bottom bar on mobile and works okay on my desktop browsers. Someone with more knowledge might want to weigh in.

src/haz3lweb/view/SchoolMode.re
cyrus-: yes tying langdoc visibility and ci visibility is fine since the toggle for langdocs is in the ci

src/haz3lweb/Main.re
```diff
@@ -123,13 +123,27 @@ module App = {
...
+      ) => {
     open Incr.Let_syntax;
-    let%map model = model;
+    let%map model = model
+    and old_model = old_model;
     Component.create(
       ~apply_action=apply(model),
       model,
       Haz3lweb.Page.view(~inject, ~handlers, model),
+      ~on_display=(_, ~schedule_action as _) => {
```
cyrus-: on_display hook runs on every *frame* even when there isn't anything new to paint (so 30+ times per second ideally), so even just doing a diff between old and new zippers in the model is going to cause performance drain. I think we can still use this trigger but we have to only do the work here once per model change, not once per frame.

cyrus- (dismissed):

src/haz3lweb/view/SchoolMode.re
7h3kk1d: I had just moved the comment since the code got slightly rearranged. I'll probably leave it unless you think it's better removed.

src/haz3lweb/Main.re
```diff
@@ -121,13 +124,29 @@ module App = {
...
       ~apply_action=apply(model),
       model,
       Haz3lweb.Page.view(~inject, ~handlers, model),
+      ~on_display=(_, ~schedule_action as _) =>
+      if (action_applied.contents) {
+        let old_zipper = Editors.get_editor(model.editors).state.zipper;
+        let new_zipper = Editors.get_editor(old_model.editors).state.zipper;
+
+        action_applied := false;
+        if (old_zipper != new_zipper) {
```
disconcision, line 146: this comparison might be costly for an obligate per-frame action, but david's hash consing stuff should take care of it eventually. performance impact seems negligable atm so i'm fine with it

disconcision (approved):
lgtm. forgot how much i missed this behavior...

PR #938, opened 2022-11-27
Go to definition when ctrl-clicking a term

7h3kk1d:
There might be a more efficient way to do this but it seemed too easy/useful not to at least do this.


disconcision (approved):
lgtm!

PR #943, opened 2022-12-18
Empty list literals and tuples

disconcision:
This adds support for 'dual' syntactic forms which have both mono- and duo-token versions, such as (empty) list literals and (empty) tuples. New such forms can be added by adding cases to Form.duosplits and Form.duomerges functions, as well as adding the forms themselves to Form.forms and Form.atomic_forms as usual. Updating Form in this way should be sufficient to support proper insertion and deletion behavior, including splitting and merging where appropriate.

Replaces:
- nil with [] in expressions, patterns, and docs.
- triv with () in expressions, patterns, and docs.
- Unit with () in types and docs

Behavior:
- Typing "[]" will result in a (monotile) empty list
- More generally, typing "]" when on "[|" will result in an empty list
- Typing "[ ]" will result in a singleton list with hole
- Typing space (or anything else) on "[|]" will split the (mono) empty list into a (duo) list literal
- Typing backspace on "[|]" will convert the (mono) empty list to a (duo) list literal and put the "[" in the BP
- Typing backspace on "[]|" will convert the (mono) empty list to a (duo) list literal and put the "]" in the BP


dm0n3y (approved):

disconcision:
@dmoon1221 there's a remolding-related bug happening here... if I have "let [|] = ...", and then insert anything that causes a split, say a "1", then the newly-separated list literal brackets get assigned expression sort for some reason. any thoughts?

cyrus-:
@disconcision needs a merge conflict resolved

disconcision:
@cyrus do you still want this in (for A1) despite the issue I showed you? (changing from a pattern [] to a nonempty list causes a sort error until you make an edit outside the list)? (this bug is beyond my ability to address atm)

cyrus-:
Already released A1, I think we can just wait for next tylr version and put this on ice.

PR #944, opened 2022-12-18
Better expansion/barfing (expansion bug fix)

disconcision:
This overhauls the entire expansion and backpack barfing system, fixing the expansion bug and attempting to make both the behavior and implementation clearer and more consistent.

As a reminder the expansion bug (as i've been calling it, but barfing bug is more accurate) affects situations like this:
"let a = b|in", where pressing space fails to drop "in" from the backpack.

First, recall some vocabulary:

- expansion: the process of a monotile expanding into a multitile. a given form can be configured to expand from either its first delimiter, its second delimiter, or both. furthermore, the expansion for each of those delmiters can be configured to be either instant or delayed 
- instant expansion: when expansion occurs immediately after a delimiter is created, e.g. for "(" or ")"
- delayed expansion: when expansion occurs only after a token boundary event, e.g. when " " (or "," or whatever) is pressed on "let|"
- barfing: when the token adjacent to the caret matches the token at the top of the backpack and is replaced by the dropped backpack token
- insertion: the action of inserting a new character. this can result in either a modification insertion, where an existing token is modified (either at the ends or internally) or a creation insertion, aka a token boundary event. modification insertions can result in instant expansions, but only token boundary events result in delayed expansions.
- token boundary event: when insertion would not result in a valid token, so a new token must be created. this occurs in two different cases: (a) the caret is at an outer position and the new character cannot be appended to either neighbour, or (b) the caret is at an inner position, resulting in the token being split.

Previously, token boundary events only triggered delayed expansion on the token to the left. Now, they can trigger up to three expansions/barfings: one for the left neighbor of the caret, one for the newly inserted thing, and one for the right neighbor. for example, pressing "(" on "if|if" will trigger three expansions. It's debatable whether the expansion of the last "if" is desired behavior, but I think there's a consistency case to be made, since we definitely do want to trigger a barf if we press " " on let a = b|in", and I feel the behavior is easier to predict if expansion and barfing happen in exactly the same circumstances I think... please debate if you disagree.

Another similar (but technically independent) consistency change  is that now barfings use the same delayed/instant behavior as expansions. Since parentheses are still instant, typing ")" on "(|" results in an instant drop. however, on "if 1 the", typing "n" no longer results in a drop; you need to press the space first. I think that this feels better... the instant drop feels to me a little jarring. It also seems to be necessary to permit entry of variables which have a keyword as a prefix.

One current sub-optimal but farily niche behavior: consider "(if|if" with ")" in the backpack. if ")" is pressed, i would guess that the first "if" would be expanded, and then the ")" would be barfed, and then the second "if" would be expanded. however, because the expansion of the first if, the ")" is no longer top of the backpack, so the ")" is treated as fresh and a "(" is added to the backpack. This is addressable in principle but I'm not sure whether it's worth it at this point.

Implementation note, mostly for @dmoon1221: backpack selection foci direction is no longer used at all, I felt they were complicating things. construct now takes two directional parameters, one reflecting the direction you want to insert it wrt to the caret, and the other reflecting what order you want to add shard to the backpack in. This means that after you insert a fresh ")" you end up on the 'wrong' side, but i think this is probably principle-of-least-surprisey.

### DIFF
--- a/src/haz3lcore/zipper/action/Insert.re
+++ b/src/haz3lcore/zipper/action/Insert.re
@@ -17,51 +17,62 @@ let sibling_appendability: (string, Siblings.t) => appendability =
     | _ => AppendNeither
     };
 
-let replace_from_backpack =
-    (d: Direction.t, (z, id_gen): state): option(state) =>
+let barf = (d: Direction.t, (z, id_gen): state): option(state) =>
+  /* Removes the d-neighboring tile and drops from backpack;
+     precondition: the d-neighbor should be a monotile
+     string-matching to the dropping shard */
   select(d, z)
   |> Option.map(Zipper.destruct)
   |> OptUtil.and_then(Zipper.put_down)
   |> Option.map(z => (z, id_gen));
 
-let will_expand = kw =>
-  List.length(Molds.delayed_completion(kw, Left) |> fst) > 1;
-
 let expand =
     (kw: Token.t, d: Direction.t, (z, id_gen): state): option(state) => {
-  let (new_label, new_dir) = Molds.delayed_completion(kw, d);
+  /* Removes the d-neighboring tile and reconstructs it, triggering
+     keyword-expansion; precondition: the d-neighbor should be a monotile
+     string-matching a keyword of an expanding form */
+  let (new_label, new_dir) = Molds.delayed_expansion(kw, d);
   select(d, z) |> Option.map(z => construct(new_dir, new_label, z, id_gen));
 };
 
-let expand_keyword = ((z, _) as state: state): option(state) =>
-  /* NOTE(andrew): We may want to allow editing of shards when only 1 of set
-     is down (removing the rest of the set from backpack on edit) as something
-     like this is necessary for backspace to act as undo after kw-expansion */
+let expand_or_barf_neighbors = ((z, _) as s: state): option(state) =>
+  /* If either neighbor is a monotile (a) string-matching the shard at the top
+     of the backpack, barf it, or (b) an expansing keyword, expand it. Only one
+     neighbor is affected; this can probably be relaxed. TODO(andrew) */
   switch (neighbor_monotiles(z.relatives.siblings)) {
-  | (Some(kw), _) when Backpack.is_first_matching(kw, z.backpack) =>
-    replace_from_backpack(Left, state)
-  | (_, Some(kw)) when Backpack.is_first_matching(kw, z.backpack) =>
-    replace_from_backpack(Right, state)
-  | (Some(kw), _) when will_expand(kw) => expand(kw, Left, state)
-  | (_, Some(kw)) when will_expand(kw) => expand(kw, Right, state)
-  | _ => Some(state)
+  | (Some(kw), _) when Backpack.will_barf(kw, z.backpack) => barf(Left, s)
+  | (_, Some(kw)) when Backpack.will_barf(kw, z.backpack) => barf(Right, s)
+  | (Some(kw), _) when Molds.will_expand(kw) => expand(kw, Left, s)
+  | (_, Some(kw)) when Molds.will_expand(kw) => expand(kw, Right, s)
+  | _ => Some(s)
+  };
+
+let expand_or_barf_neighbors2 = ((z, _) as s: state): option(state) =>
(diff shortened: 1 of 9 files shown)


dm0n3y:
> Previously, token boundary events only triggered delayed expansion on the token to the left. Now, they can trigger up to three expansions/barfings: one for the left neighbor of the caret, one for the newly inserted thing, and one for the right neighbor. for example, pressing "(" on "if|if" will trigger three expansions. It's debatable whether the expansion of the last "if" is desired behavior, but I think there's a consistency case to be made, since we definitely do want to trigger a barf if we press " " on let a = b|in", and I feel the behavior is easier to predict if expansion and barfing happen in exactly the same circumstances I think... please debate if you disagree.

agree

dm0n3y:
> Another similar (but technically independent) consistency change is that now barfings use the same delayed/instant behavior as expansions. Since parentheses are still instant, typing ")" on "(|" results in an instant drop. however, on "if 1 the", typing "n" no longer results in a drop; you need to press the space first. I think that this feels better... the instant drop feels to me a little jarring. It also seems to be necessary to permit entry of variables which have a keyword as a prefix.

agree, this feels better

dm0n3y:
> One current sub-optimal but farily niche behavior: consider "(if|if" with ")" in the backpack. if ")" is pressed, i would guess that the first "if" would be expanded, and then the ")" would be barfed, and then the second "if" would be expanded. however, because the expansion of the first if, the ")" is no longer top of the backpack, so the ")" is treated as fresh and a "(" is added to the backpack. This is addressable in principle but I'm not sure whether it's worth it at this point.

ok good to know. agree this is not worth addressing at this time.

dm0n3y:
> Implementation note, mostly for @dmoon1221: backpack selection foci direction is no longer used at all, I felt they were complicating things. construct now takes two directional parameters, one reflecting the direction you want to insert it wrt to the caret, and the other reflecting what order you want to add shard to the backpack in. This means that after you insert a fresh ")" you end up on the 'wrong' side, but i think this is probably principle-of-least-surprisey.

yes this feels better. the whole 'right/wrong' side thing was an artifact of ye olde restricted movement, no need for it now.

src/haz3lcore/zipper/action/Insert.re
```diff
@@ -1,11 +1,90 @@
...
+     the new tile because barfing is limited to the top of the backpack,
+     and I wanted things like "if|then", when you enter a "(", to
+     barf the "then", before it is buried by the ")" added to the BP.
+     The order here could be revisited if barfing was more sophisticated.
+     */
+  let* (z, id_gen) = expand_or_barf_left_neighbor(state);
+  let (z, id_gen) = regrout(Left, z, id_gen);
+  /* Note to david: I'm not sure why the above regrout is necessary.
+     Without it, there is a Nonconvex segment error thrown in exactly
+     one case, the double barf case: insert space on "if then|else" */
```
dm0n3y, line 72: not sure either... my first thought was that this would be due to some new bidelimited container being created that the final regrouting doesn't reach into (eg newly created if-then container, or maybe the segment on right side of else), but I'm not sure how this regrout operation at this stage would affect that

src/haz3lcore/zipper/action/Insert.re
```diff
@@ -1,11 +1,90 @@
...
+  switch (right_neighbor_monotile(z.relatives.siblings)) {
+  | Some(t) when Backpack.will_barf(t, z.backpack) => barf(Right, s)
+  | Some(t) when Molds.is_delayed(t) => delayed_expand(t, Right, s)
+  | _ => Some(s)
+  };
+
+let make_new_tile = (t: Token.t, caret: Direction.t, z: t): IdGen.t(t) =>
+  /* Adds a new tile at the caret. If the new token matches the top
+     of the backpack, the backpack shard is dropped. Otherwise, we
+     construct a new tile, which may immediately expand. */
```
dm0n3y, line 45: docs 🙏

dm0n3y (approved):

src/haz3lcore/zipper/action/Insert.re
disconcision: Ended up undoing this for now... see issue #953

PR #955, opened 2023-01-14
Haz3l checkbox and slider livelits

7h3kk1d:
Open questions:

- Do we want to have parens after livelit applications in this pass that doesn't support livelit params?
- Should the rendered livelits have a span or other bounding box that fixes the width/height?
  - Currently I'm just styling the input elements


src/haz3lcore/Measured.re
```diff
@@ -97,6 +97,7 @@ type t = {
   secondary: Id.Map.t(measurement),
   rows: Rows.t,
   linebreaks: Id.Map.t(rel_indent),
+  livelits: Id.Map.t(measurement),
```
7h3kk1d: This is unused. Not sure where this map will live.

src/haz3lcore/lang/Form.re
```diff
@@ -103,6 +103,9 @@ let is_bad_lit = str =>
    there are at most two quotes, in order to prevent merges */
 let is_string = t =>
   regexp("^\".*\"$", t) && List.length(String.split_on_char('"', t)) < 4;
+
+let is_livelit = t => regexp("^[\\^]\\([a-z][A-Za-z0-9_]*\\)?$", t);
```
7h3kk1d: This allows for a zero-character livelit just so you can type `^`

src/haz3lcore/statics/TermBase.re
```diff
@@ -84,6 +84,14 @@ and UExp: {
...
     | String(op_bin_string);
 
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type livelit_record = {
+    name: string,
+    expansion_type: UTyp.t,
+    // init: UExp.t,
+    // TODO: The rest of the stuff
+    body: UExp.t,
+  };
```
7h3kk1d: Unused so far.

src/haz3lweb/view/Code.re
```diff
@@ -5,30 +5,86 @@ open Util;
...
-        | [_] when !is_consistent => "mono-inconsistent"
-        | [s] when Form.is_string(s) => "mono-string-lit"
-        | [_] => "mono"
-        | _ when !is_consistent => "delim-inconsistent"
-        | _ when !is_complete => "delim-incomplete"
-        | _ => "delim"
-        };
-      [
-        span(
+    ((sort, is_consistent, is_complete, label, i, inject, livelit_state)) => {
```
7h3kk1d: Not sure what to do about this. `inject` is a function therefore unhashable so this function doesn't get memoized. There's probably a smarter way to do the event handling I'm not thinking of. I'd rather not have to do something that reflects over the dom and captures the input elements.

src/haz3lcore/Measured.re
```diff
@@ -375,12 +376,25 @@ let of_segment = (~old: t=empty, ~touched=Touched.empty, seg: Segment.t): t => {
             let map = map |> add_g(g, {origin, last});
             (contained_indent, last, map);
           | Tile(t) =>
+            let livelit_padding =
```
disconcision: break this off into a fn pls

src/haz3lcore/lang/Form.re
disconcision: So right now there's no cursor-info or error-hole feedback if you have an unbound livelit name. Eventually we'll want a legit error status for this... This isn't too hard to do now, but it's more streamlined after the ADT merge goes in (which rewrites statics and cursorinfo, so probably easiest to wait regardless).

For a workaround at the moment lets ensure that unbound livelits get highlighted in red as invalid syntax. To do this, add the current is_livelit predicate as a clause of is_bad_lit, and make is_livelit match only the two exact strings of the current two livelits.

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -135,7 +135,9 @@ let wrap = (u, mode, self, d: DHExp.t): option(DHExp.t) =>
   | InHole(_) => Some(NonEmptyHole(TypeInconsistent, u, 0, d))
   };
 
-let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) => {
+let rec dhexp_of_uexp =
+        (m: Statics.map, uexp: Term.UExp.t, livelit_state: Id.Map.t(DHExp.t))
```
disconcision: make a type alias for Id.Map.t(DHExp.t))... something like Livelit.state

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -273,13 +278,23 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
           DHExp.InconsistentBranches(id, 0, d)
         | _ => ConsistentCase(d)
         };
+      | LivelitAp({livelit_name}) =>
```
disconcision, line 279: make this whole case a function in Livelit.re i think

src/haz3lcore/zipper/Editor.re
```diff
@@ -7,6 +7,7 @@ module Meta = {
     measured: Measured.t,
     term_ranges: TermRanges.t,
     col_target: int,
+    livelit_state: Id.Map.t(DHExp.t),
```
disconcision: alias

src/haz3lcore/zipper/Editor.re
```diff
@@ -16,6 +17,7 @@ module Meta = {
       measured: Measured.of_segment(unselected),
       term_ranges: TermRanges.mk(unselected),
       col_target: 0,
+      livelit_state: Id.Map.empty,
```
disconcision: alias

src/haz3lweb/Update.re
```diff
@@ -415,6 +416,29 @@ let apply =
     | DebugAction(a) =>
       DebugAction.perform(a);
       Ok(model);
+    | LivelitStateChange(livelit_id, livelit_state) =>
+      let (id, ed) = Editors.get_editor_and_id(model.editors);
+      let editor =
+        Editors.put_editor_and_id(
```
disconcision: i feel these highly nested expansions are hard to read... maybe break this out into a couple appropriately-chosen lets. alternatively worth checking if someone already made getters/setters for these subcomponents, probably not worth adding if there arent any though

src/haz3lweb/view/BackpackView.re
```diff
@@ -32,7 +33,14 @@ let backpack_sel_view =
         ),
       ]),
     // zwsp necessary for containing box to stretch to contain trailing newline
-    Text.of_segment(~no_sorts=true, content) @ [text(Unicode.zwsp)],
+    Text.of_segment(
+      ~no_sorts=true,
+      content,
+      ~inject=_ => Ui_effect.Ignore,
+      ~font_metrics,
+      ~livelit_state=Id.Map.empty,
```
disconcision: alias

src/haz3lweb/view/Code.re
```diff
@@ -135,16 +225,18 @@ let view =
       ~unselected,
       ~measured,
       ~settings: ModelSettings.t,
+      ~inject,
+      ~livelit_state: Id.Map.t(DHExp.t),
```
disconcision: i think in general with an appropriately named alias type we can get away with renaming livelit_state everywhere to simply 'livelits'. debatable tho your call

src/haz3lweb/view/Code.re
disconcision: @dmoon1221 any thoughts?

disconcision (changes requested):
Pretty solid. A couple of functional issues that need addressing, some optional aesthetic stuff, and some minor code cleanup.

Functional issues: 
1. livelits seem to be missing a caret position, i.e. pressing right on "^slide|r(-------)" jumps to "^slider(-------)|". The livelit UI should be treated as a single character. Not sure how straightforward this is; I can imagine this complicating the interaction model, e.g. do we need to special-case what happens when we backspace to the right of a livelit? One general solution I can think of is to have a special terminating character on livelits which is treated as 'virtual'. Like lets say livelits required a "^" at the end (maybe instead of the start?), and that "^" is rendered in a different faded color, so it feels less weird that the caret skips over it... maybe try to make it feel stylistically more part of the livelit UI than the name. I sort of like how (I imagine) this feels... right now it feels mildly jarring when the livelit UI appears; if the appearance was triggered by pressing the special livelit character I could see that feeling quite nice. alternatively maybe the special terminating character could literally just be a space, but i think our current whitespace special-casing probably precludes that. in any case, there are some options here; the current caret movement feels a bit too janky though. @cyrus- do you have an opinion here? i think the postfix-^ is worth considering
2. unbound livelits currently trigger a (caught) elaboration exception. this should be fixed by the unbound livelit workaround suggested in the comments.

aesthetic quibbles:
Screen Shot 2023-03-16 at 12 44 35 PM
we might consider:
1. giving the UIs more room to breath by adding a half-space margin on each side
2. doing a small manual top- or bottom- margin adjustment to center them a bit better with the text
3. giving the livelit names a more UI-element-aligned color
4. maybe change the UI-element greys to browns to be more in-line with the color scheme (not pictured; debatable)
5. having some slightly punchier livelit names

all of these are marginal/debatable so don't feel obligated though i do think at least 1 is a good idea

src/haz3lcore/lang/Form.re
7h3kk1d: Fixed in https://github.com/hazelgrove/hazel/pull/955/commits/0495cc50adabd9d3a7447f71eaf258dc5d8fd9bd.

src/haz3lcore/zipper/Editor.re
7h3kk1d: This ones a value unless you were thinking of something like `Livelit.empty_livelit_state`

src/haz3lweb/Update.re
7h3kk1d: I simplified it a bit in https://github.com/hazelgrove/hazel/pull/955/commits/0abde16e81a1f11a4a076b1ba2439cdc9ba5a3e7. I was looking for a better way to do nested record updates and didn't find anything. Not sure if there's something like lenses or if I should just add more update functions.

src/haz3lweb/view/Code.re
7h3kk1d: I did end up refactoring this a bit in the livelit params branch https://github.com/hazelgrove/hazel/commit/2b8b9a2e2adef8575237f633b40cc275d99bef07#diff-98923b69b71582fa885895acfb8efef0a607453ecf764a5cedd4021db6521d3b. I was trying to separate out those changes but I can see if that parts easy to reason in and if it would simplify.

src/haz3lweb/Update.re
7h3kk1d: Addressed by moving https://github.com/hazelgrove/hazel/pull/955/commits/d52669ea55633a8be6a376b244cf1a088b6d03cb outside of the span. I haven't performance-tested anything in the PR though so I'm not sure how expensive the livelit rendering is.

src/haz3lweb/Update.re
```diff
@@ -425,6 +426,16 @@ let apply =
     | DebugAction(a) =>
       DebugAction.perform(a);
       Ok(model);
+    | LivelitStateChange(livelit_id, livelits) =>
+      let editor =
+        Editors.update_editor(
+          ed => {
+            Editor.update_livelits(Id.Map.add(livelit_id, livelits), ed)
+          },
```
disconcision: `ed => { <a>(ed) }`, could just be `<a>` bc currying

src/haz3lweb/view/LivelitView.re
```diff
@@ -0,0 +1,93 @@
...
+      };
+
+    [
+      Node.input(
+        ~attr=
+          Attr.many(
+            [
+              Attr.create("type", "checkbox"),
+              livelit_style(font_metrics, float_of_int(1)),
+              Attr.on_change((_evt, _str) => {
```
disconcision: is there a particular reason you use on_change for the checkbox but on_input for the slider? I don't totally grasp the implications of using one versus the other; i ask mostly because i'm thinking about generic ways to automatically reconcile DOM and model state in more generic contexts

7h3kk1d: I had changed the range to use `on_input` so that the changes are reflected while you're sliding which is really convenient if you're trying to pick a value. I had tried it with the checkbox and had a bug, but now I think that's unrelated. I just tested both with `on_input` and it seems to work.

Addressed in https://github.com/hazelgrove/hazel/pull/955/commits/58eb82ba66e6969bffe2e710918d2f917cecf5ff

cyrus-:
closing in favor of newer efforts in this direction based on #1218

PR #957, opened 2023-01-18
Partial application of non-curried functions for Haz3l

### LINKED ISSUE #795: Partial application for uncurried functions using placeholder
Hazel 2 effort: #514. See also the PHI.

alan123piano:
Continuation of #514 (partial application for Hazel 2)

This feature introduces partial application of functions that accept tuple arguments. This is implemented with a UExp `_`, which does not elaborate a DHExp or synthesize a type. Deferral expressions may only be elaborated in the context of a function application. If the number of arguments on a partial application does not match the signature of the function, then a runtime exception is raised. Upon elaboration, a new function is created which accepts any non-deferred arguments. For example:

```
let div = fun (a, b) : (Int, Int) -> a / b in
let halve = div(_, 2) in
halve(4)
```

The expression yields 2. In the elaboration semantics, `div` has type `(Int, Int) -> Int` and `halve` has type `Int -> Int`.

Implementation:

* Added two new UExps: Deferral and DeferredAp.
* `_` introduces a Deferral expression. A Deferral on its own does not type-check nor elaborate.
* Function applications on tuple arguments including Deferrals are interpreted as DeferredAp instead of Ap.
* DeferredAp is elaborated into a Lambda containing an Ap. For example, the DHExp for `halve` (from above) is `fun ~defer0 -> div(~defer0, 2)`.
* The DHExps for DeferredAps take the unapplied arguments as a tuple, unless there is only one unapplied argument, in which case they take a single argument of that type.

Static error:
- `Free(Deferral)` is assigned to a deferral expression when:
* The deferral expression is neither e_arg itself nor a projection of an e_arg tuple.
* The deferral expression is enclosed in brackets that are not part of the syntax of an application.
- `TypeInconsistent` is assigned to e_arg when:
* `? -> _`/not-a-function case: no errors
* For the case when the input type of e_fun is neither a product type nor the type hole: e_arg is not a single deferral expression
* `() -> _` case: e_arg is not a single deferral expression
* For the case when the input type of e_fun is non-trivial product: when e_arg is a single deferral expression, an incorrect number of arguments is supplied, resulting in `InHole(TypeInconsistent(Unknown(Internal), ty_ana))` (not the most precise error, maybe adding a new error variant is better?); otherwise, the error checking has the same behavior as the error checking of a full application, in which deferrals synthesizes the type hole.

Dynamic error:
Unbounded deferrals are elaborated into `InvalidText`.


tonyfettes:
> There is an issue with how the desugaring introduces new variable names, as shadowing can result in bugs

I guess it's ok use `__arg__` to prefix the arguments, because entering a variable prefixed with underscores is disallowed in the editor I think.

Maybe we should come up with a cleaner way to handle these internal variables (de burijn index)? Using underscores looks good to me though...

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,72 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
             tl |> List.fold_left((acc, d) => DHExp.Sequence(d, acc), hd)
           };*/
         Some(EmptyHole(id, 0))
+      | Deferral => Some(InvalidText(0, 0, "Unbound deferral"))
+      | DeferredAp(fn, args) =>
```
disconcision: this case is long, split some of it or even all of it into a fn

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +321,52 @@ and uexp_to_info_map =
     let (free, maps) = tms |> List.map(any_to_info_map(~ctx)) |> List.split;
     add(~self=Multi, ~free=Ctx.union(free), union_m(maps));
   | EmptyHole => atomic(Just(Unknown(Internal)))
+  | Deferral => atomic(Just(Unknown(Internal))) // Deferral does not define a type
+  | DeferredAp(fn, args) =>
+    // ((a, b, c) -> d)(~, b, c) assumes type (a -> d)
+    let is_deferral = (e: Term.UExp.t) =>
```
disconcision: also defined in elaborator. factor out

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +321,52 @@ and uexp_to_info_map =
...
+      switch (e.term) {
+      | Deferral => true
+      | _ => false
+      };
+    let n_ary_tuple = (tuple, xs) => {
```
disconcision: same

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,72 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
...
+               },
+               Some(([], [])),
+             );
+        let (pat', arg') = (
+          n_ary_tuple(x => DHPat.Tuple(x), pats),
+          n_ary_tuple(x => DHExp.Tuple(x), d_args),
+        );
+        let* d_fn = dhexp_of_uexp(m, fn);
+        let ty_fn = Statics.exp_self_typ(m, fn);
+        let+ (ty_arg, ty_ret) =
```
disconcision: elaborator should only return none if an id lookup fails, otherwise elaboration should be total.
clarify what the failure case is here.... if it should be impossible, add a comment and throw an exception, otherwise it should be handled somehow

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -259,7 +260,21 @@ and exp_term: unsorted => (UExp.term, list(Id.t)) = {
     | ([(_id, t)], []) =>
       ret(
         switch (t) {
-        | (["(", ")"], [Exp(arg)]) => Ap(l, arg)
+        | (["(", ")"], [Exp(arg)]) =>
+          let is_deferral = (e: Term.UExp.t) =>
```
disconcision: dupe, also defined in elaborator. factor out

src/haz3lweb/LangDocMessages.re
```diff
@@ -104,6 +104,15 @@ let mk_example = str => {
   };
 };
 
+let expect_opt = (err_msg, opt) => {
```
disconcision: this already exists in OptUtil (one of the get fns). but i'm not sure why you're using it? not obvious to me why the below modifications to get_group etc are necessary?

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,72 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
...
+          | Typ.Prod(tys) => tys
+          | _ => List.init(List.length(args), _ => Typ.Unknown(Internal))
+          };
+        let ty_arg': Typ.t =
+          Arrow(
+            n_ary_tuple(
+              x => Typ.Prod(x),
+              List.combine(arg_tys, args)
+              |> List.filter(((_ty, arg)) => is_deferral(arg))
+              |> List.map(((ty, _arg)) => ty),
```
disconcision: List.map(fst)

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,72 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
...
+            tuple(xs);
+          };
+        };
+        // substitute all deferrals for new variables
+        let* (pats, d_args) =
+          args
+          |> List.fold_left(
+               (acc, e: Term.UExp.t) => {
+                 switch (acc) {
+                 | None => None
```
disconcision: use let*/+ notation here to handle the None case automatically

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -259,7 +260,21 @@ and exp_term: unsorted => (UExp.term, list(Id.t)) = {
...
+            switch (e.term) {
+            | Deferral => true
+            | _ => false
+            };
+          switch (arg.term) {
+          | Tuple(es) =>
```
disconcision: this case can be replaced by:

| Tuple(es) when List.exists(is_deferral, es) => DeferredAp(l, es)

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +321,52 @@ and uexp_to_info_map =
...
+      uexp_to_info_map(~ctx, ~mode=Typ.ap_mode, fn);
+    let (ty_in, ty_out) = Typ.matched_arrow(ty_fn);
+    let tys =
+      switch (ty_in) {
+      | Typ.Prod(tys) => tys
+      | _ =>
+        print_endline("Inconsistent type for defer ap");
+        List.init(List.length(args), _ => Typ.Unknown(Internal));
+      };
+    let ty_in' =
```
disconcision: this also seems similar to a part in elaborator. maybe try to factor out

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +321,52 @@ and uexp_to_info_map =
...
+      |> List.filter(((arg, _ty)) => is_deferral(arg))
+      |> List.map(((_arg, ty)) => ty)
+      |> n_ary_tuple(tys => Typ.Prod(tys));
+    let (_, free_arg, m_arg) =
+      List.combine(args, tys)
+      |> List.filter(((arg, _ty)) => !is_deferral(arg))
+      |> List.map(((arg, ty)) =>
+           uexp_to_info_map(~ctx, ~mode=Ana(ty), arg)
+         )
+      |> List.fold_left(
```
disconcision: what this doing? isnt it a no-op?

disconcision (changes requested):
not sure this was meant to be marked as ready for review given that theres some chunks of commented out code in langdocs, but i marked off some stuff that could use a little TLC

pigumar1:
The current static self type for a bounded deferral is `Just(Unknown(Internal))` as implied by the original pull request. This brings the following issue:

`let x = fun (x, y) -> x + y in x(_)` will not end up in a static error though the incorrect number of arguments are supplied for partial application, since the deferral can successfully analyze against the type `(?, ?)`.

How should I work on this? I'm thinking about introducing a self type called `Deferral` to distinguish it with `Just(Unknown(Internal))` and modify `Statics.error_status`. Appreciate any help!

pigumar1:
Here is the updated implementation for partial application of functions.

In order to detect more errors statically, such as using an incorrect number of arguments for partial application, I made changes to the implementation of the term `DeferredAp`. Instead of `DeferredAp(t, list[t])`, it is now `DeferredAp(t, t)` in an attempt to assign an error info to `e_arg`.

The implementation of the term `Deferral` is also modified to include two pieces of information: `(is_bounded: bool)` and `(is_singleton: bool)`.

A deferral expression should only appear as a placeholder for an argument in a partial application. If it is used in other contexts (as an expression), it should be considered as an unbounded deferral. In this case, a `free_error` should be reported. `(is_bounded: bool)` and the new variant of `Typ.free_errors` is introduced for this purpose.

The use of `is_singleton` is for the special case when the argument of a partial application is just a deferral instead of a tuple containing deferrals.

I am considering replacing these two boolean variables with a variant type value, but I am unsure of where to define the new variant type. Creating a separate file is one possible option. Additionally, the implementation of deferral bounding in `MakeTerm.re` is currently quite inelegant, and I haven't found a better implementation method yet.

Furthermore, I introduced a variant of `Typ.self` called `Deferral(is_singleton: bool)` to differentiate the self type of a deferral with `Just(Unknown(Internal))`. This allows the deferral expression to be less coupled with the unknown type and provides more information to `Statics.error_status` than `Just(Unknown(Internal))`, so that `let x = fun (x, y) -> x + y in x(_)` results in a static error instead of using `?` to analyze against `(?, ?)`.

In `Statics.re`:
When a deferral is successfully analyzed against `ty_ana`, `Statics.error_status` outputs `NotInHole(AnaConsistent(ty_ana, **ty_ana**, ty_ana))` instead of `NotInHole(AnaConsistent(ty_ana, **Unknown(Internal)**, ty_ana))`. This is done to emphasize that the deferral is just a placeholder. This can be changed if required.
If a static error is detected, the input type of the entire partial application becomes `?`, but the output type remains the same as the original function, serving as the final result of the application. Currently, the type of `1(_, _)` is `(?, ?) -> ?`. It can be changed to `?`, but I believe the former provides more information, so I haven't made the change for now.

In `elaborator.re`:
When a partial application can be performed normally, the deferral expressions act as placeholders and are not elaborated. Unbounded deferrals are elaborated into `InvalidText`.
If the **function expression** is **not** a function or if the function expression is a function but its **input type does not match the argument type**, variable substitution cannot be performed. To provide more reference information to programmers, a **full application** will be performed instead. This means that deferral expressions will be further elaborated, and bounded deferrals will be elaborated into `EmptyHole`, distinguishing them from unbounded deferrals that would cause an error.
`elaborator.re` is written to resemble `Statics.re` as much as possible, for the purpose of subsequent code simplification. However, it seems that most of the code cannot be reused in unrelated contexts with deferrals, so it may not be suitable to place the extracted code in any existing file. Creating a separate file is one possible option.

A special case is the partial application of a function with an input type of `()`. The allowed argument is an empty tuple, which is not destructured like a non-empty tuple, so it cannot be treated in the same way. The partial application of such a function can be performed, but it does not perform any variable binding. This means that statically, `(fun () -> 1)(_)` retains the type of the original function, while dynamically, `(uexp_fun)(_)` is elaborated into `(dexp_fun)`.

cyrus-:
@pigumar1 can we take an approach where there is no self type for a deferral expression at all, and the DeferredAp iterates through the arguments manually, only recursing if it is not a deferral?

pigumar1:
> @pigumar1 can we take an approach where there is no self type for a deferral expression at all, and the DeferredAp iterates through the arguments manually, only recursing if it is not a deferral?

@cyrus 
I apologize for any confusion. If you are referring to the iteration and recursion in `Statics.error_status`, what I actually intended was that a deferral shouldn't be synthesized or analyzed against any specific type. However, the current implementation of `Statics.error_status` only considers the mode and self type, without taking into account whether the term associated with the self type is a deferral expression.

To obtain specific information about the term, I believe iteration and recursion would need to be performed on `Statics.uexp_to_info_map`. However, it seems that there is currently no explicit method to assign an `error_status` to an expression in `Statics.uexp_to_info_map`, as Andrew marked it as a TODO in the definition of `info_exp`.

Nevertheless, it appears that this only affects the specific edge case of `let x = fun (x, y) -> x + y in x(_)`.

src/haz3lcore/statics/TermBase.re
```diff
@@ -115,6 +117,8 @@ and UExp: {
     | Invalid(parse_flag)
     | EmptyHole
     | MultiHole(list(Any.t))
+    | Deferral(bool, bool) // (IsBounded, IsSingleton)
```
disconcision, line 120: rather than using boolean flags here, probably best to define new datatypes eg. type = IsBounded | IsNotBounded. this way it's obvious what it means at the use site (see: boolean blindness). such a type signature is also an obvious place for documentation (i.e. as i read this i'm not sure why these flags are needed, though i assume i'll figure it out after reading the rest of the implementation)

src/haz3lcore/statics/Typ.re
```diff
@@ -52,7 +53,8 @@ type self =
   // TODO: make it so that joined applies only to inconsistent types; rename NoJoin
   | Joined(t => t, list(source))
   | Multi
-  | Free(free_errors);
+  | Free(free_errors)
+  | Deferral(bool); // is_singleton: bool
```
disconcision, line 57: similar to comment in TermBase

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +331,48 @@ and uexp_to_info_map =
     let (free, maps) = tms |> List.map(any_to_info_map(~ctx)) |> List.split;
     add(~self=Multi, ~free=Ctx.union(free), union_m(maps));
   | EmptyHole => atomic(Just(Unknown(Internal)))
+  | DeferredAp(fn, arg) =>
+    let (ty_fn, free_fn, m_fn) =
+      uexp_to_info_map(~ctx, ~mode=Typ.ap_mode, fn);
+    let (ty_in, ty_out) = Typ.matched_arrow(ty_fn);
+    let self: Typ.self = {
+      let ty_in': Typ.t =
+        if (ty_in == Prod([]) && arg.term == Deferral(true, true)) {
```
disconcision, line 336: feels like it should be pattern-matching rather than if

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +331,48 @@ and uexp_to_info_map =
...
+            | Prod(ty_ins) => ty_ins
+            | Unknown(_) as ty_unknown =>
+              List.init(List.length(args), _ => ty_unknown)
+            | _ => [ty_in]
+            };
+          if (List.length(ty_ins) == List.length(args)) {
+            let ty_ins =
+              List.combine(args, ty_ins)
+              |> List.filter(((arg, _ty)) => Term.UExp.is_deferral(arg))
+              |> List.map(((_arg, ty)) => ty);
```
disconcision: List.map(snd)

src/haz3lcore/statics/Statics.re
```diff
@@ -152,6 +152,16 @@ let error_status = (mode: Typ.mode, self: Typ.self): error_status =>
     | None =>
       NotInHole(AnaInternalInconsistent(ty_ana, Typ.source_tys(tys_syn)))
     }
+  | (Syn | SynFun, Deferral(_)) =>
+    NotInHole(SynConsistent(Unknown(Internal)))
+  | (Ana(ty_ana), Deferral(true)) =>
```
disconcision: can eliminate inner switch my making this case
| (Ana(Prod([_, ...]))), Deferral(true)) =>
and the other case
| (Ana(Prod([_, ...]))), Deferral(_)) =>

I don't totally understand this code though; why does analysis against a nonempty product imply inconsistency here?

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -258,7 +259,32 @@ and exp_term: unsorted => (UExp.term, list(Id.t)) = {
     | ([(_id, t)], []) =>
       ret(
         switch (t) {
-        | (["(", ")"], [Exp(arg)]) => Ap(l, arg)
+        | (["(", ")"], [Exp(arg)]) =>
+          let bound_deferral = (arg: UExp.t, is_singleton): UExp.t => {
+            ids: arg.ids,
+            term: Deferral(true, is_singleton),
+          };
+          switch (arg.term) {
```
disconcision, line 267: I wonder if some of the upstream special-casing in statics and elaborator could be handled here by changing the second argument in DeferredAp into a list (a singleton if the argument is just a singleton, multiple elements if the argument is a tuple. Not sure this would help, just a thought

src/haz3lcore/statics/Statics.re
```diff
@@ -321,6 +331,48 @@ and uexp_to_info_map =
...
+        if (ty_in == Prod([]) && arg.term == Deferral(true, true)) {
+          Prod([]);
+        } else {
+          let args =
+            switch (arg.term) {
+            | Tuple(args) => args
+            | _ => [arg]
+            };
+          let ty_ins =
+            switch (ty_in) {
```
disconcision: this block (duplicated in elaborator) should probably be broken off into a helper, probably in Typ, something like matched_prod: t=>list(t)

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,68 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
             tl |> List.fold_left((acc, d) => DHExp.Sequence(d, acc), hd)
           };*/
         Some(EmptyHole(id, 0))
+      | Deferral(true, _) => Some(EmptyHole(id, 0))
+      | Deferral(false, _) => Some(InvalidText(0, 0, "_")) // Unbounded deferral
+      | DeferredAp(fn, arg) =>
```
disconcision, line 169: feel like this implementation is a bit awkward... see other comments in maketerm and statics. a lot of syntactic complexity here comes from negotiating the singleton versus tuple case... it might be worth trying to either regularize representation in maketerm (see comment there), or else maybe push the singleton/tuple to the top of the decision tree; like just implement the case of 
`| DeferredAp(fn, arg) when !is_tuple(arg) => ...`
see if there are any obvious helpers, and then hopefully the 
`| DeferredAp(fn,Tuple(es)) => ...`
case just uses the same helpers but maps over things

src/haz3lcore/dynamics/elaborator.re
```diff
@@ -164,6 +164,68 @@ let rec dhexp_of_uexp = (m: Statics.map, uexp: Term.UExp.t): option(DHExp.t) =>
...
+              | _ => [ty_arg]
+              };
+            if (List.length(ty_ins) != List.length(args)) {
+              let+ d_arg = dhexp_of_uexp(m, arg);
+              DHExp.Ap(d_fn, d_arg);
+            } else {
+              // Substitute all deferrals for new variables
+              let+ (pats, d_args, ty_args) =
+                List.combine(args, ty_ins)
+                |> List.fold_left(
```
disconcision, line 192: there's no actual data dependency between elements here (i think) so instead of folding, just map, and then use OptUtil.sequence (see other examples in this file)

disconcision (changes requested):
Looking good... could still use a bit of cleanup and some docs on the Deferral flags.
When I try to run this I'm still getting a langdocs exception when i try to select the parentheses around a deferral; not sure if this was resolved? I think I'm resetting localstorage state appropriately.

pigumar1:
> Looking good... could still use a bit of cleanup and some docs on the Deferral flags. When I try to run this I'm still getting a langdocs exception when i try to select the parentheses around a deferral; not sure if this was resolved? I think I'm resetting localstorage state appropriately.

Thank you for your suggestions! The strange thing is that my updates to `LangDocMessages.re` are not consistently synchronized with my builds, both locally and on the branch. For instance, I removed some spaces on line 146 to align `(fun (x, y) -> x + y)` with the `let` keyword in the example, but this change is not appearing on the branch.

pigumar1:
Follow up: #1059

PR #958, opened 2023-01-19
Explicit Polymorphism with typfun and typap 

AlienKevin:
What’s working:
1. Type checking for typfun and typap
2. Recursive typfuns
What’s not working:
1. Type alias in typfun
2. ADT (Built-in ADTs and sum...end)

Working examples:
1. Polymorphic map
```
let map : forall X -> forall Y -> (X -> Y) -> [X] -> [Y]
 = typfun X -> typfun Y -> fun f -> fun xs ->
  case xs
    | nil => nil
    | x::xs => f(x)::map@<X>@<Y>(f)(xs)
  end
in
let even = fun x -> x / 2 * 2 == x in
let id = fun x -> x in
map@<Int>@<Bool>(even)([1, 2, 3, 4, 5])
```

2. Polymorphic fold left
```
let fold_left : forall X -> forall Y -> (Y -> X -> Y) -> Y -> [X] -> Y =
  typfun X -> typfun Y -> fun f -> fun acc -> fun xs ->
    case xs
      | nil => acc
      | x::xs => fold_left@<X>@<Y>(f)(f(acc)(x))(xs)
    end
in
let add_bools = fold_left@<Bool>@<Int>(fun acc -> fun x -> acc + if x then 1 else 0)(0) in
add_bools([true, false, true])
```

3. Polymorphic fold right
```
let fold_right : forall X -> forall Y -> (X -> Y -> Y) -> [X] -> Y -> Y
= typfun X -> typfun Y -> fun f -> fun xs -> fun acc ->
  case xs
    | nil => acc
    | x::xs => f(x)(fold_right@<X>@<Y>(f)(xs)(acc))
  end
in
let add_bools = fun xs:[Bool] ->
  fold_right@<Bool>@<Int>(fun x -> fun acc -> acc + if x then 1 else 0)(xs)(0)
in
add_bools([true, false, true])
```


cyrus-:
archiving, see #1030

PR #960, opened 2023-01-20
Algebraic Datatypes

cyrus-:


disconcision:
Large update.

1. Now current with dev, including improved casts.
2. Tags are now stored as context entries. I created a new entry type and retained the existing Tag syntactic sort; technically all this could be flattened to var, with capitalization acting as the disambiguator, but this feels messier to me, so unless it would make other things easier in some way I can't see, I'd as soon leave it as such. Right now Tags are filtered out of the context display; easy to modulate this.
3. Tags and tag application now have different logic in analytic position. Not arbitrary analytic position though: In particular, when a tag or tag application is analyzed against a sum type for which that tag is a variant, the self type of that tag will be considered to be that assigned to it by that sum type. The effect is that this 'shadows' constructors of aliases sum types. See screenshot at bottom  for the basic ''''test suite''''' I was using to check behavior:
4. The logic I'm using for the above is a bit subtle, which probably means it's wrong, or at least there's a better way of doing it. Partially this may be due to the fact that I'm still using a generic Ap form for both function and constructor application. The logic for both seems like it almost entirely overlaps, with the exception of the mode to be used in function position. In any case, all the logic for both tags and tag aps for expressions and patterns are governed by Statics.tag_ana_typ, Statics.tag_ap_mode, and Statics.tag_self... those are the places to look closely at. Additionally, close inspection of the below screenshot reveal some interesting edge cases; worth checking out the CI feedback I'm giving for some of these.
5. After merging in improved casts, there were still casting issues involving tags. For example, the function at the end of the above screenshot was getting stuck; B(1) would be substituted in as the scrutinee of the case, but it would have a cast to unknown on it, which i guess stops the case from digesting it. I addressed this by adding Ap(Tag(_),_) to the list of forms with special ana rules (which don't get casts), and this addressed the behavior, but I don't know if it's the correct solution. I also added Tag(_) to that list, since it also has special Ana rules, but not for any behaviorally motivated reason.
6. Builtin ADTs are removed.

Screen Shot 2023-01-22 at 6 12 14 PM

disconcision:
Also this still has the outstanding bug that the type definition of type aliases remains ill-sorted until you conclude the definition by dropping the in. I'm not sure what causes this behavior to differ from type annotations; probably needs some @dmoon1221 guidance

disconcision:
1. Merged in some of eric's casting code and made some other related corrections. All evaluation examples I've tried work now.
2. Improved syntactic model. No more sum..end. There's a prefix + for singletons and optionally for other sums. There's one weirdness with this though... if you insert a new "+" inside an existing sum next to another "+", you won't get a hole between them, because the second + just gets parsed as a prefix operator. Right now this results in the prefixed part being error-holed... it's easy to not treat it as an error and just splice these things together, but it still looks weird. No strong opinion about this at the moment. Another option would be to pick a different symbol for the prefix op.
3. Better errors for sum type definition. Better docs. New errors include duplicate constructors, constructor application definition outside of Sum type, non-constructor in constructor position in constructor application definitions.

Screen Shot 2023-01-26 at 11 47 33 PM

cyrus-:
image

can't press backspace here

cyrus-:
also can't press `+` here to create a new first variant 
image

PR #963, opened 2023-01-21
multiholes elaborate to empty holes for now

disconcision:


cyrus-:
merging, can do something more elegant with the next version of tylr

PR #965, opened 2023-01-22
Haz3l binding uses

agrsh:
Intended behavior is to highlight all uses of a bound variable when the cursor is over a pattern. At the moment, there is code which prints the uses, but the UI portion hasn't been done yet.

### DIFF
--- a/src/haz3lcore/statics/Statics.re
+++ b/src/haz3lcore/statics/Statics.re
@@ -38,7 +38,8 @@ type info_pat = {
   term: Term.UPat.t,
   mode: Typ.mode,
   self: Typ.self,
-  ctx: Ctx.t // TODO: detect in-pattern shadowing
+  ctx: Ctx.t, // TODO: detect in-pattern shadowing
+  binding_exp: list(Id.t),
 };
 
 /* (Syntactic) Types are assigned their corresponding semantic type. */
@@ -415,7 +416,8 @@ and uexp_to_info_map =
     );
   | Fun(pat, body) =>
     let (mode_pat, mode_body) = Typ.matched_arrow_mode(mode);
-    let (ty_pat, ctx_pat, m_pat) = upat_to_info_map(~mode=mode_pat, pat);
+    let (ty_pat, ctx_pat, m_pat) =
+      upat_to_info_map(~mode=mode_pat, ~binding_exp=ids, pat);
     let ctx_body = VarMap.concat(ctx, ctx_pat);
     let (ty_body, free_body, m_body) =
       uexp_to_info_map(~ctx=ctx_body, ~mode=mode_body, body);
@@ -425,12 +427,14 @@ and uexp_to_info_map =
       union_m([m_pat, m_body]),
     );
   | Let(pat, def, body) =>
-    let (ty_pat, ctx_pat, _m_pat) = upat_to_info_map(~mode=Syn, pat);
+    let (ty_pat, ctx_pat, _m_pat) =
+      upat_to_info_map(~mode=Syn, ~binding_exp=ids, pat);
     let def_ctx = extend_let_def_ctx(ctx, pat, ctx_pat, def);
     let (ty_def, free_def, m_def) =
       uexp_to_info_map(~ctx=def_ctx, ~mode=Ana(ty_pat), def);
     /* Analyze pattern to incorporate def type into ctx */
-    let (_, ctx_pat_ana, m_pat) = upat_to_info_map(~mode=Ana(ty_def), pat);
+    let (_, ctx_pat_ana, m_pat) =
+      upat_to_info_map(~mode=Ana(ty_def), ~binding_exp=ids, pat);
     let ctx_body = VarMap.concat(ctx, ctx_pat_ana);
     let (ty_body, free_body, m_body) =
       uexp_to_info_map(~ctx=ctx_body, ~mode, body);
@@ -443,7 +447,10 @@ and uexp_to_info_map =
     let (ty_scrut, free_scrut, m_scrut) = go(~mode=Syn, scrut);
     let (pats, branches) = List.split(rules);
     let pat_infos =
-      List.map(upat_to_info_map(~mode=Typ.Ana(ty_scrut)), pats);
+      List.map(
+        upat_to_info_map(~mode=Typ.Ana(ty_scrut), ~binding_exp=ids),
+        pats,
+      );
     let branch_infos =
       List.map2(
         (branch, (_, ctx_pat, _)) =>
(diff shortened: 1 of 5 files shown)


agrsh:
Largely finished. Also incorporates some bug fixes to the co-ctx. At the moment, all uses of any binding in a pattern are highlighted orange -- it may be better to have different bindings highlighted different colors. The code itself can be polished a little (in particular, there are still some print statements littered around).

src/haz3lweb/view/Cell.re
```diff
@@ -155,6 +155,18 @@ let test_result_layer =
   );
 };
 
+let get_usage_ids = (zipper, info_map) => {
+  /* bad style, does reason have let*? */
```
disconcision, line 159: it does yes; search the codebase. you'll need to import OptUtil.Syntax

src/haz3lweb/view/Cell.re
```diff
@@ -168,6 +180,16 @@ let deco =
       ~color_highlighting: option(ColorSteps.colorMap),
     ) => {
   let unselected = Zipper.unselect_and_zip(zipper);
+  let use_ids = get_usage_ids(zipper, info_map);
+  let use_highlights =
+    List.flatten(
```
disconcision, line 185: ListUtil.flat_map

src/haz3lweb/view/CursorInspector.re
```diff
@@ -250,7 +250,24 @@ let view =
     | Some(index) =>
       switch (Haz3lcore.Id.Map.find_opt(index, info_map)) {
       | Some(ci) =>
-        inspector_view(~inject, ~settings, ~show_lang_doc, index, ci)
+        let _ =
+          List.map(
+            usage => {
+              print_string(fst(usage));
```
disconcision, line 256: print cleanup

src/haz3lcore/statics/Term.re
```diff
@@ -319,6 +319,29 @@ module UPat = {
       }
     };
   };
+
+  let rec get_all_bindings = (pat: t) => {
```
disconcision, line 323: technically this can be obtained by just extracting the variables names from the pattern's context. but this is going to change a bit with ADTs, so it's fine to just leave it for now. maybe add a "TODO(andrew)" comment on it for me to remind me to update it to the new co-ctx system once its fully in place.

src/haz3lcore/statics/Statics.re
```diff
@@ -160,6 +161,90 @@ let is_error = (ci: t): bool => {
...
+          let updated_map: list(usage_info) =
+            List.map(
+              (usage_pair: usage_info) => {
+                let (var, ids) = usage_pair;
+                let new_ids: list(Id.t) =
+                  List.map(
+                    (item: Ctx.co_item) => {item.id},
+                    List.flatten(VarMap.lookup_all(info_exp.free, var)),
+                  );
+                print_string("co_ctx | ");
```
disconcision, line 194: cleanup (multiple prints)

src/haz3lcore/statics/Statics.re
```diff
@@ -160,6 +161,90 @@ let is_error = (ci: t): bool => {
...
+      );
+    let empty_bindings_map: list(usage_info) =
+      List.map(var => (var, []), bindings);
+    let bindings_map =
+      List.fold_left(
+        (binding_map: list(usage_info), info_exp: info_exp) => {
+          let updated_map: list(usage_info) =
+            List.map(
+              (usage_pair: usage_info) => {
+                let (var, ids) = usage_pair;
```
disconcision, line 188: just destructure this immediately ie `((var, ids): usage_info) => {...`

src/haz3lcore/statics/Statics.re
```diff
@@ -459,24 +570,33 @@ and uexp_to_info_map =
...
+    /* let free = Ctx.union([free_scrut] @ branch_frees); */
+    let free = Ctx.union([free_scrut, ...branch_frees]);
     add(~self, ~free, union_m([m_scrut] @ pat_ms @ branch_ms));
   };
 }
 and upat_to_info_map =
     (
       ~ctx=Ctx.empty,
       ~mode: Typ.mode=Typ.Syn,
+      ~body_ids: list(Id.t)=[Id.invalid],
```
disconcision, line 591: these are the uses, correct? if so, i think they should just be called that, body_ids is vague. also why does it contain Id.invalid by default instead of being empty?

src/haz3lcore/statics/Statics.re
```diff
@@ -459,24 +570,33 @@ and uexp_to_info_map =
...
+      List.map2(
+        ((_, body_free, _), (_, ctx_pat, _)) => {
+          Ctx.subtract_typ(ctx_pat, body_free)
+        },
+        branch_infos,
+        pat_infos,
+      );
     let self = Typ.Joined(Fun.id, branch_sources);
-    let free = Ctx.union([free_scrut] @ branch_frees);
+    /* let free = Ctx.union([free_scrut] @ branch_frees); */
```
disconcision, line 582: cleanup

src/haz3lcore/statics/Statics.re
```diff
@@ -425,25 +512,49 @@ and uexp_to_info_map =
...
     let (ty_def, free_def, m_def) =
       uexp_to_info_map(~ctx=def_ctx, ~mode=Ana(ty_pat), def);
     /* Analyze pattern to incorporate def type into ctx */
-    let (_, ctx_pat_ana, m_pat) = upat_to_info_map(~mode=Ana(ty_def), pat);
+    let pat_body_ids =
+      if (is_rec) {
+        body.ids @ def.ids;
+      } else {
+        body.ids;
+      };
```
disconcision, line 527: consider `body_ids @ (is_rec ? def_ids : [])`

src/haz3lcore/statics/Statics.re
```diff
@@ -425,25 +512,49 @@ and uexp_to_info_map =
       union_m([m_pat, m_body]),
     );
   | Let(pat, def, body) =>
-    let (ty_pat, ctx_pat, _m_pat) = upat_to_info_map(~mode=Syn, pat);
-    let def_ctx = extend_let_def_ctx(ctx, pat, ctx_pat, def);
+    let (ty_pat, ctx_pat, _m_pat) =
+      upat_to_info_map(~mode=Syn, ~body_ids=body.ids @ def.ids, pat);
+    let (def_ctx, is_rec) = extend_let_def_ctx(ctx, pat, ctx_pat, def);
```
disconcision, line 517: i feel like this approach with is_rec is a little indirect. not sure if it's the best way, but you could just do a comparison ctx == def_ctx and avoid having to alter extend_let_def

disconcision (changes requested):
Looking pretty good... some small fixes. I think we might want to iterate a bit on the style though... the orange doesn't do a great job of visually associating the use sites with the pattern selection. I think we should maybe use the same color as for the pattern term indicator, but further differentiate the use decoration by maybe making it an outline instead of a solid box, or something like that. maybe a different stroke style. not sure how that will look but worth trying, or maybe some other variation I think.

cyrus-:
this now exists through a different effort, but thanks for your work on this @agrsh

Issue #966, opened 2023-01-22
keywords at the end of a paste don't expand


disconcision:
like if you paste "let x = 1 in" now, the "in" doesn't expand... could cause issues. not sure about principled fix tho

PR #971, opened 2023-01-24
paste fix 2: return to instant expansion on non-leading keyword

disconcision:
This is one of two alternate fixes for #966, then other one being #972; this one is currently my recommendation (despite when we talked about it @dmoon1221) because, after playing around, and independent of #966, i feel increased probability of unintentionally leaving unexpanded keywords is likely to result in more frustration than the variable name insertion limitations of the delayed expansion approach.


disconcision:
Merged #972 instead

PR #973, opened 2023-01-25
c/p gradescope grading scripts into haz3lschool

hejohns:
Gradescope-Utils is built like a cpan module, although I wouldn't recommend actually installing it lest PATH pollution


cyrus-:
@mirryi is this PR up-to-date with what we use in 490, and if so, thoughts on putting this in the Hazel repo vs. a separate repo?

cyrus-:
Closing in favor of #2053

Issue #969, opened 2023-01-24
[meta] analyze data from previous semesters


cyrus-:

cyrus-:
Interesting paper from the Learn OCaml folks on finding clusters in student editing behavior. Would be nice to reproduce / follow up on this in the Hazel data. https://arxiv.org/abs/2301.02611

PR #976, opened 2023-01-29
Add button to reconstruct state from a session export via log replay

evanweissburg:
**Summary**

Meta issue: https://github.com/hazelgrove/hazel/issues/969

It's currently possible to save a snapshot export of a user session as a file, then import it later. However, this import functionality simply loads the exact snapshot rather than building the state procedurally using the action log. For data analysis purposes, it would be nice to be able to build the state incrementally, which this PR introduces.

**Implementation**

First, load the export file. Next, parse the log and turn the list of timestamps into a list of time diffs. Then, kick off the first action in the log. Each action is responsible for executing, delaying, then kicking off the subsequent action.

**Caveats**

- Button has same icon as import button.
- Since the log does not include information about whether the original action succeeded, on replay, we don't know if the action failed because of a state mismatch or because the original action failed.
- `Update.apply` is now recursive.
- `Update.t` and `Log.t` are vaguely mutually recursive now, since logs contain actions, but certain actions (like `StartReplay`) reference logs. I get around this in a slightly hacky way, but I don't think it's a big deal.

**Testing**

Manually tested with some replays I generated. 

Eventually, we might want to set up an automated check to see if the log-replay state matches the imported final state at some point, but unless there's an easily accessible "session state ~matches" function that already exists, I'm not sure it's worth it right now.

**Example**

Other than triggering the replay, I do not take any action (eg: typing or clicking) during this clip.

https://user-images.githubusercontent.com/87710069/215351495-8944491e-0436-4488-b37d-6ab33831c80c.mov

### DIFF
--- a/src/haz3lweb/Log.re
+++ b/src/haz3lweb/Log.re
@@ -26,7 +26,8 @@ let is_action_logged: UpdateAction.t => bool =
   | Set(_)
   | FinishImportAll(_)
   | FinishImportScratchpad(_)
-  | FinishReplay(_)
+  | StartReplay(_)
+  | StepReplay(_)
   | ResetSlide
   | ToggleMode
   | SwitchSlide(_)
--- a/src/haz3lweb/Log.re
+++ b/src/haz3lweb/Log.re
@@ -19,12 +19,14 @@ let is_action_logged: UpdateAction.t => bool =
   | SetShowBackpackTargets(_)
   | InitImportAll(_)
   | InitImportScratchpad(_)
+  | InitReplay(_)
   | UpdateResult(_)
   | DebugAction(_) => false
   | ResetCurrentEditor
   | Set(_)
   | FinishImportAll(_)
   | FinishImportScratchpad(_)
+  | FinishReplay(_)
   | ResetSlide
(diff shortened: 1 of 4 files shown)


cyrus-:
Nice!

Seems to not handle edits after I do a replay very well though, or I'm not sure what it's doing. Seems to just replay the previous log instead of capturing the new actions?

cyrus-:
Also an enhancement suggestion: list the edit actions in e.g. the right sidebar and have it so you can scrub through them. You can look at the history in the left sidebar of Hazel 2 (click tan bar to open) for inspiration: https://hazel.org/build/hazel2/

evanweissburg:
> Nice!
> 
> Seems to not handle edits after I do a replay very well though, or I'm not sure what it's doing. Seems to just replay the previous log instead of capturing the new actions?

@cyrus- Can you clarify what you mean? I can't seem to reproduce editing issues locally.

https://user-images.githubusercontent.com/87710069/215592163-82819f21-4508-498b-9484-8eb483cebbe2.mov

evanweissburg:
Closing in favor of a later (and more complete) PR

PR #978, opened 2023-02-06
Fix cyclical dependency issues

evanweissburg:
**Summary**

This PR makes no changes to functionality; it's pure refactors. Practically, this PR helps break dependency cycle issues, and stylistically, I believe that one type per module is also generally preferred.

**Implementation**

- `Log.Entry` is now `LogEntry` (breaks cycle `Model -> Replay -> Log(.Entry) -> Model(.timestamp)`)
- `Model.settings` is now `ModelSettings` (breaks cycle `Model -> Replay -> LogEntry -> UpdateAction -> DebugAction -> LocalStorage -> Model`)

Extra context: https://hazelgrove.slack.com/archives/C038TL4C2KG/p1675628741667559

**Testing**

Code still compiles, and my `Replay` module no longer causes dependency cycles 🎉


cyrus-:
@evanweissburg fails to build, type error

evanweissburg:
@cyrus- Sorry about that, trailing reference. All fixed!

PR #981, opened 2023-02-06
Add support for replays from student logs, along with a step-through/scrubbing control panel

evanweissburg:
**Summary**

It would be nice to be able to step through a student log by hand for data analysis purposes, both in real-time and incrementally. This PR provides support for this functionality.

**Implementation**

The new module `Replay` is essentially a log with a position pointer. It supports three main operations: forward, backward, and step. Step is distinguished from forward because it also returns a delay, which is used for real-time playback from a student log.

The new module `ReplayControls` allows a `Replay` to be visualized on the front-end, and provides buttons for interactive replay.

A couple minor issues:

- Backward does not work properly in the presence of user actions. As a concrete example, if the log contains actions 1, 2, 3, 4, 5, and you play through 1, 2, 3, then move the cursor (action A), then press backward, we will undo the cursor movement and also scroll one position backwards through the log (so now the replay incorrectly believes we have only finished actions 1 and 2). The cleanest approach to fix this might just be preventing user actions during a replay.
- I am not super happy with the UI/UX of the control panel, but I am also not an expert in CSS.

**Video**

https://user-images.githubusercontent.com/87710069/217106548-c6a2199f-1497-40ed-9758-466d7fdf64e8.mov


disconcision (changes requested):
This looks really cool and is very well structured. One minor structural change: The top-level action list is getting crowded; I want to start grouping lists of related actions similarly to is currently done for settings. Suggest creating a replay_actions type analogous to settings_actions and just have Replay(replay_actions) as the top-level UpdateAction. then update can just call out to a split-off update_replay fn.

Behaviourally: I got a bit confused while testing this.

1. Because replayed actions can be happening on a different slide than you're on, I originally though it wasn't doing anything. We might want to change the slide view when a slide change action occurs; though this might be more complicated to implement than I'd like.

2. I'm also confused about the consequences of interleaving my own actions with replay actions. It looks to me like if I do my own action, and then advance the replay, it will keep doing the replay actions against my modified state, but if i rewind past my own action, then it will replay as-original. this makes sense, but can get confusing i think... probably need to think more about the UX of such cases. I'm tempted to say there should at least be a clear visual indication in the replay UI when you 'take if off-road'; some kind of marker in the list maybe. another option could just be disallowing all user actions during replay, or maybe only allowing movement and slide change so you can inspect cursorinfo, if we can find some principled way to rig it so that the cursor/slide gets reset to the proper position if we resume replay.

Or on the other hand perhaps being able to modify stuff arbitrarily and still keep replaying actions on top of the new state is worth the complexity, and it we just need to find a way to make it clear what's happening in the UI. As a first pass though my thought would be that the simplest, still useful, and non-confusing thing to do though would be to lock all user actions during replay; pausing replay lets you do arbitrary actions again, but resuming replay would either be impossible (easy), or (better but harder) would roll you back to the state at which it was paused.

3. I'm also somewhat confused about the list of actions displayed...I think it's showing the (up to) 10 actions before and after the currently selected one. the problem is, when you're on one of the first 10 actions, this means that it only shows 10, and advancing grows the list, which made me think I was somehow causing new actions to be added. I'd suggest changing the view logic to always show a fixed number of actions... if you're on first one, show the next 19 or whatever, and also show (something like) a "..." at the top and/or bottom when there's more actions in the list. (More ambitiously, perhaps the whole thing could just be scrolling subpanel (which would automatically reposition as you replay) where you could also click around to move between states... probably not worth worrying about this at the moment though)

evanweissburg:
@disconcision Thanks for your thorough and thoughtful review!

On your point (1), I agree this can be confusing. Actually, we do change the slide view when this occurs in a replay; the issue you're noticing occurs in the more general case where the current initial state differs from the replay's initial state. Since (as far as I know), the logs don't contain information about their initial state, this is difficult to solve, but worth calling out explicitly.

On (2), totally agreed. I think the best course of action is probably to freeze all non-replay actions during a replay. This is a bit tricky to implement given the wide range of actions available to users, but it's also worth noting explicitly. Like (1), I think it's worth thinking about this more rather than implementing now.

On (3), that makes sense! I've improved the implementation to match your suggestion where we always try to display a constant number of actions, even at the beginning or end of a replay.

To help prevent user confusion on (1) and (2), I've added a detailed comment in the code and a warning label in the replay control viewer that results may be unstable. Let me know if you'd like further changes!

cyrus-:
subsumed by #2026

PR #983, opened 2023-02-09
Fix term/tile decoration lengths on chrome

disconcision:
This fixes a longstanding bug with long term/tile decorations and error holes being too short or long on chrome. Introduces a new svg wrapper, code_svg_sized, which actually sets the exact viewbox size needed for its content. Previously the viewbox for all elements was a fixed size. When the elements were much bigger than the viewbox, this resulted in them becoming increasingly mis-sized in chrome. I'm not sure whether this constitutes a bug per se, or just a case of firefox being more robustly designed in the face of viewbox sizes being set smaller than the content. This had been partially addressed in the past by setting the fixed viewbox size to be very large, but this causes runoff issues, including making it awkward to inspect decoration elements, as the viewbox is what determines the bounding box of the element shown in devtools.


cyrus-:
Very broken for me... 
image

disconcision:
huh weird its fine locally for me but i'm seeing what you're seeing on the server... looking into it

disconcision:
@cyrus- fixed

PR #985, opened 2023-02-10
Add Tooltip

### LINKED ISSUE #984: feature request: submenus
just noticed we have the same icon representing three different actions. a new submenu widget would be nice for these; heres a quick mockup:

Screen Shot 2023-02-09 at 4 13 23 PM

tonyfettes:
<details>
<summary>Old PR Summary</summary>
See issue #984 

Support submenus.

[submenu-demo.webm](https://user-images.githubusercontent.com/29998228/218228956-6d75377b-7b46-4beb-a040-9cbbfadf34c1.webm)

## Roadmap

- [x] text button
- [x] toggle button
- [x] fix baseline/padding
- [x] fix menu gap
- [x] animate expansion 
</details>

Better looking tooltips.

[recording.webm](https://github.com/hazelgrove/hazel/assets/29998228/71c3d93e-5326-4f6c-84d5-4512cb609a4b)

delay: 1000ms

## Usage

Wrap your widgets in `Widgets.tooltip("Tooltip content", <your widgets>)`, or
```
Widgets.button(...) |> Widgets.tooltip("Hello, world!");
```

For now, they only work if they're in the menu (on the left, drop-down style) or in the toolbar (on the top).

## Roadmap
- [ ] Make background (css: `tooltip-background`) clickable
- [ ] Clean up CSS magic dimensions (20px, 1.2em, 2.75em, etc)

## Future Work
- Share delay between buttons
- Make the `select` dropdown menu (might be another PR)


tonyfettes:
This PR only add sub-menu as a widget. It does not solve the issue of using the same icon for different exports, neither adds the evaluation options.

cyrus-:
@tonyfettes can you resolve the merge conflict when you get a chance

tonyfettes:
@cyrus- Sure, now it should be free of conflicts.

cyrus- (changes requested):
@tonyfettes could we do something similar showing up below the icons in the top bar as well? and go ahead and remove the CSS titles since they are now redundant.

tonyfettes:
@cyrus- I have added the submenus for the toolbar and remove the titles/tooltips. Shall we merge the export/import buttons, as mentioned in the original PR #984 ?

https://github.com/hazelgrove/hazel/assets/29998228/29897d89-a2a0-44bb-a9e6-7fc1b41c9c79

disconcision:
I think that we should add submenus to the main menu; the point of this feature (as i envisioned anyway) was to clear infrequently-used icons out of the top bar.

Also I think that if we're going to do pop-up labels like this, we need to put a delay on them; see for example vscode which does ~500ms by default, otherwise they're quite distracting to see appearing all the time

tonyfettes:
> Also I think that if we're going to do pop-up labels like this, we need to put a delay on them; see for example vscode which does ~500ms by default, otherwise they're quite distracting to see appearing all the time

Agree. I guess it might be better if we're not using submenus for tooltips...

cyrus- (changes requested):
there are regions in between buttons where the dropdown appears but you cannot yet click. let's make those regions clickable.

and I agree, let's add a delay -- I think you can do that in pure CSS now?

tonyfettes:
> there are regions in between buttons where the dropdown appears but you cannot yet click. let's make those regions clickable.

@cyrus- Can I instead let the dropdown menu don't appear when user hovers over these regions? It will be a lot easier. Otherwise I have to either move the onclick callback to submenu, or have to deal with the wired interplay of css margin/padding between submenu and the buttons.

tonyfettes:
> there are regions in between buttons where the dropdown appears but you cannot yet click. let's make those regions clickable.

I don't think do this for `#editor-mode`: image would be a wise option.

Other parts is made clickable by now. Also a delay of 500ms is added.

cyrus-:
@disconcision @tonyfettes can one of you rebase this on `adt-defs-after` since there are a bunch of changes to the hazel chrome there?

disconcision:
with regards to the current nomenclature in the code: this PR is basically tooltips rather than submenus, correct? or is the infrastructure designed to be for both?

i think the delay should still be longer... it's hard to use a button without actually activating it. i think vscode is actually closer to 2000ms; though it also switches to no delay if you've already triggered it for another menu entry, but this is a bit more work to achieve (doubt it's possible without adding the ui state to the model) so maybe not worth it

disconcision:
@tonyfettes happy to do a quick merge on top of adt-defs-after if thats helpful

tonyfettes:
@disconcision This PR is originally is intend to submenu. Currently I implement 2 widget `submenu_switch` and `submenu_label`, and here is how the switch looks.
image

tonyfettes:
@disconcision This is how it looks like now:

[recording.webm](https://github.com/hazelgrove/hazel/assets/29998228/55f686ea-39e4-4221-9e9c-8a6ecbda86a2)

Any suggestions?

-----

Just for documentation purpose, I change follow part:
1. Delay of tool bar buttons is now 1000ms
2. The area around the icon is **NOT** clickable not. Make them clickable is a bit tricky, and since now the toolbar icons don't show a borderline when user hover over them for aesthetic reasons, it might be better not make them clickable.
3. I merged the two buttons for exporting scratchpad and persistent state value into one submenu

disconcision:
@tonyfettes looks good! I think we should consistently keep everything left-aligned though instead of centered (referring to both the sidemenu submenus and the dropdown submenus). (optional: i think we should also consider leaving room for icons associated with each entry to the left of the text labels, even if we don't use an icon for each entry as in this case)

also for the sidemenu submenu example it feels a bit awkward to have the top toggle to the left and then subsequent toggles to the right... not totally sure but I think we should have something more like a single icon representing the whole submenu, maybe a gear with a tau on it to represent 'dynamics settings', and then have both toggles inside the submenu. i'd keep both toggles to left of the labels as well so that theyre (a) closer to the cursor and (b) vertically aligned.

one final issue: while i think the longer delay is appropriate for a tooltip, it runs counter-purposes for the drop-down submenu. this is one reason why i'm wary of conflating these features; they have very different usability requirements, to the point where my inclination is to have them as separate implementations (or simply continue to use the native tooltips for now) as opposed to combining them). to be concrete, the reason for the delay in the tooltips it to accommodate both expert and beginner users. for experts who know what the button does, the appearance of the tooltip is a distraction; however they are unlikely to linger over the button, they will simply click it or not, which affords us the opportunity to 'overload' the action of lingering over the button to provide added context to beginners, who are likely uncertain what the button does, while minimizing the chance of distracting the experts. however, there is no advantage for either experts or beginners in delaying the appearance of the submenu, (beyond a very short base delay <250ms to account for incidental mouseovers) as this is the only way to access the contained affordances.

tonyfettes:
@disconcision It looks a bit strange that so many knobs are crowded over a small area...
image

Putting them on the right hand side make the layout more balanced, in my opinion.
image
(The padding seems to be a bit too narrow, and I'll just it to be wider)

The dropdown submenu looks like this now:
image
This design doesn't give user a clear clue on which button this submenu belongs to...

Thank you for the explanation on the delay of tooltip and submenu, it's detailed and helpful.

I'm re-evaluating my design & implementation here, and I think they're not quite yet ready for providing the functionality of a submenu... Shall we make this PR just do tooltip instead?

tonyfettes:
@disconcision Hi, sorry for the long silence. I just change this PR to just add tooltips because I don't think my current solution looks OK to myself. You can refer to the modified PR summary at the beginning of the thread to see a preview video of the newest tooltips, any suggestions/opinions are welcomed.

cyrus-:
image
weird box shows up temporarily when opening the left sidebar

tonyfettes:
@cyrus- The wired box is fixed now.

cyrus-:
Subsumed by #1165 which has been merged, so closing. Thanks for working through these design iterations @tonyfettes @disconcision. We can continue to iterate on Hazel UI things as we progress.

PR #990, opened 2023-02-15
Algebraic Datatypes II

### LINKED ISSUE #608: Algebraic Datatypes with pattern matching
Hazel 2 effort: https://github.com/hazelgrove/hazel/pull/491

### LINKED ISSUE #626: Better exception reporting

### LINKED ISSUE #860: invalid syntax errors not being decorated, getting weird values
the 'assignment' "n=0" isn't getting the nul-sort red highlighting, and is apparently evaluating to the value of n:
Screen Shot 2022-09-24 at 3 19 32 PM

### LINKED ISSUE #909: no syntax error shown for bad operators
2022-10-10 22_11_51-hazel

string comparison operator is wrong

### LINKED ISSUE #1002: Molding issue when changing empty pattern/type list/tuple to nonempty
This is an outstanding remolding issue from #943. Basically whenever an insertion happens on "[|]" or "(|)", and the expected sort is anything other than expression, the inserted tokens won't be remolded to the appropriate sort until an action is taken outside the list/tuple, e.g. pressing space here: "[molded_wrong]|" => "[molded_right]| ".

I'm merging #943 as I think it's a good idea to have in as a functional reference, and this should be fixed with the new syntax model, but if we end up doing a study or whatever with the current version we can probably rig up a hacky hotfix for this.

### LINKED ISSUE #1036: Proposal to clarify naming around co-ctx/free
Currently the co_ctx instance in cursorinfo is called 'free', which is sub-optimal as it can be misleading (my bad). I propose (as a separate PR from #1033 for git clarity) that we just rename 'free' to 'co_ctx' everywhere. Then, we can use the following vocabulary to make precise the notions of free/bound and used/unused:

- A locally free variable (in an expression) is one that occurs in the co_ctx of that expression
- A global free variable (in the program) is one that occurs in the co_ctx but not the ctx of _some_ expression
- A locally unused variable (in an expression) is one the occurs in the ctx but not the co-ctx of that expression
- A global unused variable (in the program) is one that occurs in the ctx but not the co_ctx of _some_ expression

theorems(?):
- to determine if a variable is globally free, it suffices to consider expressions which are variable references (locus of ctx lookups)
- to determine if a variable is globally unused, it suffices to consider expressions which are the bodies of binding forms (locus of ctx extensions)

@cyrus above is maybe a stilted way of looking at things but seems sort of interesting in a kind of galois connectiony kind of way?

### LINKED ISSUE #1041: Can't move caret up/down in certain postions
In this state I can't move the caret up from the 1 to the (, or if on the (, i can't move down to the 1.

My assumption is that this is approximately about the edge case where the rightmost valid position on the line above is left of the starting position, or vice-versa.

Screen Shot 2023-06-06 at 12 54 36 PM

### LINKED ISSUE #1044: Functions defined in list patterns cause exceptions
The input 
`let [a] = [fun x -> x+1] in a(1)`
will cause (Exception "Haz3lcore.EvaluatorError.Exception(_)")

On the other hand, it works well if we type-annotate either the list or `a` itself. But bad syntax in the annotation will cause the same exception.

### LINKED ISSUE #1045: casting bug when double-annotating function parameters
I don't know how to fully characterize this bug; here is a minified example. A CastBVHoleGround exception is thrown if I press ":" to insert a type annotation on the parameter 'b':

Screen Shot 2023-06-08 at 10 43 57 PM

Here is the elaboration and the specific cast triggering the exception:
Screen Shot 2023-06-08 at 10 44 22 PM
Screen Shot 2023-06-08 at 10 44 57 PM

@cyrus- any thoughts?

### LINKED ISSUE #1065: Fix capitalization of `elaborator.re` module

### LINKED ISSUE #1072: Type system error with internally inconsistent list literals in analytic position
Since the beginning of hazel 3 i think. listlits with inconsistent elements are essentially treated as if they have unknown type instead of list(unknown)

Screen Shot 2023-08-04 at 10 07 55 PM

### LINKED ISSUE #1073: Extraneous space appears when inserting space into an empty list/tuple
Screen Shot 2023-08-05 at 9 38 14 PM

should only be the hole

### LINKED ISSUE #1074: Pressing newline in string results in internal arrow symbol insertion
Screen Shot 2023-08-05 at 9 39 59 PM

### LINKED ISSUE #1075: Caret positioning and remolding bug for application on empty tuple
As mentioned in #1048. Currently, when deleting the sole argument of an application, the application is replaced by an infix hole, and the caret position becomes invalid, appearing visually to be offset to the left relative to expectation, and with subsequent rightward caret movement behaving unexpectedly until a vertical move is made:
Screen Shot 2023-08-05 at 9 49 32 PM
Relatedly, inserting an argument in that empty tuple fails to trigger a remolding into the now-viable ap: 
Screen Shot 2023-08-05 at 9 49 49 PM

disconcision:
Replaces #960.
Closes #608, Closes #626, Closes #860, Closes #909, Closes #1002, Closes #1036, Closes #1041, Closes #1044, Closes #1045, Closes #1065, Closes #1072, Closes #1073, Closes #1074, Closes #1075.

**Abstract**

This extends Hazel's type system with type aliases, anonymous labelled sum types, and recursive types, although the latter are not surfaced in the syntax, and are created implicitly in named sum type definitions. Additionally, this PR makes a number of other significant functional and internal changes as detailed below, as well as the bug fixes indicated above.

Screen Shot 2023-08-06 at 2 57 02 PM

**Sum types**
1. Sum type constructors are either nullary (have no argument; their type is the sum type), or unary (having arrow type from their parameter type to the sum type).
2. Unary constructors are valid on their own as expressions; you can pass them as arguments and map over them. However we have restricted pattern-matching on bare unary constructors to avoid confusion.
3. If a sum type is defined in an alias, its constructors are added to the context (a new type of context entry, albeit one that behaves the same as regular variable entries). This allows its constructors to be used, even in synthetic position, anywhere in the scope of the relevant type alias.
4. OTOH, if an anonymous sum type is created as a type annotation, its constructors are NOT added to the context, and are only valid in positions which are analytic against that type.
5. Two sum types are considered equal if they have the same constructors with the same types; order doesn't matter.
6. Two sum types are considered consistent if they have the same constructors with consistent types.
7. Constructor name must be unique within a type, but can shadow the constructors of previous types.
8. There is currently no way refer to a shadowed constructor in synthetic position; you must make the position analytic against the appropriate type through a type annotation

**Type aliases**

1. A type alias must have a capitalized name.
2. Type aliases can shadow other aliases, but do not shadow base types
3. Type aliases do not escape their scope; the returned type of a type alias expression has the type alias substituted away.

**Deprecations**

- Previous built-in ADTs removed (Color, ALFExpr, ALFTyp, Option_ALFTyp)

**Other Functional Changes**
- New editor mode: Examples. This is a collection of named editors, each consisting of examples intended either for exposition or testing.
Screen Shot 2023-08-06 at 2 39 02 PM

- New operators:  New unary operator `!` for boolean negation. New binary operators: `@` for list concatenation, `++` for string concatenation, `!=` for integer inequality, `!=.` for float inequality, and `\/` for boolean disjunction. The latter is temporary syntax since in current syntax engine, leading delimiters must have unique prefixes, and `||` overlaps with `|` used in case rules.
Screen Shot 2023-08-06 at 2 51 41 PM

- Improved Builtins: There are now ~40 builtin functions for int, floats, and strings, all pass-throughs to their OCaml equivalents, extending those added in #1058 and #1062. The process of adding builtins has been refined to require fewer changes. Builtin functions are now added directly to the initial environment instead of spliced into the elaboration, so they no longer clutter the elaboration display (what you see when dynamics are turned off).
Screen Shot 2023-08-06 at 2 45 13 PM

- Elaboration/Evaluation exception reporting: Internal exceptions (implementation errors) during elaboration and evaluation are now caught and will now be reported in the results area under the cell. Previously these either crashed the editor or were inconsistently reported to the console. Pic-related is about-to-be-fixed issue #1044.
Screen Shot 2023-08-06 at 2 35 14 PM

- Improved Cursor Inspector messages: Messages now take into account the specific syntactic form as well as the typing information to provide more user feedback. Adding more specific feedback should now be more straightforward for implementors. Error and ok status messages are more consistent, and some technical details regarding unknown type provenance have been relegated to tooltips.
Screen Shot 2023-08-06 at 2 48 35 PM

- Improved Context Inspector: Visual overhaul to support heterogenous context entries, which now include type aliases; long types are now truncated by default, but are visible on-hover.
Screen Shot 2023-08-06 at 2 42 43 PM

- Hole-based navigation: You can use TAB to move to the next hole, and SHIFT-TAB to move to the previous. If the top backpack item it currently droppable, the drop action will preempt the move action.
- Token merging change: Previously, if you deleted an infix operator/grout, all whitespace between the enclosing operands would be removed. This feels overly aggressive, and confusing in some situations, so I've removed this behavior. The one downside is that, in situations like `1 +|2`, pressing backspace will result in  `1 |><2` and again will result in `1|><2`, so to remove the infix grout, we'd need to move right and backspace. This was the primary reason for instituting the aggressive merge behavior, but I now believe that, seeing as there are downsides to both options, we should default to the more predictable one.
- New menu action: Export persistent data: This saves a file called Init.ml which serializes settings and the states of the example and scratch pad editors. This file can be used to replace the identically-named file in the Haz3lweb directory in order to specify initial app state. Note that due to technical issues Exercises mode editors and Langdocs settings are not currently serialized.

**Internal Changes**

- Statics Refactoring: Statics is now divided into Statics.Info, which defines the basic info data structured, and Statics.Map, which concerns the buildup of the Id map. Although this is not a perfectly realized division, the intention is that Statics.Info focusses on concentrates the logic for calculating statics attributes, and Statics.Map focusses on the tree traversal dataflow to collate the information necessary for calculation. Furthermore, Self and Mode have been split off into separate modules; all non-trivial logic related to calculating the self type of a term, or the mode of its children, should be located in those modules. See the tops of all of these modules for more documentation. This refactor includes converting the buildup of the map to a traversal from its previous bottom-up union-based approach, resulting in a 30x-speedup to statics generation. 
- Code views can now specify a custom root sort (besides just Exp); useful for e.g. Langdocs
- School Mode is internally renamed to Exercises Mode
- New sort, TPat, to support type alias definitions
- The 'free' entry in CursorInfo has been renamed 'co_ctx'; see CoCtx.re for documentation
- DHExp now uses same binary operator types as UExp, reducing redundancy
- All implementations in Typ and Ctx modules relocated to TypBase for dependency reasons


disconcision:
The only bugs I know about in this branch are syntactic ones involving remolding, including:
https://github.com/hazelgrove/hazel/pull/960#issuecomment-1415043321
https://github.com/hazelgrove/hazel/pull/960#issuecomment-1415044730
from the previous PR. These trigger a bug in remolding to which there is no obvious remedy, caused by using the same symbol for both the singleton sum type prefix and the binary sum type operator. The only workaround at this time would be using a different symbol for the prefix.

disconcision:
Test suite I:
`#Non-recursive sum/alias tests#
#all lines with trailing err comment should have 1 error#
#no other lines should have errors#

#type definitions: no errors#
type   =   in
type SingleNull = +One in
type Single = +F(Int) in
type GoodSum = A + B + C(Int) in
type Partial = Ok( ) +   in
type DoubleAlias = GoodSum in
type VerticalLeading =
  + A
  + B(GoodSum)
  + C(Bool->Bool) 
in

#incorrect or incomplete type definitions#
type badTypeName =   in #err: invalid type name#
type ( ,  ) =   in #err: invalid type name#
type   = badTypeToken in #err: invalid type token#
type NotASum = NotInSum(Bool) in #err: cons not in sum#
type Int =    in #err: shadows base type#
type Dupes =
  + Guy(Bool) #no err#
  + Guy(Int) #err: already used#
+ Guy in #err: already used#
type BadCons =
  + Um(Unbound) #err: unbound type var#
  + invalid #err: invalid#
  + Bool #err: expected cons found type#
  + Int(Int) #err: expected cons found type#
  + ( )(Int) #err: expected cons found type#
+ A(Bool)(Int) in #err: expected cons found app#

#nested sum defs dont add tags to scope#
type NestedAlias = (Int, Anonymous + Sum) in 
let x:NestedAlias = (1, Sum) in #err: not defined#

#unbound tyvars treated as unknown-typehole#
let a:Bad = 0 in a == 0; #err: not bound#

#non-sum-types cant be recursive#
type Lol = Lol in #err: not bound#

#no errors: analytic shadowing#
type Tork1 = +Blob in
type Tork2 = +Blob in 
let x:Tork1 = Blob in

#exp tests: happy#
type YoDawg =  Yo(Int) +Bo(Int)+ Dawg(Bool) in
let _ = Yo(1) in
let _ : YoDawg = Yo(2) in
let _ : +Yo(Bool) = Yo(true) in
let _ : (Yo + Dawg, Int) = (Dawg,5) in
let _ : DoubleAlias = C(4) in

#exp tests: errors#
let _ = 2(1) in #err: incons with arrow#
let _ = Undefined(1) in #err: cons undefined#
let _ = B("lol") in #err: type incons#
let _ : +Yo(Bool) = Yo in #err: type incons#
let _ : +Yo = Yo("lol") in #err: type incons#
let _ : +One = Yo(1) in #err: type incons#

#pat tests: happy (but refutable patterns so weird)#
let Yo = Bo in #kind of a weird edge#
let Yo(1) = Dawg(true) in
let Yo(1): YoDawg = Yo(1) in
let Yo(1): +Yo(Int) = Yo(1) in 
let Yo: +Yo = Yo in

#pat tests: errors#
let 2(1) = 3 in #err: incons with arrow#
let NotDefined(1) = 3 in #err: cons undefined#
let Yo = Dawg in #err: type incons#
let Yo(true) = Dawg(true) in #err: type incons#
let Yo: YoDawg = Yo(1) in #err: type incons#
let Yo(1): +Yo = Yo in #err: type incons#
let Yo(1): +Yo(Bool) = Yo(true) in #err: type incons#
"Thats all, folks"
`

disconcision:
More tests:
`#x type is normalized before leaving scope of P#
let escape = type In = Arp(In)+Bop in
  let a_in = Arp in 
  let Arp(_) = Bop in 
  (a_in,Bop)
in
let (a_out,b_out) = escape in
#happy 4x#
let Bop = a_out(Bop) in
let Arp = a_out in
let Bop = b_out in 
let Arp(Bop) = b_out in
case b_out
  | Bop => 1 
end`

`#recursive sum type dynamics tests#
#all calls should evaluate fully with no exns or cast fails#
type Exp = Var(String) + Lam(String, Exp) in

let s0 : ( ,  ,  ) ->   = fun (e,x,v) ->
  case e
    | Var(y) => (if y $== x then v else e)
    | Lam(y, e1) => (if y $== x then e else Lam(y, s0(e1,x,v))) 
end in 
let s1 : ( ,  ,  ) -> Exp = fun (e,x,v) ->
  case e
    | Var(y) => (if y $== x then v else e)
    | Lam(y, e1) => (if y $== x then e else Lam(y, s1(e1,x,v)))
end in 
let s2 : (Exp,  ,  ) -> Exp = fun (e,x,v) ->
  case e
    | Var(y) => (if y $== x then v else e)
    | Lam(y, e1) => (if y $== x then e else Lam(y, s2(e1,x,v)))
end in 
let s3 : (Exp,  , Exp) -> Exp = fun (e,x,v) ->
  case e
    | Var(y) => (if y $== x then v else e)
    | Lam(y, e1) => (if y $== x then e else Lam(y, s3(e1,x,v)))
end in

let in = Lam("b", Var("a")),"a",Var("x") in
(s0(in), s1(in), s2(in), s3(in))
`

cyrus-:
This is not currently shown as an error or warning, because `x` can match a pair, but I think it should show a warning when in the arg position of an ap because probably you don't want to match on the pair here. 
image

cyrus-:
The fact that this evaluates to 10 is I think very confusing. Let's discuss this in our meeting.
image

cyrus-:
Too many parens in DHExp view here
image

cyrus-:
Shadowing of type names does not play nicely with the type inspector / context inspector feedback. Continues to use a name that is not actually available anymore. Need to either handle this by expanding out the type alias to the structural type in this situation, or not allow type aliases.

image

cyrus-:
The type here is just straight up incorrect due to the issue above:

image

cyrus-:
Some things other than ADTs are bundled into this PR. I suppose that's fine, but small tweaks would be good there:

It would be good for the context inspector to only show up when hovering over the Gamma symbol, rather than anywhere in the type inspector. Currently the hover indicator only shows up over the Gamma but the context inspector shows up when hovering over the type inspector.
image

disconcision:
@cyrus- Fixed context hover target and dhexp over-parenthesization (although neither are specific to this PR; the latter occurs for stuck function terms. The latter fix may have unintended consequences; previously tuples were obligate-parenthesized; not sure if this is load-bearing).
After some thought, I think disallowing type alias shadowing is the way to go. I've done this for now, we can talk about it this afternoon along with warnings.

disconcision:
@cyrus on reflection i don't think the tuple thing (first comment) is that bad. if they try to use x they'll immediately find out its a tuple. if they don't, they'll get (eventually) an unused variable warning. if they use a wildcard... we do want to allow a wildcard in such situations i think, or at least i don't find it problematic when ocaml/reason does it.

If we decide these kinds of things are issues, the principled solution would seem to be to be making tuples obligate parenthesized a la reason. but not sure the this is strictly necessary.

also if we make this a warning for patterns we'd probably want to make a warning for the equivalent expression situation.

i think to be convinced this is worth a specific intervention i'd want to see a concrete example where confusion would arise that would not produce a warning which already on the roadmap

disconcision:
For the shadowing issue, I'm still of the opinion that we should disallow for now. It's not too hard to do the gamma-level replacement, but it feels sort of dirty to me, and pre-module there aren't really any desirable cases for type-name shadowing. Worth revisiting for modules though. If you're still keen on allowing shadowing I'll make the change though.

Actually in OCaml I don't think I can actually shadow a type via a module? Because one will always be qualified:
```
module AA = {
  type qq = A;
  let a:qq = A;
};
type qq = B;
let c:qq = B;
let b:qq = AA.a; //This expression has type AA.qq but an expression was expected of type qq
let d:qq = c;

module AA = {
  type qq = A;
  let a:qq = A;
};
type qq = B;
let c:qq = B;
open AA;
let b:qq = AA.a;
let d:qq = c; //This expression has type qq but an expression was expected of type AA.qq
```

disconcision:
For capturing (non-nullary) constructors with pattern variables, leaning reluctantly towards disallowing it. Added machinery to address cases like I think you were mentioning eg
```
type T = A(Int) + B(String) in
case A(4)
  | _(x) => x+1
end
```
not obviously worth it. tho seems possible (put an inconsistent-branches-style error on pattern x, which can be resolved via annotation).

src/haz3lweb/www/index.html
```diff
@@ -21,6 +21,15 @@
         <p style="font-size:4em; color: black;">loading</p>
       </div>
     </div>
+    <svg style="visibility: hidden; position: absolute;" width="0" height="0" xmlns="http://www.w3.org/2000/svg" version="1.1">
```
cyrus-: can you leave a brief comment explaining what this is for

src/haz3lcore/lang/Form.re
```diff
@@ -84,17 +84,16 @@ let is_reserved = str => is_bool(str);
 let is_var = str => !is_reserved(str) && regexp("^[a-z][A-Za-z0-9_]*$", str);
 let is_capitalized_name = regexp("^[A-Z][A-Za-z0-9_]*$");
 let is_tag = is_capitalized_name;
-let is_concrete_typ = str =>
+let is_base_typ = str =>
   str == "String" || str == "Int" || str == "Float" || str == "Bool";
```
cyrus-: I believe correct way to do string comparisons to avoid physical equality issues is using String.equal. Could also do a dysjunction in a regexp.

src/haz3lcore/lang/Form.re
```diff
@@ -84,17 +84,16 @@ let is_reserved = str => is_bool(str);
...
-let is_typ_var = t => is_capitalized_name(t) && !is_concrete_typ(t);
-let is_partial_concrete_typ = x =>
-  !is_concrete_typ(x) && is_capitalized_name(x);
+let is_typ_var = t => is_capitalized_name(t) /*&& !is_base_typ(t)*/;
```
cyrus-: can remove this comment

src/haz3lcore/tiles/Skel.re
```diff
@@ -170,7 +171,11 @@ module Stacks = {
     | (_, Some((l, r))) =>
       let is = List.map(fst, chain);
       let split_kids = n =>
-        ListUtil.split_n(n, stacks.output) |> PairUtil.map_fst(List.rev);
+        try(ListUtil.split_n(n, stacks.output) |> PairUtil.map_fst(List.rev)) {
+        | _ =>
+          print_endline(show(stacks));
+          failwith("blah");
```
cyrus-: either remove or more clear message

src/haz3lcore/tiles/Skel.re
```diff
@@ -88,6 +88,7 @@ let rel = (p1: Piece.t, p2: Piece.t): option(rel) =>
         lbl1(case) && lbl2(rule),
         lbl1(rule) && lbl2(rule),
         lbl1(comma) && lbl2(comma) && t1.mold == t2.mold,
+        lbl1(["+"]) && lbl2(["+"]) && t1.mold == t2.mold,
```
cyrus-, line 91: would be good to put special cases like this in lang rather than hard coding them here, but okay for now until next tylr

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+    | Var(Token.t)
+    | List(t)
+    | Arrow(t, t)
+    | Sum(sum_map)
+    | Prod(list(t))
+    | Rec(Token.t, t)
+  and sum_map = VarMap.t_(option(t));
+
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type sum_entry = (Token.t, option(t));
```
cyrus-: would be good to define a separate type for sum type constructors rather than using Token everywhere

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+    | Int
+    | Float
+    | Bool
+    | String
+    | Var(Token.t)
+    | List(t)
+    | Arrow(t, t)
+    | Sum(sum_map)
+    | Prod(list(t))
+    | Rec(Token.t, t)
```
cyrus-: would be good to define a separate type for type variables rather than just using bare token (also for Var above)

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type sum_entry = (Token.t, option(t));
+
+  /* MODE: The (analytic) type information derived from a term's
+     syntactic context. This can either Syn (no type expectation),
+     or Ana (a type expectation). It is conjectured [citation needed]
+     that the Syn mode is functionally indistinguishable from
+     Ana(Unknown(SynSwitch)), and that this type is thus vestigial. */
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type mode =
```
cyrus-: perhaps mode shouldn't be in the Typ module since its used in info maps / type checking only

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+    id: Id.t,
+    mode: Typ.mode,
+  };
+
+  /* Each co-context entry is a list of the uses of a variable
+     within some scope, including their type demands */
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type co = VarMap.t_(list(co_entry));
+
+  let extend: (entry, t) => t;
```
cyrus-: why is the t argument second here?

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+  let lookup: (t, Token.t) => option(entry);
+  let add_abstract: (t, Token.t, Id.t) => t;
```
cyrus-: why do we need this helper function instead of just constructing the appropriate kind?

src/haz3lcore/statics/Typ.re
```diff
@@ -82,77 +17,6 @@ let join_type_provenance =
   | (SynSwitch, SynSwitch) => SynSwitch
   };
```
cyrus-: I think Internal should dominate over TypeHole, since a joined Unknown is an internal operation

src/haz3lcore/statics/Typ.re
```diff
@@ -82,77 +17,6 @@ let join_type_provenance =
...
-  | Just(t) => t
-  | Joined(wrap, ss) =>
-    switch (ss |> List.map(s => s.ty) |> join_all) {
-    | None => Unknown(Internal)
-    | Some(t) => wrap(t)
-    }
-  | Multi
-  | Free(_) => Unknown(Internal);
-
 /* MATCHED JUDGEMENTS: Note that matched judgements work
```
cyrus-: hm, I think this won't cause any bugs but in terms of clarity, it seems to wrap up two separate mechanisms. would be better for the caller to have the handle the situation where a non-empty hole is presumably going to be inserted and pass in the correct type (Unknown) to matched_arrow/list/etc.

src/haz3lcore/statics/Typ.re
```diff
@@ -164,6 +28,12 @@ let matched_arrow: t => (t, t) =
   | Unknown(prov) => (Unknown(prov), Unknown(prov))
   | _ => (Unknown(Internal), Unknown(Internal));
 
+let matched_list: t => t =
+  fun
+  | List(ty) => ty
+  | Unknown(prov) => Unknown(prov)
```
cyrus-: Shouldn't maintain prov here, since its a new internally generated type. Should be Unknown(Internal). (same with matched_arrow above)

src/haz3lcore/statics/Typ.re
```diff
@@ -173,30 +43,24 @@ let matched_arrow_mode: mode => (mode, mode) =
       (Ana(ty_in), Ana(ty_out));
```
cyrus-: why do you need a mode for the return type?

src/haz3lcore/statics/Typ.re
```diff
@@ -173,30 +43,24 @@ let matched_arrow_mode: mode => (mode, mode) =
...
-  fun
-  | List(ty) => ty
-  | Unknown(prov) => Unknown(prov)
-  | _ => Unknown(Internal);
-
 let matched_list_mode: mode => mode =
   fun
   | SynFun
   | Syn => Syn
   | Ana(ty) => Ana(matched_list(ty));
```
cyrus-: shouldn't this be `Ana(List(matched_list(ty))`?

src/haz3lcore/statics/Typ.re
```diff
@@ -173,30 +43,24 @@ let matched_arrow_mode: mode => (mode, mode) =
...
+    List.map(ty => Ana(ty), ana_tys)
+  | Ana(Unknown(prod)) => List.init(length, _ => Ana(Unknown(prod)))
+  | _ => List.init(length, _ => Syn)
+  };
 
-/* Legacy code from HTyp */
+let matched_list_lit_modes = (mode: mode, length): list(mode) =>
+  List.init(length, _ => matched_list_mode(mode));
 
+/* Legacy precedence code from HTyp */
```
cyrus-: This is still used by HTypDoc I believe, so don't need to say it is legacy

src/haz3lcore/statics/Typ.re
```diff
@@ -209,18 +73,44 @@ let precedence = (ty: t): int =>
...
+  | Rec(x, ty_body) => subst(ty, x, ty_body)
+  | _ => ty
+  };
+
 /* equality
    At the moment, this coincides with default equality,
    but this will change when polymorphic types are implemented */
-let rec eq = (t1, t2) =>
+let rec eq = (t1: t, t2: t): bool => {
+  let eq' = eq;
```
cyrus-: guessing this is left over from debugging

src/haz3lcore/statics/Typ.re
```diff
@@ -209,18 +73,44 @@ let precedence = (ty: t): int =>
...
+  };
+};
+
+let unroll = (ty: t): t =>
+  switch (ty) {
+  | Rec(x, ty_body) => subst(ty, x, ty_body)
+  | _ => ty
+  };
+
 /* equality
```
cyrus-: alpha-equivalence not (syntactic) equality

src/util/TagMap.re
```diff
@@ -0,0 +1,98 @@
...
+[@deriving sexp]
+type key = string;
+
+[@deriving sexp]
+type binding('a) = (key, 'a);
+
+[@deriving sexp]
+type t('a) = list(binding('a));
+
+let tag_equal = (==);
```
cyrus-: should be String.equal?

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+  | Arrow(t1, t2) => free_vars(~bound, t1) @ free_vars(~bound, t2)
+  | Sum(sm) =>
+    ListUtil.flat_map(
+      fun
+      | None => []
+      | Some(typ) => free_vars(~bound, typ),
+      List.map(snd, sm),
+    )
+  | Prod(tys) => ListUtil.flat_map(free_vars(~bound), tys)
+  | Rec(x, ty) => free_vars(~bound=[x] @ bound, ty)
```
cyrus-: `[x] @ bound` is more efficiently written `[x, ...bound]`

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+  | (Unknown(_), ty)
+  | (ty, Unknown(_)) => Some(ty)
+  | (Var(n1), Var(n2)) =>
+    if (n1 == n2) {
+      Some(Var(n1));
+    } else {
+      let* ty1 = Ctx.lookup_alias(ctx, n1);
+      let* ty2 = Ctx.lookup_alias(ctx, n2);
+      let+ ty_join = join'(ty1, ty2);
+      resolve ? ty_join : Var(n1);
```
cyrus-: I believe it is incorrect to do this unless the variable is equivalent to the join (join might in general cause Unknowns to be inserted, and returning the Var could not reflect that)

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+    }
+  | (Var(name), ty)
+  | (ty, Var(name)) =>
+    let* ty_name = Ctx.lookup_alias(ctx, name);
+    let+ ty_join = join'(ty_name, ty);
+    resolve ? ty_join : Var(name);
+  /* Note: Ordering of Unknown, Var, and Rec above is load-bearing! */
+  | (Rec(x1, ty1), Rec(x2, ty2)) =>
+    let+ ty_body = join(ctx, ty1, subst(Var(x1), x2, ty2));
```
cyrus-: need to add x1 to ctx here

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+  | (Prod(tys1), Prod(tys2)) =>
+    let* tys = ListUtil.map2_opt(join(ctx), tys1, tys2);
+    let+ tys = OptUtil.sequence(tys);
+    Prod(tys);
+  | (Prod(_), _) => None
+  | (Sum(sm1), Sum(sm2)) =>
+    let* ty =
+      ListUtil.map2_opt(
+        join_sum_entries(ctx),
+        TagMap.sort(sm1),
```
cyrus-: would be good to retain the order of keys (for UI's sake) if it happens to be the same on both sides

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+  };
+
+let join_all = (ctx: Ctx.t, ts: list(t)): option(t) =>
+  List.fold_left(
+    (acc, ty) => OptUtil.and_then(join(ctx, ty), acc),
+    Some(Unknown(Internal)),
+    ts,
+  );
+
+let rec normalize_shallow = (ctx: Ctx.t, ty: t): t =>
```
cyrus-: this is called (weak) head normalization

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+       as in current implementation Recs do not occur in the
+       surface syntax, so we won't try to jump to them. */
+    Rec(x, normalize(Ctx.add_abstract(ctx, x, -1), ty))
+  };
+};
+
+let sum_entry = (tag: Token.t, tags: sum_map): option(sum_entry) =>
+  List.find_map(
+    fun
+    | (t, typ) when t == tag => Some((t, typ))
```
cyrus-: see above re:tag equality

src/haz3lcore/statics/Typ.re
```diff
@@ -231,16 +121,214 @@ let rec eq = (t1, t2) =>
...
+    | Some(ty_in) => Arrow(ty_in, ty_ana)
+    };
+  | _ => None
+  };
+};
+
+let tag_mode = (ctx: Ctx.t, mode: mode, name: Token.t): option(mode) =>
+  switch (tag_ana_typ(ctx, mode, name)) {
+  | Some(Arrow(_) as ty_ana) => Some(Ana(ty_ana))
+  | Some(ty_ana) => Some(Ana(Arrow(Unknown(Internal), ty_ana)))
```
cyrus-: not sure I understand what's going on here?

cyrus-: oh, this is only used for aps -- but why is it Unknown?

src/haz3lcore/statics/Typ.re
cyrus-: substition is with a open term so need to do what we discussed

src/haz3lcore/statics/Ctx.re
```diff
@@ -1,70 +1,71 @@
-open Sexplib.Std;
+include TypBase.Ctx;
```
cyrus-, line 1: why is some stuff in TypBase.Ctx and some stuff here? makes it hard to predict where to look for a definition / put a new definition. (also relevant to some other things)

I'd suggest we just put it all in Ctx / Typ and these files just include those

src/haz3lcore/statics/TypBase.re
```diff
@@ -0,0 +1,186 @@
...
+  type t = list(entry);
+
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type co_entry = {
+    id: Id.t,
+    mode: Typ.mode,
+  };
+
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type co = VarMap.t_(list(co_entry));
```
cyrus-: I think CoCtx should be its own module

src/haz3lcore/statics/TermBase.re
```diff
@@ -311,17 +318,20 @@ and UTyp: {
     | String
     | List(t)
     | Var(string)
+    | Tag(string)
     | Arrow(t, t)
     | Tuple(list(t))
     | Parens(t)
+    | Ap(t, t)
+    | USum(list(t))
```
cyrus-: What is the "U" in USum?

Also not clear to me we should have it be a `list(t)` in this representation -- a sum type isn't a list of types, it's a list of sum type entries which are either unary tags or tags with arguments. why can't that be the structure here?

src/haz3lcore/statics/Term.re
```diff
@@ -92,9 +102,97 @@ module UTyp = {
...
+      | Bool => Bool
+      | Int => Int
+      | Float => Float
+      | String => String
+      | Var(name) =>
+        //Var(name)
+        //Ctx.is_alias(ctx, name) ? Var(name) : Unknown(TypeHole)
+        switch (Ctx.lookup_tvar(ctx, name)) {
+        | Some(_) => Var(name)
+        | None => Unknown(TypeHole)
```
cyrus-: I think we should have a separate representation in Typ for this case, to retain the name of the unbound type

src/haz3lcore/statics/Statics.re
```diff
@@ -328,192 +141,188 @@ and uexp_to_info_map =
...
-      ~self=Just(List(ty1)),
-      ~free=Ctx.union([free1, free2]),
-      union_m([m1, m2]),
+      ~self=Info.join(ty => List(ty), tys, ids, ctx),
+      ~free=Ctx.union(List.map(Info.exp_free, es)),
+      m,
     );
+  | Cons(e1, e2) =>
+    let (e1, m) = go(~mode=Typ.matched_list_mode(mode), e1, m);
+    let (e2, m) = go(~mode=Ana(List(e1.ty)), e2, m);
```
cyrus-: why isn't this `Ana(List(Typ.matched_list_mode(mode)))`?

src/haz3lcore/statics/Statics.re
```diff
@@ -328,192 +141,188 @@ and uexp_to_info_map =
...
+         in the definition would be oblierated. But we need to check for free
+         variables to decide whether to make a recursive type or not. So we
+         tentatively add an abtract type to the ctx, representing the
+         speculative rec parameter. */
+      let (ty_def, ctx_def, ctx_body) = {
+        let ty_pre =
+          UTyp.to_typ(Ctx.add_abstract(ctx, name, Id.invalid), utyp);
+        switch (utyp.term) {
+        | USum(_) when List.mem(name, Typ.free_vars(ty_pre)) =>
+          let ty_rec = Typ.Rec("α", Typ.subst(Var("α"), name, ty_pre));
```
cyrus-, line 309: why can't we use the name instead of α?

src/haz3lcore/statics/Statics.re
```diff
@@ -328,192 +141,188 @@ and uexp_to_info_map =
...
+      ~self=Info.join(Fun.id, e_tys, branch_ids, ctx),
+      ~free=Ctx.union([scrut.free] @ e_frees),
+      m,
+    );
+  | TyAlias(typat, utyp, body) =>
+    let m = utpat_to_info_map(~ctx, ~ancestors, typat, m) |> snd;
+    switch (typat.term) {
+    | Var(name)
+        when !Form.is_base_typ(name) && Ctx.lookup_alias(ctx, name) == None =>
+      /* NOTE(andrew): This is a slightly dicey piece of logic, debatably
```
cyrus-: seems okay to me

src/haz3lcore/statics/Statics.re
```diff
@@ -328,192 +141,188 @@ and uexp_to_info_map =
...
+        | Some(sm) => Ctx.add_tags(ctx_body, name, UTyp.rep_id(utyp), sm)
+        | None => ctx_body
+        };
+      let (Info.{free, ty: ty_body, _}, m) =
+        go'(~ctx=ctx_body, ~mode, body, m);
+      /* Make sure types don't escape their scope */
+      let ty_escape = Typ.subst(ty_def, name, ty_body);
+      let m = utyp_to_info_map(~ctx=ctx_def, ~ancestors, utyp, m) |> snd;
+      add(~self=Just(ty_escape), ~free, m);
+    | _ =>
```
cyrus-: not a good place for a catchall since we might add other cases to type patterns in the future

src/haz3lcore/statics/Statics.re
```diff
@@ -522,141 +331,142 @@ and upat_to_info_map =
...
-    }
+    let (hd, m) = go(~ctx, ~mode=Typ.matched_list_mode(mode), hd, m);
+    let (tl, m) = go(~ctx=hd.ctx, ~mode=Ana(List(hd.ty)), tl, m);
+    add(~self=Just(List(hd.ty)), ~ctx=tl.ctx, m);
   | Wild => atomic(Just(unknown))
   | Var(name) =>
-    let typ = typ_after_fix(mode, Just(Unknown(Internal)));
-    let entry = Ctx.VarEntry({name, id: Term.UPat.rep_id(upat), typ});
-    add(~self=Just(unknown), ~ctx=Ctx.extend(entry, ctx), Id.Map.empty);
+    /* Note the self type assigned to pattern variables (unknown)
```
cyrus-: not sure I understand this -- shouldn't the expected type be added to the context when the mode is analytic?

cyrus- (changes requested):
In addition to comments, the context inspector is now in all caps for some reason

cyrus-:
@disconcision I made a 75% review of the code -- didn't look at elaboration or haz3lweb stuff, but there is a bunch here. can you review and we can discuss this week?

src/haz3lcore/statics/Statics.re
disconcision: Error prone; easy to make substitution-related implementation errors which are hard to track. AFAICT the only disadvantage is CI, and this should be dealt with at another level of abstraction

src/haz3lcore/statics/Statics.re
disconcision: That is what's happening, it's just slightly indirect. ty_after_fix takes the mode and handles this logic; it also handles the Syn/SynFun case, and the logic is generic.

disconcision: Yes. Clarified the comment, the point was supposed to be simply that SynSwitch is never added to the context

src/haz3lcore/statics/Typ.re
disconcision: See the statics for function literals

src/haz3lcore/statics/Statics.re
disconcision: because in the synthetic case, the mode for e2 is based on the synthesized type of e1

src/haz3lcore/statics/Ctx.re
disconcision: This organization is inherited from the type aliases branch. Presumably done to avoid circular dependency. Agreed that it's confusing but not immediately sure how to address it, other than to stick everything in TypBase?

src/haz3lcore/statics/TermBase.re
disconcision: Legacy naming. Renamed to Sum.
Yes, it should be implemented the way you suggest, but for things to be structured that way, it would require the introduction of a new sort; when I tried to do that, it was causing molding problems I wasn't able to resolve. My intention is that this should be reimplemented post-new-tylr.

src/haz3lcore/statics/Typ.re
disconcision: The current version seems right to me, in the sense of consistency within the code, but perhaps I'm interpreting matched list judgements in a different way than you are?

src/haz3lcore/statics/Typ.re
disconcision: Honestly not sure, it's been too long... I remember this and related fns being the diciest part of the logic, wasn't quite sure what I was doing here

src/haz3lcore/statics/Typ.re
disconcision: Changed this to only return Var(n1) if ty1 and ty_join are equal

src/haz3lcore/statics/Typ.re
disconcision: Don't totally follow this one, let's talk about it Wednesday

src/haz3lcore/statics/Typ.re
disconcision: Don't remember the details here, let's talk Wednesday.

disconcision (commented):
Addressed most comments. A few I've responded to; let's follow up tomorrow.

cyrus- (changes requested):
* The scratch slides with examples are great, but I think it would be better to create another scratch mode clone but called "Examples" mode for this kind of thing.
* Can you add examples that cover cons in analytic mode.
* `Singleton Sum {F(Int)}` seems unnecessary, why not just use a prefix `+F(Int)` which is how it is written in the code anyway?
* The error message "Not a valid type name" could be improved to "Not a valid type name (type names must be capitalized)" in cases where that is the issue, especially since many students will be familiar with OCaml's opposite convention.
* `isn't a valid type token` -- I don't think we want to use the word "token" in this way in error messages. Either say just "type" or "type variable". (lowercase type variables will become valid once we add polymorphism)
* Why isn't this example (on Slide 3) not working? Won't the right side of the pair be in analytic position and therefore go through the analytic rules for constructors?
image

(partial review, only looked at functionality)

gensofubi:
During my testing of the module branch, I have encountered numerous situations where it is necessary to determine whether two types are consistent. Seems that exception happens when judging consistency with sum types containing different number of constructors. For example, When I attempt to input the following content, it triggers (Exception "Invalid_argument(\"List.for_all2\")")

image


And I've noticed that recursive types are not consistent with their unrolled form. I wonder if this is an expected behavior.

image

cyrus-:
marking as draft pending resolution of above issue ^ @disconcision

disconcision:
@cyrus on your last two points: i'm not sure "type" is quite right in errors here as this error also shows up (for example) if you have a "*" operator in a type... perhaps "type or type operator"?

For your last point, the example dates from when this PR didn't support anonymous sums... there seems to be no reason why we can't support the example now... it looks like the reason it doesn't isn't because of anything to do with sum types, but rather at the type alias level, because I don't currently have a special case for analysis against type vars... so here, the var just passes through matched product, hits the fallthrough case, and hence the components get analyzed against unknown; consequently, if analysis is against a type alias, the error will show up on whole thing instead of being passed down to the component. I guess we should do (weak head) normalization before doing matched arrow, prod, and list? Does that make sense? Or do we need to think about a more general approach as to when to normalize?

disconcision:
@gensofubi the for_all2 bug should be fixed now, thanks for the bug report. the join behavior is intentional (@cyrus- iirc this is what we decided should be the behavior for isorecursive types?) some context here: https://cstheory.stackexchange.com/questions/38407/whats-the-definition-of-join-on-iso-recursive-types. i don't have this fully loaded now but iirc this is required for join to be consistent with the usual subtyping relation on isorecursive types. for equirecursive types, we would indeed have this join between two types which you can make equal by doing some amount of unrolls on both

disconcision:
All issues addressed; see specific comments. Added new examples editor mode.

src/haz3lcore/statics/Term.re
```diff
@@ -74,10 +80,13 @@ module UTyp = {
     | String
     | Bool => "Base Type"
     | Var => "Type Variable"
+    | Tag => "Sum Constructor"
```
cyrus-: since we call it a "Constructor" in user-facing code, might be worth renaming `Tag` to `Constructor` in the code as well?

src/haz3lcore/statics/Term.re
```diff
@@ -157,18 +245,18 @@ module UPat = {
     | EmptyHole => "Empty Pattern Hole"
     | MultiHole => "Multi Pattern Hole"
     | Wild => "Wildcard Pattern"
-    | Int => "Integer Literal"
-    | Float => "Float Literal"
-    | Bool => "Boolean Literal"
-    | String => "String Literal"
-    | Triv => "Trivial Literal. Pathetic, really."
+    | Int => "Integer Pattern Literal"
```
cyrus-: I think Pattern should be at the end here, "Integer Literal Pattern"

src/haz3lweb/Export.re
```diff
@@ -14,23 +14,26 @@ type all = {
 type all_f22 = {
   settings: string,
   scratch: string,
```
cyrus-, line 17: are we intentionally not exporting the state of Examples mode?

src/haz3lcore/statics/Term.re
disconcision: @cyrus- done

src/haz3lweb/ModelSettings.re
```diff
@@ -1,5 +1,20 @@
...
+  | Scratch
+  | Examples
+  | Exercise;
+
+let mode_of_string = (s: string): mode =>
+  switch (s) {
+  | "Scratch" => Scratch
+  | "Examples" => Examples
+  | "Exercise" => Exercise
+  | _ => Scratch
```
cyrus-: failwith rather than defaulting here to catch potential future bugs

src/haz3lweb/view/dhcode/layout/HTypDoc.re
```diff
@@ -105,17 +105,21 @@ let rec mk = (~parenthesize=false, ~enforce_inline: bool, ty: Typ.t): t => {
...
+          )
+          |> pad_child(~enforce_inline),
+          mk_delim("}"),
         ]),
         parenthesize,
-      );
+      )
+    | Sum(_) =>
+      //TODO: hdocs for labelled sum types
+      (text("(sum ...)"), parenthesize)
```
cyrus-: need to resolve this todo

src/util/Constructor.re
```diff
@@ -0,0 +1,7 @@
+//open OptUtil.Syntax;
+open Sexplib.Std;
```
cyrus-: not sure this should be in util since its related to a specific haz3lcore concept

src/util/ConstructorMap.re
```diff
@@ -0,0 +1,101 @@
+open OptUtil.Syntax;
```
cyrus-: same

cyrus- (changes requested):
* when the langdocs sidebar is visible and the code overflows underneath it, there is no horizontal scrolling
* error message for inconsistent element types in a list literal uses the word "branches" which is not correct:
image
* pending final fix for casting issue that we discussed

cyrus-:
@disconcision looks like some conflicts with `dev` have popped up

disconcision:
@cyrus- conflicts and comments addressed:

The principled fix for the horizontal scroll issue required upgrading top-level page layout to a css grid. The main #page element is now a 3 row 2 column grid, and the top-level elements (#top-bar, #bottom-bar, #side-bar, and #main) are all aligned to that grid. This does cause a variety of minor cascading changes so I'd suggest messing with things a bit and seeing if you're okay with the resulting scroll and visibility behaviors.

The 'branches' issue required exposing syntax class information in cursor inspector; I took advantage to revamp cursor inspector messages, emphasizing the syntax class more than before. I added some syntax-class specific statuses for e.g. empty holes, and also did a consistency and completeness pass over existing error and okay statuses. My goal for the former was to maintain readability for beginners while also exposing a more precise reflection of the underlying type system; let me know what you think.

cyrus- (dismissed):
Scrolling behavior is good, couple of small tweaks would be nice:

 - Give the langdocs sidebar a fixed width so it isn't changing width depending on the message inside (distracting)
 - Give code cells in scratch mode a minimum width, so they don't look odd when the code itself isn't very wide (e.g. an empty scratchpad)
 - Don't know how difficult this is, but when the backpack has multiple items and shows up on the first line of a scratch slide, it disappears under the header. It would be nice if the backpack showed up above everything.

The cursor inspector messages also look good overall. Some small suggestions:
- Instead of "new variable" I would say "variable binding"
- Pattern variables saying "? trivially satisfies expected type ..." reads weird to me. It can just give the expected type in that case I think.
- The phrase "trivially satisfies" in other cases also reads weird. Maybe "is consistent with"?
- The ExplainThis toggle could maybe be visually better conveyed as a "more details" icon of some sort next to the syntactic form

Other comments:
- Can some of the top bar instructor-only export buttons be removed now that we have the "Export Persistent Data" button in the left bar?

disconcision:
@cyrus-

layout:
1. langdocs constant width: That was the goal; there are complications but will fix.
3. code cell min width: already done
4. longstanding issue. AFAICT this is very difficult. the central problem is in both having the main panel scrolling without clipping the backpack. my research indicates that this is simply impossible if the backpack element is both logically inside and positioned with respect to the containing element. The only way I can see to do it would be to refactor such that the backpack element is logically outside of the main panel. This would necessitate some complex dataflow changes, in particular for multi-editor layouts like exercises mode. A project unto itself, in any case.

ci:
variable binding: done

"trivially satisfies": so previously we had special-case messages for when either the syn or ana type was unknown, suppressing the display of those types. unfortunately, this hides the unknown provenance which is functionally load-bearing. A case could be made for an expert mode which shows the provenance, hiding it otherwise. would you be satisfied with a single generic message "<syn type> satisfies expected type <ana type>" showing in all instances including unknowns (except still having a special cased "<syn type> equals expected type")? if you're keen on suppressing the unknowns i guess a quick alternative would be just having the unknown prov show up as a hover. I sort of feel though that even from a novice perspective there's a virtue to always displaying both the given and expected type, in that order, affirming a kind of regularity to the system. In any case those changes are  pushed, see how they feel.

"satisfies" versus "is consistent with": I'm fine with either of these but want to use one consistently. I prefer the former, entirely due to its shorter length.

explain this: I feel it already does this, more so than before as it's next to the form name now. maybe changing the "i" icon to a "?"?

others:

do you mean all three of the buttons? ie export/import scratchpad and export persistent state? I don't know what the precise use cases for those are now so i'm not sure if they're deprecated by the new feature?

cyrus-:
the word "satisfies" is usually used when you are discussing an expression satisfying a type, not for a relationship between two types...

cyrus-:
changing the "i" to a "?" seems fine. I was thinking maybe positioning it to the right or somehow affiliating it visually with the form rather than the sort name to its left. that would break up the form name from the cursor inspector message perhaps, not sure if that's a problem.

cyrus-:
"do you mean all three of the buttons? ie export/import scratchpad and export persistent state? I don't know what the precise use cases for those are now so i'm not sure if they're deprecated by the new feature?"

I'm referring in scratch mode to the "Export Slide Persistent State Value" (which is used for initializing slides when the UExp representation changes by exporting a version of the initial slide definition with string representations rather than UExp representations). Is scratch mode slide initialization centralized now or no?

I think the buttons in exercise mode are still necessary.

disconcision:
@cyrus: my intended reading of the inspector is `<sort> <syntax class> <message>` i.e. the syntax class is the subject, not the self type, so for expressions/patterns that have a self type it would read as e.g. "expression list literal has type [Int] satisfying type [?]". (in the representation 'has' becomes ':' and 'type' is omitted as implicit). I admit this isn't totally transparent; if you want the type to be the subject I'm happy to change it to "is consistent with"

for the langdocs toggle, visually speaking I think it works well where it is now, as a break between the sort and the syntax class... I think that positioning adequately conveys the relationship, but I'm not married to it.

scratch mode slide initialization is indeed handled centrally with settings + examples via the Ini.ml menu option now, so I've now removed "Export Slide Persistent...". but I'm not sure whether the still-extant always-visible import/export buttons for the current scratch slide are still useful?

cyrus-:
import/export buttons for the current scratch slide are still useful in that they allow sharing scratch slides with other people / your future self, so yeah we can keep those

disconcision:
@cyrus- g2g

disconcision:
@cyrus actually one problem with the 'satisfies' phrasal structure for okay messages is that it's inconsistent with the error messages with say "inconsistent with". could switch those to "does not satisfy" though that feels slightly awkward

cyrus-:
Yeah, I think I'm leaning back toward changing "satisfies expected type" to "consistent with expected type" -- it can still read as a sentence that way

cyrus-:
Also I do think that some of the superscripts on the `?` type can be a bit confusing and it would be better to have them only show up on hover.

cyrus-:
Synthetic empty holes say "Fillable by any expression" but analytic holes say "Expecting type t" -- would be good to make these more consistent, e.g. have it say "Fill with an expression of any type" and "Fill with an expression of type t"?

PR #992, opened 2023-02-15
Live Aid

disconcision:
NB: This is based off of adt-defs-after.

This adds some UI for viewing live data, currently based on cursor position, but soon with arbitrary watchpoints. The first two images below are mockups, and the last is the current state of the UI.

screen_shot_2023-01-31_at_4 44 30_pm

hazel-live-aid-mockup

Screen Shot 2023-02-14 at 8 00 45 PM


cyrus-:
incorporated into #1091

PR #993, opened 2023-02-18
Refactor ExplainThis code to be more easily extensible

hannahpotter:


src/haz3lweb/explainthis/ExplainThisForm.re
```diff
@@ -0,0 +1,106 @@
...
+  | ThumbsDown
+  | Unselected;
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type example_id =
+  | IntList
+  | TupleList
+  | Cons1
+  | Cons2
+  | BasicFun
```
disconcision: minor but for this kind of type I feel it's nice to break it up into related chunks, ie
| Fun(fun_examples);

type fun_examples =
| Basic
| Wild
...

Not sure if strictly useful here but sometimes helps for dividing functions into more manageable, related chunks

src/haz3lweb/explainthis/ExplainThisMessages.re
```diff
@@ -0,0 +1,1289 @@
+open Sexplib.Std;
+open Haz3lcore;
+open ExplainThisForm;
+
+//Add a selected field to the form_option
+// Need a separate structure or something to list the options for the group instead of in the form_option?
+[@deriving (show({with_path: false}), sexp, yojson)]
+type messages = {
```
disconcision: not sure its strictly necessary to have "_group" in each name now that these are disambiguated by type. i would be tempted just to pluralize them instead; not a big deal tho

src/haz3lweb/explainthis/ExplainThisMessages.re
```diff
@@ -0,0 +1,1289 @@
...
+  | MultiHoleExp => (messages.multi_hole_exp_group.multi_hole_exp, [])
+  | TrivExp => (messages.triv_exp_group.triv_exp, [])
+  | BoolExp => (messages.bool_exp_group.bool_exp, [])
+  | IntExp => (messages.int_exp_group.int_exp, [])
+  | FloatExp => (messages.float_exp_group.float_exp, [])
+  | StringExp => (messages.string_exp_group.string_exp, [])
+  | VarExp => (messages.var_exp_group.var_exp, [])
+  | TagExp => (messages.tag_exp_group.tag_exp, [])
+  | FunctionExp => (messages.function_exp_group.function_exp, [])
+  | FunctionEmptyHole =>
```
disconcision: all of these cases could be abstracted into a general fn, something similar to:
```
(group, base) => {
  let general = group.function_exp;
  let options = [
      (general.form.id, general.expansion_label),
      (base.form.id, base.expansion_label),
    ];
  (general.selected?general.form:base.form, options)
}
```

src/haz3lweb/explainthis/ExplainThisMessages.re
```diff
@@ -0,0 +1,1289 @@
...
+            "Group id "
+            ++ Sexplib.Sexp.to_string(
+                 ExplainThisForm.sexp_of_group_id(group_id),
+               )
+            ++ "does not handle group selections",
+          ),
+        )
+      | FunctionEmptyHole =>
+        let group = messages.function_empty_hole_group;
+        print_endline(
```
disconcision: debug print?

src/haz3lweb/explainthis/FunctionExp.re
```diff
@@ -0,0 +1,634 @@
...
+  function_exp: form_option,
+  function_floatlit_exp: form_option,
+};
+[@deriving (show({with_path: false}), sexp, yojson)]
+type function_bool_group = {
+  id: group_id,
+  function_exp: form_option,
+  function_boollit_exp: form_option,
+};
+let function_str_group = "function_str_group";
```
disconcision: are we still using these strings for anything? thought they were being moved into the types

src/haz3lweb/view/ExplainThis.re
```diff
@@ -0,0 +1,2217 @@
...
+                List.map(
+                  ((id: ExplainThisForm.form_id, segment: Segment.t)): Node.t => {
+                    let map = Measured.of_segment(segment);
+                    let code_view =
+                      Code.simple_view(~unselected=segment, ~map, ~settings);
+                    let classes =
+                      id == form_id
+                        ? ["selected"] @ get_clss(segment)
+                        : get_clss(segment);
+                    Node.div(
```
disconcision, line 311: i feel when things start to get to the point where indentation is deep enough that we start seeing one token per line its worth breaking off some fns/lets along semantic boundaries. like here maybe specificity could be a top-level fn (might be too annoying if theres too many parameters tho), and internally maybe
```
let update_group_selection =
  _ => inject(Update.UpdateExplainThisMessages(
            ExplainThisMessages.UpdateGroupSelection(group_id, id))
         );
```
cuz then the handler just reads as `Attr.on_click(update_group_selection)` and its easy to see at a glance what it does (perhaps name could be better, just guessing what the most pertinent aspect is)

src/haz3lweb/explainthis/ExampleUtil.re
```diff
@@ -0,0 +1,68 @@
+open Haz3lcore;
```
disconcision: I think just merge this file into Example... we might want to rename Example too, I dunno. I don't think it's used elsewhere atm, it was kind of a transitional construct; open to a more use-descriptive name. Does all of your syntax construction go through here? In any case, I think we should consolidate all manual syntax construction in one file, and Example is probably the place for now. The string Form names were also kind of a transitory thing and I didn't expect them to become load-bearing like this. But I think it's fine for now if we just consolidate all this kind of thing in Example, and then it can be update en-mass if I find a better way to handle manual syntax construction, as I hope to do for assistant stuff.

src/haz3lweb/explainthis/FunctionExp.re
```diff
@@ -0,0 +1,634 @@
...
+  function_intlit_exp: form_option,
+};
+[@deriving (show({with_path: false}), sexp, yojson)]
+type function_float_group = {
+  id: group_id,
+  function_exp: form_option,
+  function_floatlit_exp: form_option,
+};
+[@deriving (show({with_path: false}), sexp, yojson)]
+type function_bool_group = {
```
disconcision: i'm a bit confused by these types... they all seem to be isomorphic (all of their fields have the same types). usually in the case where record fields share common types you'd break those out to a top-level record. in this case - without me fully understanding the use - i'd expect to se something like:
```
type base =
| MultiHole
| Wild
| Float ...

type function_group = {
  base,
  id: group_id,
  exp: form_option,
  base_exp: form_option
};
```
I don't think this is quite right though, as I don't understand the use exactly. But this seems like it should be split into a variant part and a record part.

src/haz3lweb/explainthis/ExplainThisMessages.re
```diff
@@ -0,0 +1,1289 @@
...
+       (TupleTyp.tuple_typ.id, [typ("ty1"), comma_typ(), typ("...")]),
+       (
+         TupleTyp.tuple3_typ.id,
+         [typ("ty1"), comma_typ(), typ("ty2"), comma_typ(), typ("ty3")],
+       ),
+     ]),
+   ),*/
+};
+
+[@deriving (show({with_path: false}), sexp, yojson)]
```
disconcision: update is pretty important, i think this is worth breaking out into a separate file. also, the explainthis folder has a lot of files... i think it's worth adding a subfolder called maybe 'data' for the files that are basically just the data for different kinds of messages, and having the top-level files be the logic + utilies

disconcision (changes requested):
Solid progress in general. Some stylistic quibbles in the comments, I feel like we can still abstract and condense things a bit. A few points of uncertainty around some of the types.

There's a remaining larger-scale structure issue which is at least worth thinking about addressing. Right now, the entire langdocs is getting put in the model. Really though it's conceptually two pieces: The messages, which are statically invariant, and the UI state (three globals (show docs, show highlights, specificity), and three more specific options for groups, forms, and examples respectively).

The model should ideally only contain state, both conceptually and practically (we'll likely end up de/serializing a bunch of redundant content). Not a huge deal at the moment but it's the kind of baggage I'm wary about accumulating.

I'm guessing it ended up this way because otherwise it would seem like we have to duplicate the entire data hierarchy of groups and examples, but I think we can probably avoid this with clever types.. wondering if you've thought about this; if this isn't clear we could try to work up a spec together, i think it could potentially end up simplifying other things, but there's some design complexity as well. i think it's at least worth chatting about this option but it's not an outright blocker from my perspective.

To facilitate this it would be helpful I think to have a (quick, basic) conceptual diagram (either drawn or just prose) which describes the architecture of the system at a high level, i.e. describes the relation between groups, explanation/forms, and examples, and their associated kinds of UI state.

cyrus-:
@hannahpotter this has some conflicts with dev. what is the status of resolving the things @disconcision mentioned?

src/haz3lweb/explainthis/ExplainThisMessages.re
hannahpotter: Done. See ExplainThisUpdate and explainthis/data

src/haz3lweb/explainthis/ExampleUtil.re
hannahpotter: Done. All moved to Example

src/haz3lweb/explainthis/ExplainThisMessages.re
hannahpotter: Yes. Removed

src/haz3lweb/explainthis/ExplainThisMessages.re
hannahpotter: Code changed in ExplainThisModel so this isn't relevant anymore

src/haz3lweb/explainthis/FunctionExp.re
hannahpotter: Addressed. See ExplainThisForm.

src/haz3lweb/view/ExplainThis.re
hannahpotter: Updated.

src/haz3lweb/explainthis/FunctionExp.re
hannahpotter: Correct. They aren't being used anymore. They should all be removed now.

src/haz3lweb/explainthis/ExplainThisMessages.re
hannahpotter: Group names changed.

src/haz3lweb/explainthis/ExplainThisForm.re
hannahpotter: Broken up more,

hannahpotter:
Conflicts with dev fixed and comments addressed.

hannahpotter:
ExplainThisForm - defines the structure of the different messages, examples, groups, etc. This would be where people adding new messages would "register" them.
ExplainThisModel - defines the structure of the stateful parts of the messages, groups, etc. This is connected to ExplainThisForm by the IDs. People just adding new messages shouldn't have to touch this. I wrote in some checking code that should only be hit if someone either makes a mistake in defining a message in a way that won't be checked statically or if someone messes up the model.
ExplainThisMessages is no more
ExplainThisUpdate - handles all the updates to the model. People just adding new messages shouldn't have to touch this.
HoleExp - has some examples of how someone would define simple messages
FunctionExp - has some examples of how someone would define more complicated messages/groups
ExplainThis - all of the drawing code. People just adding new messages would just need to define the pattern that needs to be matched on and call the correct message (simple example ~line 587, more complicated example ~lines 604-650)

src/haz3lweb/LocalStorage.re
```diff
@@ -41,51 +41,51 @@ module Settings = {
...
+module ExplainThisModel = {
+  let save_ExplainThisModel_key: string = "ExplainThisModel";
 
-  let serialize = langDocMessages =>
-    LangDocMessages.serialize(langDocMessages);
+  let serialize = explainThisModel =>
+    explainThisModel |> ExplainThisModel.sexp_of_t |> Sexplib.Sexp.to_string;
 
   let deserialize = data =>
-    try(LangDocMessages.deserialize(data)) {
```
hannahpotter: This will mean that if a form is removed from ExplainThisForm that had saved data for it, the ExplanThisModel won't be loaded in.

src/haz3lweb/Example.re
```diff
@@ -70,3 +70,75 @@ let zipper: Zipper.t = {
   },
   caret: Outer,
 };
+
```
disconcision: This Example file is only used for explainthis now; let's move it to the explainthis directory and maybe rename it

src/util/ListUtil.re
```diff
@@ -53,6 +53,20 @@ let rec mk_frame = (n: int, xs: list('x)): frame('x) => {
   };
 };
 
+let rec split =
```
disconcision, line 56: stick this with the other split fns below, or if possible just make it a wrapper for one of them

src/haz3lweb/explainthis/data/CaseExp.re
```diff
@@ -0,0 +1,89 @@
...
+      case_example_int,
+      case_example_bool,
+      case_example_wild_simple,
+      case_example_wild_tuple,
+    ],
+  };
+};
+
+let case: group = {id: CaseExp, forms: [case_exp]};
+/*let case_exp_rules: form = {
```
disconcision: rm or document

src/haz3lweb/explainthis/ExplainThisForm.re
```diff
@@ -0,0 +1,313 @@
...
+  | Ctr
+  | Ap;
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type un_op_sub_form_id =
+  | BoolNot
+  | IntMinus;
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type bin_op_sub_form_id =
```
disconcision: this type seems like it's exactly isomorphic to the existing bin_op type in TermBase? is there a reason for having it separate?

disconcision:
going to do a review, but general question: could you provide a short step-by-step guide to the full process of adding docs for a new syntactic form? just want something to contextualize the new setup for an implementor pov

disconcision (approved):

disconcision:
@cyrus- this is ready to be merged. I made some changes, mostly cosmetic.

Screenshot 2024-01-19 at 2 54 47 AM


In re-adding the Pipeline operator docs, I streamlined the process a bit for adding new simple docs (simple == one-element groups). PipelineExp.re serves as an exemplar. Adding a new language forms should now be as simple as imitating that file and following the types. (Existing docs could be refactored along the same lines to make them much shorter)

disconcision:
This is (basically) all you need to specify to add a new form now:
Screenshot 2024-01-23 at 6 43 48 PM

PR #999, opened 2023-03-03
Fantasyland Hazel

disconcision:
This is an 𝑒𝓍𝓅𝑒𝓇𝒾𝓂𝑒𝓃𝓉𝒶𝓁 hazel branch to try variations on the core type system. It is not intended for production, just to try things out for discussion. At the moment, it:

1. Eliminates the synthetic/analytic distinction. There are no modes, only an **expected type**, which may be unknown.
2. Every expression/pattern has three type-like things associated with it. Besides the aforementioned **expected type**, each expression has:
3. a **given/self type** which can be either none (represented as《》, covering errors like unbound variables), Just a type, or inconsistent types (represented as《ty1,ty2,...》)
4. and an **is/actual type** which can be either unknown (if the given type is《》or《tys...》, or if it's《ty》and ty doesn't join with the expected type. otherwise it's the join
5. for debugging/experimentation purposes, there is a precise and complete provenance tracking system for all possible origins and operations on unknown types:`  type type_provenance = | ErrorHole | EmptyExpHole | EmptyPatHole | EmptyList | PatVar | TrueSyn | TypeHole | EmptyJoin | Join(type_provenance, type_provenance) | MatchedArrow(type_provenance) | MatchedProd(type_provenance) | MatchedList(type_provenance);`
6. Inconsistent branches errors are styled as a warning. They are shown in all cases where branches are inconsistent (after fixing), except they are suppressed when the expected type is unknown-typehole.


cyrus-:
Closing this since it is 𝑒𝓍𝓅𝑒𝓇𝒾𝓂𝑒𝓃𝓉𝒶𝓁 and not being kept up to date with `dev` changes, but we can revisit or reopen this in the context of #1234.

PR #1004, opened 2023-03-27
Update ocaml to 5.0.0

coned:
We can update ocaml directly to 5.0.0.


disconcision:
@coned confused that this seems to lose the ocaml-lsp-server dependency? are you using editor lsp integration? noticed this because it was going to suggest upgrading it as well to support the features i posted about in the dev-discuss channel

coned:
> @coned confused that this seems to lose the ocaml-lsp-server dependency? are you using editor lsp integration? noticed this because it was going to suggest upgrading it as well to support the features i posted about in the dev-discuss channel

Yeah. I forgot to install lsp integration. I will add it.

PR #1006, opened 2023-03-31
Add evaluation filter

tonyfettes:
# Evaluation Filter

`BinIntOp`, `Ap` works for now.

- Skip over multiplication.
image
(the expression `3 + 4` is not steppable, since the expression is parsed as `(((1 + 2) + 3) + 4)`)

- Skip over addition.
image

- Skip all
image

- Step all
image

- step into functions
image

## Difficulties

1. It's getting more and more tedious when managing multiple evaluation context (environment, filter).

## Roadmaps

- [x] It's reasonable to skip over the `3 * 4` here since we only care about addition here.
image
image

## Future works

- Use alternate syntax/parser/editor for filter language? (I'm not sure if this is necessary)


cyrus-:
@tonyfettes I think this isn't the branch you are now working off of?

tonyfettes:
@cyrus- This is the branch I'm working on. I have some unstaged changes locally, so I'm afraid that the behavior of the deployed instance is not most up to date.

PR #1010, opened 2023-05-08
Add jump-to-tests 

fanbaoAA:
Jump to tests from tests panel

### DIFF
--- a/src/haz3lschool/SchoolExercise.re
+++ b/src/haz3lschool/SchoolExercise.re
@@ -268,12 +268,7 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     };
 
   let switch_editor = (idx: int, {eds, _}) => {
-    pos:
-      if (idx == (-1)) {
-        pos_of_idx(eds, 5 + List.length(eds.hidden_bugs));
-      } else {
-        pos_of_idx(eds, idx);
-      },
+    pos: pos_of_idx(eds, idx),
     eds,
   };
 
--- a/src/haz3lschool/SchoolExercise.re
+++ b/src/haz3lschool/SchoolExercise.re
@@ -255,6 +255,7 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     | 2 => YourTestsTesting
     | 3 => YourTestsValidation
     | 4 => YourImpl
+    | (-1) => HiddenTests
     | _ =>
       if (idx < 0) {
         failwith("negative idx");
--- a/src/haz3lschool/SchoolExercise.re
+++ b/src/haz3lschool/SchoolExercise.re
@@ -267,10 +267,16 @@ module F = (ExerciseEnv: ExerciseEnv) => {
       }
     };
 
-  let switch_editor = (idx: int, {eds, _}) => {
-    pos: pos_of_idx(eds, idx),
-    eds,
-  };
+  let switch_editor = (~pos, instructor_mode, ~exercise) =>
+    if (!instructor_mode) {
+      switch (pos) {
+      | HiddenTests
+      | HiddenBugs(_) => exercise
+      | _ => {eds: exercise.eds, pos}
+      };
+    } else {
+      {eds: exercise.eds, pos};
+    };
 
   let zipper_of_code = (id, code) => {
     switch (Printer.zipper_of_string(id, code)) {
--- a/src/haz3lschool/SchoolExercise.re
+++ b/src/haz3lschool/SchoolExercise.re
@@ -255,7 +255,6 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     | 2 => YourTestsTesting
     | 3 => YourTestsValidation
     | 4 => YourImpl
-    | (-1) => HiddenTests
     | _ =>
       if (idx < 0) {
         failwith("negative idx");
@@ -269,7 +268,12 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     };
 
   let switch_editor = (idx: int, {eds, _}) => {
-    pos: pos_of_idx(eds, idx),
+    pos:
+      if (idx == (-1)) {
+        pos_of_idx(eds, 5 + List.length(eds.hidden_bugs));
+      } else {
+        pos_of_idx(eds, idx);
+      },
     eds,
   };
 
--- a/src/haz3lweb/Grading.re
+++ b/src/haz3lweb/Grading.re
@@ -111,12 +111,12 @@ module MutationTestingReport = {
     );
   };
 
-  let individual_report = (i, ~inject, ~hint: string, ~status) =>
+  let individual_report = (i, ~inject, ~hint: string, ~status: TestStatus.t) =>
     div(
       ~attr=
         Attr.many([
           Attr.classes(["test-report"]),
-          Attr.on_click(TestView.jump_to_test(~inject)),
+          Attr.on_click(TestView.jump_to_test(~inject, i)),
         ]),
       [
         div(
@@ -270,7 +270,7 @@ module ImplGradingReport = {
       ~attr=
         Attr.many([
           Attr.classes(["test-report"]),
-          Attr.on_click(TestView.jump_to_test(~inject)),
+          Attr.on_click(TestView.jump_to_test(~inject, i)),
         ]),
       [
         div(
--- a/src/haz3lweb/Grading.re
+++ b/src/haz3lweb/Grading.re
@@ -270,12 +270,15 @@ module ImplGradingReport = {
   //   );
   // };
 
-  let individual_report = (i, ~inject, ~hint: string, ~status, (id, _)) =>
+  let individual_report =
+      (i, ~inject, ~hint: string, ~status, (id, _), ~num_hidden_bugs: int) =>
     div(
       ~attr=
         Attr.many([
           Attr.classes(["test-report"]),
-          Attr.on_click(TestView.jump_to_test(~inject, -1, id)),
+          Attr.on_click(
+            TestView.jump_to_test(~inject, 5 + num_hidden_bugs, id),
+          ),
         ]),
       [
         div(
@@ -302,7 +305,7 @@ module ImplGradingReport = {
       ],
     );
 
-  let individual_reports = (~inject, ~report) => {
+  let individual_reports = (~inject, ~report, ~num_hidden_bugs: int) => {
     switch (report.test_results) {
     | Some(test_results) =>
       div(
@@ -313,6 +316,7 @@ module ImplGradingReport = {
                ~inject,
                ~hint,
                ~status,
+               ~num_hidden_bugs,
                List.nth(test_results.test_map, i),
              )
            ),
@@ -321,7 +325,7 @@ module ImplGradingReport = {
     };
   };
 
-  let view = (~inject, ~report: t, ~max_points: int) => {
+  let view = (~inject, ~report: t, ~max_points: int, ~num_hidden_bugs: int) => {
     Cell.panel(
       ~classes=["cell-item", "panel", "test-panel"],
       [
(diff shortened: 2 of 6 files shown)


src/haz3lweb/view/TestView.re
```diff
@@ -88,7 +90,7 @@ let test_bar_segment = (~inject, (_id, reports)) => {
     ~attr=
       Attr.many([
         clss(["segment", status]),
-        Attr.on_click(jump_to_test(~inject)),
+        Attr.on_click(jump_to_test(~inject, _id)),
```
cyrus-: variables with leading underscores indicate an unused variable, so this can be renamed to `id` in both places

src/haz3lweb/Grading.re
```diff
@@ -270,7 +270,7 @@ module ImplGradingReport = {
       ~attr=
         Attr.many([
           Attr.classes(["test-report"]),
-          Attr.on_click(TestView.jump_to_test(~inject)),
+          Attr.on_click(TestView.jump_to_test(~inject, i)),
```
cyrus-: since it is an id, use the variable name `id`

cyrus- (changes requested):
Functionality looks good, just fix the small variable naming issues and we can merge!

cyrus- (changes requested):
OK, so one remaining problem is that when you click on a test bar for an editor that isn't currently focused, it doesn't seem to work. We need to set focus on the correct editor when clicking too...

cyrus- (changes requested):
Seems to crash when you click one of the green bars under Mutation testing when in student mode, since the editor isn't available...

src/haz3lschool/SchoolExercise.re
```diff
@@ -255,6 +255,7 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     | 2 => YourTestsTesting
     | 3 => YourTestsValidation
     | 4 => YourImpl
+    | (-1) => HiddenTests
```
cyrus-: I don't think this is not correct -- should be the inverse of idx_of_pos, which is already handled in the catchall case below

src/haz3lweb/Grading.re
```diff
@@ -265,12 +270,12 @@ module ImplGradingReport = {
...
   // };
 
-  let individual_report = (i, ~inject, ~hint: string, ~status) =>
+  let individual_report = (i, ~inject, ~hint: string, ~status, (id, _)) =>
     div(
       ~attr=
         Attr.many([
           Attr.classes(["test-report"]),
-          Attr.on_click(TestView.jump_to_test(~inject)),
+          Attr.on_click(TestView.jump_to_test(~inject, -1, id)),
```
cyrus-: ah I see, it was a hack to not have to compute the number of editors above. let's not do a hack like this -- you should define a helper function to compute the correct index

src/haz3lweb/Grading.re
```diff
@@ -297,13 +302,25 @@ module ImplGradingReport = {
       ],
     );
 
-  let individual_reports = (~inject, ~hinted_results) =>
+  let individual_reports = (~inject, ~report) => {
+    let test_results =
+      switch (report.test_results) {
+      | Some(x) => x
+      | None => failwith("No test results")
```
cyrus-: this shouldn't be a failwith when there are no tests -- seems to cause a crash when there are zero tests.

src/haz3lweb/Grading.re
```diff
@@ -334,7 +351,7 @@ module ImplGradingReport = {
               @ Option.to_list(
                   report.test_results
                   |> Option.map(test_results =>
-                       TestView.test_bar(~inject, ~test_results)
+                       TestView.test_bar(~inject, ~test_results, -1)
```
cyrus-: fix this -1 as well

src/haz3lweb/Update.re
```diff
@@ -312,12 +312,15 @@ let apply =
       | DebugLoad => failwith("impossible")
       | Scratch(_) => Error(FailedToSwitch) // one editor per scratch
       | School(m, specs, exercise) =>
-        let exercise = SchoolExercise.switch_editor(n, exercise);
+        let exercise = ref(exercise);
```
cyrus-: why are you creating a ref here? not sure why this was changed...

src/haz3lweb/view/TestView.re
```diff
@@ -23,23 +23,27 @@ let test_instance_view =
     ],
   );
 
-let jump_to_test = (~inject as _, _) => Effect.Ignore;
+let jump_to_test = (~inject, pos, id, _) => {
+  let effect1 = inject(Update.SwitchEditor(pos));
+  let effect2 = inject(Update.PerformAction(Jump(TileId(id))));
+  Effect.bind(effect1, ~f=_result1 => effect2);
```
cyrus-, line 29: I think there is something like `Effect.many` you can use here instead

cyrus- (changes requested):
see comments

src/haz3lweb/Update.re
fanbaoAA: So it's to solve the issue that it crashes when jump to an non-existing editor under the student mode.
The logic is:
if it's student mode and we are jumping to the hidden editors, don't modify exercise.
else change exercise to SchoolExercise.switch_editor(n, exercise);.
So the ref here is made to change the exercise inside the if statement and be used later.

src/haz3lweb/Grading.re
fanbaoAA: fix

src/haz3lweb/view/TestView.re
fanbaoAA: Seems there isn't Effect.many in ReasonML?
Found Effect.all which "Combines the given effects into a single effect that runs them in parallel." Can't be applied here since the return type is 'a list Effect.t not compatible with unit Effect.t.

src/haz3lschool/SchoolExercise.re
fanbaoAA: Removed.

src/haz3lweb/Grading.re
fanbaoAA: Didn't change the -1 here because the module here doesn't have the access to the information necessary to determine the number of editors (this module only handles the ImplementationGradingReport, we need the number of WrongImplementation to determine the number of editors). So I postpone this responsibility to switch_editor function.

src/haz3lweb/Grading.re
fanbaoAA: Same as the previous one.

fanbaoAA:
Fix some of the issues. See above

src/haz3lweb/Update.re
cyrus-: You don't need a ref here, you can just bind to a variable...

```
let exercise = 
  if (model.settings...) { 
    SchoolExercise.switch...
  } else { exercise } 
```

src/haz3lweb/Grading.re
cyrus-: If there is information that is needed here, add it as an argument to the function so the caller is responsible for providing that information. Might require threading it through a few levels, but better than using `-1`. If that really isn't feasible, you should use an option type rather than a special value.

cyrus- (changes requested):

fanbaoAA:
Removed the unnecessary ref, eliminate all -1 and pass the info from caller instead.

src/haz3lschool/SchoolExercise.re
```diff
@@ -268,12 +268,7 @@ module F = (ExerciseEnv: ExerciseEnv) => {
...
   let switch_editor = (idx: int, {eds, _}) => {
-    pos:
-      if (idx == (-1)) {
-        pos_of_idx(eds, 5 + List.length(eds.hidden_bugs));
-      } else {
-        pos_of_idx(eds, idx);
-      },
+    pos: pos_of_idx(eds, idx),
     eds,
   };
```
fanbaoAA, line 280: remove the things dealing with -1

src/haz3lweb/Grading.re
```diff
@@ -350,7 +354,11 @@ module ImplGradingReport = {
...
                   |> Option.map(test_results =>
-                       TestView.test_bar(~inject, ~test_results, -1)
+                       TestView.test_bar(
+                         ~inject,
+                         ~test_results,
+                         5 + num_hidden_bugs,
+                       )
                      ),
                 ),
             ),
```
fanbaoAA, line 361: Pass num_hidden_bugs through functions.

src/haz3lweb/Update.re
```diff
@@ -312,15 +312,17 @@ let apply =
...
         LocalStorage.School.save_exercise(
-          exercise^,
+          exercise,
           ~instructor_mode=model.settings.instructor_mode,
         );
-        Ok({...model, editors: School(m, specs, exercise^)});
+        Ok({...model, editors: School(m, specs, exercise)});
       }
     | ToggleMode =>
       let new_mode = Editors.rotate_mode(model.editors);
```
fanbaoAA: Remove the unnecessary ref.

src/haz3lweb/view/SchoolMode.re
```diff
@@ -325,6 +325,8 @@ let view =
         ~inject,
         ~report=grading_report.impl_grading_report,
         ~max_points=grading_report.point_distribution.impl_grading,
+        ~num_hidden_bugs=
+          List.length(grading_report.mutation_testing_report.results),
       ),
     );
```
fanbaoAA: The caller provides num_hidden_bugs.

src/haz3lweb/Grading.re
cyrus-: instead of passing in num_hidden_bugs directly, let's use SchoolExercise.{idx_of_pos/pos_of_idx} and pass in the exercise, so we aren't hard coding an indexing scheme everywhere that might change in the future

src/haz3lweb/view/SchoolMode.re
cyrus-: see above

cyrus- (changes requested):

fanbaoAA:
Replaced all hard code idx with SchoolExercise.pos type.

Not sure why the check below failed.

fanbaoAA:
Replaced all hard code idx with SchoolExercise.pos type.

src/haz3lweb/Grading.re
```diff
@@ -72,13 +76,20 @@ module MutationTestingReport = {
...
+  let bar = (~inject, instances) =>
     div(
       ~attr=Attr.classes(["test-bar"]),
-      List.map(
-        ((status, _)) =>
+      List.mapi(
+        (id, (status, _)) =>
           div(
-            ~attr=Attr.classes(["segment", TestStatus.to_string(status)]),
+            //~attr=Attr.classes(["segment", TestStatus.to_string(status)]),
```
cyrus-: don't leave things commented out

src/haz3lweb/Update.re
```diff
@@ -307,12 +307,21 @@ let apply =
...
+    | SwitchEditor(pos) =>
       switch (model.editors) {
       | DebugLoad => failwith("impossible")
       | Scratch(_) => Error(FailedToSwitch) // one editor per scratch
       | School(m, specs, exercise) =>
-        let exercise = SchoolExercise.switch_editor(n, exercise);
+        let exercise =
+          if (!model.settings.instructor_mode) {
+            switch (pos) {
+            | SchoolExercise.HiddenTests
```
cyrus-: this logic should be in SchoolExercise.switch_editor instead of Update itself, since it is school mode specific

cyrus- (changes requested):
just a couple last cleanup items, looks good!

cyrus-:
@fanbaoAA can you wrap up those last couple todos?

fanbaoAA:
Done. Sorry for the delay.

cyrus- (approved):

Issue #970, opened 2023-01-24
Parsing ambiguity for "fun x: Int->Int->x"


disconcision:
Linearly entering the titular expression will result in the first arrow being associated with the fun, and the second arrow having a sort error, whereas the alternative parse, achievable only non-linearly, is sort-correct. I don't know an easy fix for this; I think maybe this has always been the case and not a result of my expansion changes?

disconcision:
To add some context to our current approach here post-backpack-II; now we have special-case logic, where arrows only associate with funs if they are on the top of the backpack; most other forms have no such top-only restriction. The reason here is that this permits writing the above annotation by parenthesizing it, by which I mean specifically inserting an opening parens after the `:` **before entering** the `->`. This means the closing parens will be on the top of the backpack, blocking the -> from joining the fun.

However, this has an ancillary consequence: Given `fun x: Int -> Int -> x`, as typed from left to right, you can't parenthesize `Int -> Int`: Inserting an open parens before the first Int, and a closing parens after the last Int, is not successful in turning `Int -> Int` into a type annotation, as the closing parens does not break the association with fun and -> to become associated with the opening parens. 

For this to work, i think we would have to extend the special case logic that governs -> in a way that can break parents.

PR #1012, opened 2023-05-14
Haz3l user defined operators

### LINKED ISSUE #828: User-defined operators

shawn-stovall:
Adding the ability for users to define their own operators.

This pull request makes changes to the external language's pattern and expression forms so that custom infix operators can be defined and applied. This then elaborates to an applied `DHExp.t` of the form `Ap(operator_variable, Tuple([arg1, a rg2]))`. The evaluator then continues unmodified to process the function application.

Features:

- `_op_` can be used in a normal let to define a new operator
- `arg1 op arg2` can then be used without the underscores to apply the operator's function

Potential Future Features:

- `_op_` can be used in expressions to allow for typical function application (of the form `_op_(arg1, arg2)`).
- `_op` and `op_` used to define postfix and prefix operators, respectively.


shawn-stovall:
User-defined operators are working with the following features:

* `_op_` defines a user-defined operator, with op representing one or more of the following symbols: `~?!$%&*+\\-./:<=>@^`
Precedence and associativity are decided by the first character of the operator. I used the list of valid operator symbols from Ocaml to determine which symbols to include in this list, as well as `~?!` which are usually reserved for different fixities in Ocaml. The precedence of the first character is determined by the matching precedence of the built-in operator of the same symbol. For the symbols not in Hazel, I used Ocaml precedence as a reference. The exceptions to this are `~!?` to which I gave the precedence of equality.
* Calling the operator surrounded by parentheses in the form `_op_(arg1, arg2)` calls the operator as a normal function.
* There are user-facing errors for defining an operator that is the same as a built-in operator, defining an operator that is bound to a non-binary function definition, and and error for trying to call an undefined operator (which uses the standard "variable not defined" error).

cyrus-:
Error message here should say something about user-defined operator being unbound, not that it is not a binary function. 

image

cyrus-:
Error box here is broken (when function extends over more than one line)

image

cyrus-:
Operator definition error should appear only if the type of the bound expression is not *consistent* with a binary function, even if it isn't exactly of the form (t1, t2) -> t3. So the hole type or hole -> t3 things like that are fine. Casts will handle downstream problems.

image

cyrus-:
Let's also update the scratch slide 1 (language reference) to list out all of the operators in order of precedence and include a comment describing how user-defined operators work (precedence and associativity).

cyrus-:
Nice work, seems like just a few polish things are remaining! Marking as draft again based on incomplete functional review comments above, will continue to take a look once those things are addressed.

cyrus- (changes requested):
I can't type `f` into the second hole here:
image

cyrus-:
@shawn-stovall quite a few conflicts with `dev` after recent ADTs merge, can you take a look at those?

disconcision:
I've been thinking lately about ambiguity concerns for characters that can live in tokens of more than one grammatical class. For our purposes let's specify the following grammatical classes:

- Operand (eg `foo`, `1.0`, `_`, `Int`)
- Infix Operator (eg `+`, `->`, `!=.`)
- Prefix Operator (ie `!`, `+`, `-`)
- Postfix Operator (current none in hazel)
- Form Delimiter (eg `let`, `=`, `in`; should maybe divide into multiple classes)

Even without user-defined operators, we already have some ambiguities in our grammar, both realized and potential. This list is currently exhaustive AFAIK:

1. `.` occurs in both operands (floating points) and infix operators (`==.` `!=.` `<.` `>.` `<=.` `>=.`). Thus `1==.1` could be parsed as `(1)==.(1)` or `(1)==(.1)`. Currently the accepted parse is the first, but this is somewhat incidental, due to the fact that we attempt to append an inserted character to the left first before the right. If we wish to allow such ambiguities, we should ideally clarify our general principles for disambiguation; either formalize the left to right model, choose a prioritization based on syntactic class, special-case per-ambiguity, etc.
2. `-` occurs as both infix and prefix operator; technically this ambiguity shouldn't be a problem; there is always one valid parse which minimizes grout insertion. Nonetheless, this currently manifests in remolding bugs such as #1076, which are sensitive to which order the two alternatives appear in in Forms.re.
3. `+` occurs as both infix and prefix operator too, though only for types, and in a way that isn't semantically ambiguous
4. `!` occurs as prefix and (as part of an) infix operator, though it currently only exists in infix ops as a leading character, so this ambiguity does not currently manifest
5.  `->` occurs as both infix and form delimiter; this is already a collision between function types and function literals; see bug #970)
6. `=` occurs as form delimiter (let expressions) and is potentially infix; if realized, this would cause a similar ambiguity to `->`
7. `=>` likewise occurs as form delimiter (case rules) and is potentially infix

The above examples suggest three severity classes with respect to syntactic ambiguity:
1. Technically ambiguous (e.g. `-` and `+`): While ambiguous, these always have one grout-minimal parse (I think). Internal bugs may cause the non-grout-optimal parse to be selected, but this is an implementation error
2. Ambiguous, resolvable by inserting whitespace (e.g. `.`): Not much of a problem in practice, though some effort should be spent to make these predictable.
3. Ambiguous, requires inserting disambiguating parentheses (e.g. `->`): More confusing and more effortful to resolve; should be avoided if possible, predictable if not. 

Other than the last class, none of these are _further_ problematized by the above allowed character list of user-defined operators, so this is mostly food for thought, except that existing symbolic form delimiters (`=`, `=>`, `->`) should probably be forbidden as user-defined operators

disconcision:
I did a functional review. There are still a few outstanding bugs, and a couple other things we should probably decide about. Going to reproduce the things I tried below for reference:
Screen Shot 2023-12-17 at 7 00 57 PM

```#1. OK: dont shadow builtins#
let _*_ = in
let _!_ = in

#2. OK? Do we want to still bind failed operator defs? We do at the moment#
let _*--_ = 1 in
let _*---_ = fun x, y, z -> x + y in

#3. Bug: Should probably be a static error#
let _*-*-_ = fun x:Int -> 1 in

#4. OK: Constant operator. Pointless but fine#
let _*+*+_= fun x: -> 1 in

#5. OK?: Returns pair of inputs. Odd but fine I guess?#
let _*!*+_= fun x: -> x in

#6. OK? Runtime cast fail; not wrong but not ideal#
let _*---*_= fun x-> let a,b,c = x in a+b in
#Maybe we should require that unannotated ops analyze against (?,?)->?#
#This would make 3 and 6 static errors.#

#7. OK: Happy unannotated cases#
let _*----_ = fun x, y -> x + y in
let _+*_= fun x-> let a, b = x in a*b in

#8. OK? Notice error is localized to pattern here#
# It would be on definition if this wasnt an operator #
# This seems reasonable, but the message shown DOES involve the definitions#
# type. if were going to attribute the err to the annotation, i feel the#
# message should be more like "must be consistent with (?,?)->?" or something#
let _+***_: (Int, Int) = fun a,b-> a*b in

#9. OK: Annotated happy path #
let _*---*_: (Int, Int) -> Bool = fun a, b -> a < b in 

#10. Bug: All unnannoted operators trigger cast exceptions when used in tests:#
test 1+*1 == 1 end;

#11. OK: Use operator as fn#
_*---*_(4,4);

#12. OK? Maybe out-of-scope but would be nice to be able to do above w/builtins#
_*_(4,4)
```

Bugs:
- Unannotated ops defs which don't take a tuple should be static errors (see 3). I propose new logic in (6) which would catch this and maybe subsume the current logic. In summary, maybe the logic should be something like: For any operator definition, the definition must analyze against (?,?)->?. If this fails, then maybe we still want to bind the operator, permitting its use as a function if not an operator; maybe not though, I dunno. Further worth pointing out that this does complicated what we do in the annotated case. Naively we could simply make an operator pattern synthesize (?,?)->?. This would achieve the desired behavior in the unannoted case automatically; it even permits using an operator pattern as part of a composite pattern, e.g. defining multiple operators via a tuple. But the user could  problematically override this behavior by annotating the operator variable with less general types i.e. ? or ?->?. So it seems like we do need to retain a special error case for annotations on pattern variables. The question is how specific do we want to be; perhaps we should insist they be of the form (a,b)->c.
- There seems to be a casting bug for unannotated operators (see 10).

Things to decide:
- Should failed operator definitions (invalid type) still get bound in context (see 1)
- The 'identity operator', which just returns the pair of its inputs, feels slightly odd to me, but I don't think it's worth special-casing to avoid. (see 5)
- The error we put on bad operator annotations could maybe be more targetted (see 8)
- Would be nice to be able to use extant builtin (functional) ops as fns as well, but maybe not necessary for this PR (see 12)

cyrus-:
archiving since this is no longer active

PR #1018, opened 2023-05-16
Added some docs for Statics and LangDoc

green726:
Work done during the code cleanup/docs meeting on 5/15/23. In progress.


cyrus-:
@green726 don't worry about LangDocs for the moment since @hannahpotter is reworking it

cyrus-:
Both statics and langdoc have been substantially refactored so closing this -- but documentation for the new versions would be welcome in the future! Thanks @green726 .

PR #1019, opened 2023-05-16
Cleanup: precedence

WondAli:
Work done during cleanup meetings. In Progress.


WondAli:
5/15/23: Started cleaning up files relating to precedence.
- Less Than, Greater Than, and their float variants were previously hardcoded at 6, equivalent to the Plus operator. In OCaml, they have the same precedence as Equals.
- I believe Cell-Join should have a later precedence than Boolean-Or, but I'm not sure where to to put it.
- The Assignment operation is currently not in use, but I'm not sure what precedence that would be.
- There is a minor fix here for recognizing the Boolean-Or that was implemented in https://github.com/hazelgrove/hazel/pull/924

PR #1020, opened 2023-05-19
module system

### LINKED ISSUE #1011: Basic module system

gensofubi:
Haz3l module system. 
Current progress: 
Type Module defined (with detailed information as name-type pair), keyword **module** and operator **.** implemented, 

Type annotation for modules implemented. Use a pair of curly parentheses around a tuple of patterns. Typically the tuple should contain only TypeAnn patterns, but Var/Tag will also be accepted and treated as hole-typed members.

Type Inconsistency warnings includes both the pattern (which define the member) and the definition expression. 
For member name inconsistency, accessing members shown in module's typeann but not defined will be treated as a emptyhole (supposing users are to implement later), accessing members defined but not shown in module's typeann will be treated as free variables.

P~M0}3R$CFZ%9@ORYXX$NUH

Basic Type member implemented. Type member in module type annotation implemented. 
For constructors defined by Sum types in module, either access directly by dot access, or access under analytic mode (dot access not necessary).

8``TU` P}~5IH75ADL9NYMP

Provided warnings for using type members that are not defined, and inconsistency for type members. Created documentations for module-related keywords.

Some current rules of typing and casting:
Module type consistency: module types are consistent only if they share the same member name list, and the types of every member are consistent. Therefore, the ground type of a module is one where all members have hole types.

For module creation, the definition part will always be packaged as the exact same module type of the annotation as we mentioned amove. So nothing to worry about casting.

Casts will be added on module aliases. Cast calculation will happen on dot operators.

0c7316442d59814e378104002eb3efb8


cyrus-:
@gensofubi as you make progress on this, can you update the PR description documenting what is now possible

cyrus-:
@gensofubi there are a number of conflicts with `adt-defs-after` as well now - can you resolve those

cyrus- (changes requested):
error message for an invalid module name shouldn't say "invalid token", it should say "invalid module name" 
image

also in above screenshot, the DHExp display for a module isn't very helpful -- it would be better to actually give all the evaluated values for the module members instead of just saying `Module`.

cyrus-:
I can't press `.` here to turn this into `M.M.y`
image

gensofubi:
The issues have been addressed.
Regarding the last issue, due to the current logic of insert, if a character is inserted into a valid token to make it invalid, it automatically splits into three tokens instead of considering the newly inserted character and the subsequent string as a new token. In order to make modifications without changing this logic, I have changed 'dot' to be an independent infix instead of being a suffix with the variable name that follows it (which also aligns better with the intuition of 'dot' as an operator). In this case, however, it is difficult to determine in static check whether the Exp after 'dot' is a Var/Tag or other constant expression, hence resulting in situations like 'M.(1) evaluates to 1' (I think this somehow makes sense).

cyrus-:
@gensofubi there is a conflict with `adt-defs-after` you should take a look at

cyrus- (changes requested):
image
The way the module type is formatted in the cursor inspector message at the bottom is different than how module types are actually written in the surface syntax -- let's change the cursor inspector print to use curly braces and the same syntax.

In addition, in the cursor inspector it says that the sort of `M` is Typ and its form is "Type variable" which is not correct -- M is a module variable, and its sort should be something like "Module Path". Same issue in the lang docs sidebar.

cyrus-:
Additional small issue: in the example at the bottom, lets include a type alias, and also do not indent the last `in` by 1 space:
image

cyrus-:
The langdocs message for a module type is also confusing -- the reference to floating point is a mistake, and think of a more clear way of stating the rest of the message using the same terminology used elsewhere / in OCaml (type alias and values)

image

gensofubi:
The issue has been resolved. Regarding the first issue, setting the sort for the Module used to access Typealias to Typ was a compromise to address the problem of the module being displayed in red font. I have now changed its form from 'Type variable' to 'Module path' and made modifications to the langdoc. Similarly, the capitalized tags as Exp and Pat, which were originally considered Constructors instead of Module variables, have also been corrected. Additionally, @cyrus- , should the syntax for the module type be changed from curly brackets to something resembling OCaml's style of 'sig ... end'? This was a temporary decision at the time, with a more C-like style rather than a Hazel-style.

cyrus-:
@gensofubi No, we're not trying to mimic OCaml syntax here -- the curly braces look good to me.

Negabinary:
image

I'm getting an error when I try to use modules on this branch now - I think it's a lang doc panel thing.

gensofubi:
@Negabinary Would you try resetting local storage? This happens when switching between branches.

Negabinary:
ah yes that worked thanks!! @gensofubi

src/haz3lweb/LangDocMessages.re
```diff
@@ -1506,6 +1566,62 @@ let let_ap_exp: form = {
   };
 };
 
+let module_ex = {
+  sub_id: "module_ex",
+  term: mk_example("module M = \ntype M = Int in\nlet x : M = 1 in\nin\nM.x"),
```
Negabinary, line 1571: It's a bit confusing to have the type name and module name be the same here.

src/haz3lcore/dynamics/DH.re
```diff
@@ -36,6 +38,7 @@ module rec DHExp: {
     | Cast(t, Typ.t, Typ.t)
     | FailedCast(t, Typ.t, Typ.t)
     | InvalidOperation(t, InvalidOperationError.t)
+    | ModuleVal(Environment.t)
```
Negabinary: This might be better as a ClosureEnvironment.t (like in Closure()), `ClosureEnvironment` s are pretty much exactly the same as ClosureEnvironment.t, but can be compared quicker, which might be useful for bigger modules.

src/haz3lcore/statics/Statics.re
Negabinary, line 1: I don't understand enough about Hazel statics yet to actually read this file - but I've had a play around on the editor

```
type U = {g:Int} in
let f : Int -> U = fun x -> 
  module T = 
    let g = x in 
  in
  T
in
f(4)
```
works but
```
type U = {g:Int} in
let f = fun x -> 
  module T:U = 
    let g = x in 
  in
  T
in
f(4)
```
gives a weird `T not found` error

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -1271,7 +1327,49 @@ and evaluate_test_eq =
...
+      | BoxedValue(d1)
+      | Indet(d1) =>
+        switch (matches(dp, d1)) {
+        | IndetMatch
+        | DoesNotMatch => Indet(Closure(env, Let(dp, d1, d2))) |> return
+        | Matches(env') =>
+          let env = evaluate_extend_env(env', env);
+          let env_in =
+            Environment.union(env', env_in)
+            |> ClosureEnvironment.of_environment;
```
Negabinary: I think the   `|> ClosureEnvironment.of_enviornment |> ClosureEnvironment.map_of` is redundant here

Negabinary (changes requested):
Looks pretty close to ready to me! I've added a few small comments around the place. You've been working on Hazel a lot longer than me so please do push back on any of my comments. I don't yet know enough to fully read the statics section.

src/haz3lweb/LangDocMessages.re
gensofubi: Thanks for the feedback! Here I’m to show a special feature in Hazel Module: its inner type member with the same name as the parent module can be directly used without dot access. This is for convenience since we expect most module to have methods working on one specific type (like the type T in most Reason module). I'm still considering how to state this in brief and clear words in the document.

src/haz3lcore/dynamics/DH.re
gensofubi: Probably because old ClosureEnvironment doesn't support some methods. Now changed to ClosureEnvironment. Thanks!

src/haz3lcore/statics/Statics.re
gensofubi: Seems that modules annotated with typevars are not correctly recognized. Resolved. Thanks!

src/haz3lcore/dynamics/Evaluator.re
gensofubi: Resolved. Thanks!

src/haz3lweb/LangDocMessages.re
Negabinary: ooh that's a nice feature! Using M makes more sense now but yeah I agree it might need some more explanation.

cyrus-:
@gensofubi 
1. This PR needs a merge with dev
2. When my cursor is on a projection, the context inspector seems to be showing the module signature rather than the context at that position in the code:
image
3. The ExplainThis description for dot says that you can put an arbitrary expression `e2` to the right of the dot, but it should in fact only be a label:
image
4. The cursor inspector message says that the type of a module binding is `?` which is consistent with its signature. This should just say its type is the signature, just like a let binding with no type annotation on it:
image

cyrus-:
Nested modules do not seem to be working correctly:
image

gensofubi:
@Negabinary During my work I found a problem that later members might refer to the previously defined member variables. So instead of creating a map, I introduced an empty ModuleVal  in the elaborator, and let expression will expand it in the transition (as long as it found the body to be a ModuleVal). 
Current result: 
{34989723-FB03-41fe-AD7D-DBBACA25BE24}

Negabinary:
> @Negabinary During my work I found a problem that later members might refer to the previously defined member variables. So instead of creating a map, I introduced an empty ModuleVal in the elaborator, and let expression will expand it in the transition (as long as it found the body to be a ModuleVal).

Good point about the previously defined member variables! I like the fact that, using this method, the let expressions within modules work very similarly to normal let expressions too.

This evaluation order kind of suggests you could do something like this:

image

but y doesn't seem to get captured by N in the evaluation at the moment. Does having N be M + y here make sense?

src/haz3lweb/explainthis/data/ModuleExp.re
```diff
@@ -0,0 +1,46 @@
...
+      term: mk_example("module M =\nlet x = 1 in\nin\nM.x"),
+      message: {|
+      A module with a member x=1 is created and assigned to tag M, so the expression is evaluated to 1.
+              |},
+    },
+    {
+      sub_id: ModuleWithType,
+      term:
+        mk_example(
+          "module OptInt =\ntype OptInt = None + Some(Int) in\nin\nlet x:OptInt = Some(1) in\nx",
```
Negabinary: This bit goes off the screen on my computer - maybe add a few newlines

image

src/haz3lweb/explainthis/data/ModuleExp.re
```diff
@@ -0,0 +1,46 @@
...
+      (pat', e_def', e_body') =>
+      [
+        mk_module([[space(), pat', space()], [space(), e_def', space()]]),
+        linebreak(),
+        e_body',
+      ]
+    ),
+  explanation:
+    Printf.sprintf(
+      "Module definition expression. The variables defined in [*definition*](%s) is packaged as a module and bound to the [*module variable*](%s) in the [*body*](%s).",
```
Negabinary: "The variables defined in definition _are_ defined"

Negabinary: oh maybe also mention types as well as variables

src/haz3lweb/explainthis/data/DotExp.re
```diff
@@ -0,0 +1,27 @@
...
+  form_id: DotExp,
+  abstract:
+    Simple.mk_2(("Module_name", mod_id), ("x", mem_id), (e_mod', e_mem') =>
+      [e_mod', space(), dot_exp(), space(), e_mem']
+    ),
+  explanation:
+    Printf.sprintf(
+      "Dot access. Retrieves the value of the [*member*](%s) x of the [*module*](%s).",
+      mod_id |> Id.to_string,
+      mem_id |> Id.to_string,
```
Negabinary, line 15: colours are the wrong way round here

src/haz3lweb/explainthis/data/DotTyp.re
```diff
@@ -0,0 +1,27 @@
...
+  form_id: DotTyp,
+  abstract:
+    Simple.mk_2(("ModuleVar", mod_id), ("t", mem_id), (t_mod', t_mem') =>
+      [t_mod', space(), dot_typ(), space(), t_mem']
+    ),
+  explanation:
+    Printf.sprintf(
+      "Dot access. Access the [*type member*](%s) of the [*module*](%s).",
+      mod_id |> Id.to_string,
+      mem_id |> Id.to_string,
```
Negabinary, line 15: colours wrong here too

src/haz3lweb/explainthis/data/DotTyp.re
```diff
@@ -0,0 +1,27 @@
...
+    ),
+  examples: [
+    {
+      sub_id: DotTyp,
+      term: mk_example("module M =\ntype T = Int\nin\nlet x : M.T = 1 in\nx"),
```
Negabinary: I think this example is missing an 'in'

src/haz3lcore/dynamics/Transition.re
Negabinary, line 1: This file looks good to me - one question I had, see https://github.com/hazelgrove/hazel/pull/1020#issuecomment-1977190297

I think we'll want to eventually have an option where you can hide all steps that occur inside module bodies, but we can probably do that after this pr.

src/haz3lcore/lang/Form.re
Negabinary, line 1: The hole at the end of the module still really bugs me - I wonder whether @dm0n3y has any ideas for how to get rid of that hole?

src/haz3lcore/statics/Statics.re
Negabinary, line 1: image

```
let functor = fun X:{Type T = Int} ->
  module F =
    type T = [X.T] in 
  in
  F
in
module V =
  type T = String in 
in
module U = functor(V) 
in
let f: U.T = 3 in
f
```

ok so two things:

1. in line 10, `module U = functor(V)` it says 
image
but I think they should be inconsistent

2. on line 12 `let f: U.T = 3 in` it says 
image
is this right? I'm not sure what the desired behaviour would be here.

Negabinary (changes requested):
I think we need a way to show off all the cool features of this pr - I keep discovering new things you can do when looking through your code, like you can make functors by just having a function that takes a module and returns a module - that's awesome!

I wonder if making a "Modules" documentation example would work? (similar to how "Basic Reference" shows off most of the features and "ADT Dynamics" shows off the features of ADTs) Then, a programmer could scan through and see all of the things they can do with modules. I'd think it would include at least type definitions, variable definitions, type accesses, variable accesses, using just the module name as a type, module type annotations, how to write functors - but there's probably more cool things I still haven't found yet too

src/haz3lcore/lang/Form.re
Negabinary: I was discussing modules with @LighghtEeloo yesterday and he pointed out that the modules here are kind of like delimited closures - the contents of the module is everything that is defined "outside" of the hole, but it's delimited because it only captures things "inside" the module keyword. In that sense, it's nice to have a hole to show that that's what the module is.

So another idea to consider is maybe keeping the hole at the bottom of the module but making it look different to make it clear that this is the location that's going to be 'captured' by the module?
image

gensofubi:
> > @Negabinary During my work I found a problem that later members might refer to the previously defined member variables. So instead of creating a map, I introduced an empty ModuleVal in the elaborator, and let expression will expand it in the transition (as long as it found the body to be a ModuleVal).
> 
> Good point about the previously defined member variables! I like the fact that, using this method, the let expressions within modules work very similarly to normal let expressions too.
> 
> This evaluation order kind of suggests you could do something like this:
> 
> image
> 
> but y doesn't seem to get captured by N in the evaluation at the moment. Does having N be M + y here make sense?

I've done a few trials about this but found it tricky. To implement this, I need to evaluate the body expression of Let expression first, which is not the current practice. It may be easy to make exceptions for modules that are directly represented by labels, but it will be difficult for modules that are generated in more complex ways (such as functions).
Alternatively, a module union operator could be defined to achieve this functionality, but I'm afraid that might be too complicated. I'm currently considering modifying stepper's expression slightly to avoid this misleading intuition.

cyrus-:
The module body won't be a regular let expression after David's tylr refactor. We could fix it even now by defining a new sort for `ModuleExp` and use different keywords like `val` instead of `let`, but probably not worth the effort right now.

src/haz3lcore/statics/Statics.re
gensofubi: 1. resolved
2. I think U.T is expected to be [Int] here? 
btw making a note of new bug discovered when fixing this: The newlines in the module in the results will cause the position of the red box to shift

Negabinary: oh yeah, my bad on (2.)

cyrus-:
The types being reported for member variables of modules seem to be inconsistent.

In the following, the type is `N.N` (expecting just `N` as long as the alias `N = N.N` is in scope):

```
module N = 
  type N = String in 
  let z : N = "Abc"  in   
in 
N.z
```

In the following, where the module is nested deeper, however, the type is `String` (I would expect either `M.N.N` or `M.N`):
```
module M =
  module N = 
    type N = String in 
    let z : N = "Abc"  in   
  in 
in 
M.N.z
```

gensofubi:
@cyrus- These issues have been addressed

Negabinary:
I finally understand some amount of casting, so I found a few casting bugs - they're pretty similar to some bugs I found recently in dev so they won't be too hard to fix but casting's always annoying...

image

**Problem # 1:** This one is because of the definition of a ground type. Specifically ground types should have the property that if two ground types are consistent then they are equal.

e.g.   Int   &   Int               are consistent and equal     (fine)
        String & Int              are consistent and not equal   (fine)
        [?] & [?]                    are consistent and equal
        [Int] & [?]                 are consistent, but [Int] is not ground so this is fine
 
If you have a basic type like int, string, then it is ground, but if you have a more complex type like list, functions, sums or modules, then only the version that only has holes should be ground

i.e.    [?],   ? -> ?,  +X(?) + Y,  {T1: ?, T2:?}                              should be ground
but   [Int],   String -> ?,  +X(Int) + Y(?),   {T1:Int, T2:?}          should not be ground

you can think of the ground type as representing just the top-level type - i.e. we know it's a module but we don't know what kind of module it is.

**Problem # 2:** once you fix problem # 1, I think you'll find that this example doesn't show an error, but it does get stuck later. This is because of the "canonical forms" thingy in the appendix of the 2019 paper. I'll probably try and write it out in the codebase because a lot of us are tripping up over this, and it's kinda hidden at the bottom of that paper.

It would be nice if we could say "if it has type module, and it is a value, then it is a Module(...)"; unfortunately it's not quite that simple with casts...

"If it has type Int, and it is a value, then it is an Int(...)" is true. "If it has type [...], and it is a value, then it is a List(...)" is unfortunately not true. This is because the list might have some casts from [X] to [Y] outside it. When you try and open up a list you have to apply a cast from X to Y for every element before you can look inside. This is what I tried to do with the `unbox_list` function in Transition.re

"If it has type {...}, and it is a value, then it is a Module(...)" is also not going to be true. Once you fix problem # 1, the casts around module values will stop disappearing, and so now we'll have to deal with them. In particular if you have a module {x=4}:<{x:Int} => {x:?} > then when you access `M.x`, you need to add a cast to it and return `x<Int=>?>`

I know this stuff is complicated, and I barely understand it myself, so if you want to meet at some point to discuss I'd be happy to.

code for reproducing example:

```
module T = 
  let x = 5 in 
in
module S : {x: } = T in
module U : {x:String} = S in

U.x ++ "hello"
```

should evaluate to

FailedCast(5, Int, String) ++ "hello"

on an unrelated note:

image

idk whether the desired behaviour is to evaluate this or to get stuck with some sort of error message, but this evaluation is just hanging atm

```
module T = 
  let x = 5 in  
in
let s :   = T in
s
```

cyrus-:
marking as draft until above casting issues are resolved

cyrus-:
@gensofubi can you merge with dev when you get a chance

cyrus-:
Getting a strange run-time cast failure on this example:
image

Negabinary:
Back with another type bug sorry!!!

An important property of type consistency is that it's a symmetric relationship. (A ~ B ⇔ B ~ A) and unfortunately that doesn't hold for modules at the moment. In particular `{x:Int}` is consistent with `{?}` but not the other way around.

At first it seems fine because whenever we need a `{?}` we can always use a `{x:Int}`, but the problem comes when you look at the type that a function takes. (This problem is more generally called covariance and contravariance https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science); but we in Hazel avoid it entirely by making consistency symmetric)

```
let f : {x:Int} -> {x:Int} = fun M -> M in
let g : {?} -> {x:Int} = f in
```

This type checks fine because `{x:Int}` is consistent with `{?}`; but if you look at what the function `g` now does, it allows us to use any module as an `{x:Int}`.

Here's an example where I've tricked the type checker into letting us use {y=4} as an {x:Int} using this g function:

image

```
let f : {x:Int} -> {x:Int} = fun M -> M in
let g : { } -> {x:Int} = f in

module M =
  let y = 4 in 
in

g(M).x
```

Negabinary:
Looks good now! I wasn't able to find any more casting bugs - not letting users define rho types feels like a good practical workaround for things like {x:?, …}->{x:?, y:?}->{y:?, …} for now.

cyrus-:
@gensofubi reviewing now, but there is a merge conflict when you get a chance

cyrus-:
@gensofubi I'm a bit confused about what is the difference between a let binding with a module signature in the type position and a module binding. The error message in this screenshot is quite confusing:

image

disconcision:
@cyrus- syntax notes / questions:

1. Is the only reason for a separate 'module' definition form to avoid ambiguity with a let which uses a nullary constructor as its pattern?
2. What is your view for what a 'hazel program' should be in the world tree model? i.e. if i just want to start writing some code without regard for interface, what is my top-level? am i implicitly (or explicitly) wrapped in a module (with some placeholder name)? or implictly in some kind of block form?
3. remind me about your position on the existing pseudo-top-level definition forms (let .. in and and type .. in). do we want to separately retain (both) these? if new scratch pads as per 2 are module-wrapped, then it seems at least that we don't strictly need a pseudo-definition type form any more. (unless we're keen on retaining the ability to define types inside functions, AND are not also replacing function bodies with some kind of curly-braces form)
4.  As per the last point, it feels to me like you can basically also use curly braces to create pseudo-top-level definition contexts which would obviate the need for separate let..in, but at the cost of maybe creating some pseudo-ambiguous edge cases between module and function definitions (pursuant on above selections). 

For concrete illustration here's a version of the contents of a scratch pad which:
1. retains separate module def form
2. is implicitly module-wrapped, or implicitly block-wrapped... doesn't seem to make a difference in this model 
3. removes existing let..in and type..in forms
4. introduces a curly-braces block form which is interpreted as a block unless it's the body of a module definition form. semantics are: returns the value of the trailing expression, if any, or null if there's a trailing definition form.

```
module M : {
  type T = Int;
  let x : T;
  let y: T
} = {
  type T = String;
  let x = "yo"
};
type R = String;
let f: Int -> R = fun x -> {
  let q = string_of_int(x);
  "q: " ++ q
};
f(4)
```

in the above module blocks and non-module blocks are both semi-colon separated lists of convex forms, including prefix-shaped definition forms (ie the semicolon is not part of let/type) taking a right unidelimited child. i've opted against having an optional trailing semicolon, but this seems possible if-desired.

the above is almost achievable using the current syntax model... the only thing not trivially doable is the `let` in the module signature. if that's not possible, we could temporarily replace it with val/var/whatever.

Anyway I know the above is a bit different than we talked about, but I'm not totally clear on what your exact preferred model is, so this is intended as starting point for discussion. My goals are to (within reason) stem keyword proliferation and avoid having incidentally-different ways of doing the same thing. But I'm also ~ fine with keeping let...in and using var/val/whatever for the block-level let.

I'm curious if there's a reasonable way of finessing block versus module forms in some way which obviates the need for a separate module definition keyword, assuming some other solution were found for the constructor issue above. like perhaps curly braces is a module IFF there trailing item is a definition. Perhaps not a good idea though, would have to guard against weird error message when you accidentally do this when defining a function etc.

disconcision:
Screenshot 2024-08-17 at 10 05 42 PM

after messing around a bit i think it'd be possible (with several days of work) to get syntax in the screenshot to work (without the sort errors currently displayed). of the four issues described in the comments i know how to reasonably fix three of them, and the other seems like it must be straightforwardly possible, though not immediately obvious to me.

this uses "typ" and "val" and as such retains the old "type" and "let" without change. "typ" needs a better name if we're going to keep these as alternates and retain the existing forms; i still think it's a better idea to ditch the old ones and let typ be type and val be let.

iirc you were talking about the option of making the forms like ["let", "=", ";"] as opposed to the infix ";". I experimented with this; there doesn't seem to be any good way unless we make a subsystem where multi-delimiter forms can change to other multi-delimiter forms. i.e. we need `module M: {typ T=Bool|}` to be grow a semicolon if we want to add more definitions etc. seems like a bit more work than the other option with advantages unclear to me at the syntax level. as an alternative we could perhaps pretend that things are like this for maketerm parsing and indication decoration purposes (as we pretend for list lits and cases).

7h3kk1d:
I'm curious how having fully first-class modules will allow for type-abstraction in the future. Naively it seems like figuring out type compatibility between abstract types would require evaluation?

gcrois:
Some notes from a discussion today: https://github.com/hazelgrove/hazel/discussions/1558

cyrus-:
added via an alternative design, inspired directly by this PR. thanks @gensofubi

Issue #824, opened 2022-09-23
as patterns


cyrus-:
```
case x
  | (a, b) as tpl => ...
```

cyrus-:
@green726 experimented with this in #1021 finding we probably need to await new version of tylr to make this work cleanly.

PR #1021, opened 2023-05-20
Addition of "as" patterns. 

green726:


cyrus-:
@green726 You're getting build errors https://github.com/hazelgrove/hazel/actions/runs/5137511215/jobs/9245694191?pr=1021

green726:
@cyrus- Ah, sorry. It appears that if the code is not formatted (by make) the initial build fails. I didn't realize this, and pushed the code formatted by my language server instead of the dune format. It build correctly now.

cyrus-:
There is something wrong with token merging/splitting, e.g. pressing backspace here causes the wrong behavior:

image

green726:
Ah, ok, thank you. I didn't catch that, I'll fix it.

green726:
I'm looking into it now. Currently, my best guess is that there is an issue with the zipper for the as patterns. I was looking in to the destruct, and the behavior you described can be solved by forcing the as patterns to skip the case which performs the "merge" operation (manually adding a case for when the left neighbor is "as"). However, the same issue exists with the last character in the actual word "as". Due to this, I'm guessing that something is wrong with the zipper (possible cursor placement?) for the characters in the "middle" of the as pattern. I will keep looking in to this and trying to fix it.

green726:
@cyrus- I fixed this deleting issue with the “as” patterns. Unfortunately, I was only able to do this in a very "hard coded" way. I had to add manual checks to parts of the "Destruct.go" for deleting as patterns. As far as I can tell, this is due to the fact that "as" patterns are the first keyword-only infix operator. All other infix operators have some sort of symbol (not letter) associated with them. For those infix operators, when "Form.is_valid_token" is called (in the destruct), it fails, as the symbols combined with letters create an invalid token. With the new "as" patterns, however, the "as" (or remaining parts of it) merges with the variable name pattern to create a valid token (a word). I haven't been able to think of any quick ways to fix this (that don't require major reworks of what is considered "valid" or the destructing). Maybe the destructing (especially with infix operators) is something that needs to be reworked in a future update, specifically, to not rely on these token validity tests?

Should we merge this version of the "as" patterns, or wait until another workaround is found?

cyrus-:
Closing this as pending the next version of tylr. Left a link to it in #824.

PR #1024, opened 2023-05-22
TyDi

### LINKED ISSUE #722: Term & Tile Selection
1. Create action to select the tile containing the caret
2. Create action to select the term containing the caret
3. Pick keyboard bindings (? check prior art)
4. Pick mouse bindings (double/triple-click)

strategy for 1:
- use indicated tile id
- use measure map to return the first and last positions of the tile
- use do towards goal left to move to first position
- use do towards goal right to grow selection to first position

strategy for 2:
- similar to 1, except abstract second step into a fn that returns first and last position for a term. this should take the focal segment, generate the skel, recurse to the tile id, then continue recursing into its left/right children to find the left and rightmost tiles in each.

strategy for 4:
- Use event.detail to detect triple-click

### LINKED ISSUE #834: Restore basic autocomplete
Hazel 2 effort: #543

### LINKED ISSUE #837: Type-directed tuple completion

### LINKED ISSUE #852: cannot bind variables with initial underscore

### LINKED ISSUE #853: allow primes in variable names

### LINKED ISSUE #1013: Make statics toggle actually turn off statics

### LINKED ISSUE #1022: Invalid characters are not allowed to be put in comment (or string literals)
comments like #%^{'}# is not allowed

### LINKED ISSUE #1048: cannot apply a function to unit input
image

creates a grout in between

### LINKED ISSUE #1076: Typing "x-1" results in "x><-1"
That is, a concave grout is inserted. Perhaps introduced in int or float improvement PR?

### LINKED ISSUE #1089: Typo in CI for expression "Trivial Litera"
Needs an "l"

disconcision:
This adds type-directed autocompletion and a number of significant supporting features and related bugfixes.

Fixes:
- Closes #1013
- Closes #1076 
- Closes #1089

Abstract:
- TyDi offers per-token type-directed autocompletion for Hazel (closes #834, closes #837). This design is focused on labor-saving, with suggestion per se only a secondary goal. Currently, TyDi offers at most a single suggestion, and suggestion ranking is purely lexicographical. Implementation errors excepted, all suggestions are syntactically correct, and are type-correct in a sense that will be made precise.

Behavioral specification:
- The completion buffer appears after any `Insert` or `Destruct` action, when the caret is to the immediate right of a (non-empty) single-token term which has an error. The completion buffer displays suggested a (non-empty) affix which can be appended at the caret. For example if the token/term `sug` has an error, and the caret is to its right `sug|` then we may see `sug|gestion`, with affix `gestion`. 
- In some cases, the completion buffer may contain multiple tokens; e.g. `foo(` (2 tokens) or `foo( )::` (5 tokens), which suggest a function application, and an application and a cons operator respectively
- Note that the conditions above imply that suggestions will never be shown when the caret is on an empty hole or a type-correct token, and that single-character tokens will never be suggested.
- Pressing `tab` accepts a completion, which pops the buffer, parsing its contained tokens into the surrounding syntax. Note that this may result in tokens being added to the backpack, if for example a `)` if the buffer contains `foo(`.
- Any other edit or movement action removes the buffer; it will be regenerated only on `Insert` or `Destruct` actions.
- Completion can be disabled via a menu toggle
- In principle, TyDi has the following correctness and completeness goals: given the activation conditions above, any TyDi suggestion, if accepted, will decrease error count, and furthermore if there exist any error-decreasing completions of the existing token, then TyDi will make a suggestion
- However, the error criteria doesn't strictly hold. When the expected type is `Bool`, TyDi also suggests `Int`s, `Float`s, and `String`s, because these can become valid if followed up by a comparison operator. We don't want to fail to suggest something just because the combination of non-prefix syntax and linear typing order necessitates passing though a incorrect state. So more subtly, we wish to suggest all tokens which will result in less errors after enough subsequent tokens have been entered to definitively establish the parent of the original token, e.g. if the expected type is Bool, and we have `num: Int`, then we might suggest `n|um` since, even though this will still be an error after completion `num|`, it will become correct if we then insert a comparison `num<|`. The difficulty of this case is that there are multiple such operators which could follow, unlike the situation with cons (`::`) and function application

Supporting features:
- Type anything: Any printable ASCII character can be inserted in the editor. In particular, all characters except `"` are valid in strings; similarly for comments except with `#`. Variables can now contain `'` and begin with `_`. This change required a series of subtle adjustments to things like invalid token treatment, grout/whitespace interaction, and token split/merge logic. In particular , there is now a notion on 'default molds' which assign every token not otherwise molded by the language definition as and invalid token which is either convex or concave, depending (mostly) on its constituent characters (closes #852, closes #853, closes #1022)
- Improved Live Coding: Cursor inspector & cell evaluation feedback are no longer disabled when the backpack is nonempty. Instead, we attempt to complete the syntax by dropping everything from the backpack (using lightweight heuristics which don't always succeed) before parsing for semantics. (This will be unnecessary in the new syntax model)
- Smart Select Click: Double-clicking on a delimiter selects its tile e.g. `let x = 1 in` ; triple-clicking one selects its term e.g. `let x = 1 in body`, or if the tile and term coincide, selects parent term, e.g. triple-clicking on `foo` in `foo(bar)` selects the entire application. Similarly, double-clicking on an infix operator selects the operator, and triple-clicking selects its operands as well (closes #722)
- Nullary Ap: You can now write `f()` for `f: () -> ?`. If `f`'s argument type is not consistent with `()`, an error is placed on the whole ap. (closes  #1048)
- Toggle elaboration display: Off by default. Large elaborations present a perf issue, in some cases nullifying the advantage of turning off dynamics.
- Explicit hole '?': A temporary convenience; '?' has the same semantics as a convex hole, but is reified as a token and is persisted across the text representation. This is useful if you have long-lived holes and don't mess with reformatting due to whitespace issues, in particular when copy/pasting or otherwise parsing from text. Relatedly, the Printer now supports specifying a string (e.g. `?`) to use to print holes.

Internal changes:
- CoCtx in patterns: The CoCtx has been piped to patterns to enable unused variable suggestions.
- Core settings: A subrecord of settings for core systems (statics + dynamics); separate mostly for dependency reasons
- Interface.re Consolidation: All access to statics and dynamics in the view layer should go through Interface.re functions. These are now extended to take initial ctxs/envs as appropriate. Statics can now specify whether the buffer should be ignored or included, and whether to try to complete incomplete syntax before getting statics.
- Defunctionalization: Some info (Self.NoJoin) & actions (Move to piece goal) were taking function parameters. This is problematic as these types can end up in places where they get compared, e.g. for memoization, so these have been defunctionalized.
- Buffer selection: The selection has been co-opted as a buffer. The buffer is ephemeral in that it disappears on movement or edit. The buffer can either contain valid hazel syntax (Parsed variant) or an unparsed string (Unparsed variant). In the latter case, it will be parsed if the buffer is accepted.
- Model.meta:  A subrecord containing non-persistent content including UI state and eval results
- ModelSettings renamed to Settings


cyrus-:
@disconcision need to fix some conflicts

cyrus- (approved):
other than needing a merge, this looks good to me.

disconcision:
@cyrus- this appears to still have an outstanding bug which is a bit weird... below is completions as they should and do appear locally, next is completions as they appear on the build server. can't figure out why the discrepancy exists, code seems to have built okay. it's not a css issue, the appropriate class isn't being put on on the server... what do you see locally?

Screen Shot 2023-11-30 at 12 02 11 AM
Screen Shot 2023-11-30 at 12 01 45 AM

disconcision:
The above issue appears to be an issue with the recursive memoization of of_piece. I find it confusing how it never seems to occur locally and always seems to occur on the build server, but repeated trials show this to be the case.

The relevant change is here:
https://github.com/hazelgrove/hazel/pull/1024/commits/3b61185c0d3fec7ada5f681b9156a2a4f7007989

The relevance to the above phenomena is that buffer_ids is a list of ids which is passed down through the code render to put a style on tokens inside the buffer. This is necessary because the code rendering is based on the segment structure: there is no notion of a selection and hence a buffer after unzipping. A slightly more principled way of doing this would be to have a separate pass which calculates a syntactic infomap; this could also track sorts and other attributes which inform token styling. I've resisted doing this because it's very syntax-model dependent, and another whole tree pass will be a perf hit.

In any case, the buffer_ids are part of the memo table key, so it's very much not obvious to me why the memoization is causing the above bug, whose direct cause is that buffer_ids doesn't contain the id of the completion token in the buffer.

This is the second significant bug caused by this memoization; note that the memo key incorporates other variables that aren't strictly function arguments... this is because the whole thing is enclosed in a functor, and there was a previous bug where indents were wrong because the memoization wasn't taking into account the Measured. Again, not obvious if this is relevant here, but given this I'm inclined to remove the memoization for now as it's starting to be more trouble than it's worth.

(Unless @cyrus-  you can immediately see the issue)

This memoization offers a 100% speedup of Code.of_segment, which corresponds to an overall ~8% speedup of basic syntax operations with everything turned off, and a ~4% speedup with everything on.

(As a side note, when doing the merge with dev I found the part in school mode that was still doing unnecessary splicing and elaboration when statics is turned off, so school mode is now significantly faster with everything turned off)

cyrus-:
You're seeing a behavioral difference between your local `release` build and the one being built by GitHub actions? Very odd. Can you make sure you have the same version of everything in `make deps`?

cyrus-:
I'm okay with disabling that memoization if the fix isn't clear, but that's very odd!

Are there any other TODOs for this PR after that is resolved?

disconcision:
@cyrus- yes. is there quick way to confirm those versions? i reran make deps, turns out i had a couple things out of date; doesn't seem to make a difference for locally behavior.

anyway turns out disabling the memoization is for the best... I should have thought through the approach more, it was actually causing slowdown on large programs, i think due to the way the default ocaml hash works.

otherwise I think this is good to go, although after having done the merge I have some slight concerns about the way queue_scroll_to_caret was implemented, i ended up having to thread that parameter through some of my new stuff.. I'd like to maybe get that conceptually resolved and if possible cleaned up here before the merge... I think it should be possible to factor queue_scroll_to_caret out of update entirely, I prospectively made that change and pinged matt about in on #1138

PR #1031, opened 2023-05-24
Integer polish

xzxzlala:
Brief Description
- underscores in integer literals
  Add underscores in style.css file.

- add static error for integer literals larger than maximum int or smaller than minimum int
  When parse the integer, if it is "arbitary_int" but not a "legal" int, give the error of integer out of range.

- add builtins for conversions to other types (float, string)
  The int_of_float and float_of_int is alreay existed, so I add the int_of_string and string_of_int builtins. 

- negative integer literals should be IntLits
  I adjust the regular expression for "arbitary_int" to contain the negative integers.


cyrus-:
@xzxzlala By underscores in integer literals, I meant things like `1_000_000`, not adding underlines to integer literals (OCaml supports this for example). Although in the context of EECS 490 the underlines might be useful too.

Also you can't transition from a negative integer literal to a float -- try typing `-2` then `.` -- it should go to `-2.` but instead a grout is inserted.

Marking as draft again, you can click "ready for review" when those issues are addressed.

xzxzlala:
For the negative integer to float, I revise the regular expression of float to contain negative float.
For underscores in integer literals, now it should work. Integer starts with underscores aren't included.

cyrus-:
image

Out of range integers shouldn't cause an exception in evaluation / elaboration and the error message at the bottom shouldn't cause the cursor inspector to lose the context / sort.

xzxzlala:
Can I evaluate these bad-integers to zero, so the evaluation / elaboration is correct, but the user still gets the error message.

xzxzlala:
I elaborate the invalid integer to invalid text and the elaboration works correct now. 
And the cursor problem is fixed.

cyrus-:
@xzxzlala next time click the "Ready for review" button at the bottom of the PR rather than adding a label

cyrus- (changes requested):
Hazel is crashing when I try to call int_of_string on an invalid input, e.g. `int_of_string("1.5")`. This should cause a run-time error hole in the output.

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,7 +33,27 @@ module Pervasives = {
...
+        let f = Re.Str.string_after(f, 1);
+        let f = Re.Str.string_before(f, String.length(f) - 1);
+        if (String.length(f) == 0) {
+          BoxedValue(IntLit(0)) |> return;
+        } else if (Form.is_int(f)) {
+          let i = int_of_string(f);
+          BoxedValue(IntLit(i)) |> return;
+        } else {
+          Indet(
+            InvalidOperation(ApBuiltin(name, [d1]), InvalidIntOfString),
```
xzxzlala, line 79: here I use the invalid operator to indicate the invalid string to int conversion.

cyrus- (changes requested):
string_of_int doesn't seem to be evaluating to a string

image

cyrus-:
@disconcision all the scratch slides with tests are useful but maybe we should create another "Tests" or "Examples" mode that is a clone of Scratch mode for this kind of thing?

disconcision:
@cyrus yeah I'll try to put together something where it's easy to add new named slides, that way ppl can add one with their PRs as lightweight docs/tests

cyrus-:
@disconcision (left that comment on the wrong PR, sorry -- see ADTs)

xzxzlala:
Sry! I forget to add the double quotation marks.

cyrus-:
(waiting to review this pending float polish being merged in)

cyrus-:
@xzxzlala there is a conflict with `dev`, can you take a look?

xzxzlala:
> @xzxzlala there is a conflict with `dev`, can you take a look?

I see, I forgot to pull dev in my local repo and now it is resolved.

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,7 +33,27 @@ module Pervasives = {
   module Impls = {
     open EvaluatorMonad;
     open EvaluatorResult;
-
+    /* int_of_string implementation. */
+    let int_of_string = (name, r1) =>
+      switch (r1) {
+      | BoxedValue(StringLit(f) as d1) =>
+        let f = Re.Str.string_after(f, 1);
```
cyrus-: this is no longer necessary

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,7 +33,27 @@ module Pervasives = {
...
+        let f = Re.Str.string_before(f, String.length(f) - 1);
+        if (String.length(f) == 0) {
```
cyrus-: instead of these checks, use ocaml's int_of_string_opt (see float-polish for an example)

cyrus- (changes requested):

src/haz3lcore/dynamics/Builtins.re
xzxzlala: Got it and it has already been fixed in float-polish branch. 
Now the float-polish branch is merged. :)

cyrus- (approved):

PR #1032, opened 2023-05-26
Float polish

WondAli:
Aims:
- Add support for negative floats
- Add support for special floats (NaN, Inf)
added infinity, neg_infinity, and nan
- Add builtins for isNaN, conversions to other types
added is_nan, is_finite, is_infinite, string_of_float, float_of_string
- Full IEEE float syntax, e.g. 1.04e5


cyrus-:
@WondAli you might want to take a look at / coordinate with #1031, especially with the literal parsing stuff

cyrus-:
I can't write `4e5` -- only works when there is a decimal point there.

cyrus-:
@WondAli did you have a chance to look up the float literal regexp online?

WondAli:
> @WondAli did you have a chance to look up the float literal regexp online?

I did, and it looked very similar to the regex I wrote. The main differences were the option to explicitly use the + sign for positive floats and exponents (e.g. "+5.6e+5" for 5.6e5) which I do not have in mine, and some digits matching to 1-or-more instead of 0-or-more, but at the moment I kept 0-or-more since otherwise hazel wouldn't allow the partial state if it didn't match to anything. As another side effect, this means that ints could theoretically be matched as floats (e.g. "3" could be matched to a float type), but currently hazel will always first check that it does not match to an int when checking if it could be a float, so this should not be an issue.

cyrus- (changes requested):
currently `-1` is being parsed as a float literal instead of an integer literal on this branch

WondAli:
> currently `-1` is being parsed as a float literal instead of an integer literal on this branch

This is because currently in this branch, negative int literals are not yet supported since it is an addition of the #1031 Integer Polish. When they are supported, they will be properly identified as an int literal first.
I could add a quick regex change to support negative int literals

cyrus-:
@WondAli mark as ready for review once you've rebased on `integer-polish`

cyrus-:
Closing in favor of #1058 (in the future, you can change the base branch of a PR without opening a new one)

PR #1033, opened 2023-05-27
Unused Variable Warnings

lmulcahy21:
The basic functionality of the warnings works correctly, printing "unused [var name]" or "used [var name]" to the console. 

Would love review on these changes, as this does alter some of the core functionality of the backend, as well as the function definition for `upat_to_info_map`. 

After review and revision, next step is to merge in the `adt-defs-refutable` branch and implement a warning for this, so that the variable is highlighted in place.


src/haz3lcore/statics/Ctx.re
```diff
@@ -67,6 +67,25 @@ let free_in = (ctx_before: t, ctx_after, free: co): co => {
   );
 };
 
+let is_unused_in_free = (free: co, name: string): option(bool) => {
```
disconcision: a co_ctx is a VarMap so you can use VarMap.lookup. This fn should probably just be a one-liner,something like
let co_ctx_contains = (free, name): bool =>  VarMap.lookup(free,name) !=None, and then at the call site you not it to get the opposite

src/haz3lcore/statics/Statics.re
```diff
@@ -297,6 +300,7 @@ and upat_to_info_map =
       ~is_synswitch,
       ~ctx,
       ~ancestors: Info.ancestors,
+      ~free: Ctx.co=[],
```
disconcision: at least as an implementation strategy, i suggest removing the default value of this parameter (yes, i remember that i told you to add it :) to make sure you examine all the use sites (e.g. the logic you've added for the let should be added for case & lambda as well). if there aren't that many call sites i would suggest just leaving it non-optional and specifying an empty free for callsites that would use the default value... your call, this is a tradeoff

src/haz3lcore/statics/Statics.re
```diff
@@ -346,11 +350,26 @@ and upat_to_info_map =
...
-       always Unknown Internal */
+       always Unknown Internal
+
+       - if unused then the self type is UnusedVar
+       */
     let ctx_typ =
       Info.ty_after_fix_pat(ctx, mode, Common(Just(Unknown(Internal))));
     let entry = Ctx.VarEntry({name, id: UPat.rep_id(upat), typ: ctx_typ});
-    add(~self=Just(unknown), ~ctx=Ctx.extend(entry, ctx), m);
+    let self_pat_val: Info.self_pat =
```
disconcision: it looks like this doesn't actually do anything at the moment? the second switch just ignores the value of self_pat_val?

regardless, when this is hooked up, split this logic off into a helper in Statics called pat_var_self. (in general i try to keep the main mutually recursive statics functions terse; they should only handle the piping of data between the different parts of the syntax, and all logic should be split into helpers)

disconcision (changes requested):
Looks good! Some cleanup and case/lambda still need wiring up. Then prob good to proceed with bringing in warnings code

lmulcahy21:
Just made a commit implementing most of these suggestions, but needs more time/work for debugging case statements. They currently aren't evaluating and the use of a variable in the case statement isn't counting as a "use" for the state of the variable, and im pretty sure it has something to do with an empty free/co-ctx instead of a properly created/propagated one

lmulcahy21:
UVWs are working, with a basic warning interface outlined through the Info module. Would love to get feedback on it, as many things could use improvement, such as the names of warnings, info that each displays, how they are found, etc.

cyrus-:
Nice work! 

It isn't handling recursive functions correctly, so maybe look into that: 
image

Another thing we should think about is how to handle variables where there is an empty expression hole in the scope. We think of holes as being potentially filled by anything, including that unused variable, so in such a case I would call the variable "possibly unused" rather than "definitely unused". I think we should hold off on displaying the warning marker in that situation.

lmulcahy21:
Also not working correctly for case statements, will look into this

image

cyrus-:
@lmulcahy21 
* need to resolve some merge conflicts
* the message in the cursor inspector needs to be human-readable
* we need to have it so if there is a hole in the scope of a variable, we don't show unused variable warnings, otherwise we end up with stuff like this in exercises mode where odd is showing as unused because it isn't explicitly used in the "in" portion (but it will be referred to by the tests etc. downstream in exercises mode):
image

lmulcahy21:
Had a boxing match with git and lost (had to force push) but warnings now (inelegantly) don't display if there is a hole in the co-context; done with a `__hole__` variable in the co-context, as i didn't want to change the current co-ctx's interface

src/haz3lcore/statics/Statics.re
```diff
@@ -715,9 +769,25 @@ and upat_to_info_map =
...
+      /* Warn about unused variables
+            - if not in CoCtx, then unused
+            - Display warning iff not a special variable (starts with "_")
+              and body does not contain a hole
+         */
+      if (String.starts_with(~prefix="_", name)
+          || CoCtx.contains_hole(co_ctx)) {
+        None;
+      } else {
+        switch (VarMap.lookup(co_ctx, name)) {
```
disconcision: factor out these three conditions into a separate named predicate fn

src/haz3lcore/statics/Statics.re
```diff
@@ -55,6 +55,17 @@ module Map = {
...
+  let warning_ids = (term_ranges: TermRanges.t, info_map: t): list(Id.t) =>
+    Id.Map.fold(
+      (id, info, acc) =>
+        switch (Id.Map.find_opt(id, term_ranges)) {
+        | Some(_) when Info.is_warning(info) => [id, ...acc]
+        | _ => acc
+        },
+      info_map,
+      [],
+    );
```
disconcision: we should ideally avoid having separate passes over the entire info_map where possible. probably want to either merge this with error_ids as a single pass which returns a pair or record of two lists, or maybe make this pass return a single list of some perhaps new data type which covers multiple classes of 'distinguished' ids (right now, Errors and Warnings)

src/haz3lcore/statics/Statics.re
```diff
@@ -471,8 +505,21 @@ and uexp_to_info_map =
         p_ctxs,
       );
     let e_tys = List.map(Info.exp_ty, es);
+    // Create co-ctxs for each case
     let e_co_ctxs =
-      List.map2(CoCtx.mk(ctx), p_ctxs, List.map(Info.exp_co_ctx, es));
+      List.map2(
+        (p_ctx, e_co_ctx) => CoCtx.mk(ctx, p_ctx, e_co_ctx),
```
disconcision: cant this just be eta-reduced to previous?

src/haz3lweb/www/style.css
```diff
@@ -1313,6 +1330,12 @@ svg.tile-selected {
   /* HACK(andrew) */
 }
 
+.ci-header.warn {
+    background-color: var(--warn-color);
+    color: #ffaa0c !important;
+    /* HACK(liam copying andrew) */
```
disconcision: lmao, but fair

src/haz3lcore/statics/Info.re
```diff
@@ -128,6 +128,22 @@ type status_variant =
   | Unique
   | Duplicate;
 
+[@deriving (show({with_path: false}), sexp, yojson)]
```
disconcision: was hoping we could have warnings as part of the status type, as another top-level entry alongside InHole and NotInHole. (but not sure if this is right approach), is there a reason this doesn't work out?

src/haz3lcore/statics/CoCtx.re
```diff
@@ -68,3 +68,5 @@ let join: (Ctx.t, list(entry)) => Typ.t =
     | Some(ty) => ty
     };
   };
+let contains_hole = (co_ctx: t): bool =>
```
disconcision: fine but document

disconcision (dismissed):
looking pretty good but wondering about a couple details of the approach

disconcision:
Screenshot 2025-06-30 150422

seems to be claiming everything as unused

lmulcahy21:
@disconcision working on a toggle for warnings right now to avoid this problem, as those are correctly marked as unused in the documentation page, and having the screen be covered in yellow like that is not ideal.

lmulcahy21:
@disconcision ready for review, have added switch for warning visibility toggling, made the color less harsh, and fixed how co-contexts are handled for case statements such that the patterns correctly see variable uses from the corresponding expression

src/language/statics/Info.re
```diff
@@ -719,6 +731,20 @@ let is_error = (ci: t): bool => {
...
+let is_warning = (ci: t): bool => {
+  switch (ci) {
+  | InfoExp({warning, _})
+  | InfoPat({warning, _})
+  | InfoTyp({warning, _})
+  | InfoTPat({warning, _}) =>
+    switch (warning) {
+    | None => false
+    | _ => true
+    }
```
disconcision: `let is_warning = ci =>warning_of(ci) != None`

src/web/www/style/variables.css
```diff
@@ -198,8 +204,9 @@
   /* BELOW CODE LEVEL */
   --err-hole-arms-z: 1;
   --tile-z: 2;
-  --err-hole-z: 3;
-  --select-z: 4;
+  --err-hole-z: 4;
+  --warn-hole-z: 3;
+  --select-z: 5;
```
disconcision: why this change? ie does anything go wrong if errors and warnings have the same z-index?

src/web/www/style/dynamics.css
```diff
@@ -152,6 +152,13 @@
   stroke-width: 1.2px;
 }
 
+.DHCode svg.warning {
+  fill: var(--warning-hole-fill);
+  stroke-dasharray: 1, 1;
+  stroke: var(--warning-hole-stroke);
+  stroke-width: 1.2px;
+}
+
```
disconcision: is this actually used? i thought DHCode stuff was dead code now. maybe check this

src/web/app/editors/decoration/Deco.re
```diff
@@ -530,12 +532,53 @@ module Deco =
 
   let errors = () => div_c("errors", List.map(error_view, error_ids));
 
+  let warning_view = (id: Id.t) =>
```
disconcision: i feel like in principle all the warnings view code should use the same path as the error stuff. basically it should be the same modulo a single top-level css class that you choose to decide whether its styled as an error or a warning. can you see if there's a quick way to combine this with the error equivalent? perhaps just pass a more generic function which take a string to be used as a css class; pass "error" for errors, "warning" for warnings, something like that, put that class on the main div for the errors, then have seperate css rules for each (but leaving common stuff in a shared class). this could probably be done for `warning_term` as well.

this one isn't a big deal if you don't have the time but i think a small refactor could save some code duplication here and in a couple other places

src/language/statics/Statics.re
```diff
@@ -1424,10 +1456,12 @@ and upat_to_info_map =
           id: Pat.rep_id(upat),
           typ: ctx_typ,
         });
+      let warning: Warning.t = Warning.var_is_unused(co_ctx, name);
```
disconcision: to calculate this warning it looks like you need the co_ctx and the name; both of are already captured by the info; you can add this line to Info.derived_pat (you'll need a switch there to to get the name from Var(name)).

I'm think if you do it this way you might not need to thread `~warning` around 20 times in this file.

src/language/statics/Statics.re
```diff
@@ -318,10 +325,22 @@ and uexp_to_info_map =
...
+    | MultiHole(_)
+    | EmptyHole
+    | Invalid(_) =>
+      CoCtx.singleton(
+        "__hole__",
+        Exp.rep_id(uexp),
+        Unknown(Internal) |> Typ.temp,
+      )
+    | _ => CoCtx.empty
+    };
```
disconcision, line 343: what is this? does it serve a functional role in this PR? needs a comment and probably should be in CoCtx not here

disconcision (changes requested):
1. let's go with some slightly less aggressive colors. something like:

Screenshot 2025-07-15 211959

either use those colors, or better yet, you could look at the oklch in variables.css for the red colors used for error hole fill and stroke and just copy those leaving the L and C the same but making the hue orange.

2. for the warning message itself, how would you feel about instead of `Warning: Unused variable <x>`, `Warning: Variable <x> is never used`? minor, just feels like more of a sentence to me.

3. pending the color change, i think these should be on by default. perhaps the documentation slides that don't reference most things they define could be updated to be more naturalistic 

Looks very solid! The only things which are blockers for me is the color change and checking if that threading of ~warning can be removed from statics as per the code comments. the others are nice-to-haves

src/language/statics/Statics.re
lmulcahy21: This is a hack to have terms recognize that their co-context would contain a hole, as we currently don't mark that in the info term (as far as I'm aware). I changed it to be `$hole` such that it can't be made by a user, but let me know if there's a better way to solve this issue of "there's a hole in a term's scope and it needs to know".

lmulcahy21:
image
@disconcision made the structural changes, only thing left is the `$hole` hack you commented on, see reply + new comment added to describe that code. The only thing left is the color of the holes; I didn't know how to replicate the oklch feature you described about taking the error colors and changing to yellow so hoped you might provide guidance/your ideal colors for that.

src/language/statics/Statics.re
disconcision: Hmm, I don't immediately follow... could you clarify?

lmulcahy21: The unused variable warnings shouldn't display if the subsequent program contains a hole, since it could be used inside of the hole. In order to check this I am using the co-context map improperly by having hole terms add a "$hole" entry to the co context and having the warning check for this "$hole" entry. If there's a better way to implement this check that'd be ideal, but I used this hack to get it working.

disconcision (approved):
I tweaked the colors again and factored out the view code duplication i mentioned; please confirm the latter doesn't cause any issues I missed.

Could you update the co-context hole comment with an explanation closer to the one you replied to me with? Other than that. lgtm

lmulcahy21:
Will do as soon as I get home, thanks for the feedback and color changes.

PR #1042, opened 2023-06-06
Fix #1040

### LINKED ISSUE #1040: multi-line tile selection backing drawn wrong if starting x pos is rightwards of ending x pos
For example in this screenshot the parenthesized term is selected; note that a backing is being drawn, but it just appears as a little rectangle thing above the first parens.

@dmoon1221 any thoughts? not really familiar with the code here, wondering if its a quick fix

Screen Shot 2023-06-06 at 12 32 07 PM

dm0n3y:
Fix https://github.com/hazelgrove/hazel/issues/1040


disconcision (approved):
lgtm

PR #1050, opened 2023-06-12
Cleanup: Precedence

WondAli:
Remake of earlier precedence-cleanup draft pull request.
Old Notes:

> 5/15/23: Started cleaning up files relating to precedence.
Less Than, Greater Than, and their float variants were previously hardcoded at 6, equivalent to the Plus operator. In OCaml, they have the same precedence as Equals.
I believe Cell-Join should have a later precedence than Boolean-Or, but I'm not sure where to to put it.
The Assignment operation is currently not in use, but I'm not sure what precedence that would be.


WondAli:
6/12/23:
Cell-join in forms.re has been renamed to sequence. Its precedence is now 13; fun_ and prod were moved up to 14 and 15 respectively.

cyrus-:
@WondAli there are now conflicts with dev after the ADTs merge -- can you take a look

cyrus- (changes requested):
Getting `Unhandled exception (Failure nth)` when I switch into exercises mode. Possibly something to do with adding sequence to the forms list?

WondAli:
I noticed while updating exercises that the files generated by the export modules buttons generate ```syntax_tests = [("not is applied", <fun>); ("odd is recursive", <fun>)]``` which causes a syntax error at the ```<fun>``` for me when building.
I manually replaced the syntax_tests with the ones from the old files

cyrus-:
This would be a useful cleanup to get back up to date with `dev` if we were sticking with the current tylr system, but since new tylr is incoming going to close this for now. Thanks for the effort on this though @WondAli !

PR #1052, opened 2023-06-26
pull before final push in github actions to fix issue

### LINKED ISSUE #843: Fix Git Actions system for closely spaced pushes

cyrus-:
I *think* this should fix the issue when simultaneous build actions are running but a little tricky to test, so going to merge it in for now and can revisit the issue if it persists.


cyrus-:
I think this should fix the issue when simultaneous build actions are running but a little tricky to test, so going to merge it in for now and can revisit the issue if it persists.

PR #1056, opened 2023-06-27
Let operators

### LINKED ISSUE #1035: Let operators

xzxzlala:
We can now define let-operators(let+, let*, let-) and use them.
image
If the let-operator is not defined by user, a warning message will show. 
image

About what part of the code I modified:
1, Developed from user-defined-operators branch, extend the user-defined-operators, I revise the regexp to contain the operators start with "let".
2, add let*, let-, let+ as forms. 
3, statics and dynamic of LetOp(op, pat, def, body), where "op" is a string indicating which let-operator is used. "pat", "def", "body" is just the same as Let(pat, def, body).


disconcision:
why is the undefined let operator error a dynamic error instead of a static error?

xzxzlala:
I remove the dynamic error and add the static error. 
image

cyrus-:
Since this is based on the user-defined operators, can you change the base branch of this PR to that branch? There are also potential conflicts that need to be resolved.

cyrus- (changes requested):
It isn't working correctly when the let+ definition is annotated with a type hole:
image

cyrus-:
@xzxzlala see also conflicts with `dev` that need to be resolved.

xzxzlala:
I merged dev and user-defined-operators(set origin/haz3l-user-defined-operators as upstream).
The annotation bug is a bug of user-defined-operators, I'm still trying to find a solution.

cyrus-:
Marking as draft pending merge of #1012

cyrus-:
@xzxzlala I think this should be ready to merge -- do you have time to resolve the last merge conflicts?

Negabinary:
I'm wondering whether it might be good to add a static error as well for undefined let operators, since it can be determined at compile-time whether a let operator is in scope?

Negabinary:
> I'm wondering whether it might be good to add a static error as well for undefined let operators, since it can be determined at compile-time whether a let operator is in scope?

Sorry I didn't notice Andrew had already brought this up - is there a reason we've gone back to dynamic errors?

cyrus-:
Oh yeah, I thought that is what we had done but seems it is still a dynamic error. @xzxzlala do you have time to make this a static error?

xzxzlala:
No problem! 
I could work on it. Sorry for the late reply, I didn't see the message because I had to use VPN to check my Gmail messages.
However, I'm now quite busy with graduation/capstone design. Maybe need more time :)

cyrus-:
archiving because this is no longer an active effort, but want this in a future PR

PR #1058, opened 2023-06-30
Float polish

WondAli:
Merging Float-Polish into Integer-Polish to be merged at once to dev


WondAli:
I suppose after this merge, I can continue to make float-polish updates if needed on the float-polish branch and merge them here again when integer-polish is ready to be merged into dev.

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,6 +33,40 @@ module Pervasives = {
...
     open EvaluatorResult;
+
+    /* is_finite implementation. */
+    let is_finite = (name, r1) =>
+      switch (r1) {
+      | BoxedValue(FloatLit(f)) =>
+        let b = Float.is_finite(f);
+        BoxedValue(BoolLit(b)) |> return;
+      | BoxedValue(d1) =>
+        raise(EvaluatorError.Exception(InvalidBoxedBoolLit(d1)))
```
cyrus-: shouldn't raise an InvalidBoxedBoolLit error

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,6 +33,40 @@ module Pervasives = {
...
+      };
+
+    /* is_infinite implementation. */
+    let is_infinite = (name, r1) =>
+      switch (r1) {
+      | BoxedValue(FloatLit(f)) =>
+        let b = Float.is_infinite(f);
+        BoxedValue(BoolLit(b)) |> return;
+      | BoxedValue(d1) =>
+        raise(EvaluatorError.Exception(InvalidBoxedBoolLit(d1)))
```
cyrus-: same

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -33,6 +33,40 @@ module Pervasives = {
...
+      };
+
+    /* is_NaN implementation. */
+    let is_nan = (name, r1) =>
+      switch (r1) {
+      | BoxedValue(FloatLit(f)) =>
+        let b = Float.is_nan(f);
+        BoxedValue(BoolLit(b)) |> return;
+      | BoxedValue(d1) =>
+        raise(EvaluatorError.Exception(InvalidBoxedBoolLit(d1)))
```
cyrus-: same

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -75,16 +110,51 @@ module Pervasives = {
         raise(EvaluatorError.Exception(InvalidBoxedFloatLit(d1)))
       | Indet(d1) => Indet(ApBuiltin(name, [d1])) |> return
       };
+
+    /* float_of_string implementation. */
+    let float_of_string = (name, r1) =>
+      switch (r1) {
+      | BoxedValue(StringLit(s) as d1) =>
+        let s = Re.Str.string_after(s, 1);
```
cyrus-: hm, is this because the argument to StringLit includes the quotation marks?

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -75,16 +110,51 @@ module Pervasives = {
...
+        let s = Re.Str.string_before(s, String.length(s) - 1);
+        if (String.length(s) == 0) {
+          BoxedValue(FloatLit(0.)) |> return;
```
cyrus-: not sure we need to special case the empty string here, just give it the same behavior as OCaml

src/haz3lcore/dynamics/Builtins.re
```diff
@@ -75,16 +110,51 @@ module Pervasives = {
...
+        } else if (Form.is_float(s)) {
```
cyrus-: maybe more efficient to just catch the exception that float_of_string would throw instead of separately checking if the input is a float

src/haz3lweb/www/style.css
```diff
@@ -1532,6 +1532,12 @@ svg.expandable path {
   font-size: 75%;
   top: -10px;
 }
+.DHCode .InvalidFloatOfString{
```
cyrus-: as a cleanup can we just have a single class for styling all such run-time errors instead of copying this for each new error

cyrus- (changes requested):

src/haz3lcore/dynamics/Builtins.re
WondAli: What kind of error would it raise? Would it be an InvalidOperation error?

src/haz3lcore/dynamics/Builtins.re
WondAli: It is because of that, yes.
I copied it to mirror how int_of_string handled it, but could look into modifying both to handle it a different way if that would be better

cyrus-:
@WondAli any remaining items on this PR? if not, click Ready for review

WondAli:
> @WondAli any remaining items on this PR? if not, click Ready for review

I was thinking of updating integer-polish and float-polish with the stringlit changes in dev to make sure it works still, just in case.
Though I did test with the quotation mark modification and it worked, so maybe we could also merge first and then update with dev

src/haz3lweb/view/dhcode/DHCode.re
```diff
@@ -59,6 +59,10 @@ let view_of_layout =
                  [with_cls("InvalidIntOfString", txt)],
                  ds,
                )
+             | OperationError(InvalidFloatOfString) => (
+                 [with_cls("InvalidFloatOfString", txt)],
```
cyrus-: let's just add an OperationError class rather than a separate class for each kind of operation error -- we will want to style all errors the same

src/haz3lweb/www/style.css
```diff
@@ -1526,19 +1526,9 @@ svg.expandable path {
...
-  top: -10px;
-}
-.DHCode .DivideByZero {
-  display: inline-block;
-  position: relative;
-  font-size: 75%;
-  top: -10px;
-}
-
+.DHCode .InvalidIntOfString,
```
cyrus-: with above change, this can be simplified

cyrus- (changes requested):

cyrus- (approved):

PR #1059, opened 2023-07-03
Partial application of non-curried functions

### LINKED ISSUE #795: Partial application for uncurried functions using placeholder

pigumar1:
Continuation of #957.
This feature introduces partial application of non-curried functions. As #514 mentioned:

> During elaboration, the partial application is replaced with a new function literal that takes in the unapplied arguments, then inserts them where they belong in the application of the original function.

The partial application of non-curried functions has been mostly implemented in #957, but it violates the principle that "deferral expressions should not synthesize to or analyze against any type." Additionally, there are some design issues to address. The purpose of this pull request is to resolve these issues.

### Examples

> ```
> val f: ('a, 'b, 'c) → 'd
> val x: 'a
> val y: 'b
> val z: 'c
> ```
> 
> |External Expression|Internal Expression|Synthesized Type|
> |:--|:--|:--|
> |`f (x, y, z)`|`f (x, y, z)`|`'d`|
> |`f (x, _, z)`|`λb.{f (x, b, z)}`|`'b → 'd`|
> |`f (_, y, _)`|`λ(a, c).{f (a, y, c)}`|`('a, 'c) → 'd`|
> |`f (_, _, _)`|`λ(a, b, c).{f (a, b, c)}`|`('a, 'b, 'c) → 'd`|

A function that has an input type inconsistent with a non-trivial product type cannot be partially applied. `(fun x : Int -> x)(_)` or `(fun x : () -> x)(_)` should result in some kind of error that prompts the user to work on the function rather than meaningless partial application.

If the input type of the function is a type hole, we cannot statically analyze whether the function can be partially applied. This means that the expression `(fun x -> x + 1)(_, 1)` is statically valid, and would be elaborated as `λy.{(λx.{x + 1}) (y, 1)}`. 

Providing a deferral singleton as an argument to such functions, like `(fun x -> x + 1)(_)`, is something programmers can do. There are different ways to handle such expressions statically:
* We can enforce that the argument for partial application must be a non-empty tuple, and then have this kind of expression result in an error to alert programmers about "meaningless partial application."
* Alternatively, we can refrain from imposing restrictions on the argument for partial application, allowing such expressions to be valid from a static perspective and be elaborated accordingly.
* ~~A more extreme approach would be to further restrict the input type of the function, disallowing type holes as valid input types.~~

### Implementation
* Add two new `UExp.term`s: `Deferral` and `DeferredAp(UExp.t, UExp.t)`.
* `_` introduces a `Deferral` expression. When a `Deferral` is either **a projection of the argument of a partial applicaton**, or **the argument of a partial application itself**, it's interpreted as bounded. Any other case would result in a `FreeDeferral` error. Adding extra parentheses around `_` would also result in a `FreeDeferral` error.
* Add a new `Info.error_exp` variant called `InconsistentWithDeferrableArrow` for functions that cannot be partially applied. ~~Assigning this error as Info to a function expression is challenging. Similar to the dilemma encountered in #957, although the current implementation of `Info.exp` includes `status_exp` to allow bypassing the function `status_exp` and directly assigning `status_exp` to an expression in `Statics.uexp_to_info_map`, the function `status_exp` is also used in other places like `elaborator.re`, resulting in inconsistent `status_exp` for the same expression in different locations. It would be best to restrict the use of the function `status_exp` to `Statics.re` only, but if that is not possible, to maintain data consistency, it would still require implementing the assignment of the new error in `status_exp`. Then,~~ a new variant of `Mode.t` called `SynDeferrableFun` would be needed, which, in addition to requiring the expression type to be consistent with the arrow type in `SynFun` mode, would also require its input type to be consistent with a non-trivial product type.
* If a function cannot be partially applied, it will match the arrow type `? -> ?`. If a function can be partially applied, the argument expression (arguments applied and deferrals) may not match the input type of the function. The verification process involves analyzing each element of the argument expression against the corresponding function input type, in which `Deferral`s are skipped.
* If there is a type inconsistency between the input and the argument types, a new variant of `Info.error_exp` called `InconsistentPartialApArg` will be assigned to the argument. In this error, `Deferral`s do not synthesize to or analyze against any type and appear only as placeholders (`_`) in the error message. ~~Similar to the challenges encountered with assigning `InconsistentWithDeferrableArrow`, if we want to implement error assignment within the function `status_exp`, the information about the *term* must be part of the input to `status_exp`. A better solution is to implement this within `Statics.uexp_to_info_map`, which currently leads to data inconsistency.~~
* The entire partial application is synthesized as an arrow type. If there is a type inconsistency between the input and the argument types, the input type of the arrow type is `Unknown`. In any case, the output type of the arrow type remains the same as the output type of the original function expression, providing as much as possible information to the programmer.
* In `elaborator.re`, when a partial application can be performed normally, the deferral expressions act as placeholders and are not elaborated. When there is a static error, partial application cannot be performed. To provide more reference information to programmers, a **full application** will be performed instead. This means that deferral expressions will be further elaborated into `EmptyHole`.
```
let plus = (fun (x, y) -> x + y) in
```
|External Expression|Internal Expression|Synthesized Type|
|:--|:--|:--|
|`_`|`InvalidText("_")`|`?`|
|`1 (_)`|`1 (?)`|`?`|
|`plus (_)`|`plus (?)`|`?`|
|`plus (_, 1)`|`(fun ~deferred0 -> plus(~deferred0, 1))`|`? -> Int`|
|`plus ((_), 1)`|`plus ((?), 1)`|`?`|
|`plus (_, _)`|`(fun (~deferred0, ~deferred1) -> plus(~deferred0, ~deferred1))`|`(?, ?) -> Int`|
|`plus (_, _, 1)`|`plus (?, ?, 1)`|`?`|


pigumar1:
Based on yesterday's discussion, I have made some revisions to the parsing of `DeferredAp` so that parentheses, commas, and deferral expressions are all part of the `DeferredAp` syntax:
image
image
image
image

The `InconsistentPartialApArg` will result in the following behavior. In `Statics.uexp_to_info_map`, an inconsistent partial application is assigned the self type `IsInconsistentPartialApArg`, which is interpreted as `InconsistentPartialApArg` in any mode to avoid deferrals' type checking. 
The current implementation of partial application still retains its tuple argument. Because this tuple argument shares the same syntax as the partial application itself, and the partial application takes precedence in parsing, it is not possible to assign this error to the argument separately.
image

The issue I'm currently facing is that all syntax elements of the partial application point to the same `Info`. Consequently, if there is a static error, the error message and type information cannot be displayed simultaneously.

* Would it be beneficial to split the partial application into two parts when encountering `InconsistentPartialApArg`? The first part would include **deferrals, comma tokens, and tuple elements**, while the second part would consist of **the parentheses token and the function expression**. This approach would resemble the handling of full application, where selecting the **parentheses** provides the **type** of the application expression, and selecting the **commas** provides **error information**. This splitting would occur only when `InconsistentPartialApArg` occurs, so when there are no static errors, the "type" of the argument remains hidden, ensuring that deferrals do not need to synthesize or analyze against specific types.
* Another approach is to fully **destructure the tuple argument during parsing**. This approach may be easier to implement and syntactically more intuitive. However, it would result in fewer and less accurate information being provided to programmers, and it would not allow the error message and type information to be displayed simultaneously.

pigumar1:
Also, this design issue needs to be resolved:
> Providing a deferral singleton as an argument to such functions, like (fun x -> x + 1)(_), is something programmers can do. 
>
> There are different ways to handle such expressions statically:
> 
> * We can enforce that the argument for partial application must be a non-empty tuple, and then have this kind of expression result in an error to alert programmers about "meaningless partial application."
> * Alternatively, we can refrain from imposing restrictions on the argument for partial application, allowing such expressions to be valid from a static perspective and be elaborated accordingly.

pigumar1:
~~Furthermore, now that deferrals have become part of the partial application syntax, selecting a deferral will provide information about the partial application rather than the deferral itself. In that case, would it be better to treat unbounded deferrals as `BadToken`s as well? This way, there would be no independent deferral syntax form anymore.~~

Another question is whether we should enforce the presence of at least one non-deferral argument in a partial application, otherwise generating an error. This may facilitate handling certain edge cases.

cyrus-:
I think it's fine if the error message is all you can see when there is a static error and your cursor is on the parens or commas of a deferral.

I think making it an error to have only one deferral with no other arguments makes sense, that is never what you want.

(If you want to chat a bit more after lab meeting today let me know.)

pigumar1:
When there is no static error, selecting `_` in a partial application displays the type of the entire partial application instead of the type matched by the deferral, which is misleading. As a result, `_` is no longer part of the partial application syntax but exists as an independent deferral expression.
image

Furthermore, we have added a new restriction to partial application: the argument must include at least one non-deferral actual value; otherwise, the error `MeaninglessPartialAp` will be assigned to the entire partial application.

Finally, if possible, it would be preferable to assign a separate `ok_exp` status to ok deferral expressions. The current two `ok_exp` statuses - `SynConsistent` and `AnaConsistent` - both involve inappropriate type checking of deferral expressions, which are sometimes displayed in the cursor inspector. The **bold information** below shows examples of these inappropriate type checkings. A separate `ok_exp` status would allow us to bypass these issues in `status_exp`.

##
image

#### Case 1
image
* Error message: "~~Deferral does not appear in a partial application as a placeholder~~Deferral must appear as a function argument"

#### Case 2 (Meaningless partial application evaluates to its function expression)
image
* Select the parentheses: "~~Meaningless partial application: expected at least one non-deferral expression~~Expected at least one non-deferred argument"
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?"

#### Case 3
image
image
* Select the parentheses or the comma: "[EXP] has type ? -> ?"
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?"

#### Case 4 | 5 | 6
image
image
* Select the parentheses of the partial application: "~~Meaningless partial application: expected at least one non-deferral expression~~Expected at least one non-deferred argument"
* Select the deferral: "[EXP] satisfies expected type Int"

image
* Select the deferral: "[EXP] satisfies expected type ()"

image
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?"

#### Case 7
image
image
* Select the parentheses (or the comma) of the partial application: "Arity mismatched partial application: expected 1 ~~expression~~argument, ~~found~~got 2"
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?" (consistent with the handling of a full application when there is an "arity error")

#### Case 8
image
image
* Select the parentheses (or the comma) of the partial application: "[EXP] has type ? -> ?"
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?"

#### Case 9
image
* Select the parentheses (`InconsistentPartialAp` has higher priority): "Arity mismatched partial application: expected 2 ~~expressions~~arguments, ~~found~~got 1"
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?" ~~or "**[EXP] satisfies expected type (?, ?)**"~~ ~~(this piece of information might be misleading. I'm considering assigning an error to the deferral as well. To do so, we need to have a self type for the deferral, which will be specially treated in `status_exp`; if not, either the deferral would have the self type `Common(Just(Unknown(Internal)))`, which is consistent with the function's input type, or we need to bypass `status_exp` to assign the error.)~~

#### Case 10 | 11 | 12
image
image
* Select the parentheses or the comma: "[EXP] has type Int -> Int"
* Select the deferral: "[EXP] satisfies expected type Int"

image
* Select the deferral: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?"

image
image

#### Case 13 (This is not a partial application)
image
* Select the deferral: "~~Deferral does not appear in a partial application as a placeholder~~Deferral must appear as a function argument"

#### Case 14
image
* Select the parentheses or the comma: "Arity mismatched partial application: expected 2 ~~expressions~~arguments, ~~found~~got 3"
* Select the deferrals: "~~**[EXP] has expected & actual type ?**~~[EXP] satisfies expected type ?" (consistent with the handling of a full application when there is an "arity error")

#### Case 15
image
image

src/haz3lweb/LangDocMessages.re
```diff
@@ -164,6 +165,27 @@ let triv_exp: form = {
   };
 };
 
+let deferral_exp_group = "deferral_exp_group";
+let deferral_exp: form = {
+  let explanation = {
+    message: "Deferral expression. This marks a placeholder for some part of the argument expression in a partial application whose value supply is deferred to future applications.",
```
cyrus-: "Deferral expression. This marks an argument that has not yet been applied in a partial application."

src/haz3lweb/LangDocMessages.re
```diff
@@ -164,6 +165,27 @@ let triv_exp: form = {
...
+    feedback: Unselected,
+  };
+  let deferral_exp_ex = {
+    sub_id: "deferral_exp_ex",
+    term: mk_example("let plus1 =\n(fun (x, y) -> x + y)(_, 1)\nin plus1(2)"),
```
cyrus-: this example goes off-screen and is too convoluted. try 

```
let plus = fun (x, y) -> x + y in 
let incr = plus(_, 1) in 
incr(5)
```

src/haz3lweb/LangDocMessages.re
```diff
@@ -1541,6 +1564,12 @@ let conapp_exp_ex = {
   message: "The constructor Some is applied to 1, which evaluates to Some(1).",
   feedback: Unselected,
 };
+let deferred_funapp_exp_ex = {
+  sub_id: "deferred_funapp_exp_ex",
+  term: mk_example("(fun (x, y) -> x + y)(_, 1)"),
+  message: "The plus function is partially applied to (_, 1). Part of the tuple pattern y is bound to 1 in the function body and the whole partial application evaluates to the new function plus1.",
```
cyrus-: this message doesn't seem to be showing up correctly

src/haz3lweb/LangDocMessages.re
```diff
@@ -1581,6 +1610,44 @@ let conapp_exp: form = {
...
+let _exp_fun = exp("e_fun");
+let _exp_arg = exp("e_arg");
+let deferred_funapp_exp_coloring_ids =
+    (~x_id: Id.t, ~arg_id: Id.t): list((Id.t, Id.t)) => [
+  (Piece.id(_exp_fun), x_id),
+  (Piece.id(_exp_arg), arg_id),
+];
+let deferred_funapp_exp: form = {
+  let explanation = {
+    message: "Partial function application. Bound part(s) of the [*function*](%i) input pattern to supplied values in the [*argument*](%i).",
```
cyrus-: this message doesn't seem to be showing up correctly. just get the normal function application docs.

src/haz3lweb/view/CursorInspector.re
```diff
@@ -154,6 +154,29 @@ let typ_err_view = (ok: Info.error_typ) =>
 let exp_view: Info.status_exp => t =
   fun
   | InHole(FreeVariable) => div_err([text("Variable is not bound")])
+  | InHole(FreeDeferral) =>
+    div_err([
+      text(
+        "Deferral does not appear in a partial application as a placeholder",
```
cyrus-: "Deferral must appear as a function argument."

src/haz3lweb/view/CursorInspector.re
```diff
@@ -154,6 +154,29 @@ let typ_err_view = (ok: Info.error_typ) =>
...
+      ),
+    ])
+  | InHole(ErroneousPartialAp(Meaningless)) =>
+    div_err([
+      text(
+        "Meaningless partial application: expected at least one non-deferral expression",
```
cyrus-: "Expected at least one non-deferred argument."

src/haz3lweb/view/CursorInspector.re
```diff
@@ -154,6 +154,29 @@ let typ_err_view = (ok: Info.error_typ) =>
...
+      ),
+    ])
+  | InHole(ErroneousPartialAp(ArityMismatch({expected, actual}))) =>
+    div_err([
+      text(
+        "Arity mismatched partial application: expected "
+        ++ string_of_int(expected)
+        ++ " expression"
```
cyrus-: "argument(s)"

src/haz3lweb/view/CursorInspector.re
```diff
@@ -154,6 +154,29 @@ let typ_err_view = (ok: Info.error_typ) =>
...
+        ++ (expected == 1 ? "" : "s")
+        ++ ", found "
```
cyrus-: got

cyrus- (changes requested):
- there is a merge conflict with `dev`
- the decoration when the cursor is on a comma in a partial application are missing the `f` (it works correctly when on the parentheses):
image

src/haz3lweb/LangDocMessages.re
pigumar1: ~~I managed to let the message show up in my local build, but it doesn't work correctly on the deployed build for unknown reasons... Here comes a screenshot of the message:~~
image

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -261,6 +262,48 @@ let rec dhexp_of_uexp =
...
+        | InHole(ErroneousPartialAp(ArityMismatch(_))) =>
+          Some(DHExp.InvalidText(id, 0, "<inv partial ap>"))
+        | _ =>
+          let mk_tuple = (ctor, xs) =>
+            List.length(xs) == 1 ? List.hd(xs) : ctor(xs);
+          let args = Term.UExp.matched_args(arg);
+          let* ty_fn = fixed_exp_typ(m, fn);
+          let (ty_arg, ty_ret) = Typ.matched_arrow(ty_fn);
+          let ty_ins = Typ.matched_args(List.length(args), ty_arg);
+          /* Substitute all deferrals for new variables */
```
pigumar1, line 275: > there's no actual data dependency between elements here (i think) so instead of folding, just map, and then use OptUtil.sequence (see other examples in this file)

I gave it a try. The issue I encountered is that lines 206-207 require both pats and d_args to share the same name, and this name is determined by the length of the growing pats, so perhaps using a fold would be more suitable.

src/haz3lweb/view/Deco.re
```diff
@@ -115,7 +115,8 @@ module Deco =
         };
       let range: option((Measured.Point.t, Measured.Point.t)) = {
         // if (Piece.has_ends(p)) {
-        switch (TermRanges.find_opt(Piece.id(p), M.term_ranges)) {
+        let id = Id.Map.find(Piece.id(p), M.terms) |> Term.rep_id;
```
pigumar1, line 122: > the decoration when the cursor is on a comma in a partial application are missing the f (it works correctly when on the parentheses):
> image

The issue I encountered while fixing this bug is that the `TermRanges` seems to be related to `Skel`. Even if I can break down and reconstruct `Term` in `MakeTerm`, I'm not sure how to do the same for the corresponding `Skel`. It appears that the creation of `Skel` doesn't take into account the information from `Term`.

Here is the solution that Haoxiang came up with. I'm not sure if it will have any impact on existing programs, but it appears to be the least disruptive solution to the codebase.

cyrus- (changes requested):
crashes when I try to turn on the langdocs color toggle at this state 
image

pigumar1:
> crashes when I try to turn on the langdocs color toggle at this state image

@cyrus- how should we highlight the arguments? We could highlight all of them as a whole, highlight each of them separately, or we don't highlight any of them.

cyrus-:
> > crashes when I try to turn on the langdocs color toggle at this state image
> 
> @cyrus- how should we highlight the arguments? We could highlight all of them as a whole, highlight each of them separately, or we don't highlight any of them.

Ideally it would highlight the partially applied arguments in a different color from the other arguments, which can all have the same color. So the message can say "<regular arguments> are partially applied and <deferred arguments> are deferred".

src/haz3lweb/view/LangDoc.re
```diff
@@ -1902,6 +1915,50 @@ let get_doc =
...
+              supplied_id,
+              deferred_id,
+            ),
+            LangDocMessages.deferred_funapp_exp_coloring_ids(
+              ~x_id,
+              ~supplied_id,
+              ~deferred_id,
+            ),
+          )
+        | Colorings =>
```
pigumar1: image
Color highlighting is done though the implementation in `LangDoc.re` has become really hacky... It seems that each term should have a different color and this implementation circumvents `ColorSteps.get_color` to have all deferrals highlighted with the same color.

src/haz3lweb/view/LangDoc.re
```diff
@@ -1902,6 +1915,50 @@ let get_doc =
             LangDocMessages.funapp_exp_coloring_ids,
           );
         };
+      | DeferredAp(x, args) =>
+        let x_id = List.nth(x.ids, 0);
+        let supplied_id = (-1);
+        let deferred_id = (-2);
```
pigumar1: Hacky: the coloring ids are decoupled with terms, since there might be a case in which no arguments are supplied.

cyrus- (changes requested):
- The lang docs message should "The deferred arguments can be applied in future applications." rather than "unsupplied arguments..."

- The example in the lang doc message should use a let binding rather than applying an anonymous function for clarity, and to avoid horizontal scrolling.

pigumar1:
image
How about this?

cyrus-:
looks good! 

@pigumar1 can you resolve the conflicts with the adts branch, @disconcision is going to take a look at the code after that

src/haz3lcore/statics/Statics.re
```diff
@@ -160,6 +160,8 @@ and uexp_to_info_map =
   | Invalid(token) => atomic(BadToken(token))
   | EmptyHole => atomic(Just(Unknown(Internal)))
   | Triv => atomic(Just(Prod([])))
+  | Deferral(status_deferral) =>
+    add'(~self=IsDeferral(status_deferral), ~co_ctx=CoCtx.empty, m)
```
disconcision: use atomic here instead of add'.
just call 'status_deferral' 'status', we know the deferral part from context.

src/haz3lcore/statics/Statics.re
```diff
@@ -236,6 +238,19 @@ and uexp_to_info_map =
       ~co_ctx=CoCtx.union([fn.co_ctx, arg.co_ctx]),
       m,
     );
+  | DeferredAp(fn, args) =>
+    let fn_mode = Mode.of_ap(ctx, mode, UExp.ctr_name(fn));
+    let (fn, m) = go(~mode=fn_mode, fn, m);
+    let (ty_in, ty_out) = Typ.matched_arrow(fn.ty);
+    let (self: Self.exp, m, arg_co_ctx) = {
+      let length = List.length(args);
```
disconcision: let's call this num_args

src/haz3lcore/statics/Statics.re
```diff
@@ -236,6 +238,19 @@ and uexp_to_info_map =
       ~co_ctx=CoCtx.union([fn.co_ctx, arg.co_ctx]),
       m,
     );
+  | DeferredAp(fn, args) =>
+    let fn_mode = Mode.of_ap(ctx, mode, UExp.ctr_name(fn));
+    let (fn, m) = go(~mode=fn_mode, fn, m);
+    let (ty_in, ty_out) = Typ.matched_arrow(fn.ty);
+    let (self: Self.exp, m, arg_co_ctx) = {
```
disconcision: I don't think this needs to be a nested block

src/haz3lcore/statics/Statics.re
```diff
@@ -236,6 +238,19 @@ and uexp_to_info_map =
...
+      let length = List.length(args);
+      let ty_ins = Typ.matched_args(length, ty_in);
+      let self = Self.of_deferred_ap(args, ty_ins, ty_out);
+      let modes = Mode.of_deferred_ap_args(ty_ins, length);
```
disconcision: Typ.matched_args and Mode.deferred_ap_args have their parameters in opposite order; let's go length-first for both, same as matched_prod

src/haz3lcore/statics/TermBase.re
```diff
@@ -97,12 +97,18 @@ and UExp: {
     | Bool(op_bin_bool)
     | String(op_bin_string);
 
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type status_deferral =
+    | Used
+    | Unused;
```
disconcision: Not obvious to me what these names connote... maybe something more like InAp and OutsideAp? and something like "deferral_position" instead of "status_deferral"?

src/haz3lcore/statics/Mode.re
```diff
@@ -131,3 +131,10 @@ let of_ap = (ctx, mode, ctr: option(Constructor.t)): t =>
     }
   | None => SynFun
   };
+
+let of_deferred_ap_args = (ty_ins: list(Typ.t), length: int): list(t) =>
+  (
+    List.length(ty_ins) == length
+      ? ty_ins : List.init(length, _ => (Unknown(Internal): Typ.t))
+  )
+  |> List.map(ty => Ana(ty));
```
disconcision, line 129: no need for a separate map here, just add the Ana to the Unknown(Internal(

src/haz3lcore/statics/Self.re
```diff
@@ -35,7 +35,15 @@ type t =
 [@deriving (show({with_path: false}), sexp, yojson)]
 type exp =
   | Free(Var.t)
-  | Common(t);
+  | IsDeferral(Term.UExp.status_deferral)
+  | IsErroneousPartialAp(error_partial_ap)
+  | Common(t)
+and error_partial_ap =
```
disconcision: and is mostly used for mutually recursive types... just define this type separately above exp

src/haz3lcore/statics/Self.re
```diff
@@ -35,7 +35,15 @@ type t =
...
+  | Meaningless
```
disconcision: 'Meaningless' needs a better name... maybe NoDefferedArgs?

src/haz3lcore/statics/Self.re
```diff
@@ -35,7 +35,15 @@ type t =
 [@deriving (show({with_path: false}), sexp, yojson)]
 type exp =
   | Free(Var.t)
-  | Common(t);
+  | IsDeferral(Term.UExp.status_deferral)
+  | IsErroneousPartialAp(error_partial_ap)
```
disconcision: let's just call it BadPartialAp; for better or worse we already use Bad this way in the rest of the codebase

src/haz3lcore/statics/Self.re
```diff
@@ -85,6 +95,23 @@ let of_ctr = (ctx: Ctx.t, name: Constructor.t): t =>
       },
   });
 
+let of_deferred_ap = (args, ty_ins: list(Typ.t), ty_out: Typ.t): exp => {
+  let expected = List.length(ty_ins);
+  let actual = List.length(args);
+  if (expected != actual) {
+    IsErroneousPartialAp(ArityMismatch({expected, actual}));
+  } else if (!List.exists(arg => !Term.UExp.is_deferral(arg), args)) {
```
disconcision: eliminate the double negative... something like `List.for_all(is_deferral, args)`

src/haz3lcore/statics/Self.re
```diff
@@ -85,6 +95,23 @@ let of_ctr = (ctx: Ctx.t, name: Constructor.t): t =>
...
+    IsErroneousPartialAp(Meaningless);
+  } else {
+    let ty_ins =
+      List.combine(args, ty_ins)
+      |> List.filter(((arg, _ty)) => Term.UExp.is_deferral(arg))
+      |> List.map(snd);
+    let ty_in = List.length(ty_ins) == 1 ? List.hd(ty_ins) : Prod(ty_ins);
```
disconcision, line 125: This is basically the same as mk_tuple... might want to consider making this a Typ-level helper... not sure though your call.

src/haz3lcore/dynamics/Elaborator.re
disconcision: Makes sense

src/haz3lcore/statics/Info.re
```diff
@@ -95,7 +97,9 @@ type ok_common =
   | Ana(ok_ana);
 
 [@deriving (show({with_path: false}), sexp, yojson)]
-type ok_exp = ok_common;
+type ok_exp =
+  | AnaDeferralConsistent(Typ.t)
```
disconcision, line 103: Can you remind me why we need this additional ok status? I don't think we need if for status_exp because we already know by construction in maketerm that every deferral inside a partial ap will be a Deferral(Used))

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -241,6 +242,47 @@ let rec dhexp_of_uexp =
...
+                   };
+                 },
+                 Some(([], [], [])),
+               );
+          let (pat, c_arg, ty_arg) = (
+            mk_tuple(x => DHPat.Tuple(x), pats),
+            mk_tuple(x => DHExp.Tuple(x), c_args),
+            mk_tuple(x => Typ.Prod(x), ty_args),
+          );
+          DHExp.Fun(pat, Arrow(ty_arg, ty_ret), Ap(c_fn, c_arg), None);
```
disconcision: This is going to require some manual casting... worth talking about with Cyrus, it's kind of complicated. Here's an example that will trigger a casting exception when run (the "?" are holes):

`let f: ?  -> ?  = fun a,b -> a+ b in
let g = f(1,_) in
g(1)`

src/haz3lweb/LangDocMessages.re
```diff
@@ -1586,6 +1621,47 @@ let conapp_exp: form = {
...
+  (Piece.id(_exp_deferral), deferred_id),
+];
+let deferred_funapp_exp: form = {
+  let explanation = {
+    message: "Partial function application. Apply the [*function*](%i) to the [*supplied arguments*](%i). The [*deferred arguments*](%i) can be applied in future applications.",
+    feedback: Unselected,
+  };
+  let comma = comma_exp();
+  let or_exp = () =>
+    Example.mk_tile(Form.mk_infix("|", Rul, Precedence.or_), []);
```
disconcision: this is rendering as red (sort inconsistency error) as this is not a valid hazel operator on its own... This needs to be a concrete example, not a schema... probably just drop the "e1 |" part and just go with the "_"

src/haz3lweb/view/Deco.re
disconcision: Interesting.... TermRanges are currently not working quite right in dev for both list literals and case rules, so it's not the end of the world if it doesn't work here. I don't really know the details of TermRanges so @dm0n3y will need to weigh in on whether this change is likely to be disruptive or not

disconcision (changes requested):
Looking pretty solid. The only functional problem I could find is related to casting. See my comment in Elaborator for a problematic example. Casts are usually applied automatically but the way the rest of the code is set up right now, they need to be manually constructed in cases like this where the DHExp is manually constructed. I don't immediately see what the correct casting approach is; I'm going to defer to Cyrus on this one.

I'm also not totally clear on the functional role being played by AnaDeferralConsistent; see my comment there.

cyrus-:
@pigumar1 have you resolved all of @disconcision's comments above? if so, please mark them as resolved. marking as draft for now.

pigumar1:
> @pigumar1 have you resolved all of @disconcision's comments above? if so, please mark them as resolved. marking as draft for now.

Because the dev branch has been updated, new bugs have been exposed. I am currently addressing them...

src/haz3lweb/view/Deco.re
dm0n3y: looks good to me. in fact, probably should have been written this way originally (using the rep_id).

cyrus-:
@pigumar1 click Ready for review when this is ready again

cyrus-:
@pigumar1 can you merge with the latest changes, should be small -- mostly be the langdocs rework. will prioritize getting it merged into `dev` after that.

pigumar1:
I added cases for `Deferral` and `DeferredAp `to `SyntaxTest.re`, but have no idea whether they are correct or not...

cyrus- (approved):

cyrus- (approved):

src/haz3lcore/statics/Statics.re
```diff
@@ -243,11 +243,11 @@ and uexp_to_info_map =
     let (fn, m) = go(~mode=fn_mode, fn, m);
     let (ty_in, ty_out) = Typ.matched_arrow(ctx, fn.ty);
     let (arg, m) = go(~mode=Ana(ty_in), arg, m);
-    add(
-      ~self=Just(ty_out),
-      ~co_ctx=CoCtx.union([fn.co_ctx, arg.co_ctx]),
-      m,
-    );
+    let self: Self.t =
```
disconcision: ideally break this out into Self.of_ap

disconcision (approved):
lgtm

disconcision:
actually I feel this error could be tightened up, or at least capitalization could be more consistent:
"Arity mismatched partial application: expected 2 arguments, got 1"
'partial application' is redundant; the label right before it says that.
how about "Arity mismatch: Expected 2 arguments, got 1"

PR #1060, opened 2023-07-04
Transpiler from Hazel to Roc

yujiagao-audrey:
Merged the recent commit of dev branch and ready for review.


src/haz3lcore/statics/MakeRoc.re
```diff
@@ -0,0 +1,193 @@
+let rec generate_code = (t: TermRoc.UExp.t, i: int): string =>
+  switch (t) {
+  | Bool(true) => "Bool.true"
+  | Bool(false) => "Bool.false"
+  | Int(n) => string_of_int(n)
+  | Float(f) =>
+    let f_str = string_of_float(f);
+    let last_char = f_str.[String.length(f_str) - 1];
+    if (last_char == '.') {
+      f_str ++ "0";
```
cyrus-: looks like latest version of Roc doesn't have this issue anymore (at least on https://www.roc-lang.org/repl)

src/haz3lcore/statics/MakeRoc.re
```diff
@@ -0,0 +1,193 @@
...
+    } else {
+      f_str;
+    };
+  | String(s) => s
```
cyrus-: we now need to add quotation marks here in latest `dev`

src/haz3lcore/statics/MakeRoc.re
```diff
@@ -0,0 +1,193 @@
+let rec generate_code = (t: TermRoc.UExp.t, i: int): string =>
```
cyrus-: document what the purpose of `i` is here

src/haz3lcore/statics/Translation.re
```diff
@@ -0,0 +1,402 @@
+let rec get_roc_exp = (t: TermBase.UExp.t, m: Statics.map): TermRoc.UExp.t =>
+  switch (t.term) {
+  | Invalid(_) => String("Invalid Not Implemented")
```
cyrus-: instead of emitting a String here, define an `UnsupportedInput` exception and raise it in these cases

src/haz3lcore/statics/Translation.re
```diff
@@ -0,0 +1,402 @@
...
+  | String(s) => String(s)
+  | ListLit(l) => ListLit(get_roc_exp_list(l, m))
+  | Fun(p, t) => Fun(get_roc_pat(p, m), get_roc_exp(t, m))
+  | Tuple(l) => Record(get_roc_exp_list(l, m))
+  | Var(token) => Var(get_name(m, token))
+  | Let(pat, def, body) =>
+    let indent_flag =
+      switch (def.term) {
+      | Let(_) => true
+      | _ => false
```
cyrus-: instead of using a wildcard pattern here, it is good practice to list all the possibilities so that if something new is added to the language that requires indentation, the exhaustiveness checker forces consideration of what should happen here

cyrus-: also, does a match not require indentation? or is it only nested lets?

src/haz3lcore/statics/Translation.re
```diff
@@ -0,0 +1,402 @@
...
+      };
+    switch (get_typ_ann(pat, m)) {
+    | Some(p) =>
+      if (indent_flag) {
```
cyrus-: can you add a comment explaining why you need to check the indent flag here? maybe with an example showing the issue

src/haz3lcore/statics/Translation.re
```diff
@@ -0,0 +1,402 @@
...
+        ]);
+      } else {
+        Seq([
+          p,
+          Assign(get_pat_var(pat, m), get_roc_exp(def, m)),
+          get_roc_exp(body, m),
+        ]);
+      }
+    | None =>
+      if (indent_flag) {
```
cyrus-: same here

cyrus- (changes requested):
- move all the Roc-related things into a new top-level library alongside haz3lcore called `haz3l-roc-transpiler` rather than putting it in statics. It can call into `haz3lcore`.

(partial review)

cyrus-:
not going to maintain a Roc transpiler, but thanks for the effort on this @yujiagao-audrey

PR #1062, opened 2023-07-11
Update StringLit

WondAli:
Aim:
- remove quotation marks from string literals internally, but keep them in the hazel web display


cyrus-:
thanks @WondAli !

disconcision:
There is some potential for confusing here... After this change there are now inconsistent representations of stringlits at the zipper and term levels. I've already made this change at both levels on branch llmass so the zipper level could be cherry-picked from there. However, this changes the zipper serialization so existing serialized editors would need to be updated.

PR #1066, opened 2023-07-20
Replace integer counter IDs with random 128bit UUIDs

disconcision:
This change touches a lot of code but merges should be straightforward: any id_gens you see should now be unnecessary; any functions returning a pair of an id_gen and something (these pairs are sometimes aliased as type 'state') is now unnecessary and can just return the something without the id_gen.

(Linecount increase is just from reserializations of init data; this change strictly simplifies everything touching ids outside Id.re)


cyrus-:
~15% performance regression doesn't seem insignificant -- might be worth trying a JavaScript uuid library instead of the ocaml one?

disconcision:
@cyrus: not actually sure its id generation slowing it down as opposed to the change from ptmaps to ocaml maps for id maps... need to run more precise tests

cyrus-:
@disconcision some conflicts with `adt-defs-after` to resolve here

disconcision:
I did some simple benchmarking. Specifically I have a branch off dev, `simple-benchmark`, which adds code to do the following: When Shift-F7 is pressed, schedule a bunch of actions, then measure the elapsed time until the last of the actions is executed. This is intended to be reasonably non-synthetic; we are testing the time for the full incr_dom cycle including UI (afaict...), not merely the computation of the new model. As such, it is intended to guage gross performance effects of a code change.

Specifically, the test I ran is the execution of a few thousand insert actions correponding to character-by-character linear entry of a medium-length hazel program (see Benchmark.re in the abovementioned branch for the program string). I ran this on the blanked-out Examples/Introduction editor, with default settings, on both `dev` and `randIds`, compiled with the standard `make`, each opened in adjacent tabs. I ran the benchmark 3 times for each, alternating between each tab. I did this parallel alternation as I noticed there could be considerable variance (30%+) in times depending on what else was going on on my system.

The results were as follows:
branch `dev`:
Benchmark: Finished: time: 16.255000 seconds
Benchmark: Finished: time: 17.585000 seconds
Benchmark: Finished: time: 17.506000 seconds

branch `randIds`:
Benchmark: Finished: time: 9.926000 seconds
Benchmark: Finished: time: 9.113000 seconds
Benchmark: Finished: time: 9.690000 seconds

That is, randIds represents a 44% speedup. This contradicts my earlier informal observational evidence, so it would be ideal if someone else would try to at least re-run these benchmarks. Honestly though I'm not terribly surprised. My expectation is that the id generation itself should have ~ no performance impact: generating a random id will be slower in absolute terms relative to incrementing a counter, but both will be entirely negligible compared to everything else going on. More likely IMO the difference is dominated by map generation, which has been switched from PtMap to (standard OCaml) Map. Given that the reason for the switch was improved merge performance, which both (a) didn't pan out and (b) is no longer relevant, it doesn't surprise me that the latter outperforms the former. That said, my confidence isn't particularly high here.

disconcision:
Bug: given `let g:Int| = 1 in 1` entering `-` triggers Not_found exception
Similarly:  1111111111 >< 1
This is an interaction with a bug in dev.... badlits are getting parsed as empty holes if they are adjacent to multiholes and hence properly sorted as Any. Attempted fix in #1080

cyrus-:
@tonyfettes can you reproduce the performance numbers that @disconcision reported in the comment above?

cyrus-:
marking as draft until #1080 is resolved

cyrus-:
@disconcision some conflicts with `dev` need to be resolved. does this need anything else?

disconcision:
@cyrus: about to resolve, then should be good to go. though i am slightly concerned that there's an outstanding bug... do you remember exactly what you did during our meeting to trigger that Not_found exception? was it during the paste benchmark or something else? I originally thought it was due to the dark holes fix but on reflection I'm not sure anymore, though I've been unable to reproduce it

disconcision:
ok @cyrus- should be g2g. lmk when you merge, i'll make a comment in discuss as this will prob cause conflicts for everyone, though they should be very mechanically resolvable

cyrus-:
@disconcision I believe what I did during the meeting was just run the benchmark several times in a row.

disconcision:
@cyrus-  can't reproduce; gonna attribute it to either the dark hole thing or old app cache.

though i guess one thing to note about the benchmark is that it (by design) does not attempt to reset initial conditions (so that you can use it to test with statics/dynamics on/off, or at the beginning/middle/end of a file), which means that if you try to benchmark with (say) an unempty, unbalanced selection, it can expose extant latent syntax model issues

cyrus- (approved):

PR #1069, opened 2023-07-29
Loosened annotation requirements for recursive functions

### LINKED ISSUE #796: Loosen annotation requirements for recursive functions
Currently require an arrow type annotation, but could use unknown types to loosen this so that it works as long as there is a function literal available (especially practical once we have type hole inference).

pigumar1:
## Introduction
This pull request aims to loosen annotation requirements for recursive functions.

Currently, marking a function as recursive is achieved by adding an arrow type annotation to its corresponding variable. This makes all variables within the function body having the same name become bound, though shadowing within the body may change the expression they are bound to:
image

The new approach aims to make a function recursive as long as it's correctly annotated (specifically, normalized `Arrow`s, `Invalid`s, `MultiHole`s, `TypeHole`s, and free type variables). The new approach implements this by introducing a new function `Statics.is_recursive` that checks if the number of variables match the number of function literals in a let expression, and they are "correctly" annotated. It replaces `extend_let_def_ctx`.

The new approach also attempts to synthesize out more accurate types (at least arrow types) for unannotated recursive functions.

## Some special cases
The process of determining recursiveness doesn't consider whether the type of the definition (the function literal) matches the annotation. I believe this has already been addressed properly...?
image

The tuple pattern for mutual recursive function variables can be annotated as a whole:
image

There is also a case where the synthesize mode might be turned into the analyze mode: https://github.com/hazelgrove/hazel/pull/1069#discussion_r1320060919

## If this was merged, ...
* #965 would be affected because it modified `extend_let_def_ctx` (but in the good way, because they modified it to include the output of `is_recursive`.
* #1056 would be affected because it applied `extend_let_def_ctx`. But would let operators be related to recursions?


src/haz3lcore/statics/Statics.re
```diff
@@ -246,9 +247,22 @@ and uexp_to_info_map =
     );
   | Let(p, def, body) =>
     let (p_syn, _) = go_pat(~is_synswitch=true, ~mode=Syn, p, m);
-    let def_ctx = extend_let_def_ctx(ctx, p, p_syn.ctx, def);
-    let (def, m) = go'(~ctx=def_ctx, ~mode=Ana(p_syn.ty), def, m) /* Analyze pattern to incorporate def type into ctx */;
-    let (p_ana, m) = go_pat(~is_synswitch=false, ~mode=Ana(def.ty), p, m);
```
pigumar1, line 251: Generates a more-accurate info map for the pattern

src/haz3lcore/statics/Statics.re
```diff
@@ -246,9 +247,22 @@ and uexp_to_info_map =
...
+      if (Option.is_none(UExp.get_recursive_bindings(p, def))) {
+        let (def, m) = go(~mode=Ana(p_syn.ty), def, m) /* Analyze pattern to incorporate def type into ctx */;
+        let (p_ana, m) =
+          go_pat(~is_synswitch=false, ~mode=Ana(def.ty), p, m);
+        (def, p_ana, m);
+      } else {
+        let (def_base, _) = go'(~ctx=p_syn.ctx, ~mode=Ana(p_syn.ty), def, m) /* Analyze pattern to incorporate def type into ctx */;
+        let (p_ana, m) =
+          go_pat(~is_synswitch=false, ~mode=Ana(def_base.ty), p, m);
+        let (def, m) = go'(~ctx=p_ana.ctx, ~mode=Ana(p_syn.ty), def, m);
```
pigumar1: In the recursive case, as the part of the context associated with the pattern becomes more accurate, the context for the definition expression becomes outdated and requires updating.

cyrus-:
@pigumar1 I don't think we should be looking for uses of the self reference to determine if it is recursive, because in that case just before the variable is used it will not appear in the context (and thus not in the context inspector) and make it look like it isn't recursive. Instead we should more simply just always make a let-bound function or tuple of functions recursive / mutually recursive, with the unknown type used if there is not an explicit type annotation.

pigumar1:
@cyrus- I apologize for missing today's office hour!

* The current implementation of recursion seems to have a bug: in this state, it is not possible to input `x` or `y` into the hole marked with the cursor, and it results in a "(Exception "Stack overflow")" in the console. The reason is still under investigation (on the adt-defs-after branch):
image

* Currently, as long as the definition expression of a let expression is **a function literal / a tuple of function literals**, and the newly defined variables / tuple of variables are **not annotated with types inconsistent with arrow types**, then these functions are considered recursive / mutually recursive. Even if the number of variables is not equal to the number of function literals, referencing these variables in the bodies of these functions will not produce additional "variable not found" errors:
image
(`y` is bound in the body of the function)

   Do we need to enforce that the number of variables must be equal to the number of function literals?

* Would it be better if we synthesize an arrow type, rather than the unknown type, for the variable associated with an unannotated recursive function? While we might not achieve 100% accuracy in synthesizing the type due to self-reference being in-place, I believe we can still synthesize a consistent arrow type (better than `? -> ?`) for the recursive function.

pigumar1:
In fact, with the introduction of type aliases, relying solely on term information to determine whether a variable is annotated with an arrow type may no longer be sufficient. It might be worth considering incorporating additional static information:
image

cyrus-:
Ah good point about type aliases.

src/haz3lcore/statics/Statics.re
```diff
@@ -47,13 +47,26 @@ let map_m = (f, xs, m: Map.t) =>
...
-    (ctx: Ctx.t, pat: UPat.t, pat_ctx: Ctx.t, def: UExp.t): Ctx.t =>
-  if (UPat.is_tuple_of_arrows(pat) && UExp.is_tuple_of_functions(def)) {
-    pat_ctx;
-  } else {
-    ctx;
+let is_recursive = (ctx, p, def, syn: Typ.t) => {
+  switch (Term.UPat.get_num_of_vars(p), Term.UExp.get_num_of_functions(def)) {
+  | (Some(num_vars), Some(num_fns))
+      when num_vars != 0 && num_vars == num_fns =>
+    switch (Typ.normalize(ctx, syn)) {
```
pigumar1, line 71: The process of determining recursiveness doesn't consider whether the type of the definition (the function literal) matches the annotation. I believe this has already been addressed properly...?
image

src/haz3lcore/statics/Statics.re
```diff
@@ -47,13 +47,26 @@ let map_m = (f, xs, m: Map.t) =>
...
+    | Unknown(_)
+    | Arrow(_) => num_vars == 1
```
pigumar1, line 73: Annotations for recursive functions (normalized) can be: `Arrow`s, `Invalid`s, `MultiHole`s, `TypeHole`s, and free type variables.

src/haz3lcore/statics/Statics.re
```diff
@@ -47,13 +47,26 @@ let map_m = (f, xs, m: Map.t) =>
...
+    | Prod(syns) when List.length(syns) == num_vars =>
```
pigumar1, line 74: Mutual recursive cases: the tuple pattern always has a `Prod` type:
(Synthesizes `Prod(Unknown(SynSwitch), Unknown(SynSwitch))`)
image

(Synthesizes `Prod(? -> ?, Unknown(SynSwitch))`)
image

image

src/haz3lcore/statics/Statics.re
```diff
@@ -260,10 +273,34 @@ and uexp_to_info_map =
...
-    /* Analyze pattern to incorporate def type into ctx */
-    let (p_ana, m) = go_pat(~is_synswitch=false, ~mode=Ana(def.ty), p, m);
+    let (def, p_ana, m) =
+      if (!is_recursive(ctx, p, def, p_syn.ty)) {
+        let (def, m) = go(~mode=Ana(p_syn.ty), def, m);
+        let (p_ana, m) =
+          go_pat(~is_synswitch=false, ~mode=Ana(def.ty), p, m);
+        (def, p_ana, m);
+      } else {
+        let (def_base, _m) = go'(~ctx=p_syn.ctx, ~mode=Ana(p_syn.ty), def, m) /* Analyze pattern to incorporate def type into ctx */;
```
pigumar1: Attempts to synthesize out more accurate types for recursive functions:
e.g.: for unannotated functions `odd` and `even`, 
image
* `p_syn.ctx` has `odd: ?, even: ?`
* `def_base.ty` = `odd: ? -> Bool, even: ? -> ?`

src/haz3lcore/statics/Statics.re
```diff
@@ -260,10 +273,34 @@ and uexp_to_info_map =
...
+        let (p_ana, m) =
+          go_pat(~is_synswitch=false, ~mode=Ana(def_base.ty), p, m);
```
pigumar1: `def_base.ty` has "finalized" consistent type(s) for recursive function(s).
The info for the pat can be finalized as well.
`m` is extended.

src/haz3lcore/statics/Statics.re
```diff
@@ -260,10 +273,34 @@ and uexp_to_info_map =
...
+        let def_ctx = p_ana.ctx;
+        let (def_base2, _m) =
```
pigumar1: The following pieces of code aims to keep the synthesized types consistent.
For example, with the new context for the definition expression, `odd` has type `? -> Bool`, allowing `fun x -> odd(x - 1)` to synthesize out a more accurate type (`? -> Bool`), which is consistent but not equal to the "finalized" type for `even`, `? -> ?`. 
**Even if we only use ? as their types in the context for the definition expression, they would still synthesize out two arrow types, which is more accurate than two ? s.**
So if we can synthesize out a more accurate (than the "finalized" context) type for some variable(s), instead of doing so, we use the synthesized type to analyze against the expected type in the "finalized" context:
image

The type for a recursive function can be infinitely more accurate:
image

I'm considering factoring out this piece of code but have no idea where to place it...

pigumar1: `def_base.ty` and `def_base2.ty` might differ in positions whose corresponding variables are not annotated

src/haz3lcore/statics/Statics.re
```diff
@@ -260,10 +273,34 @@ and uexp_to_info_map =
...
+          go'(~ctx=def_ctx, ~mode=Ana(p_syn.ty), def, m);
+        let ana_ty_fn = ((ty_fn1, ty_fn2), ty_p) => {
```
pigumar1, line 370: When the variable is annotated: `ty_p != Typ.Unknown(SynSwitch)`, use `ty_p`
When the variable is not annotated: `ty_p == Typ.Unknown(SynSwitch)`
* The synthesized type does not become more accurate: use synthesize mode
* The synthesized type does become more accurate: analyze against the finalized type `ty_fn1`

image

src/haz3lcore/statics/Statics.re
cyrus-: I think we want this to work for any type that has a matched arrow type, so that should include Invalid/MultiHole/free type variables. Perhaps you can use matched arrow type directly after normalizing?

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -207,36 +207,47 @@ let rec dhexp_of_uexp =
...
+                       Substitution.subst_var(
+                         DHExp.Prj(self_var, i),
+                         f,
+                         ddef,
+                       );
+                     (i + 1, ddef);
+                   },
+                   (0, ddef),
+                 );
+            Let(dp, FixF(self_id, ty_body, substituted_def), dbody);
```
cyrus-: the type annotation on the fixpoint is incorrect (Bool here rather than the product type)
image

cyrus- (changes requested):
Looks good -- just a couple of small changes needed and should be good to merge.

src/haz3lcore/dynamics/Elaborator.re
pigumar1: Fixed. I think this is one of the bugs that Andrew fixed in TyDi.

src/haz3lcore/statics/Statics.re
pigumar1: @cyrus- I think we have already been supporting this:
image

cyrus- (approved):

PR #1070, opened 2023-07-31
Unit test framework

ruiz-m:
Add unit tests to Hazel using the Alcotest framework (https://github.com/mirage/alcotest)

Added a separate module for the elaboration tests

The intention is for others to create separate testing modules to test certain modules and create a list of test cases that is then placed in the run function in haz3ltest.re


cyrus-:
@disconcision want to take a quick look at this

disconcision (commented):
I think the basic test framework feels sound. I think we'll want to develop some test style guidelines but those will emerge in time.

Some minor things: do we really need the "let () ="? I understand it's a check that the definition actually has unit type, but won't compiler warnings flag it if it's not? ideally I'd just like to write my tests as `test("description", input, output)`. Also are we going to use the tags? I noticed the defined helper takes them as an argument, but they are never used... again, not a big deal, but if this is the canonical example then we might have people having a bunch of empty list arguments floating around for no reason. I don't particularly anticipate using tags myself, though I could be wrong.


For these specific elaborator tests I feel like they're not doing that much; arguably better to cover some cases like 'If' that aren't simply direct transliterations. Fine as simple examples though. Will have to be redone for UUID refactor but that's my problem :).

ruiz-m:
I understand. I will make those changes and make the tests less redundant.

disconcision (approved):
lgtm

cyrus- (changes requested):
I'm getting `Unimplemented Javascript primitive caml_js_expr!` when I try to run `make test` at the terminal. This may not have to do with this PR in particular -- I'd recommend removing the dependency on js_of_ocaml in hazelcore and seeing what breaks.

disconcision:
@cyrus only core Js dependencyI know is regexp in Form; not obvious why that might be getting executed. but regardless might be easiest just to run with Node?

ruiz-m:
@cyrus- @disconcision I am having an issue with the last test where the input uexp is equivalent to `let f : Int -> Int = fun x -> x+1 in 55`.
The old elaborator would give the type in the FixF dhexp Arrow(Int, Int). However, in the newer version of the elaborator that type becomes Int which causes the last test to fail now. I believe it should be Arrow(Int, Int)

cyrus-:
Yes, the fixpoint type should definitely be an arrow type. @disconcision can you take a look

disconcision:
afaik these types aren't currently touched so it should be fine just to change it from ty_body to the relevant arrow type

cyrus-:
@ruiz-m go ahead and see if you can fix the issue in the elaborator

src/regexp/regexp_ocaml/Regexp.re
```diff
@@ -0,0 +1,4 @@
+let regexp = (r, s) => {
+  print_endline("CALLING OCAML VERSION");
```
cyrus-: this isn't necessary in the final commit

cyrus-:
@ruiz-m iirc there were some issues due to the fact that JS and OCaml had slightly different regex syntax -- was that resolved here?

disconcision:
So we've determined via this semester's 490 that the regexp wrapper approach is problematic; regexp libraries have some problematic syntactic differences. Instead when necessary to run hazel from the shell we should do it via Node. I have this working here: https://github.com/hazelgrove/hazel/pull/1148/files#diff-14411b7839e8415a9552c07d18d554f54f738c55b694a3b502fa3d0e797d21db

Not sure if it's immediately relevant here, but for reference the process is:
1. Set up a new (haz3lweb) dune build target for the module you want to run from the shell
2. In the relevant module, do something like this:
`Js.Unsafe.js_expr("require('process')");
Js.Unsafe.js_expr("process.argv[2]") |> Js.to_string |> main;`
(that assumes that there is exactly one command line argument, and you've implemented a function main in the relevant module.
3. Run via `node <generated_js_file_path> <arg>` (the file will be generated in the same build folder as hazel.js. optionally add a new stanza in the haz3lweb/www dune to move/rename this file; see above link)

(For the purposes of this PR I think this just means we want to rip out the wrapper; @cyrus- can confirm/refute)

cyrus-:
Let's replace the dune virtual library stuff in this library with dune modifications to allow running the tests via node.js as described above by Andrew.

README.md
```diff
@@ -190,6 +190,11 @@ for a type named something else like `q`, it will be `show_q`.
 #### Debug Mode
 If Hazel is hanging on load or when you perform certain actions, you can load into Debug Mode by appending `#debug` to the URL and reloading. From there, you have some buttons that will change settings or reset local storage. Refresh without the `#debug` flag and hopefully you can resolve the situation from there.
 
+#### Testing
+You can run all of the unit tests located in `src/haz3ltest` by running `make test`.
+
+Unit tests are written using the [Tezt framework](https://gitlab.com/nomadic-labs/tezt/).
```
cyrus-: This needs to be updated to refer to alcotest.

README.md
```diff
@@ -190,6 +190,11 @@ for a type named something else like `q`, it will be `show_q`.
...
+
```
cyrus-, line 197: We should also mention `make reset-regression-tests` and `make fix-test-answers` (or remove them from the Makefile).

src/haz3ltest/haz3ltest.re
```diff
@@ -0,0 +1,10 @@
+open Alcotest;
+open Js_of_ocaml;
+open Test_Elaboration;
+
+let main = (_s: string) => {
+  run("Dynamics", [("Elaboration", elaboration_tests)]);
+};
+
+Js.Unsafe.js_expr("require('process')");
+Js.Unsafe.js_expr("process.argv[2]") |> Js.to_string |> main;
```
cyrus-: what is this code doing? the argument to main doesn't seem to be used so is this necessary?

Makefile
```diff
@@ -50,11 +51,8 @@ repl:
 	dune utop src/haz3lcore
 
 test:
-	dune build @src/fmt --auto-promote || true
-	dune exec src/hazeltest/hazeltest.exe -- --regression-dir src/hazeltest/regressions
-
-reset-regression-tests:
-	dune exec src/hazeltest/hazeltest.exe -- regression --regression-dir src/hazeltest/regressions --reset-regressions
+	dune build @src/fmt --auto-promote src --profile dev || true
+	node $(TEST_DIR)/haz3ltest.bc.js test
```
cyrus-, line 56: We should add instructions on installing `node` to the install instructions since it is now a dependency.

cyrus- (changes requested):
In addition to the small comments in the code, we should think about how we want to do continuous integration with the testing. Maybe a second GitHub Action that runs the tests and fails if any of the tests fail?

.github/workflows/deploy_branches.yml
```diff
@@ -24,12 +24,13 @@ jobs:
...
       - name: Install dependencies and build hazel # STEP 3
         run: |
-          sudo apt --assume-yes install curl m4 opam
+          sudo apt --assume-yes install curl m4 opam nodejs
           export OPAMYES=1
           opam init --compiler=ocaml-base-compiler.5.0.0
           eval $(opam env)
           make deps 
           make release
+          make test
```
7h3kk1d: Can we put this in another step so it's clear from the email if it fails?
```suggestion
      - name: Run tests
          make test
```

.github/workflows/deploy_branches.yml
```diff
@@ -24,12 +24,13 @@ jobs:
           key: ${{ runner.os }}-modules-${{ hashFiles('./source/opam.export') }}
       - name: Install dependencies and build hazel # STEP 3
         run: |
-          sudo apt --assume-yes install curl m4 opam
+          sudo apt --assume-yes install curl m4 opam nodejs
```
7h3kk1d: It might be worth using https://github.com/actions/setup-node in a previous step so that we're not accidentally upgrading the node version with Ubuntu. I could take it or leave it though.

.github/workflows/deploy_branches.yml
7h3kk1d: Outside the scope of this PR but placing here for a reminder that we should consider using
https://github.com/mirage/alcotest/issues/130 and https://github.com/marketplace/actions/test-reporter so we can get actual test report files.

src/haz3ltest/Test_Elaboration.re
```diff
@@ -0,0 +1,290 @@
+open Alcotest;
```
7h3kk1d: Looking at other ocaml projects it looks standard to put your tests in `test/. I'm not sure if it's worth doing that sooner rather than later if we want to avoid causing a ton of merge conflicts. @cyrus- do you have thoughts?

7h3kk1d:
> In addition to the small comments in the code, we should think about how we want to do continuous integration with the testing. Maybe a second GitHub Action that runs the tests and fails if any of the tests fail?

@cyrus- I'm assuming by CI you mean merging mainline into the feature branch and running tests again? Just clarifying because [the term](https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration) has started to just mean automated tests on push.

Assuming that's what you meant I think it makes sense to do but break into a separate PR.

cyrus-:
> > In addition to the small comments in the code, we should think about how we want to do continuous integration with the testing. Maybe a second GitHub Action that runs the tests and fails if any of the tests fail?
> 
> @cyrus- I'm assuming by CI you mean merging mainline into the feature branch and running tests again? Just clarifying because [the term](https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration) has started to just mean automated tests on push.
> 
> Assuming that's what you meant I think it makes sense to do but break into a separate PR.

Yes, I just meant running the test suite on push and reporting the results on GitHub. We can put that in a separate PR.

cyrus- (approved):

PR #1079, opened 2023-08-10
Perfmageddon

### LINKED ISSUE #883: editor performance
editor performance degrades dramatically as the program gets bigger, even at ~100 lines of code it is unusably slow even on modern machines.

disconcision:
Currently offers an up to an order-of-magnitude speedup on best-case edit action keystrokes

The biggest gains are due to eliminating extraneous Incr_dom cycles happening on ~ every keystroke:

- Due to the legacy doubletap detection, every KeyUp event was triggering a new cycle. Removing this gives a ~50%+ overall speedup
- Saving was happening every keystroke, in fact sometimes twice (on keyup as well). This was a double-whammy, as saving on edit actions was being done in Keyboard by scheduling an additional event. The ability of the key handlers to return a list of updates, each of which triggers a full incr_dom cycle, is a moral hazard and has been removed; each key event now returns an option(Action) instead. Saving also no longer happens after every edit action; rather it happens 1 second after the last edit action happened. Together this is a ~200-400% overall speedup.
- There was a check in the on_display hook which scrolled the caret into view which was gated on detecting a change in the model. This is expensive, not because of the model comparison, but simply because the old_model is in the Incr monad, and mapping into Incr triggers a cycle. Removing this check gives a ~50% speedup.

Other gains:

- 5x speedup to Paste by bypassing Editor Meta recalculation
- Huge speedup in very imbalanced states by memoizing Backpack.shard_info
- 10x speedup in Holes Deco by just scanning info_map instead of recursing
- 2x speedup for Termranges.mk; was called twice with no memoization. I now passing it through directly as it's clean to do so, but I also memoized it defensively
- 60% speedup for Code.view by recursively memoizing Text.of_tile
- 30% speedup for MakeTerm by memoizing Skel.mk

Remaining possible speedups:

- Logging: Gets slow with lots of entries. With localstore you can't just append, you need to set. Could switch to a system for which each entry gets its own a KV pair of something.
- Saving Editors to LocalStorage: I caused a bit of a regression when I moved to serializing the entire Editors. Probably not a priority with the new save delay system, but could use some TLC eventually.
- Backpack: If we are still experiencing too much slowdown in unbalanced states, shard_info could be greatly speeded up if we were to eliminate unrestricted pick-up functionality (i.e. the ability to have backpack selections containing multiple shards)
- Maketerm: The recursion structure in Maketerm makes recursive memoization difficult. We want to memoize something that's likely to have physical equality i.e. substructures which are likely literally preserved over edits, like segments or tiles. But the core Maketerm functions (go_s, unsorted) take both a segment and a skel as arguments; since the skel is derived, we're never going to have physical equality for substructures if the superstructure changes. I'm wondering if it's possible to internalize skel generation to these functions; instead of sending in a skel, send in a subsegment
- PtMaps/Maps/Hashtbls: Replace remaining ptmap use with plain maps and/or replace maps use with hashtbls everywhere. I'm unsure as to the overall impact but it could be significant; probably worth doing more bottom-up profiling. As a stopgap we're still unioning in some places which could be turned into traversals (e.g. TermRanges)
- Elaboration Views: Right now turning off dynamics can slow things down as it takes time to render large elaborations. This is partially addressed in `llmass` as it makes it possible to actually turn of statics, resulting in no elaboration being generated. There should probably be a separate toggle for elaboration though.
- Zipper pass fusion: In general we could probably consider some more structured architecture for passes over the zipper. Currently we have MakeTerm, Measured, TermRanges, and Code.view which all do zipper passes. Two or more of these could potentially be fused into a single 'SyntaxInfo' pass. SyntaxInfo could potentially collect some information which is either being reproduced ad-hoc in many places (expected sort) or currently lives in static info (term, cls, list of ancestor ids)

### DIFF
--- a/src/haz3lcore/Measured.re
+++ b/src/haz3lcore/Measured.re
@@ -223,27 +223,6 @@ let find_by_id = (id: Id.t, map: t): option(measurement) => {
   };
 };
 
-let union2 = (map: t, map': t) => {
-  tiles:
-    Id.Map.union((_, ms, ms') => Some(ms @ ms'), map.tiles, map'.tiles),
-  grout: Id.Map.union((_, m, _) => Some(m), map.grout, map'.grout),
-  secondary:
-    Id.Map.union((_, m, _) => Some(m), map.secondary, map'.secondary),
-  rows:
-    Rows.union(
-      (_, s: Rows.shape, s': Rows.shape) =>
-        Some({
-          indent: min(s.indent, s'.indent),
-          max_col: max(s.max_col, s'.max_col),
-        }),
-      map.rows,
-      map'.rows,
-    ),
-  linebreaks:
-    Id.Map.union((_, i, _) => Some(i), map.linebreaks, map'.linebreaks),
-};
-let union = List.fold_left(union2, empty);
-
 let post_tile_indent = (t: Tile.t) => {
   // hack for indent following fun/if tiles.
   // proper fix involves updating mold datatype
--- a/src/haz3lcore/TermRanges.re
+++ b/src/haz3lcore/TermRanges.re
@@ -54,3 +54,5 @@ let rec mk = (~map=empty, seg: Segment.t) => {
        union(map, snd(go(Segment.skel(seg)))),
      );
 };
+
+let mk = Core.Memo.general(~cache_size_bound=1000, mk);
--- a/src/haz3lcore/TermRanges.re
+++ b/src/haz3lcore/TermRanges.re
@@ -6,6 +6,14 @@ type nonrec t = t(range);
 
 let union = union((_, range, _) => Some(range));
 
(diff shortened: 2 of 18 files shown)


cyrus- (changes requested):
getting `Not_found` exceptions when editing some school mode editors, e.g. the Prelude. Maybe having to do with serialization issues?

disconcision:
@cyrus- fixed. the issue wasn't specific to exercises mode, it was something I had noticed and fixed a while ago, the fix just ended up on the wrong side of the permageddon/llmass fix. I realize the comment I've added to the fix isn't totally satisfying; it's a complex/finicky syntax issue that predates this branch; it is only surfaced here because previously we've never been iterating over the info_map directly, only recursing though the AST

cyrus- (approved):

PR #1080, opened 2023-08-10
Undarken holes

disconcision:
@dm0n3y could you take a look at this? It's an attempt to remove dark holes while retaining the behavior as I understand it (which is not very well). Basically in the case where I can't figure out a reasonable sort I just go with Exp instead of the dark hole thing; not sure if this might cause problems I'm not anticipating.

The reason I'm trying to make this change is that in my UUID branch I'm getting Not_found exceptions due to the fact I had to remove dark hole filtering. I seem to be getting these in a pretty specific case: Exactly when you have an Any-sorted operand inside a multihole. These are pretty liminal cases; the representative examples are:

- `11111111111><1` (first operand is too big to be an integer; is classified by bad_lit mold and gets assigned Any sort)
- `.><1` (we need to allow a dot on its own to be able to enter floats; is classified by bad_float mold with Any sort)

### DIFF
--- a/src/haz3lcore/lang/Form.re
+++ b/src/haz3lcore/lang/Form.re
@@ -69,7 +69,7 @@ let mk_nul_infix = (t: Token.t, prec) =>
 /* Token Recognition Predicates */
 let is_arbitary_int = regexp("^-?\\d+[0-9_]*$");
 let is_arbitary_float = x =>
-  x != "." && regexp("^-?[0-9]*\\.?[0-9]*((e|E)-?[0-9]*)?$", x);
+  x != "." && x != "-" && regexp("^-?[0-9]*\\.?[0-9]*((e|E)-?[0-9]*)?$", x);
 let is_int = str => is_arbitary_int(str) && int_of_string_opt(str) != None;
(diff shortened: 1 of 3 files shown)


dm0n3y:
Does this PR resolve those exceptions? Or are they remaining?

Assuming the former, I think this looks fine. My use of dark holes in cases like (https://github.com/hazelgrove/hazel/pull/1080/files#diff-db7a3f96ed680c747b5dbed63fad57a48276cbe28696aacf3d662f52f215d8ffL168) was not at all principled, just a hacky fallback thrown in during last summer's rush. I think I introduced the dark hole stuff originally for use in an early attempt at ghost-style term completion (see the `complete_root` function (https://github.com/hazelgrove/hazel/pull/1080/files#diff-db7a3f96ed680c747b5dbed63fad57a48276cbe28696aacf3d662f52f215d8ffL55) that was abandoned bc it broke other things.

dm0n3y (approved):

dm0n3y:
linking to original PR that introduced dark ids https://github.com/hazelgrove/hazel/pull/760 for future reference

PR #1084, opened 2023-08-18
parameterized types

xzxzlala:
Based on branch poly-adt-after2, I implemented the parameterized types. 

- I merged the forall type and typAp/typFun.

- Add the forms tpatAp and let the var could appear in tpat and typ.

- Add a new type "Ap" and its rules.

- Add the statics of the parameterized types. (tyalias of Ap)

- Revise the pattern match rules in evaluator to fit the "Forall" constructors.

SV0UKX3G7YF4TI5VMINYP0A

Here is an example of the parameterized types.
Note that the type of constructor of the parameterized types is Forall type, so we need to first use @<> to apply the type.


cyrus-:
way out of date at this point, so archiving for reference in a future effort

PR #1092, opened 2023-08-26
Support for Explicit Polymorphism

Crazycolorz5:
See #1030. Changed target branch after rebase as to not lose code in poly-adt-after relevant to possible implementation of Debrujin Indices.


Crazycolorz5:
To account for an analyzed type having a different binding than the expression being analyzed against, we perform alpha renaming via capture-avoiding binding. This can expose internal fresh type variable names to the user (see the screenshot below). This _might_ be able to be remedied by having the Forall constructor have an extra field that tracks the original binding name, and using the original name for all user-facing output (and possibly the same for Var).

image


Edit: Or by implementing DeBrujin indices. Thoughts from the reviewers on preference? We could also try to merge this sooner and then I can make a separate PR to implement DeBrujin (which, I may note, has most of the code relevant to the implementation in poly-adt-after, so it's a matter of making it work with the ADT code).

Edit2: Alternatively, we could require the same binding name in a type annotation and a type function. This seems to be the approach to most System F formalizations, though I'd say it's less free for the programmer. This amounts to requiring an equivalence between the typefun X and forall Y in the analytic rules for checking type functions.

Crazycolorz5:
image

Use of type variable names can cause incorrect substitution when shadowing occurs. It's unclear to me how to sanely resolve this aside from alpha renaming on the spot if a typfun introduces shadowing, or to just use Debrujin indices.

Crazycolorz5:
image

Printing does not properly paranthesitize foralls. Also, printed foralls are "Forall" with capital F, but the syntax is with lowercase.

Also create an example page.

cyrus-:
What necessitated moving to `.` instead of `->` in forall types?

cyrus-:
> we could require the same binding name in a type annotation and a type function

I think this is the best solution here -- needing / exposing internally generated names is not ergonomic.

cyrus-:
It would be nice to allow lowercase type variable names with this change -- let's go ahead and change the regexp for type variable names. It would be good actually to only allow them in foralls (with implicit polymorphism, lowercase names will be implicitly quantified).

cyrus-:
Still need to add lang docs for the things in this PR.

Crazycolorz5:
> What necessitated moving to . instead of -> in forall types?

It's not a necessary change, but I think it's useful for the reasons of:
- More closely matches how it's typically presented in theory. (In fact, I was motivated to change this initially because the type output printed them as such, and I didn't notice that it differed from how we input it until I was specifically trying to make them match)
- Disambiguates forall quantification from function types
- Makes a syntactic distinction between type functions (typfun, ->) and the type of type functions (forall, (.)). (I had actually found myself typing foralls where a typfun was expected. I suspect requiring a different separator for foralls will help with this, especially with those less familiar with operating on types).

> same binding name in a type annotation and a type function

> lowercase type variable names 

> lang docs

Sure, I can look at getting these done before Tuesday.

Crazycolorz5:
Re: same binding name in type annotation and type function.
Supporting this is more difficult than I initially thought. This is because we would like to require the strict equality in the analysis case. However, with the way Info is structured, we'd need to pass down information that it was in the analysis of a typfun (as opposed to the analysis for the argument of a function). Otherwise we end up excluding cases we don't want to, such as `let id : forall A . (A -> A) = in (fun x : forall X . (X -> X) -> x(id))` (which we don't want to exclude since this may be hidden behind several substitutions through which we may not be able to guarantee agreeing binding names). I've implemented this as-is in poly-adt-after3 if you're curious as to how that looks.

This is reflected in the formalism in Ap() and Cast() allowing the first type to be up to alpha. We don't have a good way to specifically allow this in Ana here, unless we add that to the mode, which seems a bit heavy for a (possibly?) niche issue that can be fixed with Debrujin indices.

For these reasons I'm going to leave this out of the current set of edits from the review.

src/haz3lweb/view/LangDoc.re
```diff
@@ -672,20 +672,78 @@ let get_doc =
...
-        get_message(
-          doc,
-          options,
-          LangDocMessages.triv_exp_group,
-          doc.explanation.message,
-          [],
-        );
+      | TypFun(_, _ /* tpat, body */) =>
+        basic_info(LangDocMessages.typ_function_group)
+      /* TODO: Cannot quite get the below to work. The EmptyHole case throws an unhandled exception Not_found.
```
Crazycolorz5: Might need some help here from someone that understands the LangDoc infrastructure better than me?

src/haz3lcore/statics/Term.re
```diff
@@ -128,6 +200,23 @@ module UTyp = {
...
       | List(u) => List(to_typ(ctx, u))
       | Parens(u) => to_typ(ctx, u)
+      | Forall({term: Var(name), _} as utpat, tbody) =>
+        let ctx =
+          Ctx.extend_tvar(
+            ctx,
+            {name, id: UTPat.rep_id(utpat), kind: Abstract},
+          );
+        Forall(name, to_typ(ctx, tbody));
+      | Forall(_, tbody) => to_typ(ctx, tbody)
```
cyrus-: * don't use underscore here in case of future expansion of UTPat
* if there is a UTPat hole here, then I think it should still elaborate to a forall -- we can either generate a fresh variable, or preferably add UTPat holes into the Typ representation as well. It should behave just like a variable that is never used.

src/haz3lcore/statics/Term.re
```diff
@@ -128,6 +200,23 @@ module UTyp = {
...
+      // Forall is same as Rec
+      | Rec({term: Var(name), _} as utpat, tbody) =>
+        let ctx =
+          Ctx.extend_tvar(
+            ctx,
+            {name, id: UTPat.rep_id(utpat), kind: Abstract},
+          );
+        Rec(name, to_typ(ctx, tbody));
+      | Rec(_, tbody) => to_typ(ctx, tbody)
```
cyrus-: same here

src/haz3lcore/statics/Term.re
```diff
@@ -548,19 +606,21 @@ module UExp = {
...
     | ListLit => "List literal"
     | Constructor => "Constructor"
     | Fun => "Function literal"
+    | TypFun => "Type Function Literal"
     | Tuple => "Tuple literal"
     | Var => "Variable reference"
     | Let => "Let expression"
     | TyAlias => "Type Alias definition"
     | Ap => "Application"
+    | TypAp => "Type Application"
```
cyrus-: lowercase "a" in "application"

src/haz3lcore/statics/Term.re
```diff
@@ -571,9 +631,13 @@ module UExp = {
     | UnOp(op) => show_unop(op)
     | Match => "Case expression";
 
+  // TODO (poly): May need to create a separate is_typfun function,
```
cyrus-: this sounds sensible to me

src/haz3lcore/statics/TypBase.re
```diff
@@ -130,12 +133,43 @@ module rec Typ: {
...
+      ListUtil.flat_map(
+        fun
+        | None => []
+        | Some(typ) => free_vars(~bound, typ),
+        List.map(snd, sm),
+      )
+    | Prod(tys) => ListUtil.flat_map(free_vars(~bound), tys)
+    | Rec(x, ty) => free_vars(~bound=[x, ...bound], ty)
+    | Forall(x, ty) => free_vars(~bound=[x, ...bound], ty)
+    /* TODO: check that these are correct. */
```
cyrus-: lgtm

src/haz3lcore/statics/TypBase.re
```diff
@@ -601,4 +648,33 @@ and Kind: {
   type t =
     | Singleton(Typ.t)
     | Abstract;
+  /* Is not needed as not using debrujin indices */
```
cyrus-: can delete if not used in this PR

src/haz3lcore/statics/TypBase.re
```diff
@@ -175,40 +224,22 @@ module rec Typ: {
...
+    | (Prod(tys1), Prod(tys2)) => List.equal(eq_internal(n), tys1, tys2)
     | (Prod(_), _) => false
-    | (List(t1), List(t2)) => eq(t1, t2)
+    | (List(t1), List(t2)) => eq_internal(n, t1, t2)
     | (List(_), _) => false
     | (Sum(sm1), Sum(sm2)) =>
-      ConstructorMap.equal(Option.equal(eq), sm1, sm2)
+      ConstructorMap.equal(Option.equal(eq_internal(n)), sm1, sm2)
     | (Sum(_), _) => false
     | (Var(n1), Var(n2)) => n1 == n2
```
cyrus-, line 240: this does not normalize singletons, so that should be noted in a comment -- this is only type equivalence on already normalized types

src/haz3lcore/statics/TypBase.re
```diff
@@ -244,18 +275,17 @@ module rec Typ: {
       !resolve && eq(ty_name, ty_join) ? Var(name) : ty_join;
     /* Note: Ordering of Unknown, Var, and Rec above is load-bearing! */
     | (Rec(x1, ty1), Rec(x2, ty2)) =>
-      /* TODO:
```
cyrus-, line 248: did this TODO get resolved? still looks like we're doing substitution with an open term, `Var(x1)`...

src/haz3lcore/statics/TypBase.re
```diff
@@ -365,6 +397,13 @@ module rec Typ: {
     | _ => (Unknown(Internal), Unknown(Internal))
     };
 
+  let matched_forall = (ctx, ty) =>
+    switch (weak_head_normalize(ctx, ty)) {
+    | Forall(t, ty) => (t, ty)
+    | Unknown(SynSwitch) => ("matched_forall", Unknown(SynSwitch))
```
cyrus-: this could be resolved if we added a TPat hole in the Typ representation -- don't need a fresh/dummy variable

src/haz3lcore/statics/TypBase.re
```diff
@@ -442,6 +482,7 @@ and Ctx: {
   let lookup_var: (t, string) => option(var_entry);
   let lookup_ctr: (t, string) => option(var_entry);
   let is_alias: (t, TypVar.t) => bool;
+  let is_tvar: (t, TypVar.t) => bool;
```
cyrus-: should be called is_abstract not is_tvar

src/haz3lcore/statics/Kind.re
```diff
@@ -1 +0,0 @@
-include TypBase.Kind;
```
cyrus-: don't need to delete this, useful to have as a convenience

cyrus- (changes requested):
partial review -- see comments

src/haz3lcore/statics/Term.re
Crazycolorz5: Do you just mean give the variable a name? I think the linter complains of it being an unused variable since we don't use it in the case where it's not a type variable name.

src/haz3lcore/statics/TypBase.re
Crazycolorz5: I think this is resolved since subst() itself performs capture-avoiding substitution. I'll note that in a comment and replace this one.

cyrus-:
@Crazycolorz5 when you get a chance, this needs some merge conflicts resolved.

Crazycolorz5:
<s>I addressed all the merge conflicts, but there seem to be a few subtle bugs with respect to argument casts on function applications after casts. See the following:
image

The reason this errors is because (fun x -> string_of_int(x+1))  is cast to `? -> ?`, so the literal `1` in the list is cast to `?`, but it then errors when it is plugged into `x+1`, as there is no cast added to turn it from `? => Int`. 

I'm not immediately sure how to fix this. Doing any of the following removes the error:
 - annotate x in the last line with `: ?`
 - fill in the type annotation for f as `X -> ?`
 - fill in the type annotation for f as `X -> Y` and annotate x as `: Int`
 
The following do not remove the error:
 - fill in the annotation for x as `: Int`
 - annotate x as `: ?` and f as `X -> ?`

The minimal example I can get to error from something like this is as follows:
image

Notably, this does not error on `dev`.

Given that Elaborator.re looks almost identical, I suspect it may be a bug in TypeBase.re? </s>


I fixed it in the commit just below.

cyrus- (changes requested):
We're getting `Warning: Molds.get: can't decide if operand or operator: '.'` in the console on load, which doesn't happen on `dev`. Not sure what happened with dot, maybe from a previous iteration where we used dot as the forall separator?

Crazycolorz5:
I've been looking into this warning for a few hours. I suspect it might have to do something with float tokenization in langdocs or init? I know for sure it's not caused by the forall stuff; I triple checked every where that that's parsed and it uses `->`. It could perhaps be a bug due to parser precedence as well? But I don't think I'm able to figure out what exactly is causing it (without better debugging tools for Reason). See the attached files for the console output when I print all tiles being re-molded (i.e. adding `print_string(Tile.show(t));` at line 129 in Segment.re, one for when I do this in `dev` and one here.
[out-dev.txt](https://github.com/hazelgrove/hazel/files/13847731/out-dev.txt)
[out-poly-adt-after2.txt](https://github.com/hazelgrove/hazel/files/13847732/out-poly-adt-after2.txt)

cyrus-:
As discussed, this needs to be brought up-to-date with dev, which has a new version of the LangDocs (ExplainThis) system that might fix the above issue. Marking as draft for now.

Crazycolorz5:
Updated to merge with dev. Note that #1215 breaks the current Polymorphism entry in the Documentation tab in Init. The same `Molds.get` warning from before is present; this seems to be benign and doesn't cause any errors.

Crazycolorz5:
image
Was playing around with this example and this output is wrong. The variable is getting treated as a hole, which is correct, but it's not marked as such in the UI.


Edit: Seems like there's also an error with a forall type hole when rendering its text description
image

Crazycolorz5:
Fixed the errors noted in the previous comment.

cyrus-:
Few small issues:

1. Type functions do not indent their bodies when you create a newline, the way regular functions do.

2. The ExplainThis description for type functions uses function syntax at the top, and let's use lowercase type variables in the examples.
image

3. The ExplainThis description for type application shouldn't use `_typ` for the type part, and the example is falling off the right side of the sidebar. Would be better to bind polymorphic identity to a variable and then instantiate it on the next line.

4. Need to add examples of type functions and type application to the Language Reference page in Documentation. (I can do this before PR merge if you'd like.)

cyrus-:
Also would be good to give polymorphic functions optional names in the internal language for display in the same way functions do, so it doesn't just always say anonymous typfun:
image

(also one more merge conflict to resolve)

cyrus-:
The name still doesn't seem to show up in the result?
image

Crazycolorz5:
I think your screenshot is missing the result line, unless I'm misinterpreting where you mean for the name to show up?
This is what the version on build/poly-ady-after2 looks like for me:
image

cyrus-:
Marking as draft pending resolution of stepper issue we discussed today

cyrus-:
It would be nice if type application didn't erase the name of the function, e.g. if `<map>@<Int>` stepped to `<map@<Int>>` (essentially just updating the string field of the underlying function or type function on type app).

Crazycolorz5:
image
It has been done.

cyrus-:
looks like even non-recursive polymorphic functions are getting fixpoints placed around them:
image

also the `Forall` there should not be capitalized and should not have the `.{tau}` notation

Crazycolorz5:
> looks like even non-recursive polymorphic functions are getting fixpoints placed around them: image
> 
> also the `Forall` there should not be capitalized and should not have the `.{tau}` notation

image
It appears that all functions that are type annotated have the fixed points -- this screenshot is from dev? As far as I can tell this is intended behavior,

I'll address the internal printing being Forall momentarily.

cyrus-:
Ah okay, I'll make a separate issue.

cyrus-:
@Crazycolorz5 looks nearly ready for merge -- just needs the merge conflict resolved!

Crazycolorz5:
Yep, just pushed that. Though I'm concerned about the potential for a large number of merge conflicts for others since we need to re-order the modules in `Term.re` since UTyp depends on UTPat in this branch. Nevertheless, this should be good, pending tests.

cyrus-:
Yeah no worries about merge conflicts with other branches, that's just part of the cost of having a lot of projects going on.

cyrus- (approved):

PR #1099, opened 2023-08-31
Better Indent

### LINKED ISSUE #882: Fix indentation for case rules
Case rules are indented but their bodies are not indented so it ends up quite awkward looking: 

Image

### LINKED ISSUE #887: Indentation of case rules when previous rule's expression is not bidelimited is incorrect
Image

### LINKED ISSUE #1087: Indentation triage
@dm0n3y how involved would it be to do a little indentation TLC? currently indentation in the complete syntax case can get somewhat confusing; I can't quite tell what's happening in screenshot-related. A patch here should probably also address #882, #887.

Screen Shot 2023-08-21 at 10 14 12 PM

disconcision:
- Closes #882 
- Closes #887 
- Closes #1087

In addition to fixing the preceding complete-case indentation bugs, this introduces a new system for indentation in the incomplete case, whereby indentations are based on the same best-effort syntax completion scheme used in TyDi to provide better semantic information to the assistant.

The result is the while adding new code to a new line above existing complete code, writing left-to-right, any partially-entered forms should get correct indentation (as in this case the completion scheme will always succeed), whereas all pre-existing syntax (below the newline following the new code) should retain its previous indentation.

Things get more interesting however when you move around with an unbalanced backpack; in such situations indentation may shift as you move the caret vertically, as the best-effort put-down make succeed in dropping the backpack contents, which may create or remove an indentation context.


cyrus- (changes requested):
@disconcision this looks good to me, just needs a merge in `Code.re`

disconcision:
@cyrus I've brought this up to date with dev. As we talked about, I have some misgivings about merging this in, due to (a) likely minor but complex-to-reason-about performance hit, (b) debatably confusing behavior when indentation changes with caret movement. As a result, I am ambivalent about whether this should be merged or not, though I do intend to maintain it for now as a reference point.

cyrus-:
Closing in anticipation of new tylr implementation, will have to revisit this topic.

disconcision:
Closed in favor of #1761

PR #1101, opened 2023-09-02
Invalid type casting fix + Nested Annotation casting candidate fix

### LINKED ISSUE #1100: Certain multiply-annotated patterns lead to internal exceptions
image
image

Similarly, annotating both the variable bound to a function and the input variable of that function (which must be an `Unknown(Internal)`) results in a CastBVHoleGround exception:
image
In most cases, users won't go into this. However, if a user annotates a function that takes a function as an input, and then attempts to annotate the variable type of that function, they may encounter this issue (still not likely).

disconcision:
Closes #1100.

Two separate fixes here. The first is simple, and addresses the last screenshot in #1100. Previously, invalid types (as created by the minus here) were interpreted as Unknown(Internal) instead on Unknown(TypeHole), causing them to bypass the special fix case in join added to address the previous function casting bug. (This is the change in Term)

The second is a bit more substantive:

Screen Shot 2023-09-02 at 12 53 02 AM

The problem fundamentally is existing annotation semantics mean that nested annotations interfere with information flow between variable bindings and their definitions.

Currently, the self type of an annotation is the annotated type, and the mode passed to the pattern is the annotated type. In the first line, b says it's unknown, but this self type is entirely hidden by the Int annotation, so 2 thinks that b says it's an Int. In the second line, b says it's an Int, but this self type is entirely hidden by the unknown annotation, so 2 thinks b says it's unknown. Thus the casts are inconsistent.

My attempted fix is to propagate type information bidirectionally through annotations. That is, the self type of an annotation is now the join of the pattern's type and the annotation, and the mode of the pattern is join of the analytic type and the annotation. (This is the change in Statics+Info+Self)

This ended up requiring a supporting change (in TypBase) to ensure that TypeHoles were not dominated by Internals, as otherwise this change (similar to the first fix) interferes with the join fix. I don't think this provenance dominance case was particularly principled so this is a neutral change I think.

Anyway, not sure if this is the right approach.

### DIFF
--- a/src/haz3lcore/statics/Self.re
+++ b/src/haz3lcore/statics/Self.re
@@ -104,3 +104,9 @@ let list_concat = (ctx: Ctx.t, tys: list(Typ.t), ids: list(Id.t)): t =>
   | None => NoJoin(ty => List(ty), add_source(ids, tys))
   | Some(ty) => Just(ty)
   };
+
+let of_ann = (ctx: Ctx.t, ~p: Typ.t, ~ann: Typ.t) =>
+  switch (Typ.join_fix(ctx, ann, p)) {
+  | Some(ty) => Just(ty)
+  | None => Just(ann)
+  };
--- a/src/haz3lcore/statics/Mode.re
+++ b/src/haz3lcore/statics/Mode.re
@@ -120,3 +120,14 @@ let of_ap = (ctx, mode, ctr: option(Constructor.t)): t =>
     }
   | None => SynFun
   };
+
+let of_ann = (ctx: Ctx.t, mode: t, ann: Typ.t) =>
+  switch (mode) {
+  | Syn
+  | SynFun => Ana(ann)
+  | Ana(ty) =>
+    switch (Typ.join_fix(ctx, ty, ann)) {
+    | Some(ty) => Ana(ty)
+    | None => Ana(ann)
+    }
+  };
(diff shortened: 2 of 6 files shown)


src/haz3lcore/statics/Self.re
```diff
@@ -104,3 +104,9 @@ let list_concat = (ctx: Ctx.t, tys: list(Typ.t), ids: list(Id.t)): t =>
   | None => NoJoin(ty => List(ty), add_source(ids, tys))
   | Some(ty) => Just(ty)
   };
+
+let of_ann = (ctx: Ctx.t, ~p: Typ.t, ~ann: Typ.t) =>
+  switch (Typ.join_fix(ctx, ann, p)) {
```
cyrus-, line 109: looking at the definition of `join_fix` more closely (which was from the previous PR), I'm confused -- it seems to be defined asymmetrically, with Unknown(TypHole) on the right behaving differently than Unknown(TypHole) on the left. I would expect a join to be symmetric. That asymmetry seems to be load-bearing here in favoring the direct annotation on the variable

disconcision:
Reminder from today's meeting:

1. Prohibit arbitrarily nested annotations
2. New error status on any annotation who pattern contains (recursively) another annotation

cyrus-:
Marking as draft until above issues are addressed.

cyrus-:
@disconcision can you follow up with this when you get the chance

cyrus-:
Closing, created a new issue for a future revisit #1233

PR #1103, opened 2023-09-05
IndexedDB Logging

### LINKED ISSUE #988: Optimize logging and state saving
Relatively easy gains to be had here.

Make writes to localstore async so that they don't block the UI. The expensive functions are Update.save_editors and Log.append_entry. The latter is only expensive when the log is large. It probably should also be refactored to no longer depend on log length, but the async change would successfully sweep that under the carpet I think.

disconcision:
Closes #988.

This replaces localstorage logging with an IndexedDB table. PKs are (string) timestamps, entries are the existing Log.Entry data structure, serialized to strings via sexps.

This passes my basic functionality and perf tests. I haven't tried to test it at scale (1000s of actions). I'm not worried about updating, that should be far more efficient now, but exporting and importing might be interesting. Importing in particular fires as many queries as there are log entries in the imported file.

I disabled the webworkers (since they're not doing anything), as otherwise the IndexedDB library fills the console with (non-fatal) errors about trying to access window (IndexedDB is accessed via window.indexedDB). This is thrown from a callback, not sure how to fix this without forking the library.

One outstanding issue: On compilation I get the message:
`Warning 58 [no-cmx-file]: no cmx file was found in path for module Ezjs_idb, and its interface was not compiled with -opaque`
Not sure how to suppress this.

### DIFF
--- a/src/haz3lweb/dune
+++ b/src/haz3lweb/dune
@@ -12,6 +12,8 @@
   \
   Worker)
  (libraries
+  ezjs_min
+  ezjs_idb
   str
   incr_dom
   virtual_dom.input_widgets
--- a/README.md
+++ b/README.md
@@ -166,6 +166,16 @@ independent of the one you sent in your home directory. This allow you to
 alternate dependencies, or test dependencies changes, without affect
 existing OCaml projects.
 
+NOTE: You may see the following warning when building:
+
+```
+Warning 58 [no-cmx-file]: no cmx file was found in path for module Ezjs_idb, and its interface was not compiled with -opaque
+```
+
+This is due to an upstream library issue and does not cause problems with Hazel: 
+
+  https://github.com/OCamlPro/ezjs_idb/issues/1
+
 ### Debugging
 
 #### Printing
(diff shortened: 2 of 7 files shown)


disconcision:
Something to note: Previously the FinishImportAll actions seems to have been logged. It takes as parameter the entire import file, so if this was being logged last semester, we may have been adding multiple megabytes to the log just from that one action

src/haz3lweb/dune
```diff
@@ -12,6 +12,8 @@
   \
   Worker)
  (libraries
+  ezjs_min
```
cyrus-, line 15: the dependencies need to be updated so that the build doesn't fail

cyrus-:
> One outstanding issue: On compilation I get the message: `Warning 58 [no-cmx-file]: no cmx file was found in path for module Ezjs_idb, and its interface was not compiled with -opaque` Not sure how to suppress this.

I filed an issue here: https://github.com/OCamlPro/ezjs_idb/issues/1.

cyrus-:
Nice work! I'm not worried about the import performance, but I suspect it isn't going to cause a serious problem.

PR #1106, opened 2023-09-08
Fix sequence stepping

Negabinary:
This fixes the stepper not showing any steps when evaluating `1;2`.

### DIFF
--- a/src/haz3lcore/dynamics/EvaluatorStep.re
+++ b/src/haz3lcore/dynamics/EvaluatorStep.re
@@ -1163,7 +1163,12 @@ module Decompose = {
         decompose(d1) >>| Return.wrap(c => Cast(c, ty, ty'))
       | BoundVar(_) => Return.mark(act)
       | Sequence(d1, d2) =>
-        decompose(d1) >>| Return.wrap(c => Sequence(c, d2))
+        let* r1 = decompose(d1);
+        switch (r1) {
+        | BoxedValue
+        | Eval(_) => Return.mark(act)
+        | _ => r1 |> Return.wrap(c => Sequence(c, d2)) |> Monad.return
+        };
       | Let(dp, d1, d2) =>
         let* r1 = decompose(d1);
         switch (r1) {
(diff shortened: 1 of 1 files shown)


src/haz3lcore/dynamics/EvaluatorStep.re
```diff
@@ -1163,7 +1163,12 @@ module Decompose = {
         decompose(d1) >>| Return.wrap(c => Cast(c, ty, ty'))
       | BoundVar(_) => Return.mark(act)
       | Sequence(d1, d2) =>
-        decompose(d1) >>| Return.wrap(c => Sequence(c, d2))
+        let* r1 = decompose(d1);
+        switch (r1) {
+        | BoxedValue
+        | Eval(_) => Return.mark(act)
+        | _ => r1 |> Return.wrap(c => Sequence(c, d2)) |> Monad.return
```
cyrus-: generally good not to use a wildcard to make future refactoring (e.g. adding new cases) easier -- list out the cases

cyrus- (changes requested):

src/haz3lcore/dynamics/EvaluatorStep.re
tonyfettes: I have used _ pattern in other places - maybe we shall remove these altogether.

I would like to change the `_` pattern to the following - Otherwise it's just plainly listing out all 4 * 4 - 4 = 12 possibility and contribute nothing to the semantic of the code.
```suggestion
        | (_, Indet)
        | (Indet, _)
        | (Step(_), _)
        | (_, Step(_))  => r1 |> Return.wrap(c => Sequence(c, d2)) |> Monad.return
```

tonyfettes (approved):
See comments above. Otherwise looks great.

src/haz3lcore/dynamics/EvaluatorStep.re
Negabinary: This switch statement actually only has one input, so there's only 4 possibilities. I've written them all out now.

PR #1113, opened 2023-09-20
code folding

gensofubi:
module view.

The target is to make a foldable view for module definitions. Currently I'm going to add a id map indicating whether a module definition is folded, which is similar to how the values of livelits are stored in #955 . (I think that binding the collapse property to an expression means that we need a new label to represent the collapsed module (since tile has no other parameters to help distinguish it), and by clicking the button to replace an exp having a certain id with another one looks tricky)

Current progress: Added a clickbox before every module defining expression. Fixed the relative position of select shadowing. Solve the problem that a clickbox also appears in the backpack view. Add a folded view which replace the definition code if checkbox is checked. 

Disabled Caret moving into folded codes. If caret at the side of a folded part, pressing left or right will move it to the other end. Other movement behavior: up/down/home/end behave normally. Clicking on the second half of the ellipsis will move the cursor to the side opposite the initial position. When continuously pressing Tab, cursor stops exactly once at the ellipsis if any holes (no matter how much) appear in the omitted part.
Another small issue: selection sometimes cause the omitted holes to appear.

Move the checkbox to the side of editor. Only the first module expression in each row will have fold button.

Change the checkbox to a button because the checkbox's default checking appearance seems not coordinate well with my code.

Automatically Move the cursor at the beginning of module statement when folding, to avoid the cursor being inside folded codes.

Create a shortcut to fold/unfold (ctrl+shift+'['/']'). 

~~Hide the fold button. Display on hovering~~

Use the botton style in top bar.
{1EDAB811-BF24-403e-AE39-F23C22CC1A7A}


cyrus-:
@gensofubi can you describe in more detail how you are going about adding the outliner support (since we had discussed a number of possible approaches when we chatted)

Negabinary:
Nice demo yesterday btw! The code looks good to me, and the actual feature feels robust to play with. 
I wonder if we can make the fold button look a little nicer though? You might have other ideas, but if you don't I'd try making it look like the buttons at the top of the screen - same colour, no rectangular border, always showing (I personally think having them hidden makes them hard to find, but maybe the screen gets cluttered if they're showing?), gets bigger when you mouse over it. And then maybe you could use a couple unicode triangle symbols for 🞂 and 🞃

gensofubi:
Thanks for the advice for apprearance. I've used the style of the buttons in the top bar, and unicode triangle symbols for 🞂 and 🞃. 
Considering that in practice the buttons won't be dense, I decide to not hide them to let user directly know that modules are foldable.

xzxzlala:
Hi Tao! I'm trying to expand the breadcrumb bar for your module view system, could u merge dev to enable me to start working on it? Or maybe I could work on this branch? Thanks a lot!

gensofubi:
@xzxzlala I've merged with dev. btw if your work is not involved with folding I'd suggest working on #1020, which includes all module functioning and is more stable.

xzxzlala:
Got it!

Negabinary (approved):
I think this is ready to merge

cyrus-:
There seem to be some issues with cut/copy when you have selected across a collapsed region:

image

I get the above Not_found exception in the console when I try to Ctrl+C in the shown state. It seems to work with some other ranges so not sure what is going on there.

cyrus-:
@disconcision can you take a look at this when you get a chance

cyrus-:
Marking this as draft and pending -- would like a folding UI like this PR, but based on @disconcision's work on #1218.

cyrus-:
closing in favor of projectors-based approach in #1218 -- thanks for the effort on this @gensofubi. I think we might want something that looks more like this at some point, even if based on projectors.

PR #1094, opened 2023-08-27
Porting implementation of incon judgment into Hazel 3.

DavidFangWJ:
This branch is created by porting `Constraint.re` and `Incon.re` into Hazel 3 (aka branch `dev`). It has not yet been tested, and the treatment of lists is potentially incorrect.

PR #1114, opened 2023-09-20
Case expression exhaustiveness & redundancy checking

pigumar1:
This pull request aims to implement the exhaustiveness checking feature mentioned in [Live Pattern Matching with Typed Holes](https://victoryyw.github.io/assets/pdfs/pattern.pdf).

If a `case` expression is _necessarily inexhaustive_ (i.e.: the expression will necessarily fail to match all values of the type of its scrutinee), the expression will be wrapped with an error hole like in Fig. 2b in the paper. This is planned to be done by introducing a new `error_exp` called `InexhaustiveMatch` _(tentative)_, which behaves uniquely when supplied to `CursorInspector.re`. Like discussed in #1059, we probably need a corresponding `Self.exp` for the new expression error. `Self.match()` might be modified accordingly.

Exhaustiveness checking on a `case` expression relies on whether the final constraint for the rules can be entailed from truth (i.e.: all expressions of the type of the `case` expression’s scrutinee “determinately or indeterminately match any of the branches”). This final constraint for the rules is discovered in the process of typing the rules of the match expression, which is to be implemented by applying the sequence of rules to a type generator called `ruls_to_info_map` _(tentative)_. It corresponds to the judgment _TRules_ in the paper (ideally). `ruls_to_info_map` is going to be based on the list scheme `List.fold_left` instead of `List.map` since the info of rules will be dependent on each other in the redundancy checking.

~~The entailment check itself has already been implemented in #1094 as the function `Incon.is_exhaustive`.~~

After checking the exhaustiveness, the branches are yet to be checked for type consistency **(should they be in sequential order? Or should two errors show up in the cursor inspector at the same time?)** So the output of `ruls_to_info_map` probably contains a list of expression types for branches and the final constraint for the rules.

(Karan Anand got involved in editing the proposal)


cyrus-:
Plan looks reasonable. If there is both a type error and an exhaustiveness error, it would be nice if we could show both but if not, let's favor the type error.

src/haz3lcore/statics/Self.re
```diff
@@ -35,6 +35,7 @@ type t =
 [@deriving (show({with_path: false}), sexp, yojson)]
 type exp =
   | Free(Var.t)
+  | InexhaustiveMatch(exp)
```
pigumar1, line 52: `Self.exp.InexhaustiveMatch` is finally designed to be a "wrapper" around (`Self.exp` assigned to a match expression when exhaustiveness checking is ignored). I wanted it to be as flexible as possible so that it might be easier to add new features to the match expression in the future.

src/haz3lcore/statics/Info.re
```diff
@@ -64,6 +64,7 @@ type error_common =
 [@deriving (show({with_path: false}), sexp, yojson)]
 type error_exp =
   | FreeVariable(Var.t) /* Unbound variable (not in typing context) */
+  | InexhaustiveMatch(option(error_exp))
```
pigumar1: `Self.exp.InexhaustiveMatch` gets translated to `Info.error_exp.InexhaustiveMatch`. There might be an additional error associated with the match expression. Currently, it seems that it can only be an inconsistent-branch error (`Common(Inconsistent(Internal(_)))`).

src/haz3lcore/statics/Info.re
```diff
@@ -318,9 +319,21 @@ let status_pat = (ctx: Ctx.t, mode: Mode.t, self: Self.pat): status_pat =>
    depending on the mode, which represents the expectations of the
    surrounding syntactic context, and the self which represents the
    makeup of the expression / pattern itself. */
-let status_exp = (ctx: Ctx.t, mode: Mode.t, self: Self.exp): status_exp =>
+let rec status_exp = (ctx: Ctx.t, mode: Mode.t, self: Self.exp): status_exp =>
   switch (self, mode) {
   | (Free(name), _) => InHole(FreeVariable(name))
+  | (InexhaustiveMatch(self), _) =>
```
pigumar1, line 413: Translates `Self.exp.InexhaustiveMatch` to `Info.error_exp.InexhaustiveMatch`. 

Currently, `additional_err` can only be `Common(Inconsistent(Internal(_)))`. Other type inconsistency errors (`Inconsistent(Expectation(_) | WithArrow(_))`) will be overrided by inexhaustiveness, which fails type checkings.

A match expression cannot be invalid tokens (`Common(NoType(_))`) or a free variable (`FreeVariable(_)`) at the same time, so currently, an error that can be shown along with an inexhaustive error can only be an inconsistent-branch error.

src/haz3lcore/statics/Statics.re
```diff
@@ -283,11 +283,12 @@ and uexp_to_info_map =
...
         p_ctxs,
       );
+    // let rec ruls_to_info_map = (xi_pre: Constraint.t, m) => (xi_pre, m);
     let e_tys = List.map(Info.exp_ty, es);
     let e_co_ctxs =
       List.map2(CoCtx.mk(ctx), p_ctxs, List.map(Info.exp_co_ctx, es));
-    add(
-      ~self=Self.match(ctx, e_tys, branch_ids),
+    add'(
+      ~self=InexhaustiveMatch(Common(Self.match(ctx, e_tys, branch_ids))),
```
pigumar1, line 291: Temporary code: just trying to get the new error showed up.

src/haz3lweb/view/CursorInspector.re
```diff
@@ -157,10 +157,19 @@ let typ_err_view = (ok: Info.error_typ) =>
     ]
   };
 
-let exp_view = (cls: Term.Cls.t, status: Info.status_exp) =>
+let rec exp_view = (cls: Term.Cls.t, status: Info.status_exp) =>
```
pigumar1, line 176: `exp_view` is modified to be recursive since `error_exp` can be somehow "recursive" now

pigumar1:
Dynamic layout: #1125

cyrus-:
The error marker seems to not just be on the case expression but also layered up a second level on the scrutinee + rules, creating a double box effect:

image

The number of these inner error boxes, judging by the opacity changes, is a function of the number of rules, so the bigger case expression above has a darker red background than the smaller expression.

pigumar1:
@cyrus- I have fixed this bug with the latest push. Thanks for Andrew's help!

cyrus-:
@pigumar1 is there anything left to do here? if not, mark as ready for review

pigumar1:
> @pigumar1 is there anything left to do here? if not, mark as ready for review

@cyrus- I listed some of my concerns in the code as github comments (which, for some reason, is only visible in the "Files changed" window). Other than that I believe it's ready.

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  | And(t, t)
+  | Or(t, t)
+  | InjL(t)
+  | InjR(t)
+  | Pair(t, t);
+// | Tuple(list(t));
+
+// Unused
+let rec constrains = (c: t, ty: Typ.t): bool =>
+  switch (c, Typ.weak_head_normalize(Builtins.ctx_init, ty)) {
```
cyrus-: shouldn't use the initial context -- take in the relevant context.

(or remove this function since it is unused)

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  // | (Tuple(cs), Prod(tys)) when List.length(cs) == List.length(tys) =>
+  //   List.for_all2(constrains, cs, tys)
+  // | (Tuple(_), _) => false
+  };
+
+let rec dual = (c: t): t =>
+  switch (c) {
+  | Truth => Falsity
+  | Falsity => Truth
+  // The complement of an indeterministic set is still indeterministic
```
cyrus-: not "indeterministic", can just remove this comment

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  | Hole => Hole
+  | Int(n) => NotInt(n)
+  | NotInt(n) => Int(n)
+  | Float(n) => NotFloat(n)
+  | NotFloat(n) => Float(n)
+  | String(n) => NotString(n)
```
cyrus-: don't use `n` for a string variable

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  | Float(float)
+  | NotFloat(float)
+  | String(string)
+  | NotString(string)
+  | And(t, t)
+  | Or(t, t)
+  | InjL(t)
+  | InjR(t)
+  | Pair(t, t);
+// | Tuple(list(t));
```
cyrus-: remove all the commented out Tuple stuff in this PR

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  | String(_)
+  | NotString(_) => c
+  | And(c1, c2) => And(falsify(c1), falsify(c2))
+  | Or(c1, c2) => Or(falsify(c1), falsify(c2))
+  | InjL(c) => InjL(falsify(c))
+  | InjR(c) => InjR(falsify(c))
+  | Pair(c1, c2) => Pair(falsify(c1), falsify(c2))
+  };
+
+// temporary name
```
cyrus-: if it's temporary, remove in the PR, else remove comment

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  | _ => failwith("input can only be pair(_, _)");
+
+let rec or_constraints = (lst: list(t)): t =>
+  switch (lst) {
+  | [] => Falsity
+  | [xi] => xi
+  | [xi, ...xis] => Or(xi, or_constraints(xis))
+  };
+
+// Temporary name
```
cyrus-: if it's temporary, remove in the PR, else remove comment

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+let rec ctr_of_nth_variant = (num_variants, nth): (t => t) =>
+  if (num_variants == 1) {
+    Fun.id;
```
cyrus-, line 114: what is Fun.id?

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -0,0 +1,173 @@
...
+  } else if (nth == 0) {
+    xi => InjL(xi);
+  } else {
+    xi => InjR(xi |> ctr_of_nth_variant(num_variants - 1, nth - 1));
+  };
+
+let of_ap = (ctx, ctr: option(Constructor.t), arg: t): t =>
+  switch (ctr) {
+  | Some(name) =>
+    switch (Ctx.lookup_ctr(ctx, name)) {
```
cyrus-: you want to base this on a provided scrutinee type, not relying on the context having the constructor names -- this doesn't work when you have anonymous sums

image

src/haz3lcore/statics/Info.re
cyrus-: this can be `error_common` instead of `error_exp` in that case

src/haz3lcore/statics/Info.re
```diff
@@ -200,6 +201,7 @@ type pat = {
   cls: Term.Cls.t,
   status: status_pat,
   ty: Typ.t,
+  constraint_: Constraint.t,
```
cyrus-, line 219: why a trailing underscore?

src/haz3lcore/statics/Statics.re
```diff
@@ -270,27 +270,39 @@ and uexp_to_info_map =
...
-    let branch_ids = List.map(UExp.rep_id, es);
-    let (ps, m) =
-      map_m(go_pat(~is_synswitch=false, ~mode=Mode.Ana(scrut.ty)), ps, m);
-    let p_ctxs = List.map(Info.pat_ctx, ps);
-    let (es, m) =
-      List.fold_left2(
-        ((es, m), e, ctx) =>
-          go'(~ctx, ~mode, e, m) |> (((e, m)) => (es @ [e], m)),
-        ([], m),
+    let rules_to_info_map = (rules: list((UPat.t, UExp.t)), m) => {
```
cyrus-: can move this to top-level rather than just in this branch

src/haz3lcore/statics/TypBase.re
```diff
@@ -440,10 +440,21 @@ and Ctx: {
     kind: Kind.t,
   };
 
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type constructor_entry = {
+    name: Var.t,
+    id: Id.t,
+    typ: Typ.t,
+    /* Temporary variables. Better implementation is a TO-DO. */
+    nth: int,
```
cyrus-: per other comment, we shouldn't have index calculations in the context because we have anonymous sums as well. we need to use the normalized type itself to determine indices wherever needed, and we should also be invariant to reordering (i.e. index is based on alphabetized position of constructor within the sum type rather than the list position)

src/haz3lcore/statics/TypBase.re
```diff
@@ -542,21 +564,61 @@ and Ctx: {
...
-            switch (typ) {
-            | None => Var(name)
-            | Some(typ) => Arrow(typ, Var(name))
-            },
-        }),
-      ctrs,
-    )
-    @ ctx;
+  let add_ctrs = (ctx: t, name: TypVar.t, id: Id.t, ctrs: Typ.sum_map): t => {
+    // let (ctx, _, _) =
```
cyrus-: don't leave large commented sections in PRs

cyrus- (changes requested):
see individual comments -- main issue is that the indexing scheme does not work for anonymous sums and does not respective equivalence of sum types up to reordering. see comments for suggestions.

otherwise looks good to me!

cyrus- (changes requested):
* small merge conflict needs to be resolved
* need to do exhaustiveness checking in lets and functions, not just case analysis:
image
and
image

pigumar1:
Not deployable yet since there is a format issue on the dev branch.

pigumar1:
@cyrus- Would you mind merging #1124 into this PR so that I can extract out reusable code without needing to update that PR as well?

pigumar1:
@cyrus- It's now working for lets and functions:
image
image

cyrus-:
@pigumar1 the error decorations for inexhaustive cases seem to be doubled up, with a second on on the scrutinee + rules. can we eliminate that? 

image

cyrus-:
When the scrutinee has unknown type, it looks like it marks the case expression non-exhaustive and every pattern redundant. I don't think this is what we want here. Ideally, we would check to see if all the patterns have a join type and check exhaustiveness and redundancy with that type I think.

image

pigumar1:
> @pigumar1 the error decorations for inexhaustive cases seem to be doubled up, with a second on on the scrutinee + rules. can we eliminate that?
> 
> image

@cyrus- I believe this is a separate issue related to parsing and can be fixed separately, and I have created a new pull request #1266 such that this issue can be solved before we merge everything about pattern matching!

src/haz3lcore/statics/TypBase.re
cyrus-: @pigumar1 there is still an index here -- what did you decide to do here?

src/haz3lcore/statics/TypBase.re
```diff
@@ -599,10 +610,21 @@ and Ctx: {
     kind: Kind.t,
   };
 
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type constructor_entry = {
+    name: Var.t,
+    id: Id.t,
+    typ: Typ.t,
+    /* Temporary variables. Better implementation is a TO-DO. */
+    nth: int,
```
cyrus-: same here

src/haz3lcore/statics/TypBase.re
cyrus-: @pigumar1 ^

cyrus- (approved):

PR #1118, opened 2023-10-08
Breadcrumb bar

xzxzlala:
Breadcrumb bar of hazel. 
Now merged the haz3l-module branch. 
could show the current function/module you are in, and the droppings contains other same level functions/modules.
The level is decided by how many funs/modules you are in.
For example, below the cursor is in f1 and g3 and finally in M2:
@Q1Y9HR T$6V0UT`E}{F3UP


src/haz3lweb/Init.ml
Negabinary, line 1: Having the breadcrumb_bars property in settings feels unintuitive to me - it's not like the other settings. I wonder whether this could be implemented using a <select> tag like how the scratch/examples/excercises button is implemented.

src/haz3lweb/UpdateAction.re
```diff
@@ -10,6 +10,7 @@ type settings_action =
   | Dynamics
   | Benchmark
   | ContextInspector
+  | Breadcrumb_bar(int, bool)
```
Negabinary: It's best to stick to the case convention for constructors of the other files and call it `BreadcrumbBar` instead of Breadcrumb_bar

src/haz3lweb/view/ScratchMode.re
```diff
@@ -98,7 +98,180 @@ let download_slide_state = state => {
...
+              ),
+            ]),
+          [text(name)],
+        ),
+      ];
+      //if (List.exists(a => a == List.hd(ids), ancestors)) {
+      //} else {
+      //  [];
+      //};
+      let funs_display = Array.make(10, []);
```
Negabinary: Generally in ReasonML we try to avoid using arrays, and use lists instead - it's because adding "mutability" makes it easy for errors to creep in - not that any mutability errors have crept in here, but just for a general rule.

src/haz3lweb/view/ScratchMode.re
```diff
@@ -98,7 +98,180 @@ let download_slide_state = state => {
...
+      let ancestors = Info.ancestors_of(ci);
+      let ancestors =
+        List.map(
+          a => {
+            switch (Id.Map.find_opt(a, info_map)) {
+            | Some(v) =>
+              switch (v) {
+              | Info.InfoExp({term, ancestors, _}) =>
+                switch (term.term) {
+                | Fun(_) => List.hd(ancestors)
```
Negabinary: Try typing just `fun x -> 5` into the hazel editor

src/haz3lweb/view/ScratchMode.re
```diff
@@ -98,7 +98,180 @@ let download_slide_state = state => {
...
+                List.rev(funs_display[level]),
+              ),
+            );
 
+          let siblings_div = {
+            let clss =
+              clss(
+                ["siblings"]
+                @ (
+                  List.nth(settings.breadcrumb_bars, level)
```
Negabinary: This will throw an error if someone has 11 nested functions (though anyone writing an 11th nested function probably deserves to have the editor give up on them because their code is too ugly...)

src/haz3lweb/www/style.css
```diff
@@ -442,7 +441,43 @@ select {
...
+  align-items: flex-end;
+  padding-left: 0.1em;
+  padding-right: 0.5em;
+  white-space: nowrap;
+  border-radius: 0 0.2em 0em 0;
+  color: #586e75;
+  border-top: 1px solid #e6ce8f;
+  border-right: 1px solid #c7b480;
+  border-bottom: 1px solid #c7b480;
+  background-color: #f7ebc5;
```
Negabinary: missing a border-left here

src/haz3lweb/www/style.css
```diff
@@ -442,7 +441,43 @@ select {
...
+  left: -1.2em;
+  max-height: 10em;
+  overflow: auto;
+  display: none;
+  flex-direction: column;
+  align-items: flex-end;
+  padding-left: 0.1em;
+  padding-right: 0.5em;
+  white-space: nowrap;
+  border-radius: 0 0.2em 0em 0;
```
Negabinary: is the rounded edge on just one corner intentional?

Negabinary (changes requested):
Cool demo yesterday!!

src/haz3lweb/Init.ml
Negabinary: \* using a `<select>` tag

src/haz3lweb/www/style.css
xzxzlala: I just copy it from context-inspector where the top-right corner is the only corner not close to other frame. So here maybe I should rounded two bottom corner.

src/haz3lweb/www/style.css
xzxzlala: Thanks! Added :)

src/haz3lweb/view/ScratchMode.re
xzxzlala: After I change into list, it doesn't need a fix integer to limit the nested functions.

src/haz3lweb/view/ScratchMode.re
xzxzlala: Hazel crashes because it has no ancestor and I perform a List.hd. 
Now I first check whether ancestors is empty.

src/haz3lweb/view/ScratchMode.re
xzxzlala: Fixed

src/haz3lweb/UpdateAction.re
xzxzlala: Thanks! I didn't notice that

src/haz3lweb/Init.ml
xzxzlala: Wonderful suggestion! It not only enables me to not adding property in settings, but also give a selecting window(html/ccss) for user to select which is a better UI than the one I designed without any consideration :)

src/haz3lweb/view/ScratchMode.re
```diff
@@ -98,7 +98,174 @@ let download_slide_state = state => {
...
+        | If(t1, t2, t3) =>
+          tag_term(t1, level) @ tag_term(t2, level) @ tag_term(t3, level)
+        | ListLit(t)
+        | Tuple(t) => List.concat(List.map(t1 => tag_term(t1, level), t))
+        | Match(t1, t) =>
+          let l1 = tag_term(t1, level);
+          let l2 =
+            List.concat(List.map(t1 => tag_term(snd(t1), level), t));
+          l1 @ l2;
+        | _ => []
```
Negabinary: Sorry, another annoying little detail!! For pattern matching over expressions like this we like to write out every case, instead of having a catch-all "_  => ..." so that if anyone else adds anything to the language, the compiler will force them to update this function.

src/haz3lweb/view/ScratchMode.re
```diff
@@ -98,7 +98,220 @@ let download_slide_state = state => {
...
+            List.remove_assoc(key, acc) @ [(key, [id, ...ids])]
+          | None => [(key, [id]), ...acc]
+          };
+        let result = List.fold_left(combine, [], lst);
+        List.map(((key, ids)) => (key, List.rev(ids)), result);
+      };
+      let tagged = tag_term(term, 1);
+      let lst = combineList(tagged);
+      let ancestors = Info.ancestors_of(ci);
+      let rec filter_ancestors = (ancestors_lst: Info.ancestors, level: int) =>
```
xzxzlala, line 157: For the code review, could someone take a look at this function("filter_ancestors") and give me some suggestions?
this function is very ugly, but I'm not sure whether there is a better way.
I try to filter the unrelated ancestors in this function by letting only the closest ancestor be the ancestor of the cursor at that level and then recursively deal with next level.
For the other cases needed to be recurse, they have exactly same code :(

cyrus-:
@xzxzlala this needs a merge now

cyrus-:
@disconcision can you do a style pass on this? (really this is meant for down the line when we have a module view)

disconcision:
how about something like this? 

Screenshot 2023-12-04 233516

(we should also just stick those icons in the nut menu)

src/haz3lweb/view/ScratchMode.re
Negabinary: sure, here's a couple tricks that might help:

1. you can combine patterns when pattern matching
2. you can add a "when" to pattern matches as well

so this whole thing could probably become just one pattern match with four cases

```reason
switch (Id.Map.find_opt(List.hd(ancestors_lst), info_map)) {
  | Some(Info.InfoExp({Fun(_), ancestors, _})) when List.length(ancestors) >= 1 => [ ... ]
  | ...
```

cyrus-:
> how about something like this?
> 
> Screenshot 2023-12-04 233516
> (we should also just stick those icons in the nut menu)

yeah this looks good!

cyrus-:
@xzxzlala I think we actually want to include let bindings of regular values, not just functions, in the list + we want to show siblings of the current position, even at the top level. we can have a special "dot" location for when you aren't on a binding.

xzxzlala:
> @xzxzlala I think we actually want to include let bindings of regular values, not just functions, in the list + we want to show siblings of the current position, even at the top level. we can have a special "dot" location for when you aren't on a binding.

Sure, I will add the "dot" for cursor not in bindings. 
I'm a little bit confused about using let binding of regular values. Won't that lead to lots of levels (fill all the room for top bar)?

cyrus-:
> > @xzxzlala I think we actually want to include let bindings of regular values, not just functions, in the list + we want to show siblings of the current position, even at the top level. we can have a special "dot" location for when you aren't on a binding.
> 
> Sure, I will add the "dot" for cursor not in bindings. I'm a little bit confused about using let binding of regular values. Won't that lead to lots of levels (fill all the room for top bar)?

It will add more items to the dropdown but it shouldn't increase the horizontal width much since that is based on depth -- it isn't common to nest let bindings very deeply.

xzxzlala:
> > > @xzxzlala I think we actually want to include let bindings of regular values, not just functions, in the list + we want to show siblings of the current position, even at the top level. we can have a special "dot" location for when you aren't on a binding.
> > 
> > 
> > Sure, I will add the "dot" for cursor not in bindings. I'm a little bit confused about using let binding of regular values. Won't that lead to lots of levels (fill all the room for top bar)?
> 
> It will add more items to the dropdown but it shouldn't increase the horizontal width much since that is based on depth -- it isn't common to nest let bindings very deeply.

I see, I misunderstand that. We want all other regular values to be siblings in same level and when in the body of new functions/modules we add another level.

cyrus-:
There would be a new level whenever one binding is nested in another, so also in 

```
let x = 
  let y = 2| in 
  y + 1
in 
x
```

disconcision:
k i'm good to restyle if you want. when this is current with dev i'll branch off and do the restyling, then PR against this

cyrus-:
Seems to be quite buggy right now -- try moving around the following program:

```
let x = 2 * 3 in 
let y = 3 * 4 in 
let z = 2 * 3 in 
x
```

It should indicate `x` only when in the expression `2 * 3` but it indicates `z` there. Similar issues elsewhere.

cyrus-:
this needs to be reworked on top of the latest updates to Hazel including the new module system implementation, so archiving. thanks for the design @xzxzlala!

PR #1121, opened 2023-10-16
type assignment for internal expressions

ruiz-m:
Adds a module that performs type assignment after conducting elaboration.


cyrus-:
@ruiz-m can you fix the small merge conflicts here

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -369,5 +390,10 @@ let uexp_elab = (m: Statics.Map.t, uexp: Term.UExp.t): ElaborationResult.t =>
       | Some(ty) => ty
       | None => Typ.Unknown(Internal)
       };
+    print_endline(DHExp.show(d));
```
cyrus-: remove prints before merge

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -369,5 +390,10 @@ let uexp_elab = (m: Statics.Map.t, uexp: Term.UExp.t): ElaborationResult.t =>
...
+    let _check = TypeAssignment.property_test(ty, d, m);
```
cyrus-: the convention in OCaml is that variables starting with `_` are unused

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -369,5 +390,10 @@ let uexp_elab = (m: Statics.Map.t, uexp: Term.UExp.t): ElaborationResult.t =>
...
+    _check
+      ? print_endline("Types are equal")
```
cyrus-: remove prints before merge

src/haz3lcore/dynamics/FilterMatcher.re
```diff
@@ -1,8 +1,8 @@
 let rec matches_exp =
         (env: ClosureEnvironment.t, d: DHExp.t, f: DHExp.t): bool => {
   switch (d, f) {
-  | (Constructor("$e"), _) => failwith("$e in matched expression")
-  | (Constructor("$v"), _) => failwith("$v in matched expression")
+  | (Constructor("$e", _), _) => failwith("$e in matched expression")
```
cyrus-: @tonyfettes why are the wildcards encoded as constructors rather than variables?

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
+open Util;
+open OptUtil.Syntax;
+
+let equal_typ_case = (l: list(Typ.t)): option(Typ.t) => {
+  switch (l) {
+  | [] => None
+  | _ =>
+    let ty = List.hd(l);
```
cyrus-: use pattern matching rather than List.hd / List.tl

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+        (ctx: Ctx.t, m: Statics.Map.t, dh: DHExp.t): option(Typ.t) => {
+  switch (dh) {
+  | EmptyHole(id, _) => delta_ty(id, m)
+  | NonEmptyHole(_, id, _, d) =>
+    switch (typ_of_dhexp(ctx, m, d)) {
+    | None => None
+    | Some(_) => delta_ty(id, m)
+    }
+  | ExpandingKeyword(_) => None
+  | FreeVar(_) => Some(Unknown(Internal))
```
cyrus-: do `FreeVar`s not have an entry in delta?

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+let rec typ_of_dhexp =
+        (ctx: Ctx.t, m: Statics.Map.t, dh: DHExp.t): option(Typ.t) => {
+  switch (dh) {
+  | EmptyHole(id, _) => delta_ty(id, m)
+  | NonEmptyHole(_, id, _, d) =>
+    switch (typ_of_dhexp(ctx, m, d)) {
+    | None => None
+    | Some(_) => delta_ty(id, m)
+    }
+  | ExpandingKeyword(_) => None
```
cyrus-: these should also have unknown type

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+  | FreeVar(_) => Some(Unknown(Internal))
+  | InvalidText(_) => None
```
cyrus-: these should also have unknown type

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+  | InconsistentBranches(_, _, Case(d_scrut, d_rules, _)) =>
+    let* ty' = typ_of_dhexp(ctx, m, d_scrut);
+    let typ_cases =
+      d_rules
+      |> List.map(rule_prj)
```
cyrus-: you don't need to do this map -- you can directly pattern match in the argument position on the next line

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+      |> List.map(((dhp, de)) => {
+           typ_of_dhexp(dhpat_extend_ctx(dhp, ty', ctx), m, de)
+         })
+      |> OptUtil.sequence;
+
+    switch (typ_cases) {
+    | None => None
+    | Some(_) => Some(Typ.Unknown(Internal))
+    };
+  | Closure(_, d) => typ_of_dhexp(ctx, m, d)
```
cyrus-: I think we need to add the items in the closure to the typing context for this to work

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+    let entry = Ctx.VarEntry({name, id: Id.invalid, typ: ty1});
+    typ_of_dhexp(Ctx.extend(ctx, entry), m, d);
+  | Fun(dhp, ty1, d, _) =>
+    let+ ty2 = typ_of_dhexp(dhpat_extend_ctx(dhp, ty1, ctx), m, d);
+    Typ.Arrow(ty1, ty2);
+  | Ap(d1, d2) =>
+    let* ty1 = typ_of_dhexp(ctx, m, d1);
+    let* ty2 = typ_of_dhexp(ctx, m, d2);
+    switch (arrow_aux(ty1)) {
+    | Arrow(tyl, tyr) when Typ.eq(tyl, ty2) => Some(tyr)
```
cyrus-, line 177: we don't do matched arrow types in the internal language -- that should all be done via casts

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+      };
+    } else {
+      None;
+    };
+  | ListLit(_, _, ty, _) => Some(List(ty))
+  | Cons(d1, d2) =>
+    let* ty1 = typ_of_dhexp(ctx, m, d1);
+    let* ty2 = typ_of_dhexp(ctx, m, d2);
+    switch (ty2) {
+    | List(Unknown(Internal)) => Some(Typ.List(ty1))
```
cyrus-: we don't need to do any consistency things here -- the head element should have the same type as the element type of the tail, which must be a list type.

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+    | List(ty3) when Typ.eq(ty3, ty1) => Some(ty2)
+    | _ => None
+    };
+  | ListConcat(d1, d2) =>
+    let* ty1 = typ_of_dhexp(ctx, m, d1);
+    let* ty2 = typ_of_dhexp(ctx, m, d2);
+    switch (ty1, ty2) {
+    | (List(Unknown(Internal)), _)
```
cyrus-: same as above

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+  | ConsistentCase(Case(d_scrut, d_rules, _)) =>
+    let* ty' = typ_of_dhexp(ctx, m, d_scrut);
+    let* typ_cases: list(Typ.t) =
+      d_rules
+      |> List.map(rule_prj)
+      |> List.map(((dhp, de)) => {
+           typ_of_dhexp(dhpat_extend_ctx(dhp, ty', ctx), m, de)
+         })
+      |> OptUtil.sequence;
+    Typ.join_all(~empty=Unknown(Internal), ctx, typ_cases);
```
cyrus-: we don't want to do a join here -- they should all be equal types

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+    let* tyd = typ_of_dhexp(ctx, m, d);
+    Typ.eq(tyd, ty1) ? Some(ty2) : None;
+  | FailedCast(d, ty1, ty2) =>
+    if (ground(ty1) && ground(ty2) && !Typ.eq(ty1, ty2)) {
+      let* tyd = typ_of_dhexp(ctx, m, d);
+      Typ.eq(tyd, ty1) ? Some(ty2) : None;
+    } else {
+      None;
+    }
+  | InvalidOperation(_) => None
```
cyrus-: this should have the type that the operation would have if it were successful

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+  | IfThenElse(ConsistentIf, d_scrut, d1, d2) =>
+    let* ty = typ_of_dhexp(ctx, m, d_scrut);
+    if (Typ.eq(ty, Bool)) {
+      let* ty1 = typ_of_dhexp(ctx, m, d1);
+      let* ty2 = typ_of_dhexp(ctx, m, d2);
+      Typ.join_all(~empty=Unknown(Internal), ctx, [ty1, ty2]);
```
cyrus-: no join here, they should be equal

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,288 @@
...
+      let* _ = typ_of_dhexp(ctx, m, d1);
+      let+ _ = typ_of_dhexp(ctx, m, d2);
+      Typ.Unknown(Internal);
+    } else {
+      None;
+    };
+  };
+};
+
+let property_test = (uexp_typ: Typ.t, dhexp: DHExp.t, m: Statics.Map.t): bool => {
```
cyrus-, line 322: it would be good to surface these tests directly around the elaboration function to the user, so we can rapidly fix any issues that come up in elaboration

cyrus- (changes requested):

src/haz3lcore/dynamics/TypeAssignment.re
ruiz-m: @cyrus- What should happen if the types are not equal like in `if true then  else 1.` ? because this is elaborated to a ConsistentCase (actually IfThenElse) but the first case type is Unknown(Internal) and the second case is Float

cyrus-: it should elaborate with a cast to the join type on the branches I believe

src/haz3lcore/dynamics/FilterMatcher.re
tonyfettes: They can be variables. This is a design choice I made a long time ago, so I don't quite remember why I did this. But if this blocks or might have semantic issue I'm happy to change it.

cyrus-: @tonyfettes not blocking but would be good to change this in a separate PR

src/haz3lcore/dynamics/TypeAssignment.re
cyrus-: @ruiz-m this still needs to be fixed I believe

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    switch (mode) {
+    | Syn
+    | SynFun => Some(Unknown(Internal))
+    | Ana(ana_ty) => Some(Typ.normalize(ctx, ana_ty))
+    }
+  | _ => None
+  };
+};
+
+let ground = (ty: Typ.t): bool => {
```
cyrus-, line 26: there should be more ground types than this -- ground products and sums, and I think there is already a function to determine this in the evaluator?

src/haz3lcore/dynamics/TypeAssignment.re
cyrus-: this is still an issue here

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    let* l =
+      env
+      |> ClosureEnvironment.to_list
+      |> List.map(((name, de)) => {
+           let+ ty = typ_of_dhexp(ctx, m, de);
+           Ctx.VarEntry({name, id: Id.invalid, typ: ty});
+         })
+      |> OptUtil.sequence;
+    let ctx' =
+      List.fold_left(
```
cyrus-, line 131: can probably combine the map above and this fold into one, but not critical

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+  | (Tuple(l1), Prod(l2)) =>
+    if (List.length(l1) == List.length(l2)) {
+      List.fold_left2(
+        (acc, dhp, typ) => {dhpat_extend_ctx(dhp, typ, acc)},
+        ctx,
+        l1,
+        l2,
+      );
+    } else {
+      ctx;
```
cyrus-: need to error out in cases like this

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    }
+  | (Ap(Constructor(_, typ), dhp), _) =>
+    let (ty1, ty2) = Typ.matched_arrow(ctx, typ);
+    if (Typ.eq(ty2, ty)) {
+      ctx |> dhpat_extend_ctx(dhp, ty1);
+    } else {
+      ctx;
+    };
+  | _ => ctx
```
cyrus-: don't use a wildcard here

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    Typ.Arrow(ty1, ty2);
+  | Ap(d1, d2) =>
+    let* ty1 = typ_of_dhexp(ctx, m, d1);
+    let* ty2 = typ_of_dhexp(ctx, m, d2);
+    switch (arrow_aux(ty1)) {
+    | Arrow(tyl, tyr) when Typ.eq(tyl, ty2) => Some(tyr)
+    | _ => None
+    };
+  | ApBuiltin(_)
+  | BuiltinFun(_) => None
```
cyrus-: need to have a way of checking and looking up the type of a builtin, not just erroring on uses of builtins

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    if (Typ.eq(ty1, String) && Typ.eq(ty2, String)) {
+      switch (op) {
+      | Concat => Some(Typ.String)
+      | Equals => Some(Typ.Bool)
+      };
+    } else {
+      None;
+    };
+  | ListLit(_, _, ty, _) => Some(List(ty))
+  | Cons(d1, ListLit(_, _, _, [])) =>
```
cyrus-: what is this case for?

cyrus-: ListLits have types so don't need to special case nil...

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    Some(Typ.List(ty1));
+  | Cons(d1, d2) =>
+    let* ty1 = typ_of_dhexp(ctx, m, d1);
+    let* ty2 = typ_of_dhexp(ctx, m, d2);
+    switch (ty2) {
+    | List(ty3) when Typ.eq(ty3, ty1) => Some(ty2)
+    | _ => None
+    };
+  | ListConcat(ListLit(_, _, _, []), d)
+  | ListConcat(d, ListLit(_, _, _, [])) =>
```
cyrus-: see above

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,293 @@
...
+    let* tyd = typ_of_dhexp(ctx, m, d);
+    Typ.eq(tyd, ty1) ? Some(ty2) : None;
+  | FailedCast(d, ty1, ty2) =>
+    if (ground(ty1) && ground(ty2) && !Typ.eq(ty1, ty2)) {
+      let* tyd = typ_of_dhexp(ctx, m, d);
+      Typ.eq(tyd, ty1) ? Some(ty2) : None;
+    } else {
+      None;
+    }
+  | InvalidOperation(d, _) => typ_of_dhexp(ctx, m, d)
```
cyrus-, line 299: this should have the return type of the underlying operation, see above comment

cyrus- (changes requested):

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,301 @@
...
+      d_rules
+      |> List.map((DHExp.Rule(dhp, de)) => {
+           let* ctx = dhpat_extend_ctx(dhp, ty', ctx);
+           typ_of_dhexp(ctx, m, de);
+         })
+      |> OptUtil.sequence;
+
+    switch (typ_cases) {
+    | None => None
+    | Some(_) => Some(Typ.Unknown(Internal))
```
cyrus-: let's actually check to make sure that there is at least one branch with an inconsistent type here

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,301 @@
...
+      equal_typ_list([ty1, ty2]);
+    } else {
+      None;
+    };
+  | IfThenElse(InconsistentIf, d_scrut, d1, d2) =>
+    let* ty = typ_of_dhexp(ctx, m, d_scrut);
+    if (Typ.eq(ty, Bool)) {
+      let* _ = typ_of_dhexp(ctx, m, d1);
+      let* _ = typ_of_dhexp(ctx, m, d2);
+      Some(Typ.Unknown(Internal));
```
cyrus-: also here -- let's actually check to make sure that there is at least one branch with an inconsistent type

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,301 @@
...
+    | Ap(Constructor(_, typ), dhp) =>
+      let (ty1, ty2) = Typ.matched_arrow(ctx, typ);
+      Typ.eq(ty2, ty) ? dhpat_var_entry(dhp, ty1) : None;
+    | EmptyHole(_)
+    | NonEmptyHole(_)
+    | Wild
+    | ExpandingKeyword(_)
+    | InvalidText(_)
+    | BadConstructor(_)
+    | IntLit(_)
```
cyrus-: we should type check in these cases

cyrus- (changes requested):

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,322 @@
...
+    typ_of_dhexp(ctx, m, db);
+  | FixF(name, ty1, d) =>
+    let entry = Ctx.VarEntry({name, id: Id.invalid, typ: ty1});
+    typ_of_dhexp(Ctx.extend(ctx, entry), m, d);
+  | Fun(dhp, ty1, d, _) =>
+    let* ctx = dhpat_extend_ctx(dhp, ty1, m, ctx);
+    let* ty2 = typ_of_dhexp(ctx, m, d);
+    Some(Typ.Arrow(ty1, ty2));
+  | TypFun({term: Var(name), _}, d, _) =>
+    let* ty = typ_of_dhexp(ctx, m, d);
```
cyrus-, line 162: this needs to add the type variable to the context

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -0,0 +1,322 @@
...
+    Some(Typ.Forall(name, ty));
+  | TypFun(_, d, _) =>
+    let* ty = typ_of_dhexp(ctx, m, d);
```
cyrus-, line 165: when would this case come up?

cyrus- (changes requested):

src/haz3lcore/dynamics/Elaborator.re
cyrus-: @ruiz-m ^

cyrus- (changes requested):
This looks good for a merge now, can you just remove the remaining prints that ended up in here?

Issue #1120, opened 2023-10-11
Add if-then-else directly to internal language


cyrus-:
Currently `if e1 then e2 else e3` desugars to the following during elaboration:

```
case e1
  | true => e2
  | false = e3
end
```

This causes confusion when there is an indeterminate result, e.g. `if ? then 1 else 2`. For that reason, it would be better to directly add if-then-else to the internal language (DHExp) instead of desugaring to case.

PR #1122, opened 2023-10-18
add IfThenElse to DHExp.t

LarWong:
#1120


Negabinary (approved):
lgtm!

src/haz3lcore/dynamics/DH.re
```diff
@@ -244,6 +254,11 @@ module rec DHExp: {
       fast_equal(d1, d2) && reason1 == reason2
     | (ConsistentCase(case1), ConsistentCase(case2)) =>
       fast_equal_case(case1, case2)
+    | (IfThenElse(d10, d11, d12, d13), IfThenElse(d20, d21, d22, d23)) =>
+      d10 == d20
```
cyrus-: minor: use `d` variable name prefix for DHExp's only, not the boolean flag

src/haz3lcore/dynamics/DH.re
```diff
@@ -36,6 +36,7 @@ module rec DHExp: {
     | Cast(t, Typ.t, Typ.t)
     | FailedCast(t, Typ.t, Typ.t)
     | InvalidOperation(t, InvalidOperationError.t)
+    | IfThenElse(bool, t, t, t) // use bool tag to track if branches are consistent
```
cyrus-: instead of a bool, define a sum type that makes clear what it represents:

```
type consistent = 
  | Consistent
  | Inconsistent
```

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -242,17 +243,15 @@ let rec dhexp_of_uexp =
...
         let+ d2 = dhexp_of_uexp(m, e2);
-        let d_rules =
-          DHExp.[Rule(BoolLit(true), d1), Rule(BoolLit(false), d2)];
-        let d = DHExp.Case(d_scrut, d_rules, 0);
+        // Use tag to mark inconsistent branches
         switch (err_status) {
         | InHole(Common(Inconsistent(Internal(_)))) =>
-          DHExp.InconsistentBranches(id, 0, d)
-        | _ => ConsistentCase(d)
+          DHExp.IfThenElse(false, c', d1, d2)
```
cyrus-: see above -- the comment seemed to indicate inconsistent should be `true` but it is false here. make sure this is doing what you intend.

src/haz3lcore/dynamics/PatternMatch.re
```diff
@@ -266,7 +266,8 @@ and matches_cast_Sum =
   | Sequence(_, _)
   | Closure(_)
   | Cons(_)
-  | ListConcat(_) => DoesNotMatch
+  | ListConcat(_)
+  | IfThenElse(_) => DoesNotMatch
```
cyrus-: elimination forms like if/then/else (and case, prj) should actually be IndetMatch. can you fix that in this PR for all of these?

src/haz3lcore/dynamics/PatternMatch.re
```diff
@@ -356,6 +357,7 @@ and matches_cast_Tuple =
   | NonEmptyHole(_) => IndetMatch
   | FailedCast(_, _, _) => IndetMatch
   | InvalidOperation(_) => IndetMatch
+  | IfThenElse(_) => DoesNotMatch
```
cyrus-: see above

src/haz3lcore/dynamics/PatternMatch.re
```diff
@@ -491,4 +493,5 @@ and matches_cast_Cons =
...
```
cyrus-: see above

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -313,6 +315,32 @@ let rec mk =
        };
        */
 
+      | IfThenElse(_, c, d1, d2) =>
+        let c_doc = (~enforce_inline) => mk_cast(go(~enforce_inline, c));
+        let d1_doc = (~enforce_inline) => mk_cast(go(~enforce_inline, d1));
+        let d2_doc = (~enforce_inline) => mk_cast(go(~enforce_inline, d2));
+        hcats([
+          DHDoc_common.Delim.mk("("),
```
cyrus-: parenthesization should be handled by the parent expression

cyrus- (changes requested):

Negabinary:
I've just noticed there's something strange going on with casting on this branch as well
image

Compared with other branches, the elaboration appears to have an extra `?` cast at the end of the if-then-else block

image

I don't know enough about casting to know where that's come from or why it's a problem though

cyrus-:
@disconcision can you take a look at the cast issue above, ideally today if you can? (we need to release A1 which uses this, I can take a look if not)

disconcision:
@cyrus- IfThenElse appears to have been added to the 'Normal cases: wrap' section (which gets casts) as opposed to the 'Forms with special ana rules but no particular typing requirements' which includes Case. I've moved it to the latter and that resolves @Negabinary's example above

cyrus- (approved):

PR #1123, opened 2023-10-19
analytic constructor casting fix

### LINKED ISSUE #1119: Casting bug at elaboration
When elaborating this expression

image

The empty hole inside the parentheses is casted twice from Unknown(Internal) to Arrow(Unknown(Internal), Unknown(Internal))

image

I believe that this is caused when elaboration pattern matches to the parentheses and then calls dhexp_of_uexp to match with the empty hole. When the wrap function is called the casting is performed and returns to the call position of the parentheses case expression which then also calls the wrap function and casts a second time

ruiz-m:
Bug 1
When performing uexp_to_info on this expression
image

The constructor Yo in the function position of the Ap is analyzed with Arrow(Unknown(Internal), Sum["Yo", None]).
In of_ctr_in_ap, the mode given is Arrow(Unknown(Internal), Sum["Yo", None])

When the expression is elaborated

image

However, when I perform the type assignment. the resulting type is None because the type inside the outer cast is evaluated to be Sum["Yo", None] instead of Unknown(Internal).
This problem seems to be a compatibility issue with the type assignment module.
I initially came up with a solution, but i recently realized that it was not correct.
The current solution by editing of_ctr_in_ap works with type assignment, but it may require further discussion.


disconcision (commented):
So for the Mode.re changes, these cause the following observable changes in Examples/ADT Statics:

1. a. in dev:
2023-10-30-15-59-firefox_FR4baej5xt
1. b: in casting-arrow-fixes:
2023-10-30-16-00-firefox_6k5golSS3S

The reason I instituted the behavior in (a) was to try and make the analytic pseudo-scope for constructors resemble their synthetic scope behavior. Your changes makes it so that nullary constructors defined in a type annotation won't be considered 'in scope' if they are used in their pattern as unary constructors; instead, such uses will be either unbound, or as is the case here, refer to their previous definition. Similarly:

2. a. in dev: 
2023-10-30-15-58-firefox_PhRRwjDDx1
2. b: in in casting-arrow-fixes: 
2023-10-30-15-59-firefox_XN2KYxEb4i

I don't have a strong feeling about this kind of behavior one way or the other... I generally didn't feel great about this special case, but the new behavior feels potentially confusing to me as well. @cyrus- any opinion?

But can you clarify for me exactly what Bug 1 is? I don't totally follow what exactly goes wrong in DHExp type assignment in this case?

Also @ruiz-m can you clarify which changes here are addressing which bug? The first elaborator change is obviously addressing #1119 aka bug 2, and the Mode change bug 1, but I'm not sure about the second elaborator change i.e. changing the FixF type annotation to ty_body? is this actually load-bearing, I didn't think anything was being done with this type?

ruiz-m:
Based on the example let_ : +Yo = Yo("lol") in, the type assignment module would first take the type of nonempty hole by returning the type that is being analyzed for the empty hole which would be ? -> Sum(["Yo", None]). The type for cast("lol", String, ?) would be ? because "lol" has type String and is casted with no issues. Since the expression in the definition of the let expression is Yo("lol") and Yo has type ? -> Sum(["Yo", None]) and the type inside the parameter is ?, the resulting type of the Ap expression is Sum(["Yo", None]). However, there is a cast outside the Ap expression from ? to Sum["Yo", None] which results in a casting error because ? and the type of Ap expression are not equal. That particular cast causes the problem and type assignment fails.

276480637-38ab2780-34da-4835-8ea7-3c281b0e685e

cyrus-:
Marking as draft for now as you work through the details here.

src/haz3lcore/statics/Statics.re
```diff
@@ -242,9 +242,13 @@ and uexp_to_info_map =
...
-      Id.is_nullary_ap_flag(arg.term.ids)
-      && !Typ.is_consistent(ctx, ty_in, Prod([]))
-        ? BadTrivAp(ty_in) : Just(ty_out);
+      switch (fn.ty) {
+      | Sum(_) => BadSumAp(fn.ty)
+      | _ =>
+        Id.is_nullary_ap_flag(arg.term.ids)
+        && !Typ.is_consistent(ctx, ty_in, Prod([]))
+          ? BadTrivAp(ty_in) : Just(ty_out)
+      };
```
disconcision: lets split this out into a fn called Self.of_ap

src/haz3lcore/statics/Statics.re
```diff
@@ -464,7 +468,12 @@ and upat_to_info_map =
     let (fn, m) = go(~ctx, ~mode=fn_mode, fn, m);
     let (ty_in, ty_out) = Typ.matched_arrow(ctx, fn.ty);
     let (arg, m) = go(~ctx, ~mode=Ana(ty_in), arg, m);
-    add(~self=Just(ty_out), ~ctx=arg.ctx, m);
+    let self: Self.t =
```
disconcision: this can be a call to the same fn as the exp version maybe

ruiz-m:
Screenshot from 2024-01-31 11-28-46

Demonstration that the unwanted cast is removed with this solution.

disconcision (changes requested):
I feel like I'm not sold on the changes in error attribution / error messages here, independent of the casting situation. From a consistency perspective all other such errors are attributed to function position, not the application itself; from the user perspective it's not obvious what this is a special case. Same for the error message, which reads as over-specialized (if I wrote `1(2)` it doesn't get an "application of Int type" error).

Also (this is maybe orthogonal) it's odd that when you put the caret on the constructor itself, it claims nothing is wrong:
Screenshot 2024-01-31 at 5 51 19 PM

@ruiz-m are we sure there isn't a way to resolve the casting issue while keeping the existing attribution in place?

ruiz-m:
@disconcision I will see if there could be another way to resolve the casting issue.

cyrus- (approved):

disconcision (approved):

PR #1124, opened 2023-10-20
Case expression redundancy checking

pigumar1:
This pull request aims to implement the redundancy checking feature mentioned in [Live Pattern Matching with Typed Holes](https://victoryyw.github.io/assets/pdfs/pattern.pdf).

~~`Incon.is_redundant` has already been implemented in #1094.~~

_ It seems that currently, there is no way to wrap rules with error holes. In our first pattern-matching group meeting, professor Omar mentioned that we can **wrap patterns instead of rules** for now, and this pull request follows this approach. _

_This means that there can be two errors for one pattern at the same time. For example, consider this example:
image
The pattern `[]` is erroneous because it's inconsistent with the scrutinee type, and it's definitely going to be redundant as well. We will re-apply the same technique discussed in #1114 to resolve this.
_


pigumar1:
Dynamic layout: #1125

cyrus- (approved):
This looks good. Will hold off merging until exhaustiveness checking issues are resolved.

PR #1125, opened 2023-10-22
Dynamic layout for inexhaustive case/indeterminately match case

pigumar1:
This pull request implements the functionality of visually crossing out mismatched rules during evaluation. It's possible for the evaluation of an **indeterminately exhaustive case expression** to get stuck when we cannot determine whether a rule is matched or mismatched. In this case, the rules preceding this rule are crossed out.
image


src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -271,7 +272,11 @@ let rec dhexp_of_uexp =
         | InHole(
             InexhaustiveMatch(Some(Common(Inconsistent(Internal(_))))),
           ) =>
+          // TODO: review
```
pigumar1, line 276: Not problematic for now since currently `DHExp.InexhaustiveCase` behaves exactly the same as `DHExp.InconsistentBranches`

src/haz3lcore/dynamics/DH.re
```diff
@@ -33,6 +33,7 @@ module rec DHExp: {
     | Prj(t, int)
     | Constructor(string)
     | ConsistentCase(case)
+    | InexhaustiveCase(MetaVar.t, HoleInstanceId.t, case)
```
cyrus-, line 36: Instead of having three different case constructs in DHExp, let's merge them into one with an additional flag parameter for when there is an error.

cyrus- (changes requested):
see comment -- I think we can unify the case constructs here rather than implementing a third variant

cyrus-:
@pigumar1 I don't recall the precise purpose of this PR -- what is it and what is the status?

pigumar1:
> @pigumar1 I don't recall the precise purpose of this PR -- what is it and what is the status?

@cyrus- I have updated the purpose of this PR at the top. The expected behavior has already been implemented, but some code simplifications can be done, as suggested by the comment above. #1154 was introduced for this code simplification, but it has its own relevant concerns, which are mentioned at the top of that PR.

cyrus-:
archiving since this is no longer an active project, but thanks for the design @pigumar1 !

PR #1132, opened 2023-11-05
Empty hole casting fix

### LINKED ISSUE #1119: Casting bug at elaboration

ruiz-m:
Bug
Closes #1119


src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -33,7 +33,11 @@ let cast = (ctx: Ctx.t, mode: Mode.t, self_ty: Typ.t, d: DHExp.t) =>
   | SynFun =>
     switch (self_ty) {
     | Unknown(prov) =>
-      DHExp.cast(d, Unknown(prov), Arrow(Unknown(prov), Unknown(prov)))
+      switch (d) {
+      | Cast(_) => d
```
cyrus-: this seems like too general of a fix to handle the specific case of empty holes in function position having casts on them. empty holes should always take on the expected type, and in function position that would be ? -> ?.

cyrus- (changes requested):

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -33,7 +33,15 @@ let cast = (ctx: Ctx.t, mode: Mode.t, self_ty: Typ.t, d: DHExp.t) =>
   | SynFun =>
     switch (self_ty) {
     | Unknown(prov) =>
-      DHExp.cast(d, Unknown(prov), Arrow(Unknown(prov), Unknown(prov)))
```
disconcision: let's add a comment here providing an explicit example of why this case is necessary (eg from the issue)

disconcision (approved):
@ruiz-m in the original bug report (#1119), and when we talked about it, you thought parentheses were involved here? did that turn out to be a red herring?

anyway I'm fine with this fix, as it's relatively lightweight and solves the issue, but I can't say i'm 100% clear on why this is correct, so it's probably worth planting a few words here or in a comment about the rational.

ruiz-m:
@disconcision yes, the issue turned out to be a combination with SynFun and parentheses

In the bug, the parentheses were in the function position of Ap and because of those parentheses the wrap function does another cast from ? to ? -> ? on the empty hole. That is why I made the fix in the SynFun section of the cast function

cyrus- (approved):

PR #1138, opened 2023-11-14
Make scroll to caret more selective

### LINKED ISSUE #1085: be more selective about scroll-to-caret
Select-all doesn't scroll to caret in vscode and probably shouldn't here. Especially true if quadruple-click selects all, as there is a significant possibility of accidental quadclick and incumbent context loss

Negabinary:
This fixes #1085. It was getting too annoying on the stepper.

### DIFF
--- a/src/haz3lcore/zipper/action/Perform.re
+++ b/src/haz3lcore/zipper/action/Perform.re
@@ -17,6 +17,24 @@ let is_write_action = (a: Action.t) => {
   };
 };
 
+let should_scroll_to_caret = (a: Action.t) => {
+  switch (a) {
+  | Move(_)
+  | MoveToNextHole(_)
+  | Jump(_)
+  | Select(Resize(_))
+  | Select(Term(_))
+  | Destruct(_)
+  | Insert(_)
+  | Pick_up
+  | Put_down
+  | RotateBackpack
+  | MoveToBackpackTarget(_) => true
+  | Unselect
+  | Select(All) => false
+  };
+};
+
 let go_z =
     (~meta: option(Editor.Meta.t)=?, a: Action.t, z: Zipper.t)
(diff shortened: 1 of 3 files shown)


Negabinary:
@disconcision do you understand what happened with the checks here?

cyrus-:
Sometimes it is weird if two pushes come in too quickly in succession? I thought I had fixed that but seems not perfectly. You can just re-run the job if that happens (already did that in this case).

disconcision:
is there a reason this couldn't be implemented as a State variable triggered via scheduled_action? as opposed to threading queue_scroll_to_caret through several things?

also i think the logic mapping actions to queue_scroll_to_caret() should probably be factored out ala reevaluate_post_update as opposed to called in in update on a per-case basis, better separation of concerns

disconcision:
I'm a bit confused about what actions queue_scroll_to_caret is called in too... like why is it called in SetShowBackpackTargets?

disconcision:
@Negabinary could you take a look at this diff? https://github.com/hazelgrove/hazel/pull/1024/commits/07aeb0b833cd432d14c276634f7f5b5e74d133ab
i removed the threading and just wrote a single predicate (Update.should_scroll_to_caret) which is called in Main. does this approach miss anything?

cyrus-:
@disconcision can you make a separate PR for that to keep track better?

disconcision:
@cyrus- there's a substantial intersection between this and TyDi, so that would require mostly redoing the change from scratch. this was an attempt to do it quickly and avoid blocking a TyDi merge. i can separate it out, but I might not get to it for a bit.

disconcision:
@cyrus- unless you mean just separate that last commit as a post-tydi PR, in which case i can do that rn

disconcision:
@cyrus- created new PR on top of TyDi #1149

PR #1140, opened 2023-11-16
Evaluator rewrite for stepper

Negabinary:
This is a re-write of the evaluator that changes it to be defined in terms of small steps .

Key changes:

* Small-step semantics is now defined in `Transition.re` separately to the evaluator in `Evaluator.re` which decides what to do with those small steps.
* EvaluatorState now uses references instead of monads

Very quick benchmarking:
 
* Stack overflows when calculating factorial of 290 on eval instead of 190 on dev
* fib(21) takes 888ms on eval and 1420ms on dev


cyrus-:
minor note: you don't need to include `haz3l` in branch names anymore

src/haz3lcore/dynamics/Transition.re
```diff
@@ -0,0 +1,557 @@
...
+    | Sequence(d1, d2) =>
+        let. _ = otherwise(d1 => Sequence(d1, d2))
+        and. _ = req_final(req(state, env), 0, d1);
+        Step({apply: () => d2, kind: Sequence, final: false});
+
+
+    Each step semantics starts with a `let. () = otherwise(...)` that defines how
+    to wrap the expression back up if the step couldn't be evaluated.
+
+    This is followed by a series of `and. _ = req_final(req(state, env), <i>, <d>)`
```
cyrus-: say what <i> should be here as well

src/haz3lcore/dynamics/Transition.re
```diff
@@ -0,0 +1,557 @@
...
+      Constructor;
+    | Prj(d1, n) =>
+      let. _ = otherwise(d1 => Prj(d1, n))
+      and. d1' = req_final(req(state, env), 0, d1);
+      Step({
+        apply: () =>
+          switch (d1') {
+          | Tuple(ds) when n < 0 || List.length(ds) <= n =>
+            // TODO(Matt): This is iconsistent - we don't deal with any other static error like this
+            InvalidOperation(d1', InvalidOperationError.InvalidProjection)
```
cyrus-: yeah you can go ahead and make this consistent with other such ill-typed states in this PR

cyrus- (changes requested):

src/haz3lcore/dynamics/Transition.re
Negabinary: sorry, I'm not sure I understand what you mean here

cyrus-: oh sorry GitHub formatting got weird, I wanted to say to include a description of the `<i>` parameter

cyrus- (approved):

src/haz3lcore/dynamics/Transition.re
```diff
@@ -0,0 +1,560 @@
...
+        and. _ = req_final(req(state, env), 0, d1);
+        Step({apply: () => d2, kind: Sequence, final: false});
+
+
+    Each step semantics starts with a `let. () = otherwise(...)` that defines how
+    to wrap the expression back up if the step couldn't be evaluated.
+
+    This is followed by a series of `and. d1' = req_final(req(state, env), <i>, <d1>)`
+    which indicate that in order to evaluate the step, <d1> must be final. (req_value
+    is also available if it needs to be a value). Note that if successful, d1' hold
```
cyrus-: "hold contain" -> "will be"

cyrus-:
just need those two comment fixes and then should be gtg unless @disconcision wants to point anything out

src/haz3lcore/dynamics/Transition.re
```diff
@@ -0,0 +1,560 @@
...
+      and. _ = req_all_final(req(state, env), 0, ds);
+      Constructor;
+    | Prj(d1, n) =>
+      let. _ = otherwise(d1 => Prj(d1, n))
+      and. d1' = req_final(req(state, env), 0, d1);
+      Step({
+        apply: () =>
+          switch (d1') {
+          | Tuple(ds) when n < 0 || List.length(ds) <= n =>
+            print_endline("InvalidProjection");
```
disconcision: lets get rid of these (i assume legacy) redundant print_endlines strewn about. i think these got put here because the console message for raise was sub-optimal, but in the current implementation these exceptions get caught and surfaced in the UI anyway

src/haz3lcore/dynamics/Builtin.re
```diff
@@ -1,39 +1,18 @@
...
-type evaluate =
-  (ClosureEnvironment.t, DHExp.t) => EvaluatorMonad.t(EvaluatorResult.t);
-
+type evaluate = (ClosureEnvironment.t, DHExp.t) => DHExp.t;
 [@deriving (show({with_path: false}), sexp, yojson)]
 type args = list(DHExp.t);
-
 [@deriving (show({with_path: false}), sexp, yojson)]
-type builtin_evaluate =
-  (ClosureEnvironment.t, args, evaluate) =>
```
disconcision, line 15: why was there an environment here before do you know?

src/haz3lcore/dynamics/Transition.re
```diff
@@ -0,0 +1,560 @@
...
+    which indicate that in order to evaluate the step, <d1> must be final. (req_value
+    is also available if it needs to be a value). Note that if successful, d1' hold
+    contain the fully-evaluated version of d1.
+
+    Finally, we have the Step construct that defines the actual step. Note "Step"s
+    should be used if and only if they change the expression. If they do not change
+    the expression, use `Constructor` or `Indet`.
+
+    The step defines firstly, a `() => ...` function giving the result of the step,
+    secondly a `kind`, that describes the step (which will be used in the stepper)
```
disconcision, line 33: can you remind me why these are thunked?

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -1,1280 +1,112 @@
...
-  evaluate(env, d, es);
+  let result = evaluate(state, env, d);
+  let result =
+    switch (result) {
+    | BoxedValue(x) => BoxedValue(x)
+    | Indet(x) => Indet(x)
+    | Uneval(x) => Indet(x)
+    };
+  (state^, result);
 };
```
disconcision, line 112: Evaluator so beautiful now 😭

disconcision (approved):
lgtm

src/haz3lcore/dynamics/Transition.re
Negabinary: two reasons:
1. so that the test side effect happens at the right time
2. avoiding unexpected delays from "speculative execution" - perhaps not a problem atm (though there might be some big data structure cases where this is a problem); but when we start combining steps in the single-stepper this will be more of an issue

src/haz3lcore/dynamics/Builtin.re
Negabinary: no idea sorry

src/haz3lcore/dynamics/Transition.re
Negabinary: done

src/haz3lcore/dynamics/Builtin.re
tonyfettes: I believe this was added by me or @DavidFangWJ so that these are no-longer a list of `let` injected to user program. Feel free to cc me if I'm the author listed on the git blame

Negabinary: @tonyfettes do you remember which builtin function it was that needed access to the environment / lets?

tonyfettes: I mis-understood the code here, which means this is not written by me. Sorry for making the mistake...

PR #1143, opened 2023-11-19
Labeled Tuples

WondAli:
Goal: implement Labeled Tuples in Hazel.

Note: current branch is a transfer of progress from the haz3l-labeled-tuples branch to a new branch, that is updated with the current dev branch (11/19/2023) and separated from the module system branch, since I branched off of the module system by accident in the beginning (probably because of dot operators, which are planned to be implemented with labeled tuples in the future).
I will be working on this branch from now on.


WondAli:
See #1235 for new implementation

PR #1144, opened 2023-11-22
Letting constraint emission for ADTs based on scrutinee type

pigumar1:
The current implementation of case expression exhaustiveness / redundancy checking (#1114) behaves incorrectly when the scrutinee type is an anonymous sum. This is due to the fact that the current implementation looks up the constructor information needed to generate constraints for ADTs (specifically, `num_variants` and `nth`) **in the Ctx of the pattern**, while it seems that anonymous sums don't have their constructor information kept track of in Ctxs.

This pull request aims to change the implementation so that `num_variants` and `nth` are **acquired from the provided scrutinee type instead**. 
image

In fact, this new implementation does not need to modify `Typ.Sum` at all. `num_variants` is equal to the size of the `sum_map`, and `nth` can also be acquired easily. Note that `nth` for each constructor should be the same even if the order they are declared is changed, so sorting might be involved. In this case, to reduce time complexity, it might be helpful to **sort the sum_map beforehand** or **memorize some information, like nth**.

The next issue is where should we obtain the scrutinee type. Ideally, this should be done in `Statics.upat_to_info_map`, in which constraints for other patterns are generated. `Statics.upat_to_info_map` has `is_synswitch`, `ctx`, `ancestors`, `mode`, `upat`, and `m` as inputs, and `mode` is the only one that might carry the scrutinee type, so it's going to be used.

The rest of the code can remain unchanged.


src/haz3lcore/dynamics/Constraint.re
```diff
@@ -119,15 +119,23 @@ let rec ctr_of_nth_variant = (num_variants, nth): (t => t) =>
...
-    | Some({num_variants, nth, _}) =>
-      arg |> ctr_of_nth_variant(num_variants, nth)
+    switch (mode) {
+    | Mode.Ana(ty) =>
+      switch (Typ.weak_head_normalize(ctx, ty)) {
+      | Sum(map) =>
+        let num_variants = ConstructorMap.cardinal(map);
+        switch (ConstructorMap.nth(name, map)) {
+        | Some(nth) => arg |> ctr_of_nth_variant(num_variants, nth)
+        | None => Falsity // TODO: review
```
pigumar1: The variant with the specified name is not found

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -119,15 +119,23 @@ let rec ctr_of_nth_variant = (num_variants, nth): (t => t) =>
...
+        };
+      | _ => Falsity // TODO: review
+      }
+    | _ => Falsity // TODO: review
```
pigumar1: For `Mode.Syn` / `Mode.SynFun`, the constraint associated with an `Ap` pattern is not used.
An example in which an `Ap` pattern synthesizes a type is shown below:

image

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -119,15 +119,23 @@ let rec ctr_of_nth_variant = (num_variants, nth): (t => t) =>
...
+    switch (mode) {
+    | Mode.Ana(ty) =>
+      switch (Typ.weak_head_normalize(ctx, ty)) {
+      | Sum(map) =>
+        let num_variants = ConstructorMap.cardinal(map);
+        switch (ConstructorMap.nth(name, map)) {
+        | Some(nth) => arg |> ctr_of_nth_variant(num_variants, nth)
+        | None => Falsity // TODO: review
+        };
+      | _ => Falsity // TODO: review
```
pigumar1: This constraint value is never used. It will be replaced with `Hole` after error checkings, since the pattern doesn't have the type of the scrutinee.

src/haz3lcore/dynamics/Constraint.re
```diff
@@ -119,15 +119,23 @@ let rec ctr_of_nth_variant = (num_variants, nth): (t => t) =>
...
-    switch (Ctx.lookup_ctr(ctx, name)) {
-    | None => Falsity // TODO: review
-    | Some({num_variants, nth, _}) =>
-      arg |> ctr_of_nth_variant(num_variants, nth)
+    switch (mode) {
+    | Mode.Ana(ty) =>
+      switch (Typ.weak_head_normalize(ctx, ty)) {
+      | Sum(map) =>
+        let num_variants = ConstructorMap.cardinal(map);
+        switch (ConstructorMap.nth(name, map)) {
```
pigumar1: The postion of the ctr specified with `name` in a **sorted list** of ctrs associated with the sum type of the scrutinee.

cyrus-:
@pigumar1 there is a conflict that needs to be resolved

src/haz3lcore/statics/ConstructorMap.re
```diff
@@ -99,3 +99,12 @@ let rec is_ground = (is_ground_value: 'a => bool, map: t('a)): bool =>
   | [(_, head), ...tail] =>
     is_ground_value(head) && tail |> is_ground(is_ground_value)
   };
+
+let nth = (ctr: Constructor.t, map: t('a)): option(int) => {
```
cyrus-: minor: by convention in OCaml, functions like this would have the map as the first argument.

src/haz3lcore/dynamics/Constraint.re
cyrus-: the TODO comments can be removed

src/haz3lcore/dynamics/Constraint.re
cyrus-: we do still need to do exhaustiveness checking for patterns in lets and functions, so the type to be used here can be the synthesized type (which in the case of constructor applications I believe is just the singleton sum)

cyrus- (changes requested):

pigumar1:
@cyrus- gotcha! Now constraint emission works for `Ap`s regardless of either it's analyzed against a type or synthesizes a type.
In this example `A` emits InjL(T) and `C(A)` emits InjL(InjL(T))
image

cyrus-:
going to merge this into parent branch, see that PR for discussion about exhaustiveness checking in let/function position (those are called irrefutable positions since there is only one pattern that can be matched, so it better by itself be exhaustive)

PR #1149, opened 2023-12-04
Scroll to caret refactor

disconcision:


Negabinary (approved):
Ah yeah this is a nice way to avoid the threaded function; I think we should merge it for now.

I've been beginning to think this is a still a temporary solution though because there might be things that aren't the caret that we want to scroll to. (e.g. on the stepper, we want to scroll to any newly-created step). This'll need some way of identifying what the "active" element is, and I'm not sure what that would look like.

PR #1153, opened 2023-12-14
Clean up Builtin code

Negabinary:
Changes I've made:

 - The previous Builtins supported Builtin's with multiple arguments, while only ever being passed a single argument (sometimes a tuple). Support for multiple arguments appears to have degraded over time and we have no intention of currying functions, so I have removed this.
 - The previous Builtins wrapped up BuiltinAps in function wrappers, I have replaced these with an explicit BuiltinFun construct so it's easier to pattern match on whether it's a wrapped builtin or a meaningful function.


disconcision (approved):
lgtm!

PR #1155, opened 2023-12-26
Type Hole Inference (post merge)

RaefM:


RaefM:
I've opened this as all core functionality from the paper is complete and seems to be working as per my testing.
Will make additional passes over this week for cleanup or stylistic changes (and address comments/requests as added).
Thanks!

cyrus-:
Nice work! Going to start with a functionality pass before doing code review...

The formatting here is a bit borked: 
image

cyrus-:
Getting odd behavior with ADTs...

image

RaefM:
> Getting odd behavior with ADTs...
> 
> image

Oof I could've sworn I tested that exact case recently and had it working...
I'll work on fixing those tonight

cyrus-:
the hole that should be after `=` has shifted left here 

image

cyrus-:
Tab doesn't seem to work to accept the suggestion...

image

RaefM:
> Tab doesn't seem to work to accept the suggestion...
> 
> image

Suggestion acceptance is currently bound to enter; do you want only tab or should I include both?

cyrus-:
> > Tab doesn't seem to work to accept the suggestion...
> > image
> 
> Suggestion acceptance is currently bound to enter; do you want only tab or should I include both?

I think with @disconcision 's TyDi autocomplete bound to Tab it makes sense for this to also be bound to Tab. Thoughts @disconcision ?

RaefM:
> > Getting odd behavior with ADTs...
> > image
> 
> Oof I could've sworn I tested that exact case recently and had it working... I'll work on fixing those tonight

Fixed issue:
image
image
Precendence of types to be preferred for a "solved" annotation are 
Unknown < ADT Var < Anything else

Will continue to review other issues now

RaefM:
> the hole that should be after `=` has shifted left here
> 
> image

Update: fixed this and other mentioned issues; will work on more testing and code cleanup now

cyrus-:
@RaefM can you have it so tab accepts the completion even on the left side? currently seems to only work on right side, it moves to the right when on the left...

cyrus-:
image

two issues:
* curious why the input type on the type annotation on the let doesn't find the Int? 
* seem to be missing the hole in the cursor inspector view

cyrus-:
Maybe a comment for @disconcision but in the following, the tydi autocomplete inserts `::` which is a bit unexpected here. I know why it's doing that, but can we override this somehow?

image

cyrus-:
General thought: we should probably use the inferred types during elaboration, rather than leaving the types unknown. How hard would that be?

cyrus-:
The inferred type here confused me... B is not applied so I was expecting an error.

image

disconcision:
> Maybe a comment for @disconcision but in the following, the tydi autocomplete inserts `::` which is a bit unexpected here. I know why it's doing that, but can we override this somehow?

straightforward to disable hackily.

been annoyed by this one too. trying to think abstractly about systematic criteria to filter this instead special-casing; it's using a refutable pattern in an irrefutable context. could downgrade these, or a least downgrade them if the expected type isn't literally list. actually that second bit should probably be treated as an independent downgrade as we'd probably want the same in refutable patterns too. 

in any case to be done properly it needs some kind of ranking system

RaefM:
> The inferred type here confused me... B is not applied so I was expecting an error.
> 
> image

This was due to a found occurs check failure getting masked by UI filters I made (solved as T and Int -> T). Pushed fix

New results:

image
image

RaefM:
> @RaefM can you have it so tab accepts the completion even on the left side? currently seems to only work on right side, it moves to the right when on the left...

Fixed in latest commit

RaefM:
> image
> two issues:
> 
> * curious why the input type on the type annotation on the let doesn't find the Int?
> * seem to be missing the hole in the cursor inspector view

Issues fixed:
1) Was caused due to missing subsumption constraints in Fun. Fixed in latest commits
image
2) The unknown case in type view was sometimes set to nothing mistakenly. Added ? as a fallback and also changed all the CI Type.view invocations to supply arguments that should make it so the hexagon is always shown instead of ? for consistency with the editor. Example with CI non unification message:
image

cyrus-:
> > Maybe a comment for @disconcision but in the following, the tydi autocomplete inserts `::` which is a bit unexpected here. I know why it's doing that, but can we override this somehow?
> 
> straightforward to disable hackily.
> 
> been annoyed by this one too. trying to think abstractly about systematic criteria to filter this instead special-casing; it's using a refutable pattern in an irrefutable context. could downgrade these, or a least downgrade them if the expected type isn't literally list. actually that second bit should probably be treated as an independent downgrade as we'd probably want the same in refutable patterns too.
> 
> in any case to be done properly it needs some kind of ranking system

Not suggesting things that would lead to a refutable pattern seems reasonable in this context. Also fine with just special casing this for now if that's easier. Can you make a quick PR?

cyrus-:
Not sure what this cursor inspector message means?

image

disconcision:
@cyrus #1162

RaefM:
> Not sure what this cursor inspector message means?
> 
> image

That text pops up when the occurs check fails (so whenever a pts is a proper subset of itself, eg being solved as T and Int -> T)

I'm definitely open to other text there, I figured 'occurs check failed' doesn't tell all that much and had gone with 'inferred type refers to self' to indicate the solution set recurses on itself and is thus not valid

cyrus-:
> > Not sure what this cursor inspector message means?
> > image
> 
> That text pops up when the occurs check fails (so whenever a pts is a proper subset of itself, eg being solved as T and Int -> T)
> 
> I'm definitely open to other text there, I figured 'occurs check failed' doesn't tell all that much and had gone with 'inferred type refers to self' to indicate the solution set recurses on itself and is thus not valid

Can we show the two possibilities here? Why is it only showing Int -> T?

RaefM:
> > > Not sure what this cursor inspector message means?
> > > image
> > 
> > 
> > That text pops up when the occurs check fails (so whenever a pts is a proper subset of itself, eg being solved as T and Int -> T)
> > I'm definitely open to other text there, I figured 'occurs check failed' doesn't tell all that much and had gone with 'inferred type refers to self' to indicate the solution set recurses on itself and is thus not valid
> 
> Can we show the two possibilities here? Why is it only showing Int -> T?

Sorry for the wall of text; tldr, fixed with new behavior illustrated below

So after unification, it's solved as pts = {?, T, Int ->(pts)}. The occurs check as implemented makes it so that becomes {?, T, Int -> ?} and the general filtering rule for making suggestions out of ptses is to prefer 'non-nodes' over nodes if possible (which is anything that isn't a plain var or hole)- that filters it down to Int -> ?

The 'solution' in this case would be the set {T, Int -> T, Int -> Int -> T, ...}. Our filtering rules reads to Int -> T, where the issue is detected, and replaces T with ?.

Before ADTs were added, so when the occurs failure could only happen on holes (eg ? or Int -> ?), saying ? was always useless so we'd just suggest Int -> ? while marking occurs failing ptses as unsolved. With type variables, (T vs Int -> T) the two possibilities are each valid suggestions as they are inconsistent with one another. 

To facilitate this, changing the filtering so that when it comes to type variables, keep at least one when filtering the pts if the occurs check has failed. When occurs hasn't failed, still filtering all type variables is non hole/var alternatives exist (eg Z, Int -> T will have the suggestion Int -> T, and A, Int will have the suggestion Int)

Some suggestions based on this handling:
image
image
image

cyrus-:
@RaefM @anandrav this needs a merge with `dev`

cyrus-:
Seems to be a bug here:
image

cyrus-:
Possibly downstream of above bug, but was expecting this to be [Int], not an error.
image

cyrus-:
Probably also related but this should be a conflict but it's solving it as `[?]`: 
image

cyrus-:
When I Tab-expand on this I get `[]` which is a syntax error rather than `[?]`:
image

cyrus-:
@RaefM I think there must be a bug in your recent fixes -- the following should certainly not infer `[T]`, as neither `A` nor `B` are used anywhere!

image

Also still seeing this issue where there isn't a conflict arising from the conflicted patterns:

image

cyrus-:
Another probably related one that is behaving weirdly:

image

thomasporter522:
After re-uploading, Cyrus's identified errors are mostly solved. The exceptions are:
 - Tab-expanding the suggestion here still produces `[]`, which is a syntax error. 



- ADTs appear to be converted into binary sums at some point, which are wrongly displayed and suggested to the user.

thomasporter522:
Here's another odd bug:

thomasporter522:
Also maybe we don't want to offer this suggestion.

thomasporter522:
Adding constructor names was easy, thanks to the well-factored code. This works fine:



And this gives an inconsistency, as desired.



However, the error information is wrong. The ? constructors are my own doing, as a placeholder.

thomasporter522:
I understand why this is the error message we give, but I think it's unintuitive. Is there a better way to convey this? Either just having the type holes in the message be red, or actually showing the two different arrow types? The latter would allow the user to select which one they want and resume bidirectional checking. 



There is also the potentially appealing possibility of displaying potential type sets in full, as nested lists of options. For example, if the PTS is {Num , Arrow({Num, Bool}, {Num, Bool})}, the user could hover over any element listed between a pair of {...}, including the sublists, rather than inconsistent sublists being displayed as a hole. 

There are a couple of issues with this. The first is that if the user selects a sub-option, like the second "Num" in the text example, it is not determinate what entire type they are selecting. Do they mean Arrow(Num, Num) or Arrow(Num, Bool)? Should we just treat it as Arrow(Num, Hole)? What if the user wants to specify both sides? Maybe we could have a tree of drop downs, but that's surely far too complex. 

The second issue is that a PTS is in fact a lossy representation of the set of potential types for a hole. In the image example, the two constrained fillings for the type hole are {Arrow(Num,Bool), Arrow(Bool,Num)}, and if this is collapsed to Arrow({Num, Bool}, {Num, Bool}), the original possibilities may not be recovered, without also including Arrow(Num,Num) and Arrow(Bool,Bool).

In light of these two issues, might it be better to forgo merging inconsistent potential types? The principal disadvantage would be the loss of compactness, and now we reach an empirical matter, but I think that merging only consistent types (which needs further specification since consistency isn't transitive) would recover much of the compactness of the general scheme, and that remaining examples would probably produce moderately sized sets most of the time. To the first point, the example of a combinatorically explosive set of types in the old draft consists of six pairwise consistent types: and thus they could be compressed to one inconsistency-free type. To the second, the number of potential types grows multiplicatively, but only upon the inclusion of other type holes, and only by the number of different _type constructors_ as which these type holes are used. 

I don't doubt that it is easy to construct cases in which this "consistent representation" is too big a set. Maybe we could pragmatically compute both the compact and the consistent representations, and display the consistent one preferentially, and the compact one if the consistent one is larger than some predefined limit and larger than the compact one.

This was a bit of a ramble, sorry. Thoughts welcome.

cyrus-:
archiving in favor of the more recent effort by Alex Smart

PR #1156, opened 2024-01-08
Module alias

gensofubi:
Add module alias to work with functions.
`Module` keyword now automatically applies alias if the definition exp has module type (instead of generate from context).
{D9CF2EFF-848F-44a2-A4EF-ED822AFEB3E7}


cyrus-:
Looks good to me, merging into the main modules PR.

PR #1158, opened 2024-01-09
Re-enable workers

disconcision:
Depends on #1166.

This is good for a functional test. Not using the worker pool, just my simple worker model, but I've tweaked some behavioural things and it feels pretty good to me.

The basic behavior is this: There is one worker at a time. Whenever you do an edit, it checks if there is currently an active worker:
-   If there isn't an active worker, it starts a new worker and dispatches the job, marking it pending in the model results. While pending, the cell result display show a spinner and shows the previous result greyed out until the worker returns the result.
- If there is an active worker, it kills the worker, starts a new one, and dispatches the job as above. So if you start a large computation, you can cancel it by doing another edit. This incurs the worker startup time penalty, but that's not bad (<250ms). If you don't perform more edits, the worker just keeps on churning, though I've set a 20-second kill threshold on all worker jobs, for battery courtesy reasons if nothing else.

Screenshot: after inserting the zero:
Screenshot 2024-02-02 at 1 41 28 AM

@cyrus- : There's one significant outstanding issue here... previously I had made Closures [@ opaque], as they get absolutely huge very quickly when serialized. but i had to disable that here to send values back and forth. for example, when i am writing a short function (~3 lines), every time i have a hole, the serialized message size balloons to 50Kb. a few 100Kb is  actually fine perf wise but It's exponential so quickly becomes untenable. not sure how to proceed here exactly, though I can think of a few options. I guess I could just customize the to_sexp to strip closures?


disconcision:
@mirryi could you take a look at this when you get the chance? I tried to create a minimal webworker evaluator (see SimpleWorker), but the worker doesn't seem to be starting up right. I can tell a worker with the right filename is running from devtools, but it seems to have nothing printed to its console. Even the top-level printlns in Worker.re don't seem to be getting printed. From doing self in console its onmessage is null, so registration isn't happening either. it seems like the worker is getting initialized but the script doesn't actually run?

mirryi:
@disconcision Sorry, didn't have a chance to look at this earlier---did your recent commits fix the problem?

disconcision:
@mirryi yes it's basically fixed, at least the problem of the workers not doing anything at all. the indexedDB library is crashing them. in this simplified version the fix was simple; split the worker API into client and server, and make it so the server only depends on hazelcore not hazelweb. not sure how straightforward that is to backport to your version

disconcision:
@tonyfettes @Negabinary maybe take a look if I've broken the stepper. it seems to basically still work, though it doesn't get offloaded to the worker. I'm not sure if it makes sense to do so give that it's used synchronously? I  am noticing that when I have the stepper open and I've stepped partway though a large computation, the whole editor slows down, e.g.
Screenshot 2024-02-02 at 1 23 14 AM
which is probably sub-optimal though possibly just an edge case?

Anyway I think all the stepper work was done while webworker eval was disabled so there's some hard edges between the systems... maybe take a look and see if it's obvious how to make the connection tighter.

Also, perhaps related to the closure opacity issue I describe above, I noticed that you're now serializing (some of) the results to localstorage? This is potentially problematic ; when stepping that example I noticed that I had immediately filled localstore, maybe because of the closure serialization issue.

Negabinary:
@disconcision for some reason when I try to click on a step in this branch nothing's happening.

It's hard to know if it makes sense to offload the stepper to the worker, obviously for single-steps they're synchronous, but when there are filters that skip steps, then I can imagine that one might want to have it run asynchronously. The stepper is still a lot slower than the evaluator anyway when skipping steps so perhaps we're imagining that performance isn't important on the skip use cases.

If we wanted to make the connection tighter, my idea was to send ModelResult.t objects to the worker, and then the worker could step it if it was a stepper or evaluate it if it was an evaluation. Obviously, like you say, the web workers were disabled for that whole period, so I never tested that strategy.

Negabinary:
If we're filling the local store then yeah, that's almost definitely the same closure issue. I suspect on most of our test cases so far the majority of that space is going to come from the fact that all of the builtins are defined in every closure, I would think there must be a way to not serialize the builtins because they're always the same.

disconcision:
@Negabinary re: stepping not working what are you trying specifically?
on https://hazel.org/build/workers2/ if i put "1+1" in a scratch editor i can step it

tonyfettes:
@disconcision I can't proceed when I click the steppable area, for both "1 + 1" and the fibonacci example above. I have tried cleaning my local storage and the problem persist...

tonyfettes:
@disconcision Edit won't update the stepper, and I need to toggle-off then on the stepper for it to be updated.

Also, I guess it would be good to leave the stepper toggle button when the worker evaluator is running?

tonyfettes:
The map example is pretty slow... I know stepper filter performs pretty badly, but this one is significantly slower than the `dev` branch?
Screenshot from 2024-02-03 00-59-12

disconcision:
@tonyfettes I've re-enabled the stepper toggle to display even when evaluating or stack overflowing.

As for the slowness, i'm guessing it's due to closure environment serialization. As far as you know is closure serialization load-bearing for the stepper? I've tried to override the sexp serialization to just not serialize closure environments, but it seems to crash the stepper after a couple steps

disconcision:
@tonyfettes I notice for that program, both in dev and on this branch, activating the stepper immediately blows localstore.
(program included here for reference)
`
eval $e in
let map:    ->   =
  fun xs, f ->
    case xs
      | [] => []
      | hd :: tl => f(hd)::map(tl, f) 
    end 
in

let f = fun x -> x + x in
pause f($v) in
map([1, 2, 3], f)
`
i think it's slower on this branch because the way things are set up now, the stepper is getting round-tripped to the worker, and the slowness is in de/serializing the closure. It seems like the way things are set up now it's a bit awkward to not send the stepper to the worker. Not sure whether I should just hack it or if we can figure out a way to tackle closure issues directly

tonyfettes:
@disconcision IMO it should be solve on the stepper side - by making the closure capture only the variables it uses.

Link #1195 for reference.

disconcision:
@cyrus- this is good for another functional review. i couldn't figure out a strategy to both eliminate flickering entirely and also provide some kind of reasonably timely notification that evaluation is underway, but I think it's at a better compromise position now.

@tonyfettes stepping should now be skipping the worker and thus is hopefully no slower than on dev; lmk

cyrus-:
The delay / fade out behavior feels good to me now!

You mentioned a 20 second time out on worker tasks, but we don't seem to provide any feedback when the timeout is hit? It just continues to beachball afaict. Can we provide feedback similar to what we display for stack overflows in that situation?

cyrus-:
When I toggle dynamics off and then back on, it doesn't cause the result to re-appear until I make an edit.

disconcision:
Toggling dynamics now re-evaluates. Footer display is now updated on 20-second timeout.

cyrus-:
The whole screen is flickering on edits in exercises mode:


https://github.com/hazelgrove/hazel/assets/280638/98847eab-0362-469b-a638-4f4e0fbe880c

disconcision:
@cyrus- give it another try. The specific behavior you indicate should now be no jankier than dev, but I'm noticing dev definitely has some jank around dynamic values updating while editing in exercises mode.

disconcision:
FYI this is now downstream of #1166 as they have some nasty interactions I want to get ahead of with the functional testing here

cyrus-:
@disconcision there is a small conflict here

dm0n3y:
I can build but no longer run `dev` on my machine (Apple silicon) since this PR was merged in. Get a console exception about failure to construct a worker.
image

dm0n3y:
Learned from @disconcision that we need to run `make serve` now for the workers to get set up. I'll update the README.

edit: oh it's already updated, my b

PR #1160, opened 2024-01-12
Slides

Negabinary:
This (hastily) adds the ability to put images on scratch slides, or examples slides.

image

Images are specified using hard-coded urls in init.ml


disconcision:
vibe

cyrus-:
Is it significantly harder to have it be an arbitrary vdom node rather than just an image?

Negabinary:
> Is it significantly harder to have it be an arbitrary vdom node rather than just an image?

oh I see, probably not much harder, no; I'll take a look shortly

Negabinary:
so it's harder to save the arbitrary node because there's no sexp for Node.t; but I might be able to make it just hard-coded today, never going through the save file if that works?

cyrus-:
Yeah that should be fine. I'll spend time tomorrow getting everything
prepped.

cyrus-:
Merged as part of #1164

PR #1165, opened 2024-01-19
Nut Menu Ultra

disconcision:
This moves all stray top bar icons to the nut menu. It adds submenus to the nut menu; right now these are Settings, Import, and Export. Import includes editor reset actions with the justification that these are both clobbering current state. Which icons are in the menus are conditioned on editor mode and settings. All submenu items are labelled. This also introduces the Star icon as a default icon; this can be used if you don't want to bother finding a new svg icon. 

Screenshot 2024-01-19 at 5 12 36 PM
Screenshot 2024-01-19 at 5 12 27 PM
Screenshot 2024-01-19 at 5 12 11 PM

This likely supersedes #985; sorry @tonyfettes, we had some requirement drift. I like the tooltip style from that PR but I'm not sure that it's necessary now in the absence of top bar buttons.

Incidentally I don't have an objection to top bar buttons per se, but they should justify themselves as something that gets sufficiently frequent use; I'd rather clutter accumulate in the menu than on the bar, so let's try to err on that side for the present.


cyrus-:
Looks generally good to me. One thought: it may not be obvious that the Hazel logo is where the settings go. Should we maybe do the more conventional hamburger menu on the right?

disconcision:
I think there are few enough top level UI elements that people who are even slightly motivated will find it. UL app icon as menu isn't an entirely uncommon pattern

PR #1166, opened 2024-01-21
Cached statics

disconcision:
Depends on #1189.

This PR is intended to be more or less a pure refactor, making several related changes clarifying and homogenizing how derived editor metadata (syntactic, statics, and dynamics) data is calculated and passed through the editor front end.

1. CachedStatics system: All statics data (for the main editor, not ExplainThis) is cached on the model in model.statics. This is keyed per-editor using basically the same scheme as dynamics. This cached data is updated at the end of the main Update function, similarly to dynamics. Calculation is explicitly gated on whether or not there was an edit action; this could in the future be made more fine grain depending on action result, incrementalized, pushed to a delay after last keystroke etc for better perf.
2. Cached syntax on Editor.Meta: Many of the intermediary data structures (such as syntax) built up in the view layer are now pre-calculated and cached as part of Editor.Meta. Re-calcuation of these is gated on edits and could be further refined as-above.  This already results in much less calculation being done during caret movement.
3. ExplainThis docs now use Editors/Cells to display the form header and examples; previously this was more ad-hoc. This fixes some minor inconsistencies with decorations and sorts, and allows test checkmarks to show up in the docs for the test form.
4. DebugMode is factored out from the main app to a new top level, eliminated mixed concerns with the editor mode system.
5. Dynamics unwrapping cleanup: This removes some extraneous data types around evaluation results. It seems @Negabinary already did a good chunk of this, but this further removes TestResults.simple and TestResults.simple_data for more regular dynamics data interfaces.
7. Misc cleanup and dead code removal

@Negabinary I made some minor organizational changes to some of the stepper frontend; please take a look and let me know if there are any issues or if I broke anything; the basics seem to work.


Negabinary (approved):
dynamics / stepper look good to me

cyrus-:
Getting weird z indexing? alpha? on the context inspector... 
image

cyrus-:
The ExplainThis sidebar also doesn't scroll at the moment (e.g. see the docs for when on a function literal). This is also the case on `dev` but since you are changing ExplainThis UI things, I'll mention it here -- if this is a more significant change feel free to make an issue instead.

disconcision:
@cyrus- both the context inspector z-index and the sidebar scrolling issues are fundamentally manifestations of the same issue. both should be fixed now

cyrus-:
Minor issue that seems to arise from this PR: there is now a cell border on the exercise description but it isn't well aligned:
image

Issue #1171, opened 2024-01-24
Settings menu gone


disconcision:
After the stepper merge the settings menu in the nut menu doesn't show up on hover.

Also there are some minor graphical glitches with the stepper:
<img width="385" alt="Screenshot 2024-01-24 123555" src="https://github.com/hazelgrove/hazel/assets/22436459/738de27b-ed9c-4032-986f-fa9ad8d4fba9">

1. The cell result background only extends behind the first step
2. The cell result background looses its rounded corners
3. The cell result looses the red left border that indicates cell selection

disconcision:
also might want to consider not havings the settings gear for the stepper displace the stepper toggle... it's nice when you can toggle a toggle on and off without mouse movemnet, which you can't if it gets displaced by the gear

disconcision:
question mark on green bkg here is almost invisible:
<img width="210" alt="Screenshot 2024-01-24 124257" src="https://github.com/hazelgrove/hazel/assets/22436459/c51797a4-a860-49ed-abe7-6f278402e482">

disconcision:
in exercises mode, pressing the history toggle on e.g. this state causes the stepper to disappear: 
<img width="685" alt="Screenshot 2024-01-24 124632" src="https://github.com/hazelgrove/hazel/assets/22436459/2b736d12-e90c-4ba1-b161-fb79734db1f1">
Actually the same thing happens if you click the gear... the stepper toggle is getting activated in both cases i guess

disconcision:
The console is noisy; at least two prints need disabling

PR #1174, opened 2024-01-24
Bring back settings

Negabinary:
Fix the first issue in #1171


disconcision:
i can create separate issues for the others if helpful, i was just being lazy

PR #1175, opened 2024-01-24
Add button that reparses current editor

dm0n3y:
image


src/haz3lweb/util/JsUtil.re
```diff
@@ -122,18 +124,21 @@ let copy = (str: string) => {
...
+    let main_rect = main##getBoundingClientRect;
+    let caret_rect = caret_elem##getBoundingClientRect;
+
+    if (caret_rect##.top < main_rect##.top) {
+      caret_elem##scrollIntoView(Js._true);
+    } else if (caret_rect##.bottom > main_rect##.bottom) {
+      caret_elem##scrollIntoView(Js._false);
+    };
+  }) {
+  | Assert_failure(_) => ()
```
dm0n3y, line 140: also fixed an assertion failure here happening when caret isn't present

cyrus-:
This is currently causing additional indentation to be inserted as spaces each time I press it...

dm0n3y:
Not a perfect fix, but at least idempotent now afaict. Holes are swapped with leading space, like so.
image

PR #1176, opened 2024-01-25
renamed sections to Documentation and Exercises

cyrus-:


disconcision:
Is this just a temporary measure? Not sure exactly what the use case is but these feel like they'd more naturally be a succession of cells on the same document? They feel a bit awkwardly mixed in with the current example sides, given that they don't really provide example code

cyrus-:
Yeah by no means a final design, we can refine how the top bar is organized / add more hierarchy as needed, just wanted to have the slides from lecture available as documentation in addition to the examples.

PR #1178, opened 2024-01-25
remove context entry gradient

cyrus-:
this was causing issues with the "Correct Implementation" display in Exercises mode and doesn't seem necessary in the context inspector either


disconcision:
it's to lessen the awkwardness of how the entries get truncated by the menu list. what was the issue with Exercises mode?

cyrus-:
image

It was causing long signatures to be truncated in this view

cyrus-:
feel free to bring it back more selectively, just needed a quick fix for now

PR #1186, opened 2024-01-25
Stepper CSS fixes

### LINKED ISSUE #1171: Settings menu gone

Negabinary:
Fix most of the issues in #1171.

This one https://github.com/hazelgrove/hazel/issues/1171#issuecomment-1908632858 is still not done, it's very strange, because I can't think of anything that the stepper does after the settings update but not after the code updates.


Negabinary:
ok, I've fixed the last one - it's related to the more general problem that stepper state doesn't save in exercises mode, but I'm hesitant to make any changes to the save file while students are using it, so I've just added a temporary fix.

PR #1189, opened 2024-01-27
ExplainThis highlighting on hover

disconcision:
Note: Depends on #1188 .

This PR:

- Makes explainThis term highlighting happen only when hovering over the relevant subterm in the explanation
- Clicking on a subterm now selects the subterm instead of jumping the caret to its beginning. Still not sure this click event pays for itself, but I feel it's less confusing this way as there is a clearer relationship between the before (highlighted) and after (selected) state.
- Hover highlights now work in exercise mode. This is currently broken in dev
- Removed manual toggle for term highlighting; highlighting now always on in sidebar, and on hover only in main editor.
- Moved remaining manual toggle (show feedback) to nut menu

(Soliciting feedback on above changes)


cyrus-:
Getting `Not_found` exceptions in various situations when the color toggle is on, e.g. typing `let x ` in scratch mode and then pressing `=`. Also when clicking around in exercises mode on existing code.

disconcision:
@cyrus- appears to be an outstanding issue on dev, which surfaced with the TyDi PR, specifically that we started basing CI (and statics more generally) on a prospective completion of the code. Should be fixed now, though this aspect of the code will likely remain somewhat brittle until the new syntax model is in place.

cyrus-:
How difficult would it be to get the hover highlighting behavior to also work when you hover over the highlighted subexpressions in the syntax box? Currently only works for the words in the description.

image

cyrus-:
Also minor style thing -- would be nice to move the ExplainThis toggle + closeout to be aligned on the right next to the syntax section's header:
image

cyrus-:
Created issues for the above issues.

PR #1193, opened 2024-01-31
Add Coq export functionality for arithmetic stepper expressions

jchen-cs:
This PR will add Coq export functionality similar to the Interactive Evaluator to Hazel, for simple arithmetic expressions (without variables)

### DIFF
--- a/src/haz3lweb/Update.re
+++ b/src/haz3lweb/Update.re
@@ -528,7 +528,9 @@ let rec apply =
       switch (result) {
       | NoElab
       | Evaluation(_) => ()
-      | Stepper(stepper) => CoqExport.exportCoq(stepper.previous)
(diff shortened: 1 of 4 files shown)


src/haz3lweb/Update.re
```diff
@@ -528,7 +528,7 @@ let rec apply =
       switch (result) {
       | NoElab
       | Evaluation(_) => ()
-      | Stepper(_stepper) => print_endline("Hello world!")
+      | Stepper(_stepper) => CoqExport.exportCoq(_stepper.previous)
```
Negabinary: small style thing: an underscore at the start of a variable usually means that the variable is unused.

jchen-cs: Thanks! Fixed in the next commit

Issue #1195, opened 2024-02-02
Stepping in deep recursion blows up localstore


disconcision:
@tonyfettes @Negabinary I thought this was related to my worker changes (@1158) but it turns out this happens in dev too. basically i clicked through about ~20 random steps and this happens:

<img width="1300" alt="Screenshot 2024-02-02 at 11 12 17 AM" src="https://github.com/hazelgrove/hazel/assets/22436459/d5533ecf-c404-4d29-8d6e-9f7b8e00c019">

edit: this seems to still be a thing after clearing my localstore but apparently i'm bad at that so someone else should confirm lol

Negabinary:
can confirm this happens on my computer as well, it's giving up after the file gets to about 4MB after ~20 steps and the file is mostly nested closures

tonyfettes:
@Negabinary I'm not really clear about what are dumped into local storage, but I'm suspecting logging step actions eats up a lot of spaces really quickly.

Nested closures are indeed a problem now, considering that we can quite some prelude functions there and in stepper we're in fact duplicate these when serializing nested closures.

Negabinary:
@tonyfettes the things that are dumped into local storage are:
1. a save file of the current stepper state (a list of all previous DHExp.ts and previous steps)
2. a log of all the stepper actions taken (including EvalObj.ts)
both are going to contain a lot of prelude/builtin functions and yeah I had a brief look through and a lot of the closures have ~150 copies of the same closure in the save files, so perhaps some way of exploiting the fact that they are all the same might help.

tonyfettes:
We used to have closure capturing analysis before (it only includes
captured variables) and I think it can help?

Re-implement in new evaluator would also not be hard... I guess...

disconcision:
Stepped through fib(20) until I got bored, seems resolved

PR #1196, opened 2024-02-06
Closure saving

Negabinary:
This PR changes the way closures are saved in s-expressions to reduce file sizes.

#1195 pointed out that recursive calls can make save files blow up quite quickly. Analysing these files shows that much of the file is just closures, and some of these closures are saved 100s of time.

It turns out that removing duplicate closures was able to shrink the file in #1195 to roughly a quarter of the original size. This is less of a reduction than we were expecting. Interestingly, also, once the reduction has taken place, the closures take up a very small portion of the file, so we cannot make further progress by just changing closures alone.

### DIFF
--- a/src/haz3lcore/StructureShareSexp.re
+++ b/src/haz3lcore/StructureShareSexp.re
@@ -1,0 +1,76 @@
+/*
+     This module adds helpers for creating s-expressions that use some structural sharing
+     (https://www.ocamlwiki.com/wiki/Structural_sharing) instead of copying all instances
+     of the same record.
+ */
+
+let structure_share_map: ref(option(Id.Map.t(Sexplib.Sexp.t))) = ref(None);
+
+[@deriving sexp]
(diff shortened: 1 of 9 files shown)


tonyfettes (commented):
LGTM!

disconcision:
i'm now wondering if this is possible to do entirely or almost entirely generically wrt substructural sharing, without having to wait of dhexp uuids. like in principle if shared structure is actually referentially shared, the pointers should already act as unique ids.

so we could do something like: recurse through the data structure, comparing each subcomponent for referential equality against an accumulated list of all subcomponents seen so far. any duplicated subcomponents are assigned unique ids in a separate lookup table. the linearity of that sounds bad but i assume there's some way to actually get a representation of the pointer we could hashmap over or something equivalent

cyrus-:
There is no way to get the pointer hash in JavaScript (probably because that would limit GC from moving things), but you can use a Map/WeakMap which uses reference equality, ===. internally for equality...

cyrus- (approved):

PR #931, opened 2022-11-15
Merge DHExp into UExp

tonyfettes:
# Merge DHExp into UExp

This pull request is to merge DHExp into UExp, so that
1. Duplication code can be removed
2. Web interface can use ids to help annotating

## Roadmap

- [x] UExp definition
- [x] Statics
- [x] Dynamics
  - [x] Elaborator
  - [x] Evaluator
  - [x] Assign meaningful (?) ids to created expression in elab/eval (#857).
- [x] Web
  - [x] LangDoc
  - [x] DHDoc
- [x] Use `CExp`, instead of `UExp`.
- [ ] Namespace
  - [ ] Change Builtin Id assignment with standalone namespace.

## Status

Currently it can run some simple Hazel programs. Not thoroughly tested yet.

Known bugs:
- [ ] Occasionally the editor will stuck at loading screen. Unable to reproduce deterministically.
- [ ] Duplicated Id with holes.

## Future work

- Semantics
  - Pass Parenthesis (`Parens`) into evaluator, so that small step evaluator can enforce order according to parenthesis.
  - Remove `ApBuiltin`
  - Remove `Prj` and `Inj`
  - Remove `FixF` using `Let` and `Fun`
- Pretty Printer.

PR #1197, opened 2024-02-06
Merge DHExp and UExp

### LINKED ISSUE #878: Proposal: DHExp/UExp merge
This intended to be a running discussion on the benefits/drawback of eliminating DHExp. Will be adding things as I think of them.

PROS:

- One less data type (they are already very similar)
- Opportunity to clear/clean up quirks/redundancies in DHExp.t
- Unique ids in evaluation; useful for tracking dynamic data for things like e.g. perf heatmaps
- View layer consolidation without backwards conversion

CONS:

- conceptual: What does it mean to elaborate a cast?
- still need a separate notion of typing maybe? given unicity?

### LINKED ISSUE #1248: Casts should be inserted mid-pattern
Here's a "purer" version of the above problem that does not involve #1246
image

_Originally posted by @Negabinary in https://github.com/hazelgrove/hazel/issues/1242#issuecomment-2003946840_

### LINKED ISSUE #1260: `+ X(Int)` should not be a ground type
Breaking example on dev:
image

```
let x : + B( ) = B(1) in
let y : + B(Int) = x in
let B(n) = y in
n + 2
```

I think constructors should only be ground if all their arguments are hole type?

### LINKED ISSUE #1273: add explicit fixpoints for pedagogical purposes
we insert fixpoints automatically in elaboration, but for pedagogical purposes it would be useful to also write them explicitly

Negabinary:
## Motivation:
Implements #878, 

The advantages of merging DHExp and UExp are:
- Removing duplication of code, making it easier & quicker to add new features to the Hazel language
- Consistency between evaluation results and written code - making easier to copy outputs into code, or to edit outputs, etc.
- Creating a consistency between the way static and dynamic code are pretty-printed, hopefully eventually replacing the pretty printer
- Tracing of ids through code allows us to correspond each construct from an output with the construct in the input

This pr replaces #931 which replaced #857. I have started again (using the previous iterations as inspiration) because the codebase has changed too much since those prs.

## Key Changes

* DHExp is no longer a separate data structure to UExp.
* Fixpoints are now surfaced in user expressions
* Elaboration has been rewritten to more closely match elaboration rules from the POPL 2019 paper.
* Pattern matching has been rewritten to separate out an "unboxing" routine.
* Patterns can now have casts

### If you used any dynamic expressions (DHExp) before the merge:
You will need to replace these with terms from TermBase.re. Most constructs retain the same name, the complication is that TermBase.re terms also have Ids. The following functions should help you add ids to terms:

 - When creating a new term add `|> Exp.fresh`, `|> Pat.fresh`, or `|> Typ.temp` to the end of the construct. (e.g. `IntLit(4)` becomes `Int(4) |> Exp.fresh`). `Typ.temp` is preferred over `Typ.fresh` because fresh generates new ids which can slow down type operations where types are often created then thrown out quickly. Ids are added to Typ.t values before they are printed to the screen.
 - When pattern matching on a term, `switch (exp)` will usually become `switch (exp.term)`. If you want to pattern match on a term and then wrap it back up with the same Id, you can use `Exp.unwrap` and `rewrap`. (There are many examples of this in Term.re)

### If you have added any new constructs to DHExp:
 - First, you'll need to make sure that there are constructs in TermBase.re that can store these dynamic expressions. 
 - Then you'll need to adapt your elaborator code to use only these constructs, and to make sure that all the types match up (we're no longer relying on statics information to provide types for elaboration.)

### If you have added any new constructs to TermBase.re that are removed at elaboration:
Since the pre- and post-elaboration data structures are the same, it's now preferred not to have elaboration remove constructs. (This also makes stepper traces more readable.) You might need to add transition rules and DHDoc_Exp rules for your new constructs.

### If you have made any new "value" expressions:
I've tried to clarify how casting works in Unboxing.re. If you have a new value type, you might find it easier to write Transition.re rules if you add a new `unbox_request` constructor to Unboxing.re


cyrus-:
@Negabinary "Make Typ match UTyp" may be a bit subtle, lmk if you want to discuss

src/haz3lcore/dynamics/TypeAssignment.re
cyrus-, line 1: this should be removed since its from another PR

src/haz3lcore/zipper/EditorUtil.re
```diff
@@ -41,46 +41,56 @@ let editors_of_strings = (~read_only=false, xs: list(string)) => {
...
-    TermBase.UExp.{ids: e1.ids, term: Filter(act, econd, ebody')};
-  | Let(p, edef, ebody) =>
-    let ebody' = append_exp(ebody, e2);
-    TermBase.UExp.{ids: e1.ids, term: Let(p, edef, ebody')};
-  | TyAlias(tp, tdef, ebody) =>
-    let ebody' = append_exp(ebody, e2);
-    TermBase.UExp.{ids: e1.ids, term: TyAlias(tp, tdef, ebody')};
-  };
+let rec append_exp = {
+  Exp.(
```
cyrus-, line 45: is this necessary?

src/haz3lweb/Update.re
```diff
@@ -475,6 +475,18 @@ let rec apply =
         model.results
         |> ModelResults.find(key)
         |> ModelResult.step_forward(idx);
+      let _ = print_endline("Ouch");
```
cyrus-: Are these for debugging? If so, remove before merge.

src/haz3lcore/lang/Form.re
```diff
@@ -282,7 +282,7 @@ let forms: list((string, t)) = [
   ("cons_pat", mk_infix("::", Pat, P.cons)),
   ("typeann", mk(ss, [":"], mk_bin'(P.ann, Pat, Pat, [], Typ))),
   // UNARY PREFIX OPERATORS
-  ("not", mk(ii, ["!"], mk_pre(5, Exp, []))), //TODO: precedence
+  ("not", mk(ii, ["!"], mk_pre(P.not_, Exp, []))), //TODO: precedence
```
cyrus-: is this TODO resolved?

src/haz3lcore/statics/Term.re
cyrus-, line 1: should Term, TermBase be in lang/term instead of in statics?

src/haz3lcore/dynamics/DHExp.re
```diff
@@ -1 +1,151 @@
-include DH.DHExp;
+/* DHExp.re
+
+   This module is specifically for dynamic expressions. They are stored
+   using the same data structure as user expressions, but dynamic
+   expressions are specifically paired with a `Satic.Map.t`.
```
disconcision: typo satic

src/haz3lcore/lang/Form.re
disconcision: can rm comment now

src/haz3lcore/lang/Precedence.re
```diff
@@ -14,6 +14,8 @@ let ap = 2;
 let neg = 3;
 let power = 4;
 let mult = 5;
+
```
disconcision: is this empty line intentional?

src/haz3lcore/lang/term/IdTagged.re
```diff
@@ -0,0 +1,35 @@
...
+
+// let serialization = (f1, f2) =>
+//   StructureShareSexp.structure_share_here(
+//     rep_id,
+//     sexp_of_t(f1),
+//     t_of_sexp(f2),
+//   );
+
+// let sexp_of_t = f1 => serialization(f1, Obj.magic()) |> fst;
+// let t_of_sexp = f2 => serialization(Obj.magic(), f2) |> snd;
```
disconcision: is this commented code intentionally retained? if so needs addn comment

src/haz3lcore/lang/term/IdTagged.re
```diff
@@ -0,0 +1,35 @@
...
+[@deriving (show({with_path: false}), sexp, yojson)]
+type t('a) = {
+  [@show.opaque]
+  ids: list(Id.t),
+  /* UExp invariant: copied should always be false, and the id should be unique
+     DHExp invariant: if copied is true, then this term and its children may not
+     have unique ids. */
+  [@show.opaque]
+  copied: bool,
+  term: 'a,
```
disconcision, line 14: can we say more here (or elsewhere) about the functionality of the copied flag?

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+  | Ap(_)
+  | Sum(_)
+  | Rec(_) => false
+  };
+};
+
+/* Converts a syntactic type into a semantic type, specifically
+   it adds implicit recursive types, and removes duplicate
+   constructors. */
+let rec to_typ: (Ctx.t, t) => t =
```
disconcision: i know this wasnt any better before, but it would be nice to try and create a little more clarity around the role of this function. what exactly does 'adds implicit recursive types' mean? to me this reads as 'if a name is used in a definition of the same name, turn it into a rec', but that's not exactly what's going on. also it does more stuff besides the description, eg. eliminates parens, adds certain error cases. to what extent are all of these things similar? can we construe this as a kind of normalization?

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+       | (BadEntry(_), Variant(_))
+       | (BadEntry(_), BadEntry(_)) => false
+       }
+     );
+};
+
+/* Functions below this point assume that types have been through the to_typ function above */
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type source = {
```
disconcision, line 127: do we still need this now that types have ids? or is this a different id, i don't totally remember

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+  | List(ty) => free_vars(~bound, ty)
+  | Arrow(t1, t2) => free_vars(~bound, t1) @ free_vars(~bound, t2)
+  | Sum(sm) => ConstructorMap.free_variables(free_vars(~bound), sm)
+  | Prod(tys) => ListUtil.flat_map(free_vars(~bound), tys)
+  | Rec(x, ty)
+  | Forall(x, ty) =>
+    free_vars(~bound=(x |> TPat.tyvar_of_utpat |> Option.to_list) @ bound, ty)
+  };
+
+let var_count = ref(0);
```
disconcision, line 171: what are the relevant invariants here wrt uniqueness for this global? as is, if we call fresh var, then refresh the page, then call it again we'll get the same id. i assume this is irrelevant to the caller of fresh_var but seeing a global counter like this which looks like it could be used for a persistent data structure raises a flag

disconcision: realizing that some of my comments here are maybe about polymorphism rather than this PR, feel free to disregard if so

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+  | (List(ty1), List(ty2)) =>
+    let+ ty = join'(ty1, ty2);
+    List(ty) |> mk_fast;
+  | (List(_), _) => None
+  | (Ap(_), _) => failwith("Type join of ap")
+  };
+};
+
+/* REQUIRES NORMALIZED TYPES
+   Remove synswitches from t1 by maching against t2 */
```
disconcision: typo maching

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+let rec match_synswitch = (t1: t, t2: t) => {
```
disconcision, line 283: this function scares me, but i assume it's polymorphism related so whatever

src/haz3lcore/statics/Info.re
```diff
@@ -501,23 +517,47 @@ let is_error = (ci: t): bool => {
 };
 
 /* Determined the type of an expression or pattern 'after hole fixing';
-   that is, all ill-typed terms are considered to be 'wrapped in
+   that is, some ill-typed terms are considered to be 'wrapped in
```
disconcision, line 542: quantifier relaxations that make you go 'hmm'

src/haz3lcore/statics/Info.re
```diff
@@ -209,9 +208,10 @@ type pat = {
   ancestors,
   ctx: Ctx.t,
   co_ctx: CoCtx.t,
+  prev_synswitch: option(Typ.t), // If a pattern is first synthesized, then analysed, the initial syn is stored here.
```
disconcision, line 212: i feel like this name raises questions not addressed in the comment. like could we call this 'synthesized_type'? or does that miss the point

src/haz3lcore/statics/Info.re
```diff
@@ -501,23 +517,47 @@ let is_error = (ci: t): bool => {
...
   fun
   | Syn(syn) => syn
   | Ana(Consistent({join, _})) => join
   | Ana(InternallyInconsistent({ana, _})) => ana;
 
+let fixed_typ_err_common: error_common => Typ.t =
+  fun
+  | NoType(_) => Unknown(Internal) |> Typ.mk_fast
+  | Inconsistent(Expectation({ana, _})) => ana
+  | Inconsistent(Internal(_)) => Unknown(Internal) |> Typ.mk_fast // Should this be some sort of meet?
```
disconcision: what would it be meeted with?

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -138,18 +138,18 @@ let rec go_s = (s: Sort.t, skel: Skel.t, seg: Segment.t): any =>
 and exp = unsorted => {
   let (term, inner_ids) = exp_term(unsorted);
   let ids = ids(unsorted) @ inner_ids;
-  return(e => Exp(e), ids, {ids, term});
+  return(e => Exp(e), ids, {ids, copied: false, term});
```
disconcision, line 149: isn't there a constructor which sets this? if not there probably should be, copied seems like it shouldnt be factored into maketerm

src/haz3lcore/statics/Statics.re
```diff
@@ -36,6 +34,9 @@ module Map = {
   [@deriving (show({with_path: false}), sexp, yojson)]
   type t = Id.Map.t(Info.t);
 
+  // let (sexp_of_t, t_of_sexp) =
+  //   StructureShareSexp.structure_share_in(sexp_of_t, t_of_sexp);
+
```
disconcision: commented code

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+
+include TermBase.Typ;
+
+let term_of: t => term = IdTagged.term_of;
+let unwrap: t => (term, term => t) = IdTagged.unwrap;
+let fresh: term => t = IdTagged.fresh;
+/* fresh assigns a random id, whereas mk_fast assigns Id.invalid, which
+   is a lot faster, and since we so often make types and throw them away
+   shortly after, it makes sense to use it. */
+let mk_fast: term => t = term => {term, ids: [Id.invalid], copied: false};
```
disconcision: Maybe we should call this mk_temp or something else which explicitly connotes the fact that we don't want to re-introduce the result into syntax? don't totally love that name either, tempted to just use something like mk or mk' or Typ.temp given how often it's used

src/haz3lcore/statics/ConstructorMap.re
```diff
@@ -2,100 +2,179 @@ open Util.OptUtil.Syntax;
...
-      )
-      : bool =>
-    Constructor.equal(ctr1, ctr2) && val_equal(val1, val2);
-  map1 === map2
-  || {
-    let map1 = List.fast_sort(compare_bindings, map1);
-    let map2 = List.fast_sort(compare_bindings, map2);
-    List.equal(equal_bindings(val_equal), map1, map2);
-  };
+/* computes all three regions of a venn diagram of two sets represented as lists */
```
disconcision, line 87: feels a bit weird that this is generic but the failwith message isnt. nbd tho

src/haz3lcore/statics/Kind.re
```diff
@@ -1 +0,0 @@
-include TypBase.Kind;
```
disconcision, line 1: is this an empty file?

src/haz3lcore/zipper/EditorUtil.re
disconcision, line 45: this has like 3 levels of nesting more than it needs to

src/haz3lschool/Exercise.re
```diff
@@ -575,14 +575,21 @@ module F = (ExerciseEnv: ExerciseEnv) => {
     hidden_tests: 'a,
   };
 
-  let wrap_filter = (act: FilterAction.action, term: Term.UExp.t): Term.UExp.t =>
```
disconcision, line 578: i guess not related to this PR but i feel like this function should be in Filter or something

src/haz3lweb/Update.re
disconcision: https://knowyourmeme.com/memes/bone-hurting-juice

src/haz3lweb/Update.re
```diff
@@ -475,6 +475,18 @@ let rec apply =
...
+        print_endline(
+          Stepper.show_stepper_state(
+            (
+              fun
+              | (Stepper(s): ModelResult.t) => s
+              | _ => failwith("")
+            )(r).
+              stepper_state,
+          ),
+        );
```
disconcision: ibid

src/haz3lweb/explainthis/data/FixFExp.re
```diff
@@ -0,0 +1,57 @@
+open Haz3lcore;
+open ExplainThisForm;
+open Example;
+
+/* (A) Use this file as an example for adding a new form to ExplainThis.
+ * You should be able to copy-paste this file and modify it to add a new form */
```
disconcision: this file probably doesn't need to retain the explanatory comments from the template file

src/haz3lweb/util/WorkerClient.re
```diff
@@ -46,6 +46,8 @@ let request =
 
   setupWorkerMessageHandler(workerRef.contents);
 
+  // print_endline(Request.serialize(req));
```
disconcision: rm

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -37,117 +37,131 @@ let precedence_bin_float_op = (bfo: TermBase.UExp.op_bin_float) =>
...
+      ~hidden_steps: list((step, Id.t)), // The hidden steps between the above and the current one (an Id in included because it may have changed since the step was taken)
       ~chosen_step: option(step), // The step that will be taken next
-      ~next_steps: list(EvalObj.t), // The options for the next step, if it hasn't been chosen yet
+      ~next_steps: list((int, Id.t)), // The options for the next step, if it hasn't been chosen yet
       ~env: ClosureEnvironment.t,
+      ~infomap: Statics.Map.t,
       d: DHExp.t,
     )
     : DHDoc.t => {
+  // // print_endline("");
```
disconcision: rm

src/haz3lweb/view/dhcode/layout/DHDoc_Exp.re
```diff
@@ -300,169 +271,173 @@ let mk =
...
-        |> annot(DHAnnot.InconsistentBranches((u, i)))
-      | BoundVar(x) when settings.show_lookup_steps => text(x)
-      | BoundVar(x) =>
+      | EmptyHole =>
+        DHDoc_common.mk_EmptyHole(
+          ~selected=Some(DHExp.rep_id(d)) == selected_hole_instance,
+          env,
+        )
+      | MultiHole(_ds) =>
+        //ds |> List.map(go') |> Doc.hcats
```
disconcision: rm

src/haz3lweb/view/dhcode/DHCode.re
```diff
@@ -137,15 +134,16 @@ let view =
...
       ~pos=0,
-      ~previous_step: option(EvaluatorStep.step)=None, // The step that will be displayed above this one
-      ~hidden_steps: list(EvaluatorStep.step)=[], // The hidden steps between the above and the current one
+      ~previous_step: option((EvaluatorStep.step, Id.t))=None, // The step that will be displayed above this one
+      ~hidden_steps: list((EvaluatorStep.step, Id.t))=[], // The hidden steps between the above and the current one
       ~chosen_step: option(EvaluatorStep.step)=None, // The step that will be taken next
-      ~next_steps: list(EvaluatorStep.EvalObj.t)=[],
+      ~next_steps: list((int, Id.t))=[],
       ~result_key: string,
+      ~infomap,
```
disconcision, line 137: what is the infomap ultimately being used for here?

src/haz3lcore/statics/TermBase.re
```diff
@@ -1,138 +1,94 @@
 open Sexplib.Std;
 
+let continue = x => x;
+let stop = (_, x) => x;
+
+/* TODO[Matt]: Explain map_term */
```
disconcision: Todo it

src/haz3lcore/statics/Statics.re
```diff
@@ -546,13 +606,21 @@ and upat_to_info_map =
       ~co_ctx,
       ~ancestors: Info.ancestors,
       ~mode: Mode.t=Mode.Syn,
-      {ids, term} as upat: UPat.t,
+      {ids, term, _} as upat: UPat.t,
       m: Map.t,
     )
     : (Info.pat, Map.t) => {
   let add = (~self, ~ctx, m) => {
+    let prev_synswitch =
```
disconcision, line 687: this doesn't feel obviously consistent with the way prev_synswitch is described in its type definition

src/haz3lcore/lang/term/Typ.re
```diff
@@ -0,0 +1,591 @@
...
+  | Var
+  | Constructor
+  | Parens
+  | Ap
+  | Rec
+  | Forall;
+
+include TermBase.Typ;
+
+let term_of: t => term = IdTagged.term_of;
```
disconcision, line 28: maybe just 'term' as its used a lot

src/haz3lcore/dynamics/DHExp.re
disconcision: Also what does this comment mean exactly? how are they paired? like is there a type?

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -8,458 +18,543 @@ module ElaborationResult = {
...
-  | _ => None
-  };
+let fresh_pat_cast = (p: DHPat.t, t1: Typ.t, t2: Typ.t): DHPat.t => {
+  Typ.eq(t1, t2)
+    ? p
+    : {
+      let _ = print_endline("=====vvvvv===");
+      let _ = print_endline(Typ.show(t1));
+      let _ = print_endline(Typ.show(t2));
+      let _ = print_endline("=====^^^^^===");
```
disconcision: whats this

disconcision (commented):
I've gone over everything except I only skimmed dynamics. In general this feels pretty solid. Does some much-needed cleanup and generally avoids complecting things. There are some requests for clarifications in the comments but no major changes. I did some basic functional tests, the only issue I've found is the test checkmark thing.

Some general questions:

1. Why are we calling TypeAnns casts now? Not against it just curious as to the exact rationale
2. What exactly is the role of the layout/DH* modules now? And the DH terminology more broadly? It looks like we're still using the old view layer, except now we convert Exps to DHLayouts? Are we waiting to fully deprecate DH?
3. The proliferation of Typ.fresh, fast, etc makes me somewhat question the elimination of the internal Typ.t. We do so much syntax-irrelevant logic on types. Not asking that you roll back this part of the change, just noting that this feels debatable to me.

src/haz3lcore/dynamics/DHExp.re
Negabinary: Ah yeah sorry this comment is old from when we were using the static map through evaluation, I've removed that now, they're no longer paired.

src/haz3lcore/lang/term/IdTagged.re
Negabinary: added a motivation sentence

src/haz3lcore/lang/term/Typ.re
Negabinary: > "if a name is used in a definition of the same name, turn it into a rec"
Ah yeah that's what I thought was going on but looking again that's clearly not what's going on.

This two different normalizations actually really annoys me too - maybe it's worth spending a little time to get rid of this first one. The rest of the code is already tolerant to parentheses, and it shouldn't be too hard to make it also tolerant to duplicate constructors and unbound variables.

src/haz3lcore/lang/term/Typ.re
Negabinary: I guess not - I didn't notice that was there; will work to get rid of it

src/haz3lcore/lang/term/Typ.re
Negabinary: yeah probably a question for polymorphism folks

src/haz3lcore/lang/term/Typ.re
Negabinary: nah this is my beast - it's used in calculating an elaborated type.

let's say we're analysing `(1,2)` against type `(SynSwitch, Unknown)` - the elaborated type of this expression should be `(Int, Unknown)` which is neither the ana type, nor the self type, so we have to use match_synswitch((Syn, Unknown), (Int, Int))

src/haz3lcore/statics/Info.re
Negabinary: There's probably a better way to say now this since I've got rid of non-empty holes and replaced them all with casts / failed casts.

src/haz3lcore/statics/Info.re
Negabinary: I feel like that would be confusing because if it's only synthesized once, this will be None. Maybe `prev_syn_type`?

src/haz3lcore/statics/Info.re
Negabinary: If we have inconsistent branches where one branch has type (Int, Float) and another has (Int, Int), currently both branches will be cast to unknown and the whole case will have type unknown. If we could instead do some sort of meet we could have them cast to (Int, ?); this makes no functional difference to the program, only reduces the number of casts.

src/haz3lcore/statics/MakeTerm.re
Negabinary: I think the reason I left it like this is specifically because maketerm provides its own ids instead of letting fresh generate them. If you're going to make your own ids, I want you to pinky promise that the ids are unique by setting the copied flag. Come back at me if you disagree.

src/haz3lcore/lang/term/Typ.re
Negabinary: I like `Typ.temp`

src/haz3lcore/statics/ConstructorMap.re
Negabinary: now thinking of getting rid of the failwith as per https://github.com/hazelgrove/hazel/pull/1197#discussion_r1593103681

src/haz3lcore/statics/Kind.re
Negabinary: image
yeah it doesn't exist on my pc lol

src/haz3lschool/Exercise.re
Negabinary: agree, will move

src/haz3lweb/view/dhcode/DHCode.re
Negabinary: It's being used to highlight static pattern errors - DHCode is not long for this world though, and I think we'll be running a new statics pass on results soon.

src/haz3lcore/statics/Statics.re
Negabinary: in what way?

src/haz3lcore/lang/term/Typ.re
Negabinary: sure

Negabinary:
To your general questions, @disconcision 

1. I've called TypeAnns casts because that data structure is both used for type annotation and pattern casts. The dream is to also eventually have expression annotations which pare the parallel for expressions using the cast data structure. When used for annotations, one of the arguments to a cast is always ignored.
image

2. Yes, DH is fully deprecated in #1297 - the only reason we would keep the terminology is to make a distinction between expressions that have all their casts and expressions that have implicit recursion.

3. Yeah I think that's a valid observation, we possibly could have had a separate data structure just for the statics, but it might make it harder to share code between casting logic and statics code (not that I'm sure we do that much?). I also think I could have come up with some more consistent story for how ids flow through type checking, but I was lazy and just made them all new because I had no use for the type ids yet.

src/haz3lcore/lang/term/Typ.re
Negabinary: Actually I'm having second thoughts because then it'll conflict withe type that's called `Typ.term`

src/haz3lcore/lang/term/Typ.re
Negabinary: the answer is yes we do still need it, it's the id of the expression the type was synthesized from not the type itself

src/haz3lcore/lang/term/Typ.re
cyrus-: @Crazycolorz5 ^

cyrus-: I think this is okay since its just generating fresh variables locally?

cyrus-:
* Needs some conflicts resolved
* The polymorphism casting seems to have some issues (see Polymorphism under documentation):

image

Crazycolorz5:
The casting in the documentation page seems to be an issue with `->` associativity -- adding parens like so removes that:
image
However, after doing so, I *do* get an "InvalidBoxedIntLit" error, which I am now investigating.

Crazycolorz5:
The current branch also seems to not inherit bound names in stepped code / function output. E.g. in Cyrus's comment above you can see several <anon fns>. Additionally, see the screenshot:

image


See (c11977781970a509dab111625f1253f9083947b6) also for how this was extended to type functions.

Crazycolorz5:
image

Seems to be not a polymorphism in particular. This is a minimal reproducible example. It seems to be an error in the term_of -- somehow the internal variable binding isn't being looked up?

Negabinary:
Ah yeah thanks @Crazycolorz5 - it was an issue with the transition rules for builtin application not using the evaluated version of the argument.

cyrus-:
@Negabinary couple of conflicts to resolve here

cyrus-:
@Negabinary looks like we've got quite a few merge conflicts here -- can you resolve those?

cyrus- (approved):

PR #1200, opened 2024-02-07
Add Coq file export to stepper button instead of console logging

nskh:
add .v file export to StepperView, CoqExport, and Update.

- `CoqExport.re`: Changes return type of `CoqExport.exportCoq` to `string` from `unit`
- `StepperView.re`: Writes file `stepper-coq-export.v` using `JsUtil.download_string_file`
- `Update.re`: Pipes stepper output to `print_endline` (this can be removed later and is marked with a TODO)


jchen-cs (approved):
Looks good!

PR #1202, opened 2024-02-07
DHExp error holes fixes

### LINKED ISSUE #496: Elaborated expression decorations are misaligned
image

In the above example, it seems like the width of the elaborated holes is shorter than the incremental width added to the inconsistent type operator decoration.

### LINKED ISSUE #1201: NonEmptyHole Decoration prevents clicks
In the screenshot below, you can't click on the green box because it is under the error decoration's hitbox.

image

@disconcision also of interest: it turns out the two boxes are coming from the fact that the free variable gets elaborated to `NonEmptyHole(_, _, _, FreeVar(_))`

disconcision:
Fixes #1201.
Fixes #496.
Fixes: double hole on free vars
Fixes: error hole misalignments during stepping

 As usual the latter was caused by someone adding magic margin/padding/offset numbers to compensate for layout issues previously seemingly addressed but not really through more magic numbers. I'll remind everyone that I'm the only one allowed to add magic numbers to seemingly but not really address layout issues that are actually caused by more magic numbers.


Negabinary:
Looks like this also fixes #496?

disconcision:
lol pretty sure i fixed that one long ago and this one is but a variation on a common theme of decoration layer misalignment but sure i'll take it

cyrus-:
Still some misalignment 
image

disconcision:
@cyrus- that one's a completely different bug; seems specific to show function bodies option. it appears to be deep dhdoc esoterica... i have apparently fixed it by making a change i don't understand do a function i don't understand (the only non-css change here: https://github.com/hazelgrove/hazel/pull/1202/commits/e6edd305c635d866fa3d8993794bf061e1e58875), so take that for what it's worth.

PR #1203, opened 2024-02-08
Fix fixpoint semantics

### LINKED ISSUE #1198: Fixpoint evaluation rules have bad edge cases
I'm pretty sure this should evaluate to 3 not 13.

image

I think this can be fixed by re-implementing the rules in https://github.com/hazelgrove/phi/blob/15-environment-based-evaluation/15-environment-based-evaluation/assets/eval-environment-report.pdf

Negabinary:
Fixes #1198 

In the end I've gone for the following semantics:

With environment `s`, `fix f. e` steps to `Closure(s, fix f. e)`              [Rule FixClosure]
then `Closure(s, fix f. e)` as `d` steps to `Closure({s, f <- d}, e)`        [Rule FixUnwrap]

I wonder if @mirryi or @cyrus- might be up for trying find a way to break these semantics?

Unfortunately, this has led to a roughly 2x speed regression for fib.

I will attempt a brief not-quite-formal proof of correctness:

Denote closure-removing post-processing that takes an environment and substitutes all variables in the environment along with any variables in closures in the expression. we will denote this operation `[[s]](e)`. We wish to show that if `X` steps to `Y` with environment `s`, then `[[s]](X)` and `[[s]](Y)` evaluate to the same value, which I will denote `~=` since I haven't LaTeXed this.

Case FixClosure:

`[[s]](fix f. e)`  = `[[s]](Closure(s, fix f. e))`   just by entering the closure on the right hand side. Thus this case is trivial.

Case FixUnwrap:

Starting with X,
`[[s]](Closure(s', fix f.e))` 
= `[[s']](fix f. e)`                        by entering the closure
~= `[[s']](e[fix f. e/f])`                 by applying substitution definition of fixpoints
= `[[s', f <- [[s']](fix f. e)]](e)`        by adding an extra substitution to the closure

Starting with Y
`[[s]](Closure({s', f <- d}, e))` 
= `[[s', f <- d]](e)`                                   by entering the closure
= `[[s', f <- Closure(s', fix f. e)]](e)`         using definition of d
= `[[s', f <- [[s']](fix f. e)]](e)`                   by entering the closure

thus we have shown that X and Y evaluate to the same value

And we assume all other cases are correct because obviously the rest of Hazel is 100% bug free.


cyrus-:
So I'm wondering why this result has a free variable `f` in the function body, rather than a fixpoint? Is that having to do with post-processing or something else?

Negabinary:
@cyrus- I'm not sure I understand your question. I'm using the notation `fix f. e` where f is a new variable binding and f is bound within expression e.

i.e. `fix f. e` = `fix(fun f -> x)` in some others' notation. `fix f. e` reflects the current construct in DHExp. 

In general, I'm pretty sure we only ever come across fixpoints of the form `fix(fun f -> fun x -> e)` in hazel, because expressions like `fix(fun f -> f + 1)` would never be able to terminate, so from a backend perspective I've been tempted to change the construct to `FixF(f, x, e)` to make it safer, but I think I'm rambling and not answering your question.

cyrus-:
Sorry I forgot to attach the image with my comment 😓 : 

image

I'm wondering why `fib` here is not listed as something like `<fib>` or, when the Show Fixpoints option is on, the full fixpoint?

cyrus-:
just noting what we discussed in the meeting:
- need to fix arrow spacing
- distinguish hidden fixpoints from hidden functions syntactically using `<name*>` vs `<name>`

cyrus-:
Distinguishing the hidden fixpoints syntactically is not exactly what I wanted in retrospect. What I want is to distinguish between:

1. `<fac>` which is an abbreviation for `fix fac -> fun n -> if ... else n * fac(n - 1)`
2. `<fac+>` which is the one step unrolling of the fixpoint, i.e. an abbreviation for `fun n -> if ... else n * <fac>(n - 1)`

I'd like to be able to set it up so the stepper trace looks like the following:

```
  let fac = <fac> in fac(4)
= let fac = <fac+> in fac(4)
= <fac+>(4)
= if 4 < 2 then ... else 4 * <fac>(4 - 1)
= if false then ...
= 4 * <fac>(4 - 1)
= 4 * <fac+>(4 - 1)
= 4 * <fac+>(3)
= ...
```

(That is more or less what I do in 490 lecture.)

This may be beyond the scope of this PR, so I'm okay with moving it to an issue and for now going back to not distinguishing `<fac>` from `<fac*>`.

Negabinary:
image

ok, this is what you get now if `show fixpoints` is on and `show function bodies` is off - it's almost the opposite of the "*" thing.

cyrus-:
@Negabinary there is a failing test and a conflict to resolve

PR #1204, opened 2024-02-10
opam upgrade + reformats induced by upgraded reason formatter

cyrus-:
Noticed when working on the tests PR that our library versions are quite a bit behind, so did an upgrade. The newest Reason formatter makes some different choices than the one we have been using, so this includes those changes.


disconcision:
oo nice js of ocaml upgrade, I was going to suggest this, has some good modernizations

PR #1211, opened 2024-02-11
Continuous Integration

### LINKED ISSUE #1208: run unit tests as a GitHub Action
to save resources, it should avoid needing to build it all a second time just to run the tests -- use a cache generated by the initial build action

7h3kk1d:
~~The main thing we need to decide on is if/how we would want to break up into multiple workflows/jobs. Given that everything can run on the same runner, the only advantage in splitting into multiple jobs is to allow rerunning deployment, in the event of failures, without rerunning the build.~~

~~I can't see a reason to split the deploy/tests into separate workflows unless we plan on doing those on different events. Right now the plan is to build/deploy on every commit.~~

Using a single job and setting the test status check manually. The builds seemed really slow and not caching nicely when I split up the job.


Makefile
```diff
@@ -52,8 +52,7 @@ repl:
 	dune utop src/haz3lcore
 
 test:
-	dune build @src/fmt --auto-promote src --profile dev
-	node $(TEST_DIR)/haz3ltest.bc.js test
+	dune runtest
```
7h3kk1d: I'm not sure if this is running natively or using JS so I might need to rollback the change

.github/workflows/deploy_branches.yml
```diff
@@ -1,10 +1,49 @@
...
+        run: |
+          eval $(opam env)
+          make release
+        working-directory: ./source
+      - name: Cache the build
+        id: build-cache
+        uses: actions/cache@v4
+        with:
+          path: ./source/_build'
+          key: ${{ runner.os }}-build-${{ hashFiles('./source/**', '!./source/_build/**', '!./source/.git/**') }}
```
7h3kk1d: This isn't working. I'm guessing `_build` is cached using file timestamps instead of a hash of the source files so it rebuilds everything. Going to try https://github.com/marketplace/actions/set-up-ocaml with the dune-cache

Makefile
```diff
@@ -2,6 +2,8 @@ TEST_DIR="$(shell pwd)/_build/default/src/test"
 HTML_DIR="$(shell pwd)/_build/default/src/haz3lweb/www"
 SERVER="http://0.0.0.0:8000/"
 
+.PHONY: test clean
```
7h3kk1d: There are other phony targets. I just started with these two.

.github/workflows/deploy_branches.yml
```diff
@@ -10,39 +10,67 @@ jobs:
...
+        continue-on-error: true
+        run: |
+          eval $(opam env)
+          node _build/default/src/test/haz3ltest.bc.js > test_output
+          if [ $? -eq 0 ]; then
+            echo "::set-output name=tests_passed::true"
+          else
+            echo "::set-output name=tests_passed::false"
+          fi
+        working-directory: ./source
```
7h3kk1d: I separated the steps so that we can track durations separately. When the whole installation was one step it was hard to see how much time was on dependencies vs the build itself.

PR #1217, opened 2024-02-16
Fix missing nested instrumentation

tonyfettes:
When the instrumentations is nested, old implementation would remove the outer ones to optimize for the size of stepper expression. However, this cause the regression that nested instrumentation won't work and stepper can't fallback to correct pause/skip behavior when a instrumented sub-expression finished evaluation.


cyrus-:
@tonyfettes what is an example that would demonstrate the difference between the old and new code?

@Negabinary can you approve this PR before we merge into `dev`?

Negabinary (approved):

Negabinary:
as long as @tonyfettes is fine with the last couple changes I pushed to this branch, I'm happy with it

tonyfettes:
@cyrus- 

Example:
```
eval 1 + 2 + 3 + 4 in
pause 3 + 3 + 4 in
1 + 2 + 3 + 4
```

Actual behavior on dev:
```
== 3 + 3 + 4
== 6 + 4
== 10
```
 
image

Expected behavior:
```
== 3 + 3 + 4
== 10
```

image

Current behavior:
```
== 6 + 4
== 10
```
image

Replace `s.next` with changed `next'` would fix the issue, but I'm not sure if would break the saving of stepper state.

cyrus-:
@tonyfettes marking this as draft until you resolve the issues you mentioned on Slack

cyrus-:
`debug` seems to prevent further steps

image

tonyfettes:
@cyrus pause works, but debug don't. It works on mini-stepper, so I suspect there are some bugs when we are generating/displaying the evaluation contexts. I have been long not working on the hazel codebase and might need read some code to regain momentumn.

I'm marking this is a draft for now until the bug is fixed.

cyrus-:
Going to go ahead and merge this and we can make a separate PR for the debug issue.

PR #1218, opened 2024-02-20
Leaf Projectors

disconcision:
## Overview

This PR is intended to start (but not finish) a system for flexible projectional code-editing UI. Right now, the system should be sufficient to serve as the basis for non-splicing livelits, type hold inference UI, hiding parts of code in live views (e.g. function bodies), and at least a basic form of module collapsing. Both inline and block (multiline) GUIs are supported, though there is only a single reference implementation for the latter (`TextAreaCore`), so some details might not have ended up on the right side of the projector server / projector client divide. I want to support a range of use cases so I'm solicited feedback on the flexibility of the interface.

## Getting Started

- See the `Documentation/Projectors` slide in the editor for a functional overview
- The module signature for projectors is defined in `ProjectorBase.Projector`
- The core logic is distributed across four modules: `ProjectorBase`, `Projector`, `ProjectorPerform`, and `ProjectorView`
- The first two are divided only to avoid cycles. This induces a similar split of `Zipper` into `ZipperBase`
- All of the above modules are documented, as are the projector actions definitions in `Action.t`
- The implementations of the actual projectors are in modules suffixed with Core. These currently include `FoldCore`, `InfoCore`, `CheckboxCore`, `SliderCore`, `SliderFCore`, and `TextAreaCore`.

## How the projector system interacts with the parent editor 

- The projector system covers bits of syntax in the parent editor with embedded GUIs. Projectable syntax is currently limited to convex pieces (both tiles and grout). (In the future, this can be ~ straightforwardly relaxed to any segment which is semantically self-contained: any segment which parses, including nullary parses like runs of secondary.)
- A projection is created/removed by wrapping/unwrapping the underlying syntax with a new `Piece.t` type, `Projector(Base.projector_kind)`, which stores a record entry specifying the kind of projector, the syntax itself, and an optional serialized model for internal projector state. Projectors have an associated UUID which is set to that of their wrapped syntax. When their wrapped syntax is transformed, the id for that syntax is reset to the projector id to retain sync. (This is mostly an implementation detail; a convenience to retain an association between a projector and the id-tracked semantic information if its wrapped syntax)
- Whenever a projector needs to be seen or respond to a message, its `kind` is hydrated into a first class projector module of type ProjectorBase.Projector, whose various methods determine all projector-specific logic. Projector views have access to certain information fed to them by their parent editor, including their underlying syntax and static `Info` entry (later: dynamic info). Views are also provided with callbacks for both local (kind-specific) actions and parent editor requests.
- MakeTerm unwraps (removed) projectors as the term is constructed, producing as a side-effect a map of ids to projectors which is used for the view.
- For editor view purposes, when a Projector is encountered in code rendering, it is represented as a placeholder string consisting of only whitespace and linebreaks. The projector GUIs are drawn separately (using the projectors map generated above in maketerm) and are absolutely positioned with respect to the parent editor.
- For editor move/select/edit actions, projector pieces are handled analogously to single-character monotiles with a fixed mold. This makes them roughly analogous to a hybrid between single-character monotiles and the whitespace piece type: movement/selection/deletion treats them as atomic units, except they have a fixed convex shape.

## Caveats and Wontfixes

- Currently only convex pieces can be projected, so many terms will have to be wrapped in parens to be e.g. folded. This will be relaxed in a subsequent PR
- Projectors are persisted in the syntax, but are not preserved on actions which serialize to text like copy/cut/paste. This will be mitigated in a subsequent PR
- Substantial adjustments affecting only one specific projector (that is, requiring more internal logic but not an API change) are probably outside the scope of this PR. I already have enhancement ideas for basically all the individual projectors here, especially Fold, Info, and Slider.
- Substantial adjustments to the UI for adding/removing projectors is probably outside the scope of this PR. The keyboard shortcuts and projectors panel are intended to be basic generic invocation mechanisms; I expect that different genres of projectors will have custom invocation stories

## Adding a new kind of projector

1. Create a new core module implementing `ProjectorBase.Projector` (or copy e.g. `FoldCore`)
2. Add an entry for it in `Base.projector_kind` (used for adding/removing/updating)
3. Register the core module in `Projector.to_module`
4. If you want to expose the projector via a keyboard shortcut, see the existing entry for `Fold` in `Keyboard`
5. If you want to expose the projector in the projector panel bottom bar UI, update `ProjectorView.name`, `ProjectorView.of_name`, and `ProjectorView.Panel.applicable_projectors`
6. If you want to manually manage the projector as part of the update cycle, see the implementations of the `SetIndicated` and `Remove` actions in `ProjectorPerform` for how to manually add/remove projectors from an editor

## Other changes made in this PR

1. Many editor-specific actions have been relocated from Update to Perform
2. Statics storage and calculation has been moved to Editor (requiring settings to be piped to many modules)
3. Adding projectors to Zipper induces a serialization change!
4. Since the above broke some of the documentation slides (due to eg eval now being a reserved keyword), I updated these, in particular Basic Reference, with new language features

### TODO(andrews)

- [x] View: Text layer: Render placeholder instead of projected term
- [x] View: Deco layer: Render projected UI over placeholder
- [x] Action: Move: Skips over the projected term
- [x] Action: Select: Selects through and over the projected term
- [x] Action: Indicated projections override keyboard handler
- [x] View: Deco layer: Indicated deco for projected UI
- [x] Action: Update syntax under projector
- [x] Bug: Projector statics update is delayed one action
- [x] Bug: Clicking on projector only seems to register if caret is on right side
- [x] Bug: Action: Can't properly fold more than one term in the same segment
- [x] Bug: Action: Movement into concave side of folded infix/prefix/postfix crashes
- [x] Bug: Action: Under some conditions moving/selecting past a fold in a nested context skips to end of program
- [x] Bug: In school mode, typechecking and eval are using display version of terms... see TODOs
- [x] Cleanup: editor state/meta
- [x] Deal with indication in a more principled way than the find_p hack
- [x] Core:  Factor out projectors into first class modules
- [x] Action: Gate folding on tile shape (prevent nonconvex)
- [x] View: Folded term decoration should be hexagon shaped probably
- [x] Move positioning code outside specific view fns
- [x] Internalize actions
- [x] Add proper internal action type
- [x] Try to remove statics from action
- [x] Cleanup/rename syntax_map
- [x] Cleanup TODOs
- [x] Completion buffer doesn't show (it still works though, you can tab-complete blind)
- [x] In Exercises mode, the type info projector placeholder size is wrong
- [x] When drag-selecting, projector deco stays in indicated mode until you're entirely past it
- [x] langdocs selector UI has become misaligned (how)
- [x] redo type arrow precedence correction on polymorphism slide
- [x] indentation for last delimiter of case/lets is broken
- [x] When you undo textarea typing when textarea isn't focused, textarea content doesn't update (although the underlying syntax does)
- [x] No error hole dec lines
- [x] Tab-Completing "let" doesn't work
- [x] Click to move at end of line too enthusiastic


src/haz3lcore/JsUtil.re
```diff
@@ -160,3 +182,83 @@ module Fragment = {
...
+        let line_length = String.length(line);
+        if (cur_pos <= line_length) {
+          {row, col: cur_pos};
+        } else {
+          find_position(rest, cur_pos - line_length - 1, row + 1, 0);
+        };
+      };
+    };
+    let lines = lines(textarea);
+    let caret_position = textarea##.selectionStart;
```
7h3kk1d: Does this blow up if the textarea isn't selected?

src/haz3lcore/JsUtil.re
```diff
@@ -160,3 +182,83 @@ module Fragment = {
...
+    let {row, col} = caret_pos(textarea);
+    {
+      rows: rel(row, List.length(lines) - 1),
+      cols: rel(col, String.length(List.nth(lines, row))),
+    };
+  };
+
+  let set_caret_to_end = (textarea: t): unit => {
+    textarea##focus;
+    let content_length = String.length(Js.to_string(textarea##.value));
```
7h3kk1d: It's probably worth pulling out a `content` function to get the string out of the textarea since we're doing it above too.

src/haz3lcore/Measured.re
```diff
@@ -168,7 +163,10 @@ let singleton_s = (id, shard, m) => empty |> add_s(id, shard, m);
 
 // TODO(d) rename
 let find_opt_shards = (t: Tile.t, map) => Id.Map.find_opt(t.id, map.tiles);
-let find_shards = (t: Tile.t, map) => Id.Map.find(t.id, map.tiles);
+let find_shards = (~msg="", t: Tile.t, map) =>
```
7h3kk1d, line 145: I have a todo on my list to look into generating stack traces on exceptions so we don't have to pass messages everywhere. Is there any other reason you need these?

src/haz3lcore/Measured.re
```diff
@@ -363,14 +374,28 @@ let of_segment = (~old: t=empty, ~touched=Touched.empty, seg: Segment.t): t => {
...
             let add_shard = (origin, shard, map) => {
+              let token = List.nth(t.label, shard);
+              // Adjustment for multi-line tokens e.g. projector placeholders
               let last =
                 Point.{
-                  ...origin,
-                  col: origin.col + String.length(token(shard)),
+                  col: origin.col + StringUtil.max_line_width(token),
+                  row: origin.row + StringUtil.num_linebreaks(token),
                 };
```
7h3kk1d: Might be worth adding some more commentary as to why the last point is the max width

src/haz3lcore/Measured.re
```diff
@@ -363,14 +374,28 @@ let of_segment = (~old: t=empty, ~touched=Touched.empty, seg: Segment.t): t => {
...
+                switch (n) {
+                | 0 => map
+                | _ =>
+                  map
+                  |> add_n_rows(n - 1)
+                  |> add_row(
+                       origin.row + n - 1,
+                       {indent: row_indent, max_col: origin.col},
+                     )
+                };
```
7h3kk1d: As a matter of style it might be clearer as a reduction. I couldn't find a builtin range function though.

```reasonml
let add_n_rows = (n, map) => List.fold_left((map, i) => add_row(origin.row + i, {indent: row_indent, max_col: origin.col},
 map), range(n));
 ```

File: src/haz3lcore/statics/TypBase.re [hunk 6]
```diff
@@ -492,11 +492,12 @@ module rec Typ: {
     | Forall(_, _) => true
     | List(_) => false /* is already wrapped in [] */
     | Arrow(_, _) => true
-    | Prod(_)
+    | Prod(_) => false /* is already wrapped in () in this printer */
     | Sum(_) => true /* disambiguate between (A + B) -> C and A + (B -> C) */
     };
```
7h3kk1d: I'd prefer to either collapse the true/false cases or make them all explicit.
```reason
let needs_parens = (ty: t): bool =>
    switch (ty) {
    | Unknown(_)
    | Int
    | Float
    | String
    | Bool
    | List(_) /* is already wrapped in [] */
    | Prod(_) /* is already wrapped in () in this printer */
    | Var(_) => false
    | Rec(_, _)
    | Forall(_, _)
    | Arrow(_, _)
    | Sum(_) => true /* disambiguate between (A + B) -> C and A + (B -> C) */
    };
```

7h3kk1d: I wonder if we should move all the pretty printing code to a different module

src/haz3lcore/tiles/Segment.re
```diff
@@ -754,3 +754,23 @@ let rec get_incomplete_ids = (seg: t): list(Id.t) =>
 
 let ids_of_incomplete_tiles_in_bidelimiteds = (seg: t): list(Id.t) =>
   get_childrens(seg) |> List.concat |> get_incomplete_ids;
+
+let push_right = ((ls: t, rs: t)): (t, t) =>
+  switch (ls |> List.rev) {
```
7h3kk1d: Is there a reason we're piping as opposed to `List.rev(ls)`?

src/haz3lcore/tiles/Segment.re
```diff
@@ -754,3 +754,23 @@ let rec get_incomplete_ids = (seg: t): list(Id.t) =>
...
+  | [l, ...ls] => (ls |> List.rev, [l, ...rs])
```
7h3kk1d: It's probably worth adding an `unsnoc` to ListUtil:

```reason
let rec unsnoc = (xs: list('a)): option(('a, list('a))) => {
  switch ((xs: list('a))) {
  | [] => None
  | [head, ...tail] =>
    switch (unsnoc(tail)) {
    | None => Some((head, []))
    | Some((tail_last: 'a, tail_init: list('a))) =>
      Some((tail_last, [head, ...tail_init]))
    }
  };
};
```

src/haz3lcore/zipper/Editor.re
```diff
@@ -54,35 +98,70 @@ module Meta = {
...
+  let next_statics =
+      (
+        ~settings: CoreSettings.t,
+        a: Action.t,
+        z: Zipper.t,
+        old_statics: CachedStatics.t,
+      ) =>
+    if (!settings.statics) {
+      CachedStatics.empty;
+    } else if (!Action.is_edit(a)) {
```
7h3kk1d: I the idea here that we want to cache around cursor movement since the statics shouldn't change but the zipper will have?

src/haz3lcore/zipper/Printer.re
```diff
@@ -60,69 +60,43 @@ let to_rows =
   };
 };
 
-let pretty_print = (~measured: Measured.t, z: Zipper.t): string =>
+let pretty_print =
+    (~holes: option(string)=Some(""), ~measured: Measured.t, z: Zipper.t)
```
7h3kk1d: Seems like this doesn't need to be optional anymore

src/haz3lcore/zipper/Projector.re
```diff
@@ -0,0 +1,315 @@
...
+
+/* This handles the logic for selecting around projectors.
+ * This amounts to simply selecting over the projected syntax
+ * as if it were atomic instead of selecting into it. In principle,
+ * this should work the same as if the projected syntax was replaced
+ * by its placeholder, the selection was made on that projections,
+ * and then the placeholder was replaced by former syntax. In an
+ * updated syntax model, it should maybe work this way, or else
+ * projected syntax should be represented in the zipper in
+ * a more first-class way */
```
7h3kk1d: I think it would be nice in the future if we moved towards having the projected syntax first-class in the zipper.

disconcision: This may be forced when adding splices, or at least the materialized view / edit state in that case will be sufficiently weighty that it's required for reasonable perf. The complexity here lies in deciding what is the common persistent state versus what is view and/or user dependent, so decisions here are ultimately going to deeply involve both grove and @dm0n3y's syntax model. I've been trying to avoid rewriting the syntax layer but there's definitely some complexity cost to that here

src/haz3lcore/zipper/Printer.re
disconcision: perhaps we can agree on a canonical textual representation for holes

src/haz3lcore/zipper/Editor.re
disconcision: yes. this optimization hasn't changed, it's just relocated from Update

src/haz3lcore/tiles/Segment.re
disconcision: will do. wish pattern matching just supported this like in civilized languages...

src/haz3lcore/Measured.re
disconcision: no, only stack traces

src/haz3lcore/tiles/Segment.re
7h3kk1d: :popcorn: :popcorn: 

I get the vibe they don't want you to do it for performance reasons. We could consider an immutable vector or an immutable queue for one side if it becomes an issue.

src/haz3lcore/zipper/Printer.re
7h3kk1d: ⦿

src/haz3lcore/zipper/Projector.re
7h3kk1d: Outside the scope of the PR but it is probably useful for us to start having those conversations sooner than later. Even for the view layer I think having a data structure that represents what is viewed fully makes it easier to reason about and test. So we might just need more types.

src/haz3lcore/zipper/Projector.re
```diff
@@ -0,0 +1,315 @@
...
+      )
+    | _ => None
+    };
+};
+
+module MapPiece = {
+  type updater = Piece.t => Piece.t;
+
+  let rec of_segment = (f: updater, seg: Segment.t): Segment.t => {
+    seg |> List.map(p => f(p)) |> List.map(of_piece(f));
```
7h3kk1d:
```suggestion
    seg |> List.map(f) |> List.map(of_piece(f));
```

src/haz3lcore/zipper/Projector.re
```diff
@@ -0,0 +1,315 @@
...
+  and of_piece = (f: updater, piece: Piece.t): Piece.t => {
+    switch (piece) {
+    | Tile(t) => Tile(of_tile(f, t))
+    | Grout(_)
+    | Secondary(_) => piece
+    };
+  }
+  and of_tile = (f: updater, t: Tile.t): Tile.t => {
+    {...t, children: List.map(of_segment(f), t.children)};
+  };
```
7h3kk1d: I'm still trying to understand exactly what this module is doing but could the updater just be on `tile -> tile` as opposed to `Piece.t -> Piece.t`?

src/haz3lcore/zipper/ProjectorBase.re
```diff
@@ -0,0 +1,175 @@
...
+
+let info_init = (p: syntax) => {id: Piece.id(p), syntax: p, ci: None};
+
+/* To add a new projector, implement this module signature */
+module type Projector = {
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type model;
+  /* The internal model type of the projector which will
+   * be serialized and persisted. Use `unit` if you don't
+   * need other state beyond the underlying syntax */
```
7h3kk1d: Not trying to be nitpicky but can we commit to docstrings either above or below defns?

src/haz3lcore/zipper/ProjectorBase.re
```diff
@@ -0,0 +1,175 @@
...
+   * effects but could be extended in the future to
+   * take/return the model if the projector needs to
+   * maintain a complex internal position state */
+};
+
+/* Projector model and action are serialized so that
+ * they may be used by the Editor without it having
+ * specialized knowledge of projector internals */
+type serialized_model = string;
+type serialized_action = string;
```
7h3kk1d: I would like to talk more about this and see if there's some way we can hide the internals from the `Editor` while still having the rich types internal to the projector.

src/util/dune
```diff
@@ -1,14 +1,14 @@
...
- (libraries base ptmap incr_dom)
+ (libraries re base ptmap incr_dom)
  (js_of_ocaml
   (flags
    (:include js-of-ocaml-flags-%{profile})))
  (preprocess
   (pps ppx_let ppx_sexp_conv ppx_deriving.show ppx_yojson_conv)))
 
 (rule
- (write-file js-of-ocaml-flags-dev "(:standard --debuginfo --noinline)"))
```
7h3kk1d, line 11: Why this change?

src/util/StringUtil.re
```diff
@@ -24,3 +24,42 @@ let repeat = (n, s) => String.concat("", List.init(n, _ => s));
...
+let regexp: string => regexp = Js_of_ocaml.Regexp.regexp;
+
+let match = (r: regexp, s: string): bool =>
+  Js_of_ocaml.Regexp.string_match(r, s, 0) |> Option.is_some;
+
+let replace = Js_of_ocaml.Regexp.global_replace;
+
+let split = Js_of_ocaml.Regexp.split;
+
+let to_lines = (s: string): list(string) => split(regexp("\n"), s);
```
7h3kk1d: Marginal improvement but I'd expect String.split_on_char to be faster

src/util/StringUtil.re
```diff
@@ -24,3 +24,42 @@ let repeat = (n, s) => String.concat("", List.init(n, _ => s));
...
+  s |> to_lines |> List.map(String.length);
+
+let max_line_width = (s: string): int =>
+  s |> line_widths |> List.fold_left(max, 0);
+
+let num_lines = (s: string): int => s |> to_lines |> List.length;
+
+let num_linebreaks = (s: string) =>
+  List.init(String.length(s), String.get(s))
+  |> List.fold_left((acc, c) => c == '\n' ? acc + 1 : acc, 0);
```
7h3kk1d: I would expect this to be `num_lines - 1`.

Alternative implementation:
```reason
let num_linebreaks = (s: string) =>{
  s |> String.to_seq |> Seq.filter((==)('\n')) |> Seq.length;
  }
  ```

File: src/haz3lweb/Editors.re [hunk 19]
```diff
@@ -39,7 +39,35 @@ let put_editor = (ed: Editor.t, eds: t): t =>
     Exercises(n, specs, Exercise.put_editor(exercise, ed))
   };
 
-let get_zipper = (editors: t): Zipper.t => get_editor(editors).state.zipper;
+let update = (f: Editor.t => Editor.t, editors: t): t =>
+  editors |> get_editor |> f |> put_editor(_, editors);
```
7h3kk1d, line 43: Can be implemented with update_opt

src/haz3lweb/Model.re
```diff
@@ -110,16 +109,17 @@ let save_and_return = (model: t) => {
...
   Ok(model);
 };
+
 let reset = (model: t): t => {
   /* Reset model to default, including in localstorage,
      but don't otherwise erase localstorage, allowing
      e.g. api keys to persist */
   let settings = Store.Settings.init();
-  ignore(settings);
+  let settings = settings.core;
```
7h3kk1d: Can we combine these 2 let bindings so we're not immediately shadowing?

disconcision:
@cyrus- need to clean up from the fixes but i've resolved all the known bugs other than the undo one

src/haz3lcore/statics/TypBase.re
disconcision: Reasonable but reluctant to touch this as I don't think it will exist in this form after matt's merge

src/haz3lweb/Editors.re
disconcision: feel awkward to me. what do you do in the case where update_opt returns none? or i might be misunderstanding

src/util/dune
disconcision: accident...

src/haz3lcore/zipper/ProjectorBase.re
disconcision: that actually is done here. see the ProjectorBase.Cook functor

disconcision:
@cyrus- okay, pretty much cleaned up now. (there are still some rough patches in Deco)

src/haz3lweb/Editors.re
7h3kk1d: On the first readthrough I missed that the return types are different. This feels gross:
```reason
let map_editor = (f, editors: t) => {
  editors |> get_editor |> f
};

let update = (f: Editor.t => Editor.t, editors: t): t =>
  map_editor(f, editors) |> put_editor(_, editors)

let update_opt = (editors: t, f: Editor.t => option(Editor.t)): option(t) =>
  map_editor(f, editors) |> Option.map(put_editor(_, editors));
  ```

Comment (7h3kk1d) [reply src/haz3lweb/Editors.re:43]: What you have now feels great given the options. This problem kind of reminds me of https://reasonablypolymorphic.com/blog/higher-kinded-data/ where you want to interleave an effect through the data

File: src/haz3lweb/view/ProjectorView.re [hunk 21]
```diff
@@ -0,0 +1,309 @@
...
+/* A friendly name for each projector. This is used
+ * both for identifying a projector in the CSS and for
+ * selecting projectors in the projector panel menu */
+let name = (p: kind): string =>
+  switch (p) {
+  | Fold => "fold"
+  | Info => "type"
+  | Checkbox => "check"
+  | Slider => "slide"
+  | SliderF => "slidef"
```
Negabinary: slide sounds like PowerPoint to me, maybe "intslider" and "slider" instead of "slide" and "slidef"?

src/haz3lweb/Keyboard.re
```diff
@@ -105,13 +106,14 @@ let handle_key_event = (k: Key.t): option(Update.t) => {
...
-  | {key: D(key), sys, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
-    switch (sys, key) {
-    | (_, "ArrowLeft") => now(MoveToBackpackTarget(Left(ByToken)))
-    | (_, "ArrowRight") => now(MoveToBackpackTarget(Right(ByToken)))
-    | (_, "Alt") => Some(SetMeta(ShowBackpackTargets(true)))
-    | (_, "ArrowUp") => now(MoveToBackpackTarget(Up))
-    | (_, "ArrowDown") => now(MoveToBackpackTarget(Down))
+  | {key: D(key), sys: _, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
+    switch (key) {
+    | "ƒ" => Some(PerformAction(Project(ToggleIndicated(Fold))))
```
Negabinary: What's "ƒ"? I can't work out what the keyboard shortcut for projecting is

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -93,6 +94,12 @@ let kids_of_unsorted =
   | Post(l, tiles) => [l] @ kids_of_tiles(tiles)
   | Bin(l, tiles, r) => [l] @ kids_of_tiles(tiles) @ [r];
 
+type data = {
```
Negabinary: Is this used? if so is it possible to give it a more specific name than data?

src/haz3lcore/statics/TypBase.re
Negabinary: I haven't gotten rid of this function yet, but maybe we should. Afaict it's only used in one very specific circumstance when function bodies are hidden:

```
let f = typfun t -> fun x:t -> x in
f@<Int>
```

Will print as `f@<Int>` because it appends the serialized type "Int" to the function's name. I'm reluctant to replace this with segment code because it runs during evaluation.

src/haz3lcore/zipper/Editor.re
```diff
@@ -1,50 +1,98 @@
...
-    unselected: Segment.t,
     segment: Segment.t,
-    view_term: Term.UExp.t,
-    terms: TermMap.t,
+    measured: Measured.t,
     tiles: TileMap.t,
     holes: list(Grout.t),
-    buffer_ids: list(Id.t),
+    selection_ids: list(Id.t),
+    term: Term.UExp.t,
```
Negabinary: How's the term here different to the one in CachedStatics? (I feel like I've seen a comment explaining this at some point but I can't find it any more)

src/haz3lcore/zipper/ProjectorBase.re
```diff
@@ -0,0 +1,155 @@
...
+
+/* A cooked projector is the same as the base module
+ * signature except model & action are serialized */
+module type Cooked = {
```
Negabinary: definition could more closely match the comment with:

```
/* A cooked projector is the same as the base module
 * signature except model & action are serialized */
module type Cooked =
  Projector with
    type model = serialized_model and type action = serialized_action;
```

but it would require

```
module Cook = (C: Projector) : Cooked => {
  type model = serialized_model;
  type action = serialized_action;
  ...
```

src/haz3lcore/zipper/ProjectorBase.re
Negabinary: I really like the Cook functor solution and am weighing up whether it might have applications to UI

src/util/Web.re
```diff
@@ -12,3 +16,115 @@ let div_if = (p, ats, ns) => p ? div(~attrs=[ats], ns) : div_empty;
 let span_if = (p, ats, ns) => p ? span(~attrs=[ats], ns) : span([]);
 
 let unless = (p, a) => p ? Effect.Many([]) : a;
+
```
Negabinary, line 19: personally these feel like `view/Widgets.re` things to me instead of `util/Web.re` things

src/util/Util.re
```diff
@@ -15,6 +15,10 @@ module StringUtil = StringUtil;
 module TimeUtil = TimeUtil;
 module TupleUtil = TupleUtil;
 module Web = Web;
+module JsUtil = JsUtil;
```
Negabinary, line 18: out of interest how long/frustrating was it working out that you had to put these here?

src/haz3lcore/zipper/projectors/TextAreaCore.re
Negabinary, line 1: The TextArea has some weird behaviour when you type a speech mark into it.

src/haz3lweb/Keyboard.re
disconcision: ha forgot to change this. that's what i have to do to make alt-f work on the mac as alt changes the character at the OS level

src/haz3lcore/zipper/Editor.re
disconcision: i can add a comment. these terms are identical when the backpack is empty. if the backpack is non-empty, then when we make the term for semantics, we attempt to empty the backpack according to some simple heuristics (basically try to put it down greedily wrt the current caret position). this is necessary to have the cursorinfo (and completion) work when the backpack is nonempty.

this is a brittle part of the current implementation. there are some comments at some of the weakest joints; the biggest problem is the dropping the backpack can add/remove grout, causing certain ids to be present/non-present unexpectedly. i see no alternative to this given the current syntax mode though

src/haz3lweb/view/ProjectorView.re
disconcision: was trying keeping things short here for UI purposes, but yeah they're not ideal. will try to find a way to make keeping the length of this string short less UI-critical. (relatedly: i am also considering just folding in both sliders to a single implementation, given that their implementations almost entirely overlap and they are never both applicable)

src/haz3lcore/statics/MakeTerm.re
disconcision: would term_data be more evocative? this is an ephemeral structure that exists to gather certain metadata, integrated into maketerm for perf reasons, that has no logical role and likely wouldn't exist in a future syntax model. can add a comment to that effect though

src/haz3lcore/zipper/ProjectorBase.re
disconcision: not sure yet what the limitations of this pattern are wrt perf (due to de/serialization). definitely fine for small ui state which is update discretely. not so sure for heavier state, or state that is update through continuous polling

src/util/Web.re
disconcision: hmm, i forgot about Widgets. maybe we should just combine these modules? do you see a distinction between them? right now i can't just move things to Widgets as it's in web (renegotiating what goes where is likely going to be an ongoing process now that the core/web division is changing)

src/util/Util.re
disconcision: i understood immediately when i saw your commit on the upgrade PR, but something something nth rodeo

src/haz3lcore/zipper/projectors/TextAreaCore.re
disconcision: just to confirm that you're seeing the same weirdness: quote characters get inserted fine in the textarea, but these actions are not reflected in the underlying syntax? (this also means the shape of the textarea doesn't expand for quotes) or are you seeing something above and beyond that?

i knew about this one but was ignoring it as i don't totally understand how to address it in a principled way. the way regular string literals work in the underyling syntax model is a bit ad-hoc; quotes have special-case behavior relative to all other characters. briefly: inserting a quote must automatically insert a closing quote; otherwise the text serialization of that code state has an ambiguous lexing. but we don't want to prevent someone from just typing an empty string normally. the solution we ended up with is that typing a quote inside a string simply moves the caret to the end of the string. not pretty but in practice seems to work fine. of course ultimately we want escaped characters in strings, but ideally this will all be handled at the core syntax layer. 

could add some textarea-projector-specific code that catches keystrokes, suppresses quotes, and the refires the event to the textarea i guess. implementing the string livelit with the DOM textarea is likely more trouble than it's worth, but i wanted to do this as a trial run for embedding non-hazel-native components...

src/haz3lcore/statics/MakeTerm.re
Negabinary: term_data's a little better, I might be tempted towards `result` or maybe even `t` if it's the type that collects all the return data of the main MakeTerm functions?

src/util/Web.re
Negabinary: To me web is just like wrappers around html things and widgets has more hazel-specific logic and styles etc - maybe we move widgets to util/Widgets.re if it can't be in web? (I still find it a little concerning that we need an html definition of a slider in core)

7h3kk1d: I had some similar issues on my livelits branch. I struggle to see how the projectors can be web (platform) agnostic. So unless they're entirely configured in web I don't know how you'd excise the dependency from core.

src/haz3lcore/zipper/projectors/TextAreaCore.re
Negabinary: yeah same weirdness! Hazel afaik doesn't have any quote escape mechanism so in terms of the underlying syntax the best thing is probably just to suppress quotes for now if that's straightforward enough

src/util/Web.re
cyrus-, line 1: nit: module naming in util is inconsistent, some end in Util and others don't.

src/haz3lweb/view/ProjectorView.re
cyrus-: agreed on "slider" rather than "slide"

src/haz3lweb/dune
```diff
@@ -93,7 +92,9 @@
    (flags (:standard)))))
 
 (rule
- (write-file js-of-ocaml-flags-dev "(:standard)"))
+ (write-file
+  js-of-ocaml-flags-dev
+  "(:standard --debug-info --no-inline --pretty --source-map)"))
```
cyrus-: does this change the dev experience meaningfully?

src/haz3lweb/Keyboard.re
cyrus-: are you intending to change this before merge?

src/haz3lcore/zipper/ProjectorBase.re
```diff
@@ -0,0 +1,155 @@
...
+ * position and dimensions in base editor could be
+ * added here if needed */
+[@deriving (show({with_path: false}), sexp, yojson)]
+type info = {
+  id: Id.t,
+  syntax,
+  ci: option(Info.t),
+};
+
+/* To add a new projector, implement this module signature */
```
cyrus-: maybe say what else you have to do all in this comment

src/haz3lcore/zipper/Selection.re
```diff
@@ -2,8 +2,8 @@ open Util;
 
 [@deriving (show({with_path: false}), sexp, yojson)]
 type buffer =
-  | Unparsed
-  | Parsed;
+  //| Parsed
```
cyrus-, line 5: what's going on here?

src/haz3lcore/zipper/projectors/CheckboxCore.re
cyrus-, line 1: why do these files need a `Core` suffix? maybe use `CheckboxProj` or something more indicative?

cyrus- (changes requested):
Looks and feels great overall.

Only bug I can find is that when the projectors panel is disabled you can still click the grayed out button (and it seems to do some sort of folding):

image

Other notes in addition to a few inline comments in the code:

- The copy-and-paste and undo issues you indicate as a wontfix for this PR, which I am okay with for now, but that I don't want to leave hanging too long. What is your plan for those?
- The first scratch slide now has a bunch of projectors stuff in it, which should instead be under Documentation.
- I was expecting to be able to fold when my cursor was on something like an ap, e.g. in the example above, where it would fold the entire highlighted term, but I guess that's related to the parenthesization issue that we're going to fix later?

src/haz3lweb/Keyboard.re
disconcision: yes

src/util/Web.re
disconcision: I like the pattern matt has now established of importing util modules in Util so you can just refer to them as e.g. Util.Web; I think going forward we should migrate away from Util in the filename.

src/haz3lweb/dune
disconcision: not afaict. but it does make the flags the same across all our build targets, and the same as david's setup, so at least it eliminates that potential source of confusion

src/haz3lcore/zipper/Selection.re
disconcision: This is an artefact of separating the original TyDi implementation from the llm completion stuff. The parsed buffer option is currently meaningless so I just removed it until the LLM stuff is reimplemented

disconcision:
commenting to remind myself about the slider keyboard input focus issue

disconcision:
> * The copy-and-paste and undo issues you indicate as a wontfix for this PR, which I am okay with for now, but that I don't want to leave hanging too long. What is your plan for those?

I've been hoping inspiration will strike for a simple approach to copy/paste. The options seem to be: Implement some kind of synchronization between an internal (segment) buffer and the browser copy/paste actions. This is probably not too bad, but I'm also just not totally sure if it's going to work seamlessly syntax-wise. Reinserting a segment as a selection should work (it's how pickup/putdown used to work), but i'm wary against since-then baked in assumptions around adding incomplete syntax. At the very least paste has to be able to replace UUIDs. Hopefully not that bad but it's not something I want to get bogged down in atm. The other option is implementing some kind of attributes system, and use those to represent projectors in the text representation. This requires some more planning, and would be natural to implement at the same time as text-based livelit invocation (not currently on my agenda, but would likely do at some point). Ofc this latter option doesnt have the advantage.

src/util/Web.re
disconcision: @Negabinary i started to move widgets but looking at it the current arrangement makes sense to me. the range and TextArea components i defined in web are intended to be completely independent from hazel logic. all of the textarea logic is just to get and set text area ui state; it could all be in vdom in-principle. otoh current widgets are all related to hazel UI chrome outside the syntax-editor as such. so these modules are disjoint along two different dimensions. admittedly thats not well-reflected in the naming

disconcision: @7h3kk1d contemplating this resulted in a tweet: https://twitter.com/disconcision/status/1818812875411968085

disconcision: (via the tangent "what does it mean to be truly view independent?" => "let's take string out of Core" => "let's string out")

src/haz3lcore/zipper/projectors/TextAreaCore.re
disconcision: yup, suppressing quotes for now

disconcision:
@cyrus- et al: all issues addressed I think. function applications don't try to project now (greyed-out toggle should be in-sync).

- Ability to project terms which aren't single tiles will follow in a self-contained PR
- Better Undo/Redo will follow at some point. I'll create an issue for now. Current state is: it works sensibly, but causes textarea to lose caret position
- Copy/Paste will follow at some point. I'll create an issue for now

cyrus-:
@disconcision tab-completing let/if/etc. doesn't work still, what's the issue there? 

Also we talked about connecting error marks with the lines so it doesn't look like there are multiple errors, did you decide against that or just not done?

disconcision:
> @disconcision tab-completing let/if/etc. doesn't work still, what's the issue there? 
> 
> Also we talked about connecting error marks with the lines so it doesn't look like there are multiple errors, did you decide against that or just not done?

planning to fix both just taking longer than id like

disconcision:
@cyrus- all issues addressed. in particular, tab completions of definitions is fixed, and error holes now have child lines.

cyrus-:
odd decoration on the `end` here:
image

disconcision:
@cyrus- technically that's already there for the term decorations, you just can't see if because the term shards are opaque. I  looked briefly but I don't really know how to address it in the current approach to line drawing

cyrus-:
@disconcision maybe make the error shards opaque as well to avoid the issue for now and @dm0n3y can take a look in the next revision of the decorations?

disconcision:
@cyrus- i've addressed the error hole thing in summer-refresh; tedious to replicate here as the css is completely different. i've made the background color of the error holes the same color as the code background + the red overlay, so it looks the same. the cost is that this means we need to either move the error holes behind selections or it looks super funky. but  that's not a huge issue i think

cyrus- (approved):

PR #1220, opened 2024-02-25
Integrate menhir parser and elaborator tests

7h3kk1d:
This PR introduces a [Menhir](https://gallium.inria.fr/~fpottier/menhir/)-based parser for Hazel's textual syntax, designed to support the elaborator's unit testing. The textual representation is supposed to support pre and post elaboration, so special syntax is used for post-elaboration forms. Key components include:

- Parser.mly: Defines the Menhir parser rules for Exp syntax.
- Lexer.mll: Implements the lexical analyzer for tokenizing input.
- Interface.re: Provides functions for parsing text into the parser's AST.
- AST.re: Defines the structure of the parser's Abstract Syntax Tree (AST). This is mostly equivalent to the TermBase AST but with no IDs or parens.
- Conversion.re: Includes utilities to convert the parser's AST into Haz3lCore TermBase types..


cyrus-:
@green726 are you taking a look at this PR or what is the plan? @7h3kk1d

7h3kk1d:
> @green726 are you taking a look at this PR or what is the plan? @7h3kk1d

This was still in dev from my perspective but @green726 is going to take over it from now and let me know if they need assistance.

src/haz3lcore/dune
```diff
@@ -2,7 +2,7 @@
 
 (library
  (name haz3lcore)
- (libraries util re sexplib unionFind uuidm)
+ (libraries util re sexplib unionFind uuidm hazel_menhir)
```
cyrus-: the directory is called menhir-parser and the library is called hazel_menhir. 

it may make more sense here to just put the parser in haz3lcore under a parser directory.

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
-    | InconsistentBranches(MetaVar.t, HoleInstanceId.t, case)
+    | InconsistentBranches(MetaVar.t, HoleInstanceId.t, case) //(_HOLE (case ...))
     | Closure([@opaque] ClosureEnvironment.t, t)
-    | Filter(DHFilter.t, t)
-    | BoundVar(Var.t)
-    | Sequence(t, t)
-    | Let(DHPat.t, t, t)
-    | FixF(Var.t, Typ.t, t)
-    | Fun(DHPat.t, Typ.t, t, option(Var.t))
+    | Filter(DHFilter.t, t) //pause a == 2 b
```
cyrus-: this can have a few different concrete forms depending on the filter... @tonyfettes

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
   | InconsistentIf;
 
 module rec DHExp: {
+  //All comments show the textual syntax for DHExp when built using the menhir parser
   [@deriving (show({with_path: false}), sexp, yojson)]
   type t =
-    | EmptyHole(MetaVar.t, HoleInstanceId.t)
-    | NonEmptyHole(ErrStatus.HoleReason.t, MetaVar.t, HoleInstanceId.t, t)
-    | FreeVar(MetaVar.t, HoleInstanceId.t, Var.t)
+    | EmptyHole(MetaVar.t, HoleInstanceId.t) //()
```
cyrus-: `()` is the notation for unit values, so let's use `?` for empty holes

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | NonEmptyHole(ErrStatus.HoleReason.t, MetaVar.t, HoleInstanceId.t, t) //(_HOLE x)
```
cyrus-: let's use `{{e}}` for non-empty holes

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | FreeVar(MetaVar.t, HoleInstanceId.t, Var.t) //(_FREE x)
```
cyrus-: let's use `{{?x}}` for free variables

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
     | InvalidText(MetaVar.t, HoleInstanceId.t, string)
-    | InconsistentBranches(MetaVar.t, HoleInstanceId.t, case)
+    | InconsistentBranches(MetaVar.t, HoleInstanceId.t, case) //(_HOLE (case ...))
```
cyrus-: let's use `{{?case ...}}` for this one

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
     | Closure([@opaque] ClosureEnvironment.t, t)
-    | Filter(DHFilter.t, t)
-    | BoundVar(Var.t)
-    | Sequence(t, t)
-    | Let(DHPat.t, t, t)
-    | FixF(Var.t, Typ.t, t)
-    | Fun(DHPat.t, Typ.t, t, option(Var.t))
+    | Filter(DHFilter.t, t) //pause a == 2 b
+    | BoundVar(Var.t) //x
+    | Sequence(t, t) //a; b
```
cyrus-: let's use metavariables `e` for expressions, so `e1; e2` in the comment here (and similar elsewhere)

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | Let(DHPat.t, t, t) //let x = y in z
+    | FixF(Var.t, Typ.t, t) //_FIX f Int -> Int fun: Int x -> 1 + x f in 55
```
cyrus-: I believe the syntax here should be `fix x : ty -> e`

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | Fun(DHPat.t, Typ.t, t, option(Var.t)) //fun: Int x -> x + 1
```
cyrus-: the syntax here should be `fun p -> e`

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
     | TypFun(Term.UTPat.t, t, option(Var.t))
     | TypAp(t, Typ.t)
-    | Ap(t, t)
+    | Ap(t, t) //a(1)
```
cyrus-: `e1(e2)`

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | StringLit(string) //"hello"
+    | BinBoolOp(TermBase.UExp.op_bin_bool, t, t) //true && false
+    | BinIntOp(TermBase.UExp.op_bin_int, t, t) //1 + 2
+    | BinFloatOp(TermBase.UExp.op_bin_float, t, t) //1 +. 2
+    | BinStringOp(TermBase.UExp.op_bin_string, t, t) //
+    | ListLit(MetaVar.t, MetaVarInst.t, Typ.t, list(t)) //[a, b, c]
+    | Cons(t, t) //a :: b
+    | ListConcat(t, t) //a @ b
+    | Tuple(list(t)) //(a, b)
+    | Prj(t, int) //TODO
```
cyrus-: `e.n`

src/haz3lcore/dynamics/DH.re
```diff
@@ -1,52 +1,55 @@
...
+    | Constructor(string) //X
+    | ConsistentCase(case) //4 == 3 | true => 24 | false => false end
+    | Cast(t, Typ.t, Typ.t) //x <Unknown Internal => Int>
+    | FailedCast(t, Typ.t, Typ.t) //x ?<Int => Int>
```
cyrus-: `e<ty1 =!> ty2>` or something like that

cyrus- (changes requested):
left some notes on tweaks. this is going to need to be reworked a bit (but should be mostly similar) with @Negabinary 's upcoming refactoring of DHExp / UExp. since that is coming soon, so will hold off on merging for now but you can make the syntactic changes in the grammar in anticipation of that at least

green726:
Sounds good, thanks for the feedback. I'll make those syntactic changes and wait for the new UExp/DHExp to update then I'll mark as ready for review again.

src/haz3lcore/dynamics/DH.re
green726: Sorry, my docs were completely wrong for the syntax of that one. It was just ( case ... ) but I'll change it to the {{?case ...}}

src/haz3lcore/dynamics/DH.re
green726: I had to add in manual annotations for the type of the function to be able to construct the dhexp with the type. The current elaborator gets the type from the pattern but uses the statics map - I thought it would be best to just add in the annotations rather than try to interface with that. What do you think?

src/haz3lcore/dune
green726: @cyrus- it seems that the standard/recommended way to do this is with a separate directory/library (see https://stackoverflow.com/questions/70805926/include-mly-files-nested-in-subdirectories-when-compiling-with-dune and https://github.com/ocaml/dune/issues/4381) due to a bug in Dune - if you would prefer I could also place it in a subdirectory of haz3lcore with its own Dune file. I don't have a preference either way - what would you prefer? Thanks!

src/haz3lcore/dynamics/DH.re
green726: Ok, I fixed that but I unfortunately cant use the "->" because it seems to create a conflict in Menhir between the exp and typ (I'm pretty sure it is conflicting with the "->" in arrow types) so instead I used "=>". The syntax is now `fix x: ty => e`, are you alright with that or would you prefer I try to find some way around the conflict?

src/haz3lcore/dune
cyrus-: Can we put the necessary dune menhir invocations into the haz3lcore dune file or does it need its own in a subdirectory? No strong preference here, but would like the directory and library to at least have the same name.

src/haz3lcore/dynamics/DH.re
cyrus-: Maybe have it be parenthesized, so `fix x -> e` or `fix (x : ty) -> e`? would prefer to keep the Hazel notation consistent here so if we change it here then we should change it in Hazel

src/haz3lcore/dune
green726: Unfortunately it seems that we can't (that's a known bug with dune according to the stack overflow I linked) - we can put the parser in haz3lcore but the parser directory will also need its own dune file. I think that will keep it part of the same library as haz3lcore? I could try that and if that doesn't work I'll just change the current directory/library to have matching names?

cyrus-: sounds good

src/haz3lcore/dynamics/DH.re
green726: That worked! Thank you!

cyrus-:
@green726 can you get this up-to-date with dev?

green726:
@cyrus- sure, I can. Are there any other changes you would like me to make?

cyrus-:
@green726 getting it up to date and writing more tests demonstrating all of the syntactic forms in the parser would be great. I can give it another review once that works. maybe @7h3kk1d can take a look as well?

green726:
Ok, sounds good.

7h3kk1d:
@green726 I believe I saw failures similar to your last commit when my local reason version was out of sync with make deps. Let me know if you need any help debugging.

green726:
@7h3kk1d Ah, thank you! How were you able to fix it? It's interesting that its the formatter that throws the error rather than the compiler.

green726:
Oh, nevermind about that error. It seems it was just do to a "let" that should was not commented out and should have been. Thank you though!

Makefile
```diff
@@ -55,8 +55,8 @@ repl:
 	dune utop src/haz3lcore
 
 test:
-	dune build @src/fmt @test/fmt --auto-promote src test --profile dev
-	node $(TEST_DIR)/haz3ltest.bc.js
+	dune fmt --auto-promote
+	dune build @src/fmt @test/fmt --auto-promote src test --profile dev; node $(TEST_DIR)/haz3ltest.bc.js
```
7h3kk1d: What prompted this change?

7h3kk1d: I'm mostly curious if we want dune fmt to be passive before build in all of the make commands or if there's some other workflow

src/haz3lcore/lang/Operators.re
```diff
@@ -54,13 +54,90 @@ type op_un =
...
+  | Power => Power
+  | Divide => Divide
+  | LessThan => LessThan
+  | LessThanOrEqual => LessThanOrEqual
+  | GreaterThan => GreaterThan
+  | GreaterThanOrEqual => GreaterThanOrEqual
+  | Equals => Equals
+  | NotEquals => NotEquals
+  };
+};
```
7h3kk1d: This is just slightly annoying to maintain. Probably worth it in the short term but I wonder if having a common defn that haz3lmenhir could use would be good in the long term.

test/test.hazel
7h3kk1d, line 1: If we're going to be making a bunch of .hazel test files for menhir we should group those in a directory so it's clear that it's not tests in hazel

src/haz3lcore/lang/Operators.re
green726: Yeah, I agree with that. So should I just keep it in for now?

test/test.hazel
green726: I decided its better to just keep the parser tests defined inline like we do with the elaborator tests so I've removed the .hazel test file.

src/haz3lcore/lang/Operators.re
7h3kk1d: I think so since you've already done the hard work. My preference would be to have this explicit and once we feel the pain we can come up with something new.

Makefile
green726: Previously you would need to either run `make test` twice or `make fmt` then `make test` each time you wanted to run unit tests. This seemed to be due to the formatting causing the command to fail before the tests could be ran. This change seems to have fixed that and now runs the unit tests in the same command as formatting.

7h3kk1d: That makes sense. If we make this change for test we should probably make it for dev. @disconcision do you have any thoughts on that? Either way I think the last line can still be split onto two

```suggestion
	dune build @src/fmt @test/fmt --auto-promote src test --profile dev
	node $(TEST_DIR)/haz3ltest.bc.js
```

src/haz3lcore/lang/Operators.re
green726: Ok, sounds good to me

src/haz3lcore/statics/TermBase.re
```diff
@@ -243,6 +254,104 @@ and Exp: {
...
+    | TypFun(t, e) => TypFun(TPat.of_menhir_ast(t), of_menhir_ast(e), None)
+    | Undefined => Undefined
+    | TyAlias(tp, ty, e) =>
+      TyAlias(
+        TPat.of_menhir_ast(tp),
+        Typ.of_menhir_ast(ty),
+        of_menhir_ast(e),
+      )
+    | BuiltinFun(s) => BuiltinFun(s)
+    | DeferredAp(f, a) => DeferredAp(of_menhir_ast(f), [of_menhir_ast(a)])
```
7h3kk1d: Is there a reason the menhir ast only has a single deferral?

src/haz3lcore/statics/TermBase.re
```diff
@@ -243,6 +254,104 @@ and Exp: {
...
+      let act = FilterAction.of_menhir_ast(a);
+      Filter(StepperFilterKind.Filter({pat: dcond, act}), dbody);
+    | MultiHole(e) => MultiHole([Exp(of_menhir_ast(e))])
+    | TypAp(e, ty) => TypAp(of_menhir_ast(e), Typ.of_menhir_ast(ty))
+    | UnOp(op, e) =>
+      UnOp(Operators.op_un_of_menhir_ast(op), of_menhir_ast(e))
+    | DynamicErrorHole(e, s) =>
+      DynamicErrorHole(
+        of_menhir_ast(e),
+        InvalidOperationError.t_of_sexp(sexp_of_string(s)),
```
7h3kk1d: Not sure I understand why the sexp encoding is coming into play. Is this just to convert between constant names?

src/haz3lcore/statics/TermBase.re
```diff
@@ -818,6 +977,8 @@ and TPat: {
     | Var(string)
   and t = IdTagged.t(term);
 
+  // let term_of_menhir: Haz3lmenhir.AST.TPat
+
```
7h3kk1d: Cleanup

7h3kk1d:
Not sure if there's a reason it wouldn't work but could we move the core -> menhir conversion functions out of core and into the menhir package and flip the dependency order? 

I could also be missing something that's already been talked about.

test/dune
```diff
@@ -2,7 +2,38 @@
...
+   (flags (:standard))))
+ (release
+  (js_of_ocaml
+   (flags (:standard)))))
+
+(rule
+ (write-file js-of-ocaml-flags-dev "(:standard --debuginfo --noinline)"))
+
+(rule
+ (write-file js-of-ocaml-flags-release "(:standard)"))
```
7h3kk1d: We should update this to match the format in the other dune files or just revert to what's on dev

test/dune
```diff
@@ -2,7 +2,38 @@
...
+  base
+  ptmap
+  incr_dom
+  haz3lmenhir
+  junit
+  junit_alcotest)
  (modes js)
+ (js_of_ocaml
+  (flags
+   (:include js-of-ocaml-flags-%{profile})))
```
7h3kk1d: Also make this match dev

src/haz3lcore/statics/TermBase.re
7h3kk1d: Looks like this is true for a few other cases below. I'll have to read further to understand what's going on.

7h3kk1d (commented):
There's some merge conflicts to fix with dev.  I'm also still trying to understand the lexing and parsing code.  I left some comments on some minor stuff above. I'll try to get through the rest soon.

7h3kk1d:
It also might be useful to pull in the [coverage branch](https://github.com/hazelgrove/hazel/pull/1368) to see which forms haven't been exercised.
Picture of code coverage in the TermBase module showing which forms have been exercised

test/Test_Elaboration.re
```diff
@@ -7,74 +7,139 @@ let dhexp_typ = testable(Fmt.using(Exp.show, Fmt.string), DHExp.fast_equal);
 
 let ids = List.init(12, _ => Id.mk());
 let id_at = x => x |> List.nth(ids);
+
+//Starts at 0
+let get_id_menhir_closure = (id_index: int) => {
```
7h3kk1d: This seems unused

src/haz3lcore/statics/TermBase.re
green726: Cyrus suggested I encode the error messages in Sexp so that I don't have to create a parser for the error messages.

green726:
> Not sure if there's a reason it wouldn't work but could we move the core -> menhir conversion functions out of core and into the menhir package and flip the dependency order?
> 
> I could also be missing something that's already been talked about.

When I first added the menhir parser I believe there was a reason I ordered the dependencies in this way (iirc there was some conflict caused if it went the opposite direction). It's possible that I'm either misremembering, something has changed now that would make it possible, or I just simply did something wrong initially. I can take a look back through and see if there's any reason why it can't be in menhir (did you see anything that jumped out?)

test/Test_Elaboration.re
green726: Yes, sorry. That is unused - that's a relic from the DHExp (instead of universal Exp) where I had to generate ids. I'll delete it

src/haz3lcore/statics/TermBase.re
green726: Sorry, this may just have been an error on my part? What exactly should the deferral look like?

7h3kk1d:
> > Not sure if there's a reason it wouldn't work but could we move the core -> menhir conversion functions out of core and into the menhir package and flip the dependency order?
> > I could also be missing something that's already been talked about.
> 
> When I first added the menhir parser I believe there was a reason I ordered the dependencies in this way (iirc there was some conflict caused if it went the opposite direction). It's possible that I'm either misremembering, something has changed now that would make it possible, or I just simply did something wrong initially. I can take a look back through and see if there's any reason why it can't be in menhir (did you see anything that jumped out?)

I did not but I didn't look closely

src/haz3lcore/statics/TermBase.re
7h3kk1d: Based on my current understanding it should be the list of function parameters and `Deferral(InAp)` for all of the deferrals present.

So for `f(_, 1, _)` it would be `(Deferral(InAp), Int(8), Deferral(InAp))`. It's probably worth synthesizing a few of these in the editor and seeing the DHExp/Exp structure to compare and add tests. I'll leave comments on the other cases with a singleton list.

src/haz3lcore/statics/TermBase.re
```diff
@@ -243,6 +254,104 @@ and Exp: {
...
+    | Test(e) => Test(of_menhir_ast(e))
+    | Cons(e1, e2) => Cons(of_menhir_ast(e1), of_menhir_ast(e2))
+    | ListConcat(e1, e2) =>
+      ListConcat(of_menhir_ast(e1), of_menhir_ast(e2))
+    | Filter(a, cond, body) =>
+      let dcond = of_menhir_ast(cond);
+      let dbody = of_menhir_ast(body);
+      let act = FilterAction.of_menhir_ast(a);
+      Filter(StepperFilterKind.Filter({pat: dcond, act}), dbody);
+    | MultiHole(e) => MultiHole([Exp(of_menhir_ast(e))])
```
7h3kk1d: Check to see if this should be a singleton list

src/haz3lcore/statics/TermBase.re
```diff
@@ -501,6 +613,30 @@ and Pat: {
...
+    | ConstructorPat(x, ty) => Constructor(x, Typ.of_menhir_ast(ty))
+    | StringPat(s) => String(s)
+    | TuplePat(pats) => Tuple(List.map(of_menhir_ast, pats))
+    | ApPat(pat1, pat2) => Ap(of_menhir_ast(pat1), of_menhir_ast(pat2))
+    | ConsPat(p1, p2) => Cons(of_menhir_ast(p1), of_menhir_ast(p2))
+    | BoolPat(b) => Bool(b)
+    | EmptyHolePat => EmptyHole
+    | MultiHolePat(p) =>
+      let p = of_menhir_ast(p);
+      MultiHole([Pat(p)]);
```
7h3kk1d: Check if this should be a singleton list

src/haz3lcore/statics/TermBase.re
```diff
@@ -842,6 +1003,18 @@ and TPat: {
     | Var(string)
   and t = IdTagged.t(term);
 
+  let rec term_of_menhir_ast = (tpat: Haz3lmenhir.AST.tpat) => {
+    switch (tpat) {
+    | InvalidTPat(s) => Invalid(s)
+    | EmptyHoleTPat => EmptyHole
+    | MultiHoleTPat(t) => MultiHole([TPat(of_menhir_ast(t))])
```
7h3kk1d: Same check for singleton list

7h3kk1d:
There's also some warnings we want to address:
```
dune build @src/fmt --auto-promote src --profile dev
File "src/haz3lmenhir/Parser.mly", line 35, characters 7-9:
Warning: the token AS is unused.
File "src/haz3lmenhir/Parser.mly", line 14, characters 7-22:
Warning: the token BAD_CONSTRUCTOR is unused.
File "src/haz3lmenhir/Parser.mly", line 76, characters 7-12:
Warning: the token B_AND is unused.
Warning: 24 states have shift/reduce conflicts.
Warning: 2 states have reduce/reduce conflicts.
Warning: 454 shift/reduce conflicts were arbitrarily resolved.
Warning: 85 reduce/reduce conflicts were arbitrarily resolved.
```

Makefile
7h3kk1d: @green726 I tried addressing this here https://github.com/hazelgrove/hazel/pull/1379

Can you try those make commands out when you get a chance

src/haz3lcore/statics/TermBase.re
green726: Sorry, what do you mean by this? I'm assuming MultiHole shouldn't just be a single Exp? If so, do you happen to know what a Multihole looks like (I'll also take a look at some in normal Hazel). Thanks!

green726:
> It also might be useful to pull in the [coverage branch](https://github.com/hazelgrove/hazel/pull/1368) to see which forms haven't been exercised. Picture of code coverage in the TermBase module showing which forms have been exercised

Sorry, but what does the coverage branch do and what does it mean for a form to be "exercised"? Sorry, I'm not too good with the hazel terminology :). Thank you!

green726:
> There's also some warnings we want to address:
> 
> ```
> dune build @src/fmt --auto-promote src --profile dev
> File "src/haz3lmenhir/Parser.mly", line 35, characters 7-9:
> Warning: the token AS is unused.
> File "src/haz3lmenhir/Parser.mly", line 14, characters 7-22:
> Warning: the token BAD_CONSTRUCTOR is unused.
> File "src/haz3lmenhir/Parser.mly", line 76, characters 7-12:
> Warning: the token B_AND is unused.
> Warning: 24 states have shift/reduce conflicts.
> Warning: 2 states have reduce/reduce conflicts.
> Warning: 454 shift/reduce conflicts were arbitrarily resolved.
> Warning: 85 reduce/reduce conflicts were arbitrarily resolved.
> ```

Yes, I'll look at the grammar and see if I can fix those 24 ambiguities. I'll let you know once I've pushed a commit that fixes them.

Makefile
green726: I tried them out and approved the pull req - they worked. Thank you!

7h3kk1d:
> > It also might be useful to pull in the [coverage branch](https://github.com/hazelgrove/hazel/pull/1368) to see which forms haven't been exercised. Picture of code coverage in the TermBase module showing which forms have been exercised
> 
> Sorry, but what does the coverage branch do and what does it mean for a form to be "exercised"? Sorry, I'm not too good with the hazel terminology :). Thank you!

Just in the green screenshot above the green code is run during the tests and the red code isn't.

green726:
> > > Not sure if there's a reason it wouldn't work but could we move the core -> menhir conversion functions out of core and into the menhir package and flip the dependency order?
> > > I could also be missing something that's already been talked about.
> > 
> > 
> > When I first added the menhir parser I believe there was a reason I ordered the dependencies in this way (iirc there was some conflict caused if it went the opposite direction). It's possible that I'm either misremembering, something has changed now that would make it possible, or I just simply did something wrong initially. I can take a look back through and see if there's any reason why it can't be in menhir (did you see anything that jumped out?)
> 
> I did not but I didn't look closely

I moved the conversion logic over to haz3lmenhir. Turns out that when I tried to have conversion logic in haz3lmenhir there was an internal circular dependency error that was due to a bug in my haz3lmenhir dune file. I got it fixed now and it's all good now (tests still work).

src/haz3lcore/statics/TermBase.re
7h3kk1d: Yeah, that's what I mean. I'm not positive on what multiholes look like. I think they're caused by infix operators. I'll try to generate an example.

7h3kk1d: This is the current visual representation of multiholes. It looks like your parser is surrounding the inhabitants with curly braces. We probably need some sort of delimiter to parse the children inside.
Screenshot from 2024-08-26 14-21-55

7h3kk1d: @green726 I talked to Cyrus about this some today and I think we can probably just skip the MultiHole case for now

green726: Ok, sounds good for me. I'll remove the multihole parsing for now.

green726:
I just pushed a commit that should have resolved all of the menhir shift/reduce conflicts. I'm moving onto fixing the deferrals now.

7h3kk1d:
> I just pushed a commit that should have resolved all of the menhir shift/reduce conflicts. I'm moving onto fixing the deferrals now.

There's also some failed tests. I haven't looked to see what's going on.

green726:
Thanks for pointing that out - I had some problems with the new precedences in the parser which should now be fixed and the tests should pass.

green726:
@7h3kk1d I believe that everything should be fixed and good to go for the merge now.

src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -571,5 +572,7 @@ let fix_typ_ids =
 let uexp_elab = (m: Statics.Map.t, uexp: UExp.t): ElaborationResult.t =>
   switch (elaborate(m, uexp)) {
   | exception MissingTypeInfo => DoesNotElaborate
-  | (d, ty) => Elaborates(d, ty, Delta.empty)
+  | (d, ty) =>
+    print_endline("Elaborated: " ++ DHExp.show(d));
```
7h3kk1d: Remove debug code

src/haz3lmenhir/Conversion.re
```diff
@@ -0,0 +1,278 @@
...
+    | Constructor(x, ty) => Constructor(x, Typ.of_menhir_ast(ty))
+    | Deferral => Deferral(InAp)
+    // switch (pos) {
+    // | InAp => Deferral(InAp)
+    // | OutsideAp => Deferral(OutsideAp)
+    // }
+    | ListExp(l) => ListLit(List.map(of_menhir_ast, l))
+    | TupleExp(t) =>
+      if (List.length(t) == 1) {
+        Parens(of_menhir_ast(List.hd(t)));
```
7h3kk1d, line 209: Does this mean that in the menhir AST parenthesized expressions are singleton tuples?

test/Test_Menhir.re
```diff
@@ -0,0 +1,35 @@
...
+    a,
+    expected,
+    Haz3lmenhir.Interface.parse_program(actual),
+  );
+
+let fun_exp = AST.Fun(VarPat("x"), Var("x"), None);
+
+let fun_str = "fun x -> x";
+
+let tests = [
```
7h3kk1d: I think it's worth copying all the elaboration cases from `Test_Elaboration` and having parser specific tests for them to ensure that the parser is working separately from the elaborator.

test/Test_Menhir.re
```diff
@@ -0,0 +1,35 @@
...
+
+let testable_ast = testable(Fmt.using(AST.show_exp, Fmt.string), (==));
+
+let parser_test = (a, expected, actual, ()) =>
+  Alcotest.check(
+    testable_ast,
+    a,
+    expected,
+    Haz3lmenhir.Interface.parse_program(actual),
+  );
```
7h3kk1d: We should either make a `Test_Conversion` or have the `Test_Menhir` tests go to the regular `TermBase` types so that we can assure it all works together.

test/Test_Elaboration.re
```diff
@@ -7,90 +7,105 @@ let dhexp_typ = testable(Fmt.using(Exp.show, Fmt.string), DHExp.fast_equal);
...
-let empty_hole = () => alco_check("Empty hole", u2, dhexp_of_uexp(u2));
-
-let u3: Exp.t = {
-  ids: [id_at(0)],
-  term: Parens({ids: [id_at(1)], term: Var("y"), copied: false}),
-  copied: false,
-};
-
-let free_var = () => alco_check("free variable", u3, dhexp_of_uexp(u3));
+let alco_check_menhir = (name: string, dhexp: string, uexp: Term.Exp.t) =>
```
7h3kk1d: Can we rename `dhexp` to `expected` and `uexp` to `actual` so it's clear that the assertion is the dhexp and the thing we're testing is the uexp.

7h3kk1d: Also now that dhexp and uexp are merged how do you feel about this:

```reason
let alco_check_menhir = (name: string, dhexp: string, uexp: string) =>
  alco_check(
    name,
    Haz3lmenhir.Conversion.Exp.of_menhir_ast(
      Haz3lmenhir.Interface.parse_program(dhexp),
    ),
    dhexp_of_uexp(
      Haz3lmenhir.Conversion.Exp.of_menhir_ast(
        Haz3lmenhir.Interface.parse_program(uexp),
      ),
    ),
  );
  ```
  
  That way you can write the tests like
  ```reason
  alco_check_menhir2(
    "Let expression for a function which is not recursive (menhir)",
    "let f =
    fun x ->
        1 + x<Unknown Internal => Int>
    f
    in
    55",
    "let f =
          fun x ->
              1 + x
          f
          in
      55",
  );
  ```

File: test/Test_Elaboration.re [hunk 29]
```diff
@@ -7,90 +7,105 @@ let dhexp_typ = testable(Fmt.using(Exp.show, Fmt.string), DHExp.fast_equal);
...
-let bin_op = () =>
-  alco_check(
-    "Inconsistent binary integer operation (plus)",
-    d5,
-    dhexp_of_uexp(u5),
+let let_fun_str = "
+let f =
+    fun x ->
+        1 + x
+    f
```
7h3kk1d: I'm a bit confused syntactically why `f` is below the function. Is this to capture the name? And if so do we intend on this being incompatible with the regular hazel parser

test/Test_Elaboration.re
```diff
@@ -100,124 +115,314 @@ let u8: Exp.t =
...
-  test_case("Consistent if statement", `Quick, consistent_if),
-  test_case("An unapplied function", `Quick, unapplied_function),
-  test_case("Application of function on free variable", `Quick, ap_fun),
-  test_case("Inconsistent case statement", `Quick, inconsistent_case),
-  test_case("Let expression for a function", `Quick, let_fun),
-  test_case(
-    "Function application with a deferred argument",
-    `Quick,
-    deferral,
-  ),
```
7h3kk1d: It might be worth keeping the existing tests temporarily until we're fully confident in the menhir tests long term. Something like:
```reason
let tests = (
  "Menhir Elaboration",
  [
    test_case("Filter test (menhir)", `Quick, filter_menhir),
    test_case("Test failed (menhir)", `Quick, test_menhir),
    test_case("Built-in function (menhir)", `Quick, builtin_fun_menhir),
    test_case(
      "Dynamic error hole (menhir)",
      `Quick,
      dynamic_error_hole_menhir,
    ),
  ],
  "Elaboration",
  [
    test_case("Single integer", `Quick, single_integer),
    test_case("Empty hole", `Quick, empty_hole),
    test_case("Free variable", `Quick, free_var),
  ],
);
```

Or we can separate them into different modules. Let me know if you'd like me to make that change or if you'd want to.

7h3kk1d (commented):
Overall it looks really good to me. I have some minor points about test organization, and we probably want some more tests for the parser itself. Let me know if you need any help on the testing side.

test/Test_Elaboration.re
7h3kk1d: I'm mostly saying this because we're using the tests for the labeled tuple work to ensure we don't break anything.

7h3kk1d:
We should also add a `README` file in the menhir directory describing what the parser is and what it's for.

test/Test_Elaboration.re
green726: Sure, if you would like to make this change, that would be very helpful! Thanks!

test/Test_Elaboration.re
green726: As far as changing the alco_check_menhir to allow textual uexps, I would be ok with that but I do have a slight concern that if we do that the elaborator tests really become Hazel parser + elaborator tests and I think it might be a bit better to keep the parser/elaborator tests separate - what do you think about that?

green726: > Can we rename `dhexp` to `expected` and `uexp` to `actual` so it's clear that the assertion is the dhexp and the thing we're testing is the uexp.

Yeah totally - especially after the uexp/dhexp merge it makes more sense to do actual/expected.

test/Test_Menhir.re
green726: I'll have the `Test_Menhir` go to the regular `TermBase` - the conversion logic is very light and (for the most part) is essentially just casting from `AST` types to `TermBase` so I don't think we'd need separate tests for it.

test/Test_Menhir.re
green726: I agree - that is a good idea. I'll work on that.

test/Test_Elaboration.re
green726: That's to cover the `option(Var.t)` at the end of the `Fun` exp. I'm not too sure what that option represents/how it gets inserted in elaboration so I allowed it to be optionally specified textually for the menhir parser. Do you happen to know what it represents/how I should do this properly? Thanks!

7h3kk1d: I think that's the name of the function that gets pulled from the let binding. Let me confirm that and see where it's happening. Unfortunately I don't think there's anywhere in the existing syntax that differentiates these cases.

src/haz3lmenhir/Conversion.re
green726: Yes - how is the difference between parens/singleton tuples handled in Hazel? Is a paren not just a singleton tuple? (Sorry - I was a bit confused about this)

test/Test_Elaboration.re
7h3kk1d: > As far as changing the alco_check_menhir to allow textual uexps, I would be ok with that but I do have a slight concern that if we do that the elaborator tests really become Hazel parser + elaborator tests and I think it might be a bit better to keep the parser/elaborator tests separate - what do you think about that?

I agree that they'll be parser + elaborator tests. But I think that's true if we do it just for the actual or for the actual + expected. If we wanted to decouple the tests we'd probably need to remove the parser from both which seems _too_ expensive. If we have good enough parser tests hopefully we can build up some confidence on the parsing side and trust these test failures are mostly around the elaboration. In the mean time we can keep the old elaboration tests.

src/haz3lmenhir/Conversion.re
7h3kk1d: I think that singletons are just assumed to not be tuples unless there's a comma. I'm going to try to make some tests for the existing editor so it's easier to prove these things but it has been taking me a while.  Here are some examples just from adding prints to the editor:
(1) becomes Parens(1)
(1,2,3) becomes Parens(Tuple(1,2,3))
1,2,3 becomes Tuple(1,2,3)

test/Test_Elaboration.re
green726: Ok, that sounds good to me. I'll make this change - thanks!

src/haz3lmenhir/Conversion.re
green726: Got it - looking at the parser code I just realized that my `Paren` representation is actually flawed as a singleton tuple would still require a comma (oversight on my part). I'll try to fix this in the parser. Should all expressions wrapped in parentheses be elaborated to be within a `Paren`? Ie a `(x)` should elaborate to `Parens(Var("x"))` or just `x` after elaboration? Sorry, I can't quite tell from the logs you sent if the final state is `Parens(Int(1))` or `Int(1)`. Thanks!

7h3kk1d: > Should all expressions wrapped in parentheses be elaborated to be within a Paren

I'm logging pre elaboration so I'm not sure what elaboration does but the parsing definitely has a paren.

green726: Ah ok, got it. I'll try and check what elaboration does.

7h3kk1d: > Sorry, I can't quite tell from the logs you sent if the final state is Parens(Int(1)) or Int(1)

The result of the parsing is Parens(Int(1)) it's just logging the other one too because I'm logging at the beginning of the elaborate function

7h3kk1d: I think your parser code is the equivalent of MakeTerm and Form, not elaboration. So you don't need to do the work that happens inside elaboration

green726: > I think your parser code is the equivalent of MakeTerm and Form, not elaboration. So you don't need to do the work that happens inside elaboration

Yes, that's true, but upon looking at the elaboration code and my tests it looks like the elaborator strips parens so I don't believe that I need support for them at all?

7h3kk1d: It depends on if we're using the parser for the uexp or dhexp side of elaboration. I think we probably want to use it for all kinds of tests so it's nice to be able to use the parser for the pre elaboration syntax with parens and the post elaboration assertion.

test/Test_Elaboration.re
7h3kk1d: I did this and it's two modules inside the same file. It might make more sense to split the files up if it becomes unwieldy for you.

test/Test_Elaboration.re
7h3kk1d: Okay, so the function name gets added during elaboration https://github.com/hazelgrove/hazel/blob/dev/src/haz3lcore/dynamics/Elaborator.re#L292. So there's no place in the syntax text that represents this currently.

I'm not sure how we want to represent this in the Menhir parser. We can just remove it and have this be compatible with `MakeTerm` but then we couldn't use the parser for the expectation side of the elaboration tests (without changing the assertion equality to ignore function names). 

For now I'm okay with either ignoring the names in the parser or having some artificial syntactic representation of a named function.  I'm not sure if just putting the name after the function will be confusing or not.

@cyrus- or @Negabinary  do you have any other thoughts?

7h3kk1d: https://github.com/hazelgrove/hazel/pull/1409/files#diff-6967fe38adf76426a01468c788252c2355461e812298332ce6e3e282021ee705R46 might also be useful for any confusion on how the existing parser is handling cases

cyrus-: having an artificial syntax for the function name seems fine to me

7h3kk1d: @green726 feel free to come up with your own representation, but off my head I like `fun x -> 1 + x` for anonymous and `named_fun f x -> 1+x` for named.

green726: Got it - your representation sounds good to me. I'll make that change.

test/Test_Menhir.re
```diff
@@ -0,0 +1,194 @@
...
+
+      alco_check(
+        name ++ " matches expected type",
+        exp,
+        Haz3lmenhir.Conversion.Exp.of_menhir_ast(
+          Haz3lmenhir.Interface.parse_program(actual),
+        ),
+      );
+    },
+  );
```
7h3kk1d: @green726 I added a few tests of the parser for different forms below. These are asserting that it matches the expected value and that it's equivalent to what the tylr version of the parser does. There's a few failing cases right now:

1. Data constructors seem to be failing
2. My named_fun test is failing (I haven't verified that the test isn't wrong)
3. There's some issue with the nested integer operators.


The tests aren't conclusive. I'll try to add some more soon but I feel like we're close to covering all the frequently used cases.

src/haz3lmenhir/Parser.mly
```diff
@@ -0,0 +1,269 @@
...
+    | OPEN_PAREN; types = separated_list(COMMA, typ); CLOSE_PAREN { TupleType(types) }
+    | OPEN_SQUARE_BRACKET; t = typ; CLOSE_SQUARE_BRACKET { ArrayType(t) }
+    | t1 = typ; DASH_ARROW; t2 = typ { ArrowType(t1, t2) }
+
+pat:
+    | QUESTION; s = STRING { InvalidPat(s) }
+    | WILD { WildPat }
+    | QUESTION { EmptyHolePat }
+    | OPEN_SQUARE_BRACKET; l = separated_list(COMMA, pat); CLOSE_SQUARE_BRACKET; { ListPat(l) }
+    | c = CONSTRUCTOR_IDENT; COLON; t = typ;  { ConstructorPat(c, t) }
```
7h3kk1d: It's worth noting maketerm always parses the constructor call as unknown https://github.com/hazelgrove/hazel/blob/dev/src/haz3lcore/statics/MakeTerm.re#L193. I guess it's useful to have this with the type annotation for tests but it's incompatible syntax with the main editor.

src/haz3lmenhir/Parser.mly
```diff
@@ -0,0 +1,269 @@
...
+
+%inline binOp:
+    | i = intOp { i }
+    | f = floatOp { f }
+    | b = boolOp { b }
+
+binExp:
+    | e1 = exp; b = binOp; e2 = exp { BinExp (e1, b, e2) }
+
+typ:
```
7h3kk1d, line 222: This doesn't currently allow for Typ Variables.

src/haz3lmenhir/Parser.mly
```diff
@@ -0,0 +1,269 @@
...
+
+%inline intOp:
+    | PLUS { IntOp(Plus) }
+    | MINUS { IntOp(Minus) }
+    | TIMES { IntOp(Times) }
+    | POWER { IntOp(Power) }
+    | DIVIDE { IntOp(Divide) }
+    | DOUBLE_EQUAL { IntOp(Equals) }
+    | NOT_EQUAL { IntOp(NotEquals) }
+    (* | LESS_THAN { IntOp(LessThan) } *)
```
7h3kk1d: We probably want LESS_THAN fixed. I see that uncommenting it gets the shift reduce conflicts back.

test/Test_Menhir.re
```diff
@@ -0,0 +1,216 @@
...
+          )
+          |> Exp.fresh,
+        )
+        |> Exp.fresh,
+      )
+      |> Exp.fresh,
+      Int(8) |> Exp.fresh,
+    )
+    |> Exp.fresh,
+    "-1 + 2 - 3 / 4 * 5 ** 6 >= 8" // TODO Add the remaining operators. < is also currently broken
```
7h3kk1d: This is currently failing due to precedence  and we can't use `<` for the reason mentioned above.

test/Test_Menhir.re
green726: I've been working on adding some tests as well (sorry, I believe that I forgot to push them). I can cross-check our two results for the named functions. Yeah, it does seem that we are approaching that.

src/haz3lmenhir/Parser.mly
green726: Oh that is my bad. I must've forgotten about the LESS_THAN when fixing the precedences.

green726:
@7h3kk1d The "Let Binding with a type ascription" is currently failing (as written) because it is attempting to use the Hazel frontend type annotation syntax. Since this menhir parser was originally designed just for the DHExp (now the Exp) syntax we would need to use the actual casting menhir syntax. Would you like me to switch the test to use that or should I be adding a type annotation syntax to the parser?

7h3kk1d:
> @7h3kk1d The "Let Binding with a type ascription" is currently failing (as written) because it is attempting to use the Hazel frontend type annotation syntax. Since this menhir parser was originally designed just for the DHExp (now the Exp) syntax we would need to use the actual casting menhir syntax. Would you like me to switch the test to use that or should I be adding a type annotation syntax to the parser?

Now that DHExp and UExp are merged I think we want them to be the same as the front end type annotation.

green726:
> > @7h3kk1d The "Let Binding with a type ascription" is currently failing (as written) because it is attempting to use the Hazel frontend type annotation syntax. Since this menhir parser was originally designed just for the DHExp (now the Exp) syntax we would need to use the actual casting menhir syntax. Would you like me to switch the test to use that or should I be adding a type annotation syntax to the parser?
> 
> Now that DHExp and UExp are merged I think we want them to be the same as the front end type annotation.

Ok sounds good - then I should just be adding a `p : t` syntax to the menhir parser and a conversion to the internal pat `Cast` in the `Conversion.re`? Or I could simply parse the `p: t` syntax into the `CastPat` ast node?

7h3kk1d:
> > > @7h3kk1d The "Let Binding with a type ascription" is currently failing (as written) because it is attempting to use the Hazel frontend type annotation syntax. Since this menhir parser was originally designed just for the DHExp (now the Exp) syntax we would need to use the actual casting menhir syntax. Would you like me to switch the test to use that or should I be adding a type annotation syntax to the parser?
> 
> > 
> 
> > Now that DHExp and UExp are merged I think we want them to be the same as the front end type annotation.
> 
> 
> 
> Ok sounds good - then I should just be adding a `p : t` syntax to the menhir parser and a conversion to the internal pat `Cast` in the `Conversion.re`? Or I could simply parse the `p: t` syntax into the `CastPat` ast node?

I'm can't tell if it matters without checking on a computer. I think either would be fine

green726:
> > > > @7h3kk1d The "Let Binding with a type ascription" is currently failing (as written) because it is attempting to use the Hazel frontend type annotation syntax. Since this menhir parser was originally designed just for the DHExp (now the Exp) syntax we would need to use the actual casting menhir syntax. Would you like me to switch the test to use that or should I be adding a type annotation syntax to the parser?
> > 
> > 
> > > 
> > 
> > 
> > > Now that DHExp and UExp are merged I think we want them to be the same as the front end type annotation.
> > 
> > 
> > Ok sounds good - then I should just be adding a `p : t` syntax to the menhir parser and a conversion to the internal pat `Cast` in the `Conversion.re`? Or I could simply parse the `p: t` syntax into the `CastPat` ast node?
> 
> I'm can't tell if it matters without checking on a computer. I think either would be fine

Ok sounds good. Thanks!

green726:
Also @7h3kk1d I just looked through the basic reference test to see why that was failing - it declares an empty hole with a `let empty_hole =  in` (no syntax marking the empty hole). The Menhir parser expects a `?` to denote the empty hole. I'm not sure if there is some way I can modify the parser to fill in an empty hole when no forms are matched (I doubt it) so I think its best if we modify this test to use `?` for the empty holes. What do you think?

7h3kk1d:
> Also @7h3kk1d I just looked through the basic reference test to see why that was failing - it declares an empty hole with a `let empty_hole = in` (no syntax marking the empty hole). The Menhir parser expects a `?` to denote the empty hole. I'm not sure if there is some way I can modify the parser to fill in an empty hole when no forms are matched (I doubt it) so I think its best if we modify this test to use `?` for the empty holes. What do you think?

Yeah, that makes sense. Sorry I missed it. Can we try doing a find and replace on `= in` with `= ? in` just to get that part to pass. Otherwise we can paste the code in and fix it but it would be nice to make sure we're compatible with a relatively large hazel program.

7h3kk1d:
@green726 can we use a different symbol than `:` for the type of the `Constructor`. It would be nice if `A` would be interpreted as `Constructor("A", Unknown(Internal))` like `MakeTerm` and `A : T` would be `Cast(Constructor("A", Unknown(Internal)), T)`.

So we'd need a new symbol for that. Maybe `A :: T` or `A ~ T`? If we're not limited to ascii we can consider `A ∈ T`

src/haz3lmenhir/Parser.mly
green726: Just fixed this in the most recent commit - the parser should treat the constructors the same as maketerm now. The new constructor casting/annotation syntax uses a `~`

7h3kk1d: I think I miscommunicated what I was saying with the cast/constructor type. I pushed a new commit https://github.com/hazelgrove/hazel/pull/1220/commits/ac4e9f9267679973f10bf14bdc3b639bee2ed499 that does what I was describing. Where `A : T` is the unknown constructor A cast to T and `A ~ T` is the constructor A of type `T`

src/haz3lmenhir/Parser.mly
7h3kk1d: @green726 I added https://github.com/hazelgrove/hazel/pull/1220/commits/1aec2b9075548359d9421b143245554fb6fd1090 with some tests for type variable parsing

test/Test_Menhir.re
```diff
@@ -0,0 +1,439 @@
...
+  //         )
+  //         |> Exp.fresh,
+  //       )
+  //       |> Exp.fresh,
+  //     )
+  //     |> Exp.fresh,
+  //     Float(8.) |> Exp.fresh,
+  //   )
+  //   |> Exp.fresh,
+  //   "2. +. 3. /. 4. *. 5. **. 6. <. 8." // TODO Add the remaining operators. -. is also currently broken
```
7h3kk1d, line 389: @green726 I added the float test to make sure we don't forget the equivalent float operations. For some reason `-.` isn't working right now. I marked all the failing tests as `Alcotest.skip` so we don't have to keep commenting them out. We should make sure they're all passing before we merge though.

test/Test_Menhir.re
```diff
@@ -0,0 +1,458 @@
...
+          |> Exp.fresh,
+        )
+        |> Exp.fresh,
+      )
+      |> Exp.fresh,
+      Float(8.) |> Exp.fresh,
+    )
+    |> Exp.fresh,
+    "2. +. 3. /. 4. *. 5. **. 6. <. 8." // TODO Add the remaining operators. -. is also currently broken
+  ),
```
7h3kk1d: @green726  I added the float test to make sure we don't forget the equivalent float operations. For some reason -. isn't working right now. I marked all the failing tests as `Alcotest.skip` so we don't have to keep commenting them out. We should make sure they're all passing before we merge though.

src/haz3lmenhir/Parser.mly
green726: Ah ok, got it. Thanks!

green726:
@7h3kk1d I seem to have a reduce-reduce conflict in the parser between the `typvars` and the `constructors`. Because both are just a capitalized ident the parser can't differentiate and doesn't know whether to reduce to a typ or an exp. I think we are going to need some sort of non-hazel syntax for one of them. Do you have any preferences?

cyrus-:
Ah that's surprising, since they are in different sorts -- there shouldn't be anywhere where both a type and an expression / pattern can appear?

green726:
There is actually one place where an exp/typ appear together - in the cast annotation syntax (both the special syntax for Constructor exps and the general exp/pat casting syntax).

cyrus-:
But that is always preceded by a `:` right?

green726:
No exp casts are currently following the syntax `e<t1=>t2>` which now looking at the conflict makes sense - the `<` could also be a binary expression which is what is causing this conflict. So the better route then is to change the casting syntax I suppose?

cyrus-:
Yes I think that makes more sense. I believe @Negabinary's latest PR changes this syntax to only use `:` and the target type (with no source type)?

green726:
Ok sounds good to me - I guess I'll make some quick syntax change for now and merge his PR into this branch when its ready (or merged into main).

7h3kk1d:
> Ok sounds good to me - I guess I'll make some quick syntax change for now and merge his PR into this branch when its ready (or merged into main).

@green726 can you let me know what you're planning on using for the cast syntax?

green726:
All new casting syntax uses a `{` or `}` where there previously would've been a `<` or `>`

src/haz3lmenhir/AST.re
```diff
@@ -0,0 +1,160 @@
...
+  | EmptyHole;
+
+[@deriving (show({with_path: false}), sexp)]
+type typ =
+  | IntType
+  | StringType
+  | FloatType
+  | BoolType
+  | UnitType
+  | SumTyp(typ, option(typ))
```
7h3kk1d: @green726   Instead of this being recursive with SumTerm can we have `SumTyp(list(string, option(typ))` for the variants?

src/haz3lmenhir/Lexer.mll
```diff
@@ -0,0 +1,134 @@
...
+    | "_" {WILD}
+    | "fix" {FIX}
+    | "typfun" {TYP_FUN}
+    | "type" {TYP}
+    | "$" {DOLLAR_SIGN}
+    | "~" {TILDE}
+    | "t" {T_TYP} (* TODO These make it very easy to get an unparseable program *)
+    | "p" {P_PAT}
+    | "tp" {TP_TPAT}
+    | "e" {E_EXP}
```
7h3kk1d: @green726 Is this intended to be the final version for these tokens. It's very easy to have a program not parse if you have a variable called e, t, p, tp. Let me know if you both already discussed this format with @cyrus- before I joined

green726: @7h3kk1d I don't believe I discussed this fomat with cyrus yet. Yeah, it's definitely not the best and we can certainly change it. Do you have any ideas for what you would like it to be?

src/haz3lmenhir/AST.re
green726: @7h3kk1d Defining the `SumTyp` with a `list` created a `reduce/reduce` error that I just couldn't seem to fix. I spent a while trying to do stuff with different lists and precedences and digging into the automaton but the only way I was able to resolve it was with this recursive style.

7h3kk1d: I'll take a look as well but that makes sense.

src/haz3lmenhir/AST.re
```diff
@@ -0,0 +1,252 @@
...
+
+// ['a'-'z' '_'] ['a'-'z' 'A'-'Z' '0'-'9' '_']*
+// Can't be t, e, tp, or p because of the lexer
+let arb_ident =
+  QCheck.(
+    let arb_alpha = Gen.char_range('a', 'z'); // TODO make this support full indent instead of just lower alpha
+    string_gen_of_size(Gen.int_range(1, 5), arb_alpha)
+  );
+
+let arb_var = QCheck.(map(x => Var(x), arb_ident));
```
7h3kk1d: I need to put a guard here to remap keywords

src/haz3lcore/lang/Form.re
```diff
@@ -256,7 +256,7 @@ let forms: list((string, t)) = [
   ("divide", mk_infix("/", Exp, P.mult)),
   ("equals", mk_infix("==", Exp, P.eqs)),
   ("string_equals", mk_infix("$==", Exp, P.eqs)),
-  ("string_concat", mk_infix("++", Exp, P.plus)),
+  ("string_concat", mk_infix("++", Exp, P.concat)),
```
7h3kk1d, line 259: This makes it consistent with Precedence.re

test/Test_ExpToSegment.re
```diff
@@ -0,0 +1,75 @@
...
+            )
+            |> Exp.fresh,
+          );
+        let serialized = Printer.of_segment(~holes=Some("?"), segment);
+
+        check(
+          string,
+          "Match statement",
+          serialized,
+          "case x end" // TODO Why is this not serializing the cases
```
7h3kk1d: @Negabinary let me know if you have any thoughts on this

src/haz3lmenhir/AST.re
```diff
@@ -0,0 +1,600 @@
...
+                gen_typ_sized((n - 1) / 3),
+                gen_typ_sized((n - 1) / 3),
+              ),
+            ])
+          };
+        },
+        n,
+      )
+    );
+// TODO Printers, shrinkers stuff
```
7h3kk1d, line 588: I'm going to leave this for a separate PR

7h3kk1d:
@green726 I made a bunch more fixes in the parser, conversion, and tests. There's still come cleanup I need to do before we merge but let me know what else you want to look at before I give it to Cyrus for review.

src/haz3lmenhir/AST.re
```diff
@@ -0,0 +1,600 @@
...
+  | TyAlias(tpat, typ, exp);
+
+let arb_int = QCheck.(map(x => Int(x), small_int));
+let arb_str_exp = QCheck.map(x => String(x), arb_str); // Make strings anything other than `"`"
+
+// Floats are positive because we use UnOp minus
+let arb_float = QCheck.(map(x => Float(x), pos_float));
+let arb_lower_alpha = QCheck.Gen.char_range('a', 'z');
+
+let arb_constructor_ident =
```
7h3kk1d: All of the arbitrary/generator instances need to be cleaned up for readability.

src/haz3lmenhir/Conversion.re
```diff
@@ -0,0 +1,511 @@
...
+    | Deferral(_) => Deferral
+    | Filter(_) => raise(Failure("Filter not supported")) // TODO
+    | MultiHole(_) => raise(Failure("MultiHole not supported")) // TODO
+    | Closure(_) => raise(Failure("Closure not supported")) // TODO
+    | Parens(e) => of_core(e)
+    | Constructor(s, typ) => Constructor(s, Typ.of_core(typ))
+    | DeferredAp(e, es) =>
+      ApExp(of_core(e), TupleExp(List.map(of_core, es)))
+    | Fun(p, e, _, name_opt) => Fun(Pat.of_core(p), of_core(e), name_opt)
+    | Ap(Reverse, _, _) => raise(Failure("Reverse not supported")) // TODO
```
7h3kk1d, line 354: This needs to be done

src/haz3lmenhir/AST.re
```diff
@@ -0,0 +1,603 @@
...
+    let ident = String.make(1, leading) ++ tail;
+    // if ident is a keyword remap
+    switch (ident) {
+    | "Int"
+    | "Float"
+    | "String"
+    | "Unknown"
+    | "Internal"
+    | "Bool" => "keyword"
+    | _ => ident
```
cyrus-: put all reserved keywords in one place in the codebase

src/haz3lmenhir/Lexer.mll
```diff
@@ -0,0 +1,137 @@
...
+
+let identifier = ['a'-'z' '_'] ['a'-'z' 'A'-'Z' '0'-'9' '_']*
+let constructor_ident = ['A'-'Z'] ['a'-'z' 'A'-'Z' '0'-'9' '_']*
+let sexp_string = '`' [^'`']* '`'
+let ints = ['0'-'9']+
+
+rule token = 
+    parse 
+    | "undef" { UNDEF}
+    | "infinity" | "neg_infinity" | "nan" | "epsilon_float" | "pi" | "max_int" | "min_int" | "is_finite" | "is_infinite" | "int_of_float" | "float_of_int" | "string_of_int" | "string_of_float" | "string_of_bool" | "int_of_string" | "float_of_string" | "bool_of_string" | "abs" | "abs_float" | "ceil" | "floor" | "exp" | "log" | "log10" | "sqrt" | "sin" | "cos" | "tan" | "asin" | "acos" | "atan" | "mod" | "string_length" | "string_compare" | "string_trim" | "string_concat" | "string_sub" { BUILTIN(Lexing.lexeme lexbuf)}
```
cyrus-: add comment to Builtins file to tell folks to update this list as well

src/haz3lmenhir/Lexer.mll
```diff
@@ -0,0 +1,137 @@
...
+    | "-." { MINUS_FLOAT }
```
cyrus-: move down

src/haz3lmenhir/Lexer.mll
```diff
@@ -0,0 +1,137 @@
...
+    | "@<" {TYP_AP_SYMBOL}
+    | "@" {AT_SYMBOL}
+    | "?" {QUESTION}
+    | "_" {WILD}
+    | "fix" {FIX}
+    | "typfun" {TYP_FUN}
+    | "type" {TYP}
+    | "$" {DOLLAR_SIGN}
+    | "~" {TILDE}
+    | "?t" {T_TYP} (* TODO These make it very easy to get an unparseable program *)
```
cyrus-: remove TODO

test/Test_ExpToSegment.re
cyrus-: fix and mark this as a skipped test until we fix the issue (file the issue)

test/Test_Menhir.re
```diff
@@ -0,0 +1,860 @@
...
+ * serialize it, parse it with MakeTerm, and parse it with Menhir.
+ */
+let qcheck_menhir_maketerm_equivalent_test =
+  QCheck.Test.make(
+    ~name="Menhir and maketerm are equivalent",
+    ~count=100,
+    QCheck.make(~print=AST.show_exp, AST.gen_exp_sized(7)),
+    exp => {
+      let core_exp = Conversion.Exp.of_menhir_ast(exp);
+      // TODO Maybe only do this when necessary.
```
cyrus-: remove

test/Test_Menhir.re
```diff
@@ -0,0 +1,860 @@
...
+      // TODO Check with Matthew if I'm using this correctly
```
cyrus-: talk to matthew and then remove

test/Test_Menhir.re
```diff
@@ -0,0 +1,860 @@
...
+    },
+  );
+
+/**
+ * QCheck Test to check that menhir parses out what ExpToSegment serializes.
+ * We generate an expression, convert it to the core representation, convert it to a segment,
+ * serialize it, parse it with Menhir, and compare to the original.
+ *
+ * TODO This fails due to types not being serialized on constructors
+ *  and some other ExpToSegment inconsistencies
```
cyrus-, line 137: file bugs

cyrus- (changes requested):

test/Test_Menhir.re
7h3kk1d: @cyrus- these are the bugs I've found with the qcheck test. 
-  https://github.com/hazelgrove/hazel/issues/1452
-  https://github.com/hazelgrove/hazel/issues/1451
-  https://github.com/hazelgrove/hazel/issues/1445

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -445,11 +445,12 @@ let rec exp_to_pretty = (~settings: Settings.t, exp: Exp.t): pretty => {
...
-                  ? []
-                  : [Secondary(Secondary.mk_newline(Id.mk()))]
-                    @ [mk_form("rule", id, [p])]
-                    @ (e |> fold_if(settings.fold_case_clauses)),
+                (
+                  settings.inline
+                    ? [] : [Secondary(Secondary.mk_newline(Id.mk()))]
+                )
+                @ [mk_form("rule", id, [p])]
+                @ (e |> fold_if(settings.fold_case_clauses)),
```
7h3kk1d, line 453: This is also solved in https://github.com/hazelgrove/hazel/pull/1450

7h3kk1d: Merging either PR first is fine.

cyrus- (approved):

PR #1235, opened 2024-03-05
Implement Labeled Tuples

WondAli:
# Labeled Tuples

## Overview  
This PR introduces labeled tuples to Hazel, allowing tuple elements to be named using the `=` operator.  

> **Note:** This PR replaces an outdated implementation on a different branch, as I couldn't change the source branch on the old PR.

## What Are Labeled Tuples?  
Labeled tuples allow elements within a tuple to be named using the `=` operator. For example:  

```hazel
(x=1, y=2)
```

Here, `x=1` and `y=2` define labeled elements. These labels are only relevant within the enclosing tuple. Conceptually, `=` acts as a binary operator that pairs a label (a string) with an expression.

## Labeled Tuple Features
Labeled elements can appear in any order when matched against a labeled tuple type, while unlabeled elements must maintain relative order.
### Examples
- `(x=1, True, y=6)` correctly matches the type `(Bool, y=Int, x=Int)`.
- `let x=a, b = (y=2, x=4)` binds `4` to `a` and `2` to `b`.

### Dot Operator/Projection (`.`)
The dot operator allows extracting labeled elements from tuples:
```hazel
(x=1, 2, 3).x  # Evaluates to 1 #
```
### Singleton Labeled Tuples  

A labeled tuple with a single element, such as `(a=1)`, follows the same principles as multi-element labeled tuples but has some unique characteristics:  

- **Syntactic Form**:  
  - A singleton labeled tuple is typically written as `(a=1)`, but parentheses are **optional**, so `a=1` is also valid.  
- **Type-Checking & Label Inference**:  
  - If an unlabeled value is analyzed against a labeled tuple type, it is automatically wrapped as a singleton labeled tuple.  
  - Example: Analyzing `1` against `(a=Int)` infers `1` as `a=1`.  
- **Dot Operator Compatibility**:  
  - `(a=1).a` correctly extracts `1`.  

## Implementation Strategy

Labeled tuples are implemented by extending existing constructs in expressions, patterns, and types:  

- **`TupLabel`**: Represents a labeled tuple element in expressions, patterns, and types.  
- **`Label`**: Represents the label itself, always appearing in the label position of `TupLabel`.
  - Any other sort present in label position other than a valid label or empty hole is placed in a `MultiHole`
- **`Dot`**: In expressions represents tuple label projection (i.e., field access). The right-hand argument of `Dot` must be a `Label`.  

### Automatic label inference
- Missing labels are introduced during the elaboration phase for tuples
- Singleton labeled tuples are converted to tuples during Statics and the syntax is rewritten and stored in the info map.
  - During elaboration this syntax is pulled out of the info map to rewrite the syntax before evaluation
 
## Future work
* Introduce Label sort that keeps labels in their own syntactic context


cyrus-:
@WondAli playing with this branch, looks like the labels aren't showing up in the types of the tuples (and so you get weird run-time errors when projecting out a label that doesn't exist). is there a specific issue with adding labels to product types?

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,295 @@
+open Util;
+
+exception Exception;
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type t = string;
```
7h3kk1d: Should this type be `label` instead of `LabeledTuple.t`? I believe it's just used for the labels.

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,295 @@
...
+
+let equal: (option((t, 'a)), option((t, 'b))) => bool =
+  (left, right) => {
+    switch (left, right) {
+    | (Some((s1, _)), Some((s2, _))) => String.equal(s1, s2)
```
7h3kk1d: _Personal note_: It would be great if we had `map2`  from https://ocaml.janestreet.com/ocaml-core/109.20.00/doc/core/Option.html

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,295 @@
...
+      List.fold_left(
+        ((b, ls, ns), e) =>
+          switch (get_label(e)) {
+          | Some((s1, _))
+              when
+                b
+                && List.fold_left(
+                     (v, l) =>
+                       switch (l) {
+                       | (Some(s2), _) when v => compare(s1, s2) != 0
```
7h3kk1d: I think you can just use `s1 != s2`

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,295 @@
...
+    let (_, l2_lab, _) = validate_uniqueness(get_label2, l2);
+    // Second item in the pair is the full tuplabel
+    let l2_matched =
+      List.fold_left(
+        (l2_matched, l1_item) => {
+          let l2_item =
+            find_opt(
+              l2_item => {
+                switch (l1_item, l2_item) {
+                | ((Some(s1), _), (Some(s2), _)) => compare(s1, s2) == 0
```
7h3kk1d: Same here with `s1 == s2`

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,295 @@
...
+    'b => option((t, 'b)),
+    'c => option((t, 'c)),
+    ('a, 'b, 'c) => 'a,
+    'a,
+    'a,
+    list('b),
+    list('c)
+  ) =>
+  'a =
+  (get_label1, get_label2, f, accu, accu_fail, l1, l2) => {
```
7h3kk1d: I think these would be slightly easier to read if the types were on the params as opposed to on the function itself. I'm open to other viewpoints on it though.

src/haz3lcore/dynamics/DH.re
```diff
@@ -38,7 +38,10 @@ module rec DHExp: {
     | ListLit(MetaVar.t, MetaVarInst.t, Typ.t, list(t))
     | Cons(t, t)
     | ListConcat(t, t)
+    | Label(string)
+    | TupLabel(t, t)
```
7h3kk1d: Unclear to me why this needs to be a `t` and not a `string` for the first param.

src/haz3lcore/dynamics/DH.re
```diff
@@ -230,6 +257,12 @@ module rec DHExp: {
 
   let rec fast_equal = (d1: t, d2: t): bool => {
     switch (d1, d2) {
+    /* TODO: Labels are a special case, but should they be?*/
+    | (TupLabel(_, d1'), TupLabel(_, d2')) =>
+      LabeledTuple.equal(get_label(d1), get_label(d2))
```
7h3kk1d: It's probably easier to just match against `TupLabel(Label(name1)`

src/haz3lcore/LabeledTuple.re
7h3kk1d: This exists in OptUtil

7h3kk1d:
Evaluation should continue around:
```
let x : (a=Int, b=String)= (a=1, b=? ) in (x.b, x.a)
```

7h3kk1d:
There's another issue with the parsing for type annotations outside lets:
```hazel
((a="Foo", b=8) : (a=String, b=Int))
```


Screenshot from 2024-09-19 13-50-26

WondAli:
There is currently no conversion of standalone labeled types to a singleton tuple, so statics checks a singleton tuple expression against a labeled type (which decays to an int)
Screenshot from 2024-09-27 01-21-26

7h3kk1d:
There's some issues with functions. I'm assuming that's also why the elaboration test is failing.

Screenshot from 2024-09-30 10-10-00

7h3kk1d:
Self reminder to write tests and validate the function argument patterns like:
```
let fn : (x=Int, b=Int) -> Int = fun (x=a, b=b) ->a
in
fn(b=2,x=1)
```

7h3kk1d:
Check to see behaviors without parens

7h3kk1d:
There's still some ongoing conversations on what the behavior of singleton labels should be. Especially with:


```hazel
let x : (l=Int) = 8 in x #It's unclear whether we should be ascribing labels onto any term just based off of the type#

```

7h3kk1d:
Screenshot from 2024-10-11 14-39-42
The patterns don't work without parens. I think this is a parsing issue.

7h3kk1d:
> Screenshot from 2024-10-11 14-39-42 The patterns don't work without parens. I think this is a parsing issue.

This may be a problem with regular tuples in dev:
```
let x : Int, String = (8,"hello") in x
```

7h3kk1d:
TODO test nested dot operators

src/haz3lcore/statics/Info.re
```diff
@@ -514,8 +533,9 @@ let status_tpat = (ctx: Ctx.t, utpat: TPat.t): status_tpat =>
 /* Determines whether any term is in an error hole. */
 let is_error = (ci: t): bool => {
   switch (ci) {
-  | InfoExp({mode, self, ctx, _}) =>
-    switch (status_exp(ctx, mode, self)) {
+  | InfoExp({status, _}) =>
+    // TODO Confirm with disconcision that we can use the derived status
+    switch (status) {
```
7h3kk1d: @disconcision do you know if there's a reason we were recalculating the status here?

cyrus-:
- singleton analytic case for patterns should insert labels
```
let f : (b=Float) -> Float = 
  fun b -> b
in ...
```

- better messages in the cursor inspector for when lifting occurs

7h3kk1d:
Screenshot from 2024-11-08 10-40-51
Here's a first pass on the "better messages in the cursor inspector for when lifting occurs". We need to figure out how we want to represent this all in statics in a more disciplined way.

@WondAli let me know if you also have any ideas. If we move all the lifting/rearranging to statics maybe there's some additional data we can put on the Info.exp that can be used more generally.

src/haz3lcore/statics/Statics.re
```diff
@@ -209,457 +233,643 @@ and uexp_to_info_map =
...
+        ]),
+      );
+    // We need to reanalyze the elaborated expression to get the statics in the map for the label and tuple
+    let (info, m) =
+      uexp_to_info_map(~ctx, ~mode, ~ancestors, elaborated_exp, m);
+
+    // We need to keep the original status of the expression to get error messages on the unelaborated expression
+    let info = {
+      ...info,
+      status: original_info.status,
```
7h3kk1d, line 283: Copying the status from the inner expression is probably too clever. We may need to have a more principled status for this expression.

7h3kk1d:
> * singleton analytic case for patterns should insert labels
> 
> 
> ```
> let f : (b=Float) -> Float = 
>   fun b -> b
> in ...
> ```
> 

@cyrus- I'm working on the pattern case now and I'm running across an issue I'm curious if you have thoughts on. In the following case the `x` pattern is getting expanded to `(a=x)` as it would in the function case so x ends up just being `3`. So I'm kind of unclear how you actually would bind a variable to the singleton labeled tuple.

`let x : (a=Int) = (a=3)`

cyrus-:
Sorry I don't understand the example, it doesn't have x?

cyrus-:
Ah I see,  yeah that's awkward. Don't have an immediate solution...

cyrus-:
Maybe only patterns that are not under an ascription have the lifting behavior? That would also allow you to explicitly bind a 1-tuple when defining a function:

```
let f : (a=Int) -> Int = 
  fun x : (a = Int) -> x.a
```

and

```
let f : (a=Int) -> Int = 
  fun x -> x
```

seems natural enough to me.

7h3kk1d:
> Maybe only patterns that are not under an ascription have the lifting behavior? That would also allow you to explicitly bind a 1-tuple when defining a function:
> 

This is implemented in the latest commit.

test/Test_Elaboration.re
```diff
@@ -10,7 +10,45 @@ let id_at = x => x |> List.nth(ids);
...
+        },
+    ~f_exp=
+      (fn: Exp.t => Exp.t, t: Exp.t) =>
+        switch (t.term) {
+        | Cast(e, _, _) => strip_casts(e)
+        | _ => fn(t)
+        },
+    p,
+  );
+};
```
7h3kk1d: There has to be a better way to do this.

test/Test_Elaboration.re
```diff
@@ -283,20 +322,38 @@ let ap_of_deferral_of_hole = () =>
...
+        Prod([
           Unknown(Internal) |> Typ.fresh,
-        )
-        |> Exp.fresh,
-      ])
+          Unknown(Internal) |> Typ.fresh,
+        ])
+        |> Typ.fresh,
+      )
       |> Exp.fresh,
```
7h3kk1d: Something in the elaboration for tuples is adding a bunch more casts here

7h3kk1d:
```hazel
let y : (a=String, Float, b=Int) = (a="hello", c=1.2, 3) in
```

Bug where we're allowing extra labels. I don't think we want to allow this.

src/haz3lcore/statics/Info.re
disconcision: @7h3kk1d I don't think so, I think I probably just forgot to update this function when I added derived attributes (including status) into the Info record.

src/haz3lweb/app/explainthis/ExplainThis.re
```diff
@@ -631,6 +631,7 @@ let get_doc =
         let pat_id = List.nth(pat.ids, 0);
         let body_id = List.nth(body.ids, 0);
         switch (pat.term) {
+        // TODO (Anthony): put in a real message
```
7h3kk1d: @WondAli I haven't looked at any of this file so it would be good if you double checked to see if anything remains.

src/haz3lweb/app/explainthis/data/LabelTerm.re
```diff
@@ -0,0 +1,14 @@
+open ExplainThisForm;
+open Example;
+
+let label = (n: string): form => {
+  let explanation = "`%s` is a label (or name) for an item within a tuple.";
+  {
+    id: Label,
+    syntactic_form: [n |> abbreviate |> tpat], // TODO: Fix this
```
7h3kk1d: @WondAli

src/haz3lcore/dynamics/Transition.re
```diff
@@ -148,6 +149,7 @@ module Transition = (EV: EV_MODE) => {
     // Split DHExp into term and id information
     let (term, rewrap) = DHExp.unwrap(d);
     let wrap_ctx = (term): EvalCtx.t => Term({term, ids: [rep_id(d)]});
+    // print_endline(Exp.show(d));
```
7h3kk1d: todo remove

src/haz3lcore/dynamics/Unboxing.re
```diff
@@ -66,11 +68,48 @@ let rec unbox: type a. (unbox_request(a), DHExp.t) => unboxed(a) =
...
+        DoesNotMatch;
+      }
+    | (
+        TupLabel(tl),
+        Cast(t, {term: TupLabel(_, ty1), _}, {term: TupLabel(_, ty2), _}),
+      ) =>
+      let* t = unbox(TupLabel(tl), t);
+      let t = fixup_cast(Cast(t, ty1, ty2) |> DHExp.fresh);
+      Matches(t);
+    // | (TupLabel(tl), Cast(t, ty1, ty2)) =>
```
7h3kk1d: Remove comments

src/haz3lcore/statics/Elaborator.re
```diff
@@ -93,14 +108,28 @@ let elaborated_pat_type = (m: Statics.Map.t, upat: Pat.t): (Typ.t, Ctx.t) => {
     | Ana(ana_ty) =>
       switch (prev_synswitch) {
       | None => ana_ty
-      | Some(syn_ty) => Typ.match_synswitch(syn_ty, ana_ty)
+      | Some(syn_ty) =>
+        // Hack for singleton labeled tuples
```
7h3kk1d: Maybe motivate this more

src/haz3lcore/statics/Elaborator.re
```diff
@@ -131,8 +161,60 @@ let rec elaborate_pattern = (m: Statics.Map.t, upat: Pat.t): (DHPat.t, Typ.t) =>
...
+        |> cast_from(Prod([TupLabel(labty, pty) |> Typ.temp]) |> Typ.temp);
+      };
     | Tuple(ps) =>
-      let (ps', tys) = List.map(elaborate_pattern(m), ps) |> ListUtil.unzip;
+      let (ps', tys) =
+        List.map(elaborate_pattern(m, ~in_container=true), ps)
+        |> ListUtil.unzip;
+      let expected_labels: list(option(string)) =
+        Typ.get_labels(ctx, elaborated_type);
+      // let elaborated_labeled: list((option(string), DHPat.t)) =
```
7h3kk1d: Remove comments

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -317,6 +334,24 @@ and exp_term: unsorted => (Exp.term, list(Id.t)) = {
           | ([";"], []) => Seq(l, r)
           | (["++"], []) => BinOp(String(Concat), l, r)
           | (["$=="], []) => BinOp(String(Equals), l, r)
+          | (["="], []) =>
+            // TODO: Other cases to convert to labels
```
7h3kk1d: @WondAli is there remaining work here?

src/haz3lcore/statics/TermBase.re
```diff
@@ -699,6 +731,11 @@ and Typ: {
     switch (IdTagged.term_of(t1), IdTagged.term_of(t2)) {
     | (Parens(t1), _) => eq_internal(n, t1, t2)
     | (_, Parens(t2)) => eq_internal(n, t1, t2)
+    | (TupLabel(label1, t1'), TupLabel(label2, t2')) =>
+      eq_internal(n, label1, label2) && eq_internal(n, t1', t2')
+    // | (TupLabel(_, _), _) => false // TODO Verify this
+    // | (_, TupLabel(_, _)) => false
```
7h3kk1d: todo

src/haz3lweb/app/explainthis/ExplainThis.re
```diff
@@ -1320,6 +1376,27 @@ let get_doc =
...
+                  msg =>
+                    Printf.sprintf(
+                      Scanf.format_from_string(msg, "%s"),
+                      name,
+                    ),
+                ),
+              LetExp.lets_label,
+            );
+          } else {
+            /* TODO The coloring for the syntactic form is sometimes wrong here... */
```
7h3kk1d: todo

src/haz3lcore/statics/Statics.re
```diff
@@ -181,499 +195,857 @@ and uexp_to_info_map =
...
+        List.map(e => Exp.match_tup_label(e) |> Option.map(fst), es);
+
+      let (inferred_es, modes) =
+        Mode.of_prod(
+          ctx,
+          mode,
+          List.map(e => (None: option(string), e), es),
+          ((inferred, e)) => {
+            Exp.match_tup_label(e)
+            |> Option.map(((label, value)) => (label, (inferred, value)))
```
cyrus-: rename from `value`

src/haz3lcore/statics/MakeTerm.re
```diff
@@ -383,9 +428,30 @@ and pat_term: unsorted => (Pat.term, list(Id.t)) = {
...
+        | _ => child
+        };
+      };
+      let tuple_children = List.map(mapping_fn, tuple_children);
+      ret(Tuple(tuple_children));
+
     | None =>
       switch (tiles) {
+      | ([(_id, (["="], []))], []) =>
+        // TODO: Other cases to convert to labels
```
cyrus-: resolve these

src/haz3lcore/zipper/EditorUtil.re
```diff
@@ -11,13 +11,16 @@ let rec append_exp = (e1: Exp.t, e2: Exp.t): Exp.t => {
   | Int(_)
   | Float(_)
   | String(_)
+  | Label(_)
```
cyrus-: keep cases together

src/haz3lweb/app/explainthis/ExplainThis.re
cyrus-: remove

src/haz3lweb/exercises/SyntaxTest.re
```diff
@@ -201,6 +213,7 @@ let rec var_mention = (name: string, uexp: Exp.t): bool => {
   | BuiltinFun(_) => false
   | Cast(d, _, _) => var_mention(name, d)
   | Ap(_, u1, u2)
+  | Dot(u1, u2)
```
cyrus-, line 216: check to make sure label positions are not being recursed into for variable mentions / recursive functions / etc.

cyrus- (changes requested):
- floating point exponential notation isn't working anymore
- see inline comments

src/haz3lweb/exercises/SyntaxTest.re
7h3kk1d: @cyrus- since we don't have the label sort introduced, the variables in label position are still being evaluated:

Screenshot From 2025-02-10 10-24-37


I can try to make these indet, but it's not clear to me that it's preferable to stop dynamics for the expression. If we still allow evaluation I think we want to continue to recurse into the "label position" in SyntaxTest.

src/haz3lcore/LabeledTuple.re
```diff
@@ -0,0 +1,257 @@
...
+        | None => b
+        },
+      l2_reordered,
+    );
+  };
+
+// rearrange two other lists to match the first list of labels.
+// TODO: Ensure that the two lists match up with each other
+// TODO: Efficiency
+let rearrange2:
```
7h3kk1d: Better name/documentation

src/haz3lweb/exercises/SyntaxTest.re
7h3kk1d: This seems somewhat related to decisions for now evaluation should proceed in non-empty holes https://github.com/hazelgrove/hazel/issues/1497#issuecomment-2640785436.

7h3kk1d:
Reminder to regenerate all doc buffers

src/haz3lweb/exercises/SyntaxTest.re
7h3kk1d: I've put everything in label position in multiholes if it's not a label.

cyrus- (approved):

PR #1241, opened 2024-03-13
Module implicts

xzxzlala:
### Module implicit:

Parameters can be marked implicit which then allows them to be omitted from function calls.
Since we already have module system in hazel, so we decided to take advantage of modules to fulfill this feature. 
The functions would have an implicit module parameter and this parameter needs defined by users explicitly. 

Here is an example from https://www.cl.cam.ac.uk/~jdy22/papers/modular-implicits.pdf:
```
1 module type Show = sig
2   type t
3   val show : t -> string
4 end
5
6 let show { S : Show } x = S . show x
7
8 implicit module Show_int = struct
9    type t = int
10  let show x = string_of_int x
11 end
12
13 implicit module Show_float = struct
14   type t = float
15   let show x = string_of_float x
16 end
17
18 implicit module Show_list { S : Show } = struct
19   type t = S . t list
20   let show x = string_of_list S . show x
21 end
22
23 let () =
24   print_endline (" Show an int: " ^ show 5);
25   print_endline (" Show a float : " ^ show 1.5);
26   print_endline (" Show a list of ints : " ^ show [1; 2; 3]);
```
And here is a draft of what should be implemented in hazel:
image

### Implementation plans:

- Explicitly pass module parameters to functions could work in hazel.
- Add implicit keyword to hazel and let users could define implicit modules.
- Functions could find the corresponding implicit modules automatically and users needn't explicitly pass module parameters to functions.

### What code we may write (Not sure about this, may changes in the process of writing code):

- New types, merge parameterized_types to module system.
- New syntax and corresponding statics/dynamics. (Mold.re, Form.re, MakeTerm.re, Elaborator.re, Evaluator.re, DHExp.re)

### Checkpoint

- [ ] Mold.re
- [ ] Form.re
- [ ] MakeTerm.re
- [ ] Elaborator.re
- [ ] Evaluator.re
- [ ] DHExp.re


cyrus-:
@xzxzlala can you write up a detailed PR description describing the idea, what kind of code we want to be able to write, and what your implementation plan is

cyrus-:
archiving this early design, will revisit in the future

PR #1245, opened 2024-03-14
Fix for typehole join issue

### LINKED ISSUE #1244: Type List(Unknown(Typehole)) is inconsistent with everything
Screenshot 2024-03-14 at 6 26 56 PM

disconcision:
Closes #1244. Not totally sure how this happened so not totally confident in the fix...


cyrus-:
The way join works for unknowns of various provenances is very confusing -- let's sit down and rethink what's happening here some time soon @disconcision @Negabinary. I'll go ahead and merge this for now to fix the immediate issue.

Issue #1240, opened 2024-03-11
Evaluation finishes half way through this awkward cast example


Negabinary:
![image](https://github.com/hazelgrove/hazel/assets/31668468/dcbf7221-a08f-4424-8a1e-edc631a8c634)

`(fun (x,y) :(Int,  ) -> x+1)((fun x -> x)(3.2,  ))`

Negabinary:
so the issue is there's a closure getting in the way of the cast calculus doing its thing.

Closure(e<Ground=>Hole>)<Hole=>Ground>

I'll fix it but it's going to break hole closures, which we're not using for now, but @cyrus- we should have a chat at some point about what hole closures are useful for, so I can think about how to bring them back later.

cyrus-:
@Negabinary okay, can you add this as an agenda item for our next meeting.

PR #1247, opened 2024-03-18
Fix list append and list cons

### LINKED ISSUE #1242: List evaluation logic is missing cast rules
looks like I broke the list casting logic back in #1140 

image

Negabinary:
Fixes #1242

### DIFF
--- a/src/haz3lcore/dynamics/Transition.re
+++ b/src/haz3lcore/dynamics/Transition.re
@@ -571,7 +571,7 @@ module Transition = (EV: EV_MODE) => {
       Indet;
     | Closure(env', d) =>
       let. _ = otherwise(env, d => Closure(env', d))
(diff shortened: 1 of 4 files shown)


cyrus-:
This isn't doing what I would expect: 
image
Perhaps not handling casts from [?] to ? or ? to [?] correctly?

cyrus-:
(the manual definition of append also does something different and unexpected there?)

Negabinary:
image

eek there's a lot of problems here:

1. There should really be a cast from hole to int around the NonEmptyHole around the string; this is an elaboration problem; I'm guessing it's because we're not handing the polymorphism correctly?

2. The evaluator is behaving differently to the stepper, similar to in #1240

3. There's something going on with the serialization again, when clicking through the stepper on your manual example, it runs out of storage alarmingly quickly

PR #1251, opened 2024-03-19
Add comments to syntax tests

ruiz-m:
Add comments explaining the functions in SyntaxTest.re.


src/haz3lschool/SyntaxTest.re
```diff
@@ -30,6 +34,10 @@ let rec find_var_upat = (name: string, upat: Term.UPat.t): bool => {
   };
 };
 
+/*
+ Finds whether a variable with the same name as
+ the argument "name" is ever mentioned in uexp.
+ */
 let rec var_mention = (name: string, uexp: Term.UExp.t): bool => {
```
cyrus-, line 144: why is the upat version called `find_var_upat` rather than `var_mention_upat`?

src/haz3lschool/SyntaxTest.re
```diff
@@ -7,6 +7,10 @@ type syntax_result = {
   percentage: float,
 };
 
+/*
+ Finds whether a variable with the same name as
```
cyrus-: You can just say "the variable `name`" rather than "a variable with the same name as the argument "name"" here and similar elsewhere.

src/haz3lschool/SyntaxTest.re
```diff
@@ -130,6 +143,12 @@ let rec var_applied = (name: string, uexp: Term.UExp.t): bool => {
   };
 };
 
+/*
+ Goes into the pattern of a let expression and finds any mention of a variable
```
cyrus-: This is confusing to me, maybe say which function it is used as a helper in and comment it relative to that?

cyrus- (changes requested):
Maybe add a general comment at the top of SyntaxTest.re as well saying what syntax tests are for.

src/haz3lschool/SyntaxTest.re
ruiz-m: I think it was because the function was written to be a helper function and I did not change it afterwards

cyrus- (approved):

PR #1261, opened 2024-04-01
Command Palette

7h3kk1d:
Adds a command palette triggered by ctrl+k/cmd+k. Uses the https://github.com/ssleptsov/ninja-keys webcomponent.

command_palette

- [x]     add system settings
- [x]     add export stuff and remove the dev export stuff from sidebar
- [x]     shown only OS-relevant keyboard shortcuts
- [x]     reorder default drop down order to show things a typical user would want to do first


src/haz3lweb/Keyboard.re
```diff
@@ -77,7 +77,7 @@ let handle_key_event = (k: Key.t): option(Update.t) => {
     | "d" => now(Select(Term(Current)))
     | "p" => Some(PerformAction(Pick_up))
     | "a" => now(Select(All))
-    | "k" => Some(ReparseCurrentEditor)
+    // | "k" => Some(ReparseCurrentEditor)
```
7h3kk1d: Default keybinding clash.

src/haz3lweb/Main.re
```diff
@@ -102,17 +113,37 @@ module App = {
...
+        Dom_html.addEventListener(
+          Dom_html.document,
+          Dom_html.Event.make("myCustomEvent"),
+          Dom_html.handler(_ => {
+            print_endline("RECEIVED EVENT");
+            go := true;
+            Js.bool(true);
+          }),
+          Js.bool(false),
+        );
```
7h3kk1d: I would love if I could just call inject here but I don't think that works with incr_dom's model.

src/haz3lweb/Main.re
```diff
@@ -102,17 +113,37 @@ module App = {
...
         ~old_model as _: Incr.t(Haz3lweb.Model.t),
-        ~inject,
+        ~inject: UpdateAction.t => Ui_effect.t(unit),
       ) => {
     open Incr.Let_syntax;
     let%map model = model;
     /* Note: mapping over the old_model here may
        trigger an additional redraw */
+    let first = ref(false);
+    let go = ref(false);
```
7h3kk1d: We listen on frame refresh for this mutable value to be set and trigger the action on it being set. Definitely a gross hack but I'm not sure the correct way to deal with external js triggered events.

src/haz3lweb/util/JsUtil.re
```diff
@@ -98,7 +98,7 @@ let log = data => {
 
 let clipboard_shim_id = "clipboard-shim";
 
-let focus_clipboard_shim = () => get_elem_by_id(clipboard_shim_id)##focus;
+let focus_clipboard_shim = () => ();
```
7h3kk1d: The clipboard_shim captures the shortcut triggering. Need to address.

src/haz3lweb/view/Page.re
```diff
@@ -6,20 +6,19 @@ open Node;
...
+      | Some(action) => Many([Prevent_default, inject(action)])
       }
     );
   [
-    Attr.on_keypress(_ => Effect.Prevent_default),
-    Attr.on_keyup(key_handler(~inject, ~dir=KeyUp)),
-    Attr.on_keydown(key_handler(~inject, ~dir=KeyDown)),
+    // Attr.on_keypress(_ => Effect.Prevent_default),
+    // Attr.on_keyup(key_handler(~inject, ~dir=KeyUp)),
+    // Attr.on_keydown(key_handler(~inject, ~dir=KeyDown)),
```
7h3kk1d: Debugging key handling being overwritten.

src/haz3lweb/www/index.html
```diff
@@ -9,6 +9,26 @@
...
+    <link
+      href="https://fonts.googleapis.com/css?family=Material+Icons&display=swap"
+      rel="stylesheet"
+    />
+    <script type="importmap">
+      {
+          "imports": {
+              "https://unpkg.com/lit-html@latest/directives/ref.js?module": "https://unpkg.com/lit-html@2.2.6/directives/ref.js?module"
+          }
+      }
```
7h3kk1d, line 30: The current version of lit-html is broken so this overwrites it for the ninja-keys import.

src/haz3lweb/www/index.html
```diff
@@ -22,6 +42,33 @@
...
+        id: "Go To Definition",
+        title: "Go to Definition",
+        hotkey: "cmd+b",
+        mdIcon: "arrow_forward",
+        handler: () => {
+          console.log("Going to Definition");
+          document.dispatchEvent(customEvent);
+        }
+      }
+    ];
```
7h3kk1d: We don't want to hardcode all of these.

src/haz3lweb/www/ninja_module.js
```diff
@@ -0,0 +1,19 @@
...
+  const isInput = tagName === 'INPUT' && !['checkbox', 'radio', 'range', 'button', 'file', 'reset', 'submit', 'color'].includes(target.type);
+  // ignore: isContentEditable === 'true', <input> and <textarea> when readOnly state is false, <select>
+  if (
+    target.isContentEditable
+    || ((isInput || tagName === 'TEXTAREA' || tagName === 'SELECT') && !target.readOnly)
+  ) {
+    flag = false;
+  }
+  return flag;
+  };
```
7h3kk1d: This is the default hotkeys behavior but reenabling clipboard-shim.

src/haz3lweb/www/index.html
cyrus-: would be good to leave this as a comment in the code

cyrus-:
There are duplicate "Proceed to Next Hole" entries that are bugging out.

Otherwise, the behavior feels good. 

Ultimately, I imagine there being commands with parameters or a vim style verb/noun system, maybe with a simple structure editor -- not sure how easy it would be to adapt this to something more complex than just a textual search of a fixed set of commands?

cyrus-:
(I'm okay with merging this for now to give us some useful functionality, above is to start a discussion about the broader command language. Also relevant to accessibility work.)

cyrus-:
^ cc'ing @coned @heades

7h3kk1d:
I'll clean up the code and add some more actions. I imagine we'll just supplant it with something better in the future, but wanted something for the short term.

src/haz3lweb/Keyboard.re
7h3kk1d: @disconcision do you have an opinion on the keybinding? I'm used to ctrl+k but I'm fine with using something else.

cyrus-:
- add system settings
- add export stuff and remove the dev export stuff from sidebar
- shown only OS-relevant keyboard shortcuts
- reorder default drop down order to show things a typical user would want to do first

7h3kk1d:
@cyrus- The export/import on the sidebar seem to have contextual entries depending on where you are in Hazel. I don't have that setup for the command palette yet. Seems like it would be better to not remove the menu bars since it's useful for discoverability. I can add them all to the command palette for now but it won't disable/enable based on if you're on a scratch slide or exercise.

src/haz3lweb/UpdateAction.re
```diff
@@ -62,6 +62,7 @@ type t =
   | SetMeta(set_meta)
   | UpdateExplainThisModel(ExplainThisUpdate.update)
   | ExportPersistentData
+  | ExportScratchSlide // TODO Combine exports
```
7h3kk1d: @disconcision not sure if you have any reason I shouldn't do this and add all the other exports?

https://github.com/hazelgrove/hazel/blame/2bfaef4a36a671f9cbc0177274d6d0341292e527/src/haz3lweb/view/ExerciseMode.re#L294 and such just bypass the action flow

7h3kk1d:
Extended screenshot
full_command_palette

src/haz3lweb/Keyboard.re
```diff
@@ -3,6 +3,219 @@ open Haz3lcore;
...
+    ~mdIcon="download",
+    ~section="Export",
+    "Export Transitionary Exercise Module",
+    Export(TransitionaryExerciseModule) // TODO Would we rather skip contextual stuff for now or include it and have it fail
+  ),
+  mk_shortcut(
+    ~mdIcon="download",
+    ~section="Export",
+    "Export Grading Exercise Module",
+    Export(GradingExerciseModule) // TODO Would we rather skip contextual stuff for now or include it and have it fail
```
7h3kk1d: These are the contextual entries I'm just showing for now that will fail for now.

7h3kk1d: Remove instructor ones if built for instructor mode

cyrus-:
Everything looks good to me now. Only remaining question is discoverability: there isn't anything in the UI to tell a new user that there is a command palette available via the shortcut key. Perhaps we can add an entry to the nut menu to alert folks of this fact?

7h3kk1d:
> Everything looks good to me now. Only remaining question is discoverability: there isn't anything in the UI to tell a new user that there is a command palette available via the shortcut key. Perhaps we can add an entry to the nut menu to alert folks of this fact?

I added a nut menu entry. I got an MIT licensed SVG off the internet and copied the attribution in a comment. Not sure where else we'd need to put that or if @disconcision has a better way to source icons
[Screencast from 2024-08-08 11-18-07.webm](https://github.com/user-attachments/assets/dfd4378f-2992-4d3a-aa29-61454cd71e02)

7h3kk1d:
I've merged dev into the PR but have not verified everything works as expected with projector interactions. Trying that now. Also some of the actions changed in the projectors PR so I need to verify they still do similar things.

src/haz3lweb/Keyboard.re
```diff
@@ -4,6 +4,231 @@ open Util;
...
+      "Go to Previous Hole",
+      PerformAction(Move(Goal(Piece(Grout, Left)))),
+    ),
+    mk_shortcut(
+      ~mdIcon="swipe_right_alt",
+      ~section="Navigation",
+      "Go To Next Hole",
+      PerformAction(Move(Goal(Piece(Grout, Right)))),
+      // Tab is overloaded so not setting it here
+    ),
```
7h3kk1d, line 84: @disconcision do these names still make sense or is the action doing something different?

7h3kk1d:
The only issue I see is that the textarea projector seems to capture the ctrl+k for the command palette so it doesn't trigger if focused. I think this is fine for now and we can figure out a solution as part of future projector work when we have a more rigorous approach to events.

src/haz3lweb/Keyboard.re
disconcision: Yeah the names are fine. PS the overloading of Tab is on a conceptual level less ad-hoc than the current implementation suggests. In the terms of the new syntax model, the action of tab can be construed as "advance to next obligation, or if already on one, complete it".

7h3kk1d: I'll probably leave that out of the palette for this PR. Once I have a method for contextual actions I'll switch it to full obligation and go to next obligation

disconcision:
@7h3kk1d I've been sourcing all icons from https://thenounproject.com/, for which I've long had a subscription. Gives an attribution-free license. If you find something reasonable there I can send you the path, or I can pick something, whichever. Current is fine but we might as well avoid additional licensing.

After merge just file an issue for the ctrl+k thing. I think I want to just try and accumulate such issues with the textarea for a bit to make sure that eventual interventions are targeting a reasonable level of generality.

The look seems good to me; I'm developing slightly more rigorous design guidelines with #1343, so post-that I'll probably want to make some adjustments. If this gets merged first I can just do the adjustments myself. Basically I think we'd want to command palette to use the same styling as the new nut menu submenus, which it's not far off from

cyrus-:
> @7h3kk1d I've been sourcing all icons from https://thenounproject.com/, for which I've long had a subscription. Gives an attribution-free license. If you find something reasonable there I can send you the path, or I can pick something, whichever. Current is fine but we might as well avoid additional licensing.
> 
> After merge just file an issue for the ctrl+k thing. I think I want to just try and accumulate such issues with the textarea for a bit to make sure that eventual interventions are targeting a reasonable level of generality.
> 
> The look seems good to me; I'm developing slightly more rigorous design guidelines with #1343, so post-that I'll probably want to make some adjustments. If this gets merged first I can just do the adjustments myself. Basically I think we'd want to command palette to use the same styling as the new nut menu submenus, which it's not far off from

@disconcision Can you make a quick PR that adds a `docs/Icons.md` file with this in it for future reference by students? Does the noun project require a subscription to get the attribution-free license? If so we may need to look into getting a lab subscription to make sure we're above board.

cyrus- (approved):

disconcision:
@cyrus- yes, it requires a subscription (for non-attributed used). as in, you have to be subscribed when you download the icons, not in perpetuity. my read on the legal agreement is that we're both already in the clear, but i'm not an expert

PR #1266, opened 2024-04-14
Fix the stacked error holes for cases with multiple branches

pigumar1:
This pull request fixes the "stacked error holes" issue for case expressions with multiple branches:
image

This is what'll happen in the current Hazel. The reason is that the ids of `|` for each rule is added to `M.error_ids` before the type checking (I believe at the stage of parsing), so they can be error-wrapped. Yet the ids of the rules point to the info of the whole case expression, so when **the whole case expression is erroneous**, multiple error holes will occur.

This pull request tries to fix this by filtering out all the **unrepresentative ids** in `M.error_ids`, such that the ids of `|` for each rule will be filtered out.
image

Not sure if there is a better way to do this though.


disconcision (changes requested):
The general approach is sound, but we can probably push this upstream. There is already informap filtering going on to calculate M.error_ids in the first place, we can likely just do this filtering as part of that same pass. Relevant code should be in Editors.re or Editor.re

pigumar1:
@disconcision 
image
image
It seems that `error_ids` are generated based on the editor's `term_ranges`. Should I directly modify `Editor.re` or modify `ScratchSlide.re`? I'm a little bit cautious here...

disconcision:
@pigumar1 the fact that term_ranges is being used here is a bit of a hack, pending david's new syntax system. It may be that's it's unnecessary after this change; basically it was serving to filter out certain ids which were not present in the infomap, for reasons I don't immediately recall. it likely has to do with representative ids, but there might be a further wrinkle to do with certain forms (like list literals) which are technically composed of two different forms (the square brackets + the commas) at the syntax level, so it might be premature to simply remove it.

(the above is just for context)

In any case, you should be able to do the rep-id filtering in the Statics function called there, i.e. Statics.Map.error_ids, which folds over the info_map

pigumar1:
I upstreamed the implementation according to @disconcision 's comment (if I'm understanding correctly). This implementation uses `Info.id_of`, which implicitly applies `Term.rep_id`.

disconcision:
@pigumar1 I meant that the `id == Info.id_of(info)` check should be done within the fold inside the implementation of Statics.Map.error_ids, so that we're not unnecessarily iterating over the entire info_map more than once. (Also, the filtering being done there is important, as doing it in ScratchSlide means that the current fix doesn't work for Exercises mode)

pigumar1:
@disconcision Hopefully I got it correct this time XD
image
image

disconcision (approved):
lgtm

Issue #1252, opened 2024-03-19
polymorphic equality and inequality


disconcision:
@cyrus- are you still okay making the equality operator polymorphic? I've done this on the LLM branch for convenience, so I'll copy-paste some code here to accelerate someone doing it for real.

1. Add case to statics.re:
```
| BinOp(Int(Equals | NotEquals), e1, e2) =>
  let (e1, m) = go(~mode=Syn, e1, m);
  let (e2, m) = go(~mode=Ana(e1.ty), e2, m);
  add(~self=Just(Bool), ~co_ctx=CoCtx.union([e1.co_ctx, e2.co_ctx]), m);
```

2. Add cases to dynamics 
```
    | BinIntOp(Equals, d1, d2) =>
      let* r1 = evaluate(env, d1);
      let* r2 = evaluate(env, d2);
      switch (r1, r2) {
      | (BoxedValue(d1'), BoxedValue(d2')) =>
        let b =
          DHExp.fast_equal(DHExp.strip_casts(d1'), DHExp.strip_casts(d2'));
        return(BoxedValue(BoolLit(b)));
      | (BoxedValue(d1'), Indet(d2'))
      | (Indet(d1'), BoxedValue(d2'))
      | (Indet(d1'), Indet(d2')) =>
        return(Indet(BinIntOp(Equals, d1', d2')))
      };

    | BinIntOp(NotEquals, d1, d2) =>
      let* r1 = evaluate(env, d1);
      let* r2 = evaluate(env, d2);
      switch (r1, r2) {
      | (BoxedValue(d1'), BoxedValue(d2')) =>
        let b =
          !DHExp.fast_equal(DHExp.strip_casts(d1'), DHExp.strip_casts(d2'));
        return(BoxedValue(BoolLit(b)));
      | (BoxedValue(d1'), Indet(d2'))
      | (Indet(d1'), BoxedValue(d2'))
      | (Indet(d1'), Indet(d2')) =>
        return(Indet(BinIntOp(Equals, d1', d2')))
      };
```

3. Fix #1243 to prevent above from crashing

The above is sufficient to get it basically working. Remaining issues:

[...]

PR #1267, opened 2024-04-15
Polymorphic equality

GuoDCZ:
#1252.

Enable polymorphic equal / not equal comparison for:
 - `Int`, `Bool`, `Float`, `String`: Compare value
 - `Tuple` & `List`: Pairwise comparison. For `List`, also check internal type consistency
 - `Sum`: Same as Ocaml
We disallow comparison of inconsistent types and types contain arrow:
 - Dynamic / Static raised `Inconsistent` when inconsistent types of both side is detected.
 - Dynamic / Static raised `CompareArrow` when either a primitive is an arrow or a compound type contains an arrow
   - A `Sum` type raise `CompareArrow` if any entry contains an arrow argument.

Front-end quick tests:
```typescript
test 1 == 2 end;
test 1. == 1. end;
test false!= false end;
test "str" == "str" end;
test [1,2,3]!= [2,1] end;
test (false, "") == (false, "") end;
type T = +A+B(Int) in
test A == B(1) end;
test A == A end;
test B(2) == B(1) end;
#Static Error: type inconsistency#
1 == 1.;
"" == false;
("1", false) == (1, true);
[1.] == [1];
[1.,  ] == [ , 1];
#No Static Error: hole#
(1,  ) == ( , 2.);
[1,  ] == [ ];
#Internal Inconsistency#
[1, false] == [1, false];
type T1 = +B+C in
A == C;
#Static Error: type contains arrow#
string_compare == string_compare;
let f = fun x -> x in
f == f;
let a = (1, [f]) in
a == a; 
let a : +A+B(Int->Int) =    in
a == a;
let a : rec R -> A(Int->Int) + R =    in
a == a;
let a = typfun A -> fun x : A -> x in
a == a;
type T2 = +A(Int->Int)+B in
B == B;
#Hidden inconsistency caught in dynamic#
let x: ? = 1 in 
let y: ? = 1. in
test x == y end;
#Hidden containing arrow caught in dynamic#
let f: ? = fun x -> x in
test f == f end;
let b1 : ? = B in
test b1 == b1 end; 
```


cyrus-:
@GuoDCZ let's better handle functions:

* there should be a static error on an equivalence where either side has arrow type
* there should be a dynamic error if at run-time, either side evaluates to a function or cast to arrow type

GuoDCZ:
@cyrus- just to make sure I got the idea. 

When we refer to static error, does it mean something like:
image
If so, I don't think `Info.t.error_inconsistent` fit our situation. I am thinking about introducing a new synthesis mode `Mode.t.SynEqual`, who raises a new error `Info.t.error_unexpectedArrow` when finding an arrow type. Because it may influence existing code base, IDK if it's OK to do so.

cyrus-:
Yes. I'm not sure you need a new synthesis mode, but a new error definitely makes sense.

Negabinary:
I think equality with holes should probably not evaluate - e.g. `2 == ?` should stay as `2 == ?` rather than evaluate with the unknown, because it's possible the unknown will eventually be filled with the number `2`.

GuoDCZ:
I've encountered an issue with handling the Constructor case. The runtime (dynamic) DHExp.Constructor does not distinguish between arrow type constructors and non-arrow type constructors. Currently, DHExp.Constructor is treated as an arrow type, which results in Constructors always being identified as Indet. 

Here are the behaviors observed with the current deployment. Could anyone clarify which of these are expected?

image

cyrus-:
That behavior is fine -- prohibiting comparisons at arrow type is the intended behavior, even if that is a constructor.

cyrus-:
I think the error when the two sides are not consistent types should be on the == operation in all cases and indicate that the two sides have inconsistent types, a la the following for list literals (and similar for conditionals):
image

cyrus-:
Great. Some last polish things:
- change the cursor inspector and ExplainThis system, which currently still uses the phrase "Integer Equality". We should change it to "Structural Equality"
- the error when there is an arrow type should say "Cannot compare arrow types: ty1 -> ty2"
- When there is a runtime comparison between functions there should be a run time error with an error message, not just a stuck term.

GuoDCZ:
Remaining issue is that I raise Error for every Constructor type, even if it is not an arrow.
image
This is because I don't know how to check if a Constructor is an arrow type dynamically. `DHExp.Constructor` is defined as follows
https://github.com/hazelgrove/hazel/blob/60102ead840d7a5986ad2839d7ba9e2435121c02/src/haz3lcore/dynamics/DH.re#L42
Maybe we can handle #1313 before we forge on Constructor case? We won't like to compare Constructor of different types.

GuoDCZ:
When comparing types that contain an Arrow in static stages, it results in a dynamic error handled within `Elaborator.re`. However, there is an exception as followings: if the type contains a constructor with Arrow argument, comparing it also requires comparing Arrow, which should not happened. This can only be found out in dynamic stage.
image

cyrus-:
We should statically prevent comparison at any sum type where any constructor has an argument that contains an arrow type (recursively including in components of sums and products).

cyrus-:
We will still need run-time checking as well I suppose, due to polymorphism (we may not know the type it is instantiated with statically), but let's give errors here as early as possible.

GuoDCZ:
Workflow:
Static Check:
1. Type inconsistency
2. Type containing Arrow 
  - Recursive case: Var, List, Rec, Sum, Prod
Dynamic Check:
3. Type inconsistency 
  - No double check if checked in (1). 
  - Refer to Cast/FailCast
4. Type containing Arrow
  - No double check if checked in (2). 
  - Constructor may not able to access full type (?to be checked)
5. Evaluation
  - Write dedicated comparing function
  - Raise EvaluatorError for Type Inconsistent case
  - Raise EvaluatorError for Type That not allow to compare (or should not be evaled)
  - (No EvaluatorError should be raised if Everything is good.

GuoDCZ:
Comments for last commit:
- [x] prevent comparison at any sum type where any constructor has an argument that contains an arrow type
- [x] If A static error is found, Transistor will do nothing but `Indet` (to avoid duplicate error msg), otherwise make dynamic checks. Errors are found as early as possible.

Remaining Issues:
1. We may create a ListLit of different types without static/dynamic error. And the following comparison happens. If this is not intended, maybe we should handle it in ListLit?
image
```typescript
let a : ? = 1. in
[1, a] == [1, a]
```
2. In Hazel type keyword is just alias. In DHExp, `Constructor(_) | Ap(Constructor(_),_)` have no access to the original sum type. I don't know how to trace its sum type dynamically if errors are not caught in static phase. e.g. Types hided by holes.
```typescript
type B = +A+B(Int->Int) in
let a : ? = A in
let b : ? = B in
a == b
```
image
Good news are we can proper handle sum types if the types are known in static stage.
3. `IfThenElse` has a similar case when branches types are found and checked in dynamic phase. In case of inconsistency, it use `Indet` instead of an explicit error like I did for Polymorphic Equality. Maybe we should change either one?
```typescript
let a : ? = 2. in
1 == a
```
image

cyrus-:
@GuoDCZ the recent merge of #1121 modified elaboration to include the type in the elaborated constructor, so can you update this branch to make use of that to address the issues you identified above? marking as draft until then.

GuoDCZ:
Remaining type inconsistency checker issue. Can be solved by fixing #1370

cyrus-:
@GuoDCZ is this ready to review? if so go ahead and click that button

src/haz3lcore/dynamics/DHExp.re
```diff
@@ -166,3 +166,234 @@ let ty_subst = (s: Typ.t, tpat: TPat.t, exp: t): t => {
...
+  | Closure(_)
+  | Parens(_)
+  | Cons(_)
+  | ListConcat(_)
+  | UnOp(_)
+  | BinOp(_)
+  | Match(_)
+  | Dot(_)
+  | Cast(_) => false
+  | Fun(_)
```
Negabinary: can you add a comment explaining why fun is comparable? - or if I'm understanding it right, maybe this function could be named something like is_value?

src/haz3lcore/dynamics/Transition.re
```diff
@@ -529,6 +529,44 @@ module Transition = (EV: EV_MODE) => {
         kind: BinBoolOp(Or),
         is_value: false,
       });
+    | BinOp(Int((Equals | NotEquals) as op), d1, d2) =>
```
Negabinary: I wonder whether we should rename BinOp(Int(Equals)) to BinOp(Equals) or BinOp(Poly(Equals))

cyrus-:
@GuoDCZ can you add some unit tests for this functionality

src/haz3lcore/dynamics/Transition.re
```diff
@@ -529,6 +529,44 @@ module Transition = (EV: EV_MODE) => {
...
+        );
+      let (d1', d2') = (DHExp.strip_casts(d1'), DHExp.strip_casts(d2'));
+      // Note(zhiyao): detect either side contains a hole or intermediate type
+      if (!DHExp.ty_comparable(d1') || !DHExp.ty_comparable(d2')) {
+        Indet;
+      } else {
+        let expr: t =
+          if (!DHExp.ty_consistent(d1', d2')) {
+            DynamicErrorHole(
+              BinOp(Int(op), d1', d2') |> rewrap,
```
Negabinary: is there a reason we need to put fully-cast-stripped expressions here? It might be nice if we could keep the inner casts on output.

Negabinary (changes requested):
Playing around I found a statics bug:

`(5 == 3.0): ` doesn't have an error, when I think it should

image

Likewise this message is slightly strange

image

Also to echo Cyrus' comment, iirc you put a lot of effort in to dealing with edge cases (playing around with it now it does feel pretty robust to weird sum constructor/cast situations) it would be helpful if you could add some tests to stop anyone else from later breaking those edge cases.

GuoDCZ:
I am trying to add a test case. But I get:

```
 Received: `{ term =
  (BinOp ((Int Equals), { term = EmptyHole; annotation = None },
     { term = EmptyHole; annotation = None }));
  annotation = None }'
  ```
  
  where an error of comparing arrow is expected. I am sure this is a static error tested from frontend. Having no idea why It does not appear in testing.

```js
      test_case("Polymorphic equal compare arrow 2", `Quick, () => {
        annotated_tree_test(
          {|let a : +A+B(Int->Int) = in (a == a)|},
          FIError.(
            Exp.(
              bin_op(
                Int(Equals),
                empty_hole(),
                empty_hole(),
                ~ann=
                  Some(
                    FTemp.Typ.(
                      Exp(
                        Common(
                          Inconsistent(
                            CompareArrow(
                              sum([
                                Variant("A", [], None),
                                Variant("B", [], Some(arrow(int(), int()))),
                              ]),
                            ),
                          ),
                        ),
                      )
                    ),
                  ),
              )
            )
          ),
        )
      }),
```

cyrus-:
@GuoDCZ there are merge conflicts here again -- can you resolve those and ping me and we can try to get this merged asap?

(regarding your question about testing, not immediately sure, maybe take another look at that problem after merge?)

7h3kk1d:
@GuoDCZ let me know if you need help with tests. But how is this supposed to work for actual polymorphic values?

Screenshot From 2025-04-21 11-42-51

GuoDCZ:
> @GuoDCZ let me know if you need help with tests. But how is this supposed to work for actual polymorphic values?
> 
> Screenshot From 2025-04-21 11-42-51

I assume this behavior is intended: we avoid arrow comparison, detected in static stage.

7h3kk1d:
> > @GuoDCZ let me know if you need help with tests. But how is this supposed to work for actual polymorphic values?
> > Screenshot From 2025-04-21 11-42-51
> 
> I assume this behavior is intended: we avoid arrow comparison, detected in static stage.

We're not detecting it in the static stage for type variables though. I would think we should block those the same way we do arrows. Otherwise you get no static errors but still get the dynamic one.

GuoDCZ:
> > > @GuoDCZ let me know if you need help with tests. But how is this supposed to work for actual polymorphic values?
> > > Screenshot From 2025-04-21 11-42-51
> > 
> > 
> > I assume this behavior is intended: we avoid arrow comparison, detected in static stage.
> 
> We're not detecting it in the static stage for type variables though. I would think we should block those the same way we do arrows. Otherwise you get no static errors but still get the dynamic one.

I see. So the error only shows up at Dynamic now, but ideally it would be caught statically. If I understand you correctly, we should raise a static error when the type function is specialized with a type like `Bool -> Bool`.

The issue is that the body of the type function `typfun a -> fun x : a -> x == x` has already been type-checked before we specialize it with `foo@<Bool -> Bool>(id)`. So to catch this error statically, we’d either need to:

1. Go back and re-type-check the body once we know the concrete type of a, OR
2. Delay type-checking the body entirely until specialization.

Both options would require non-trivial changes to how `TypFun` work in the static stage. I’m not sure whether it is worth adding it.

7h3kk1d:
> I see. So the error only shows up at Dynamic now, but ideally it would be caught statically. If I understand you correctly, we should raise a static error when the type function is specialized with a type like `Bool -> Bool`.
> 
> The issue is that the body of the type function `typfun a -> fun x : a -> x == x` has already been type-checked before we specialize it with `foo@<Bool -> Bool>(id)`. So to catch this error statically, we’d either need to:
> 
>     1. Go back and re-type-check the body once we know the concrete type of a, OR
> 
>     2. Delay type-checking the body entirely until specialization.
> 
> 
> Both options would require non-trivial changes to how `TypFun` work in the static stage. I’m not sure whether it is worth adding it.

I wouldn't go back and typecheck or delay checking until specialization. I think we should just mark the statics as a failure if the operands are type variables after normalization so it would always get a mark even if it works after specialization. @cyrus- might have other thoughts though.

cyrus-:
Well, some potentially desirable uses of polymorphic equality requires that it work for a statically quantified type, e.g. using polymorphic equality to implement a Set or Dict type.

I'd suggest the semantics be that there is an error if the types involved are definitely arrows or foralls. Otherwise, it works and results in a run-time error if you end up with a function or type function on either side.

GuoDCZ:
> Well, some potentially desirable uses of polymorphic equality requires that it work for a statically quantified type, e.g. using polymorphic equality to implement a Set or Dict type.
> 
> I'd suggest the semantics be that there is an error if the types involved are definitely arrows or foralls. Otherwise, it works and results in a run-time error if you end up with a function or type function on either side.

Okay, the current behavior seems consistent with the latter one.

cyrus-:
@GuoDCZ okay, there are a few small merge conflicts now but after that if it's ready for review I can go ahead and do that

GuoDCZ:
@cyrus- @Negabinary This PR is ready for review

src/web/app/explainthis/ExplainThis.re
```diff
@@ -2029,12 +2029,12 @@ let get_doc =
...
-          | Int(Equals) => (int_equal, int_eq_exp_coloring_ids)
-          | Nat(NotEquals)
-          | SInt(NotEquals)
-          | Int(NotEquals) => (int_not_equal, int_neq_exp_coloring_ids)
+          /* | Nat(Equals)
+             | SInt(Equals)
+             | Poly(Equals) => (int_equal, int_eq_exp_coloring_ids)
+             | Nat(NotEquals)
+             | SInt(NotEquals)
+             | Int(NotEquals) => (int_not_equal, int_neq_exp_coloring_ids) */
```
disconcision: commented code

disconcision: can prob get rid of what this is referencing too

src/web/app/explainthis/ExplainThisForm.re
```diff
@@ -79,7 +79,10 @@ type numeric_bin_op_examples =
   | GreaterThanEqualEqual
   | GreaterThanEqualFalse
   | EqualFalse
-  | EqualTrue;
+  | EqualTrue
```
disconcision: are these non poly equal cases still live?

src/menhirParser/Parser.mly
```diff
@@ -181,12 +185,12 @@ program:
...
     | POWER_FLOAT { FloatOp(Power) }
     | DIVIDE_FLOAT { FloatOp(Divide) }
-    | DOUBLE_EQUAL_FLOAT { FloatOp(Equals) }
-    | NOT_EQUAL_FLOAT { FloatOp(NotEquals) }
     | LESS_THAN_FLOAT { FloatOp(LessThan) }
     | LESS_THAN_EQUAL_FLOAT { FloatOp(LessThanOrEqual) }
     | GREATER_THAN_FLOAT { FloatOp(GreaterThan) }
     | GREATER_THAN_EQUAL_FLOAT { FloatOp(GreaterThanOrEqual) }
+    | DOUBLE_EQUAL_FLOAT { FloatOp(Equals) }
+    | NOT_EQUAL_FLOAT { FloatOp(NotEquals) }
```
disconcision, line 188: do we still have float equals as a separate operator?

src/menhirParser/Lexer.mll
```diff
@@ -83,12 +84,12 @@ rule token =
...
     | "/." { DIVIDE_FLOAT }
     | "**." {POWER_FLOAT}
-    | "==." { DOUBLE_EQUAL_FLOAT }
-    | "!=." { NOT_EQUAL_FLOAT }
     | "<." { LESS_THAN_FLOAT}
     | "<=." { LESS_THAN_EQUAL_FLOAT }
     | ">." { GREATER_THAN_FLOAT }
     | ">=." { GREATER_THAN_EQUAL_FLOAT }
+    | "==." { DOUBLE_EQUAL_FLOAT }
+    | "!=." { NOT_EQUAL_FLOAT }
```
disconcision, line 89: is it a considered decision to retain this? float equals has always seemed like a slightly fraught operator semantically.

src/web/app/explainthis/data/OpExp.re
```diff
@@ -430,34 +445,74 @@ let int_eq_exp_coloring_ids =
     ~left_id,
     ~right_id,
   );
-let int_eq_exp: form = {
-  let explanation = "If the [*left operand*](%s) is equal to the [*right operand*](%s), evaluates to `true`. Otherwise, evaluates to `false`.";
+/* let int_eq_exp: form = {
```
disconcision: commented code

src/web/app/explainthis/data/OpExp.re
```diff
@@ -815,15 +870,15 @@ let int_greater_than_equal: group = {
   forms: [int_gte_exp],
 };
 
-let int_equal: group = {
-  id: BinOpExp(Int(Equals)),
-  forms: [int_eq_exp],
-};
+/* let int_equal: group = {
```
disconcision: commented code

src/web/app/inspector/CursorInspector.re
```diff
@@ -173,6 +174,11 @@ let common_err_view =
           : [text("Invalid labels: "), ...List.map(code, invalid_labels)]
       )
     | DuplicateLabel(name, _) => [text("Duplicate Label:"), code(name)]
+    | Inconsistent(CompareArrow(ty)) => [
+        text("cannot compare"),
+        view_type(ty),
+        text("with arrow type inside"),
```
disconcision: "values of arrow-containing <ty> cannot be compared" might read a little better

test/Test_Elaboration.re
```diff
@@ -401,7 +410,7 @@ module PlainTests = {
     test_case("Consistent if statement", `Quick, consistent_if),
     test_case("An unapplied function", `Quick, unapplied_function),
     test_case("Application of function on free variable", `Quick, ap_fun),
-    test_case("Inconsistent case statement", `Quick, inconsistent_case),
+    /* test_case("Inconsistent case statement", `Quick, inconsistent_case), */
```
disconcision: commented code

test/Test_Menhir.re
```diff
@@ -269,11 +269,11 @@ let tests =
         ty_alias(TPat.var("x"), Typ.int(), int(1)),
         "type x = Int in 1",
       ),
-      full_parser_test(
-        "Test",
-        test(bin_op(Int(Equals), int(3), int(3))),
-        "test 3 == 3 end",
-      ),
+      /* full_parser_test(
```
disconcision: commented code

src/language/dynamics/DHExp.re
```diff
@@ -127,6 +127,218 @@ let ty_subst = (s: Typ.t, tpat: TPat.t, exp: t): t => {
...
+  | (Cons(_), _)
+  | (ListConcat(_), _)
+  | (UnOp(_), _)
+  | (BinOp(_), _)
+  | (Match(_), _)
+  | (Dot(_), _)
+  | (LivelitName(_), _)
+  | (Fun(_), _)
+  | (TypFun(_), _)
+  | (Probe(_), _)
```
disconcision: should treat probes like parens... just compare the expression part

src/language/dynamics/DHExp.re
```diff
@@ -127,6 +127,218 @@ let ty_subst = (s: Typ.t, tpat: TPat.t, exp: t): t => {
...
+  | (Match(_), _)
+  | (Probe(_), _)
+  | (Use(_), _)
+  | (Dot(_), _)
+  | (LivelitName(_), _) => false
+  | (Parens(d1), _) => ty_consistent(d1, d2)
+  | (_, Parens(d2)) => ty_consistent(d1, d2)
+  | (Cast(d1, _, _), _) => ty_consistent(d1, d2)
+  | (_, Cast(d2, _, _)) => ty_consistent(d1, d2)
+  // TODO(zhiyao): are we allowed to compare int/sint/nat?
```
disconcision: good question... @cyrus- ?

disconcision: our treatment of e.g. 2 as a float where appropriate seems to clash a bit with our treatment here but i dunno

disconcision (approved):
lgtm pending cleanup and dev merge... one minor issue with probes to fix

src/language/dynamics/DHExp.re
cyrus-: no they should be explicitly converted to have the same types for equality to make sense

src/language/dynamics/DHExp.re
```diff
@@ -124,6 +124,234 @@ let ty_subst = (s: Typ.t, tpat: TPat.t, exp: t): t => {
...
+    | (Bool(_), _) => false
+    | (String(_), String(_)) => true
+    | (String(_), _) => false
+    | (Float(_), Float(_)) => true
+    | (Float(_), _) => false
+    }
+  | (Atom(_), _) => false
+  | (Label(_), Label(_)) => true
+  | (Label(_), _) => false
+  | (TupLabel(l1, d1), TupLabel(l2, d2)) =>
```
cyrus-, line 183: check labels for equality here

src/language/term/InvalidOperationError.re
```diff
@@ -5,7 +5,9 @@ type t =
   | DivideByZero
   | NegativeExponent
   | NegativeNat
-  | IntegerTooBig;
+  | IntegerTooBig
+  | CompareArrow
+  | Inconsistent;
```
cyrus-: see above re:Incomparable

src/language/dynamics/DHExp.re
```diff
@@ -124,6 +124,234 @@ let ty_subst = (s: Typ.t, tpat: TPat.t, exp: t): t => {
   };
 };
 
+let rec ty_consistent = (d1, d2) => {
```
cyrus-: ty_comparable

src/language/dynamics/transition/Transition.re
```diff
@@ -565,6 +565,24 @@ module Transition = (EV: EV_MODE) => {
       // Operator semantics are defined in Operators.re
       switch (Operators.semantics_of_bin_op(op)) {
       | Undefined(_) => Indet
+      | DefinedPoly(poly_op) =>
+        let expr: t =
+          if (!DHExp.ty_consistent(d1, d2)) {
+            DynamicErrorHole(BinOp(op, d1, d2) |> rewrap, Inconsistent)
```
cyrus-: rename this error to Incomparable

src/language/dynamics/transition/Transition.re
```diff
@@ -565,6 +565,24 @@ module Transition = (EV: EV_MODE) => {
...
+            |> fresh;
+          } else if (DHExp.ty_has_arrow(d1) || DHExp.ty_has_arrow(d2)) {
+            DynamicErrorHole(BinOp(op, d1, d2) |> rewrap, CompareArrow)
```
cyrus-: maybe combine with Incomparable

src/language/term/InvalidOperationError.re
```diff
@@ -15,4 +17,6 @@ let err_msg = (err: t): string =>
   | NegativeExponent => "Error: Negative Exponent in Integer Exponentiation (Consider using **.)"
   | NegativeNat => "Error: Cannot convert negative number to Nat"
   | IntegerTooBig => "Error: Integer too big"
+  | CompareArrow => "Error: Comparison of Arrow Types"
+  | Inconsistent => "Error: Inconsistent Type"
```
cyrus-: update error message to say "Incomparable types"

src/language/term/Typ.re
```diff
@@ -133,6 +133,33 @@ let rec is_arrow = (typ: t) => {
   };
 };
 
+let rec has_arrow = (ctx: Ctx.t, typ: t) =>
```
cyrus-: has_fun

src/language/term/Typ.re
```diff
@@ -133,6 +133,33 @@ let rec is_arrow = (typ: t) => {
...
+  switch (typ.term) {
+  | Parens(typ) => has_arrow(ctx, typ)
+  | TupLabel(_, typ) => has_arrow(ctx, typ)
+  | Arrow(_)
+  | Forall(_) => true
+  | Unknown(_)
+  | Atom(_)
+  | Label(_) => false
+  | Var(tvar) =>
+    switch (Ctx.lookup_alias(ctx, tvar)) {
```
cyrus-: normalize before calling this function so you don't have to handle these things -- then you don't need a context

src/web/app/explainthis/data/OpExp.re
cyrus-, line 1: add an example involving a compound type, e.g. a list / tuple

src/web/app/inspector/CursorInspector.re
```diff
@@ -160,6 +161,10 @@ let common_err_view =
           : [text("Invalid labels: "), ...List.map(code, invalid_labels)]
       )
     | DuplicateLabel(name, _) => [text("Duplicate Label:"), code(name)]
+    | Inconsistent(CompareArrow(ty)) => [
+        text("values of arrow-containing cannot be compared:"),
```
cyrus-: values cannot be compared

test/evaluator/Test_Evaluator_Poly_Equal.re
```diff
@@ -0,0 +1,96 @@
...
+          bool(false),
+          elaborate(parse_exp("false != false")),
+        );
+        evaluation_test(
+          "Equal String",
+          bool(true),
+          elaborate(parse_exp("\"str\" == \"str\"")),
+        );
+        evaluation_test(
+          "Equal List",
```
cyrus-: add a positive test

test/evaluator/Test_Evaluator_Poly_Equal.re
```diff
@@ -0,0 +1,96 @@
...
+          bool(true),
+          elaborate(parse_exp("[1,2,3] != [2,1]")),
+        );
+        evaluation_test(
+          "Equal Tuples",
+          bool(true),
+          elaborate(parse_exp("(false, \"\") == (false, \"\")")),
```
cyrus-, line 46: add a negative test

test/evaluator/Test_Evaluator_Poly_Equal.re
```diff
@@ -0,0 +1,96 @@
...
+        );
```
cyrus-, line 47: add labeled tuple tests

test/evaluator/Test_Evaluator_Poly_Equal.re
```diff
@@ -0,0 +1,96 @@
...
+      () => {
+        evaluation_test(
+          "Hidden inconsistency caught in dynamics",
+          dynamic_error_hole(elaborate(parse_exp("1 == 1.")), Inconsistent),
+          elaborate(parse_exp("let x: ? = 1 in let y: ? = 1. in x == y")),
+        );
+        evaluation_test(
+          "Hidden containing arrow caught in dynamics",
+          dynamic_error_hole(
+            elaborate(parse_exp("(fun x -> x) == (fun x -> x)")),
```
cyrus-, line 105: add typfun test

cyrus- (changes requested):

disconcision:
This will close #1814

PR #1274, opened 2024-04-18
prevent unnecessary fixpoints in elaboration

### LINKED ISSUE #1268: prevent unnecessary fixpoints in elaboration
It appears that even non-recursive functions now have fixpoints around them:
image

@pigumar1 maybe you can take a look at this since you were working on related code recently? not sure if it is directly connected to your PR

pigumar1:
This PR tries to fix #1268 by checking if there are self-references in the function body/bodies during elaboration and decide whether to include the fixpoint based on that information.

For potential mutual recursive functions, the fixpoint is included as long as one of the function names occur in the definition as a free variable (by looking at the co-context)

A non-recursive function:
image

A recursive function:
image

Non-recursive functions:
image

Mutual recursive functions:
image


pigumar1:
image
I'm not sure whether Alcotest should fail here: it seems that in this case, there is no reference of f in the body of the function (`BinIntOp (Plus, (IntLit 1), (BoundVar "x"))`). Should we wrap it with a fixpoint?

cyrus-:
@pigumar1 That test should be updated with the new behavior.

cyrus-:
@pigumar1 marking as draft until the test is updated

cyrus- (approved):

Issue #1279, opened 2024-04-23
recursive polymorphic functions not working


cyrus-:
not sure what happened here, it was working when I reviewed the PR. something must have gone wrong with the automatic merge with `dev`?

![image](https://github.com/hazelgrove/hazel/assets/280638/e4d693ee-9de8-4540-a0cc-a10a33afa880)

PR #1296, opened 2024-04-28
Make forall types be considered for recursion.

Crazycolorz5:
Addresses #1279.

image


Crazycolorz5:
Note that the way this is written, I am explicitly disallowing recursively defined polymorphic constants. I think this is in-line with the current definition which disallows recursively defined (non-polymorphic) constants.

cyrus- (approved):

cyrus-:
Thanks!

PR #1263, opened 2024-04-12
Added theorem keyword

nskh:
In-progress PR to add a `theorem` keyword in the style of `test ... end` to Hazel. 

Based on `dev` branch, not [`stepper-rewrites`](https://github.com/hazelgrove/hazel/tree/stepper-rewrites)

Need to 
- rebase on latest `dev`
- test some more

PR #1308, opened 2024-05-07
Adds theorem and proof construct

nskh:
Replaces https://github.com/hazelgrove/hazel/pull/1263.

In-progress PR to add a theorem keyword in the style of let-equals-in to Hazel.

Based on [stepper-rewrites](https://github.com/hazelgrove/hazel/tree/stepper-rewrites) this time.

### DIFF
--- a/src/haz3lcore/dynamics/Elaborator.re
+++ b/src/haz3lcore/dynamics/Elaborator.re
@@ -102,6 +102,7 @@ let cast = (ctx: Ctx.t, mode: Mode.t, self_ty: Typ.t, d: DHExp.t) =>
     | BinFloatOp(_)
     | BinStringOp(_)
     | Test(_) => DHExp.cast(d, self_ty, ana_ty)
+    // | Theorem(_, _, _)
     };
   };
 
@@ -199,6 +200,7 @@ let rec dhexp_of_uexp =
       | Test(test) =>
         let+ dtest = dhexp_of_uexp(m, test);
         DHExp.Test(id, dtest);
+      | Theorem(_, _, _) => Some(DHExp.Tuple([]))
       | Filter(act, cond, body) =>
         let* dcond = dhexp_of_uexp(~in_filter=true, m, cond);
         let+ dbody = dhexp_of_uexp(m, body);
--- a/src/haz3lcore/dynamics/Elaborator.re
+++ b/src/haz3lcore/dynamics/Elaborator.re
@@ -102,7 +102,6 @@ let cast = (ctx: Ctx.t, mode: Mode.t, self_ty: Typ.t, d: DHExp.t) =>
     | BinFloatOp(_)
     | BinStringOp(_)
     | Test(_) => DHExp.cast(d, self_ty, ana_ty)
-    // | Theorem(_, _, _)
     };
   };
 
--- a/src/haz3lcore/dynamics/Elaborator.re
+++ b/src/haz3lcore/dynamics/Elaborator.re
@@ -200,7 +200,7 @@ let rec dhexp_of_uexp =
       | Test(test) =>
         let+ dtest = dhexp_of_uexp(m, test);
         DHExp.Test(id, dtest);
-      | Theorem(_, _, _) => Some(DHExp.Tuple([]))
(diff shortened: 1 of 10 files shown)


src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -199,6 +200,7 @@ let rec dhexp_of_uexp =
       | Test(test) =>
         let+ dtest = dhexp_of_uexp(m, test);
         DHExp.Test(id, dtest);
+      | Theorem(_, _, _) => Some(DHExp.Tuple([]))
```
Negabinary: we probably want this to elaborate to the in-part of the expression, not just a tuple

src/haz3lcore/lang/Precedence.re
```diff
@@ -31,8 +31,8 @@ let rule_arr = 19;
 let rule_pre = 20;
 let rule_sep = 21;
 let case_ = 22;
-
 let min = 23;
+let theorem_ = 24;
```
Negabinary: we can probably re-use the precidence that let uses here - we probably don't want a lower precedence than the min

src/haz3lcore/statics/Statics.re
```diff
@@ -274,6 +274,9 @@ and uexp_to_info_map =
   | Test(e) =>
     let (e, m) = go(~mode=Ana(Bool), e, m);
     add(~self=Just(Prod([])), ~co_ctx=e.co_ctx, m);
+  | Theorem(_, def, _) =>
+    let (def, m) = go(~mode=Syn, def, m);
+    add(~self=Just(Prod([])), ~co_ctx=def.co_ctx, m);
```
Negabinary: this is going to be more like the Seq case now that we've got an `in` - you can pretty much copy seq exactly

Negabinary (changes requested):

src/haz3lcore/lang/Precedence.re
nskh: cool, just use `let_` directly, or do I set `theorem_ = let_`?

src/haz3lcore/dynamics/Elaborator.re
nskh: makes sense, I wasn't sure what exactly to put here so let's discuss further

src/haz3lcore/statics/Statics.re
nskh: cool, and the `e1, e2` are instead `def, body` from the theorem? will update

Negabinary: yeah because we're discarding def like how we're discarding e1

src/haz3lcore/lang/Precedence.re
Negabinary: I think you can use `let_` directly

src/haz3lcore/dynamics/Elaborator.re
nskh: resolved in latest commit

nskh:
@Negabinary: I addressed your changes, should be ready for review. We can merge as-is, or discuss how to add in some way to invoke a stepper by adding a `proof` keyword here too.

Negabinary (approved):

cyrus-:
```
theorem name = typ 
proof "serialized stepper data" 
in e

things we need in the typ syntax:

forall x -> ty
exists x : T -> typ.  [sigma types]
(x : typ) -> typ. [pi types]
   (syntactic sugar: forall x : T -> ty)
typ(typ). 
e1 = e2.  [equality types]
```

cyrus-:
If we do the syntax as:

```
theorem name : typ 
proof e1
in e2
```

then this becomes just another let expression exactly -- that should be pretty straightforward to do so let's update this PR in that direction, then make another PR for adding the new type forms.

cyrus-:
TODO:

- change sorting to `theorem pat = exp in exp`, which matches let expressions exactly
- have theorem behave exactly like a let expression everywhere else in the codebase
- at the type level, rename existing `forall` to `type` 
- add new `forall pat -> ty` construct to types
- add `{e1 = e2}` form to types

PR #1315, opened 2024-06-17
Update all modules about 'Undefined'

Mars-Cat2023:
Based on the previous two contributors’ (@agrsh and @linsyking) code, **I update them on the newest version of Hazel-haz3lcore code but with following few comments**:

(1) [Old Problem] Inconsistency of the undefined case in “UExp, UPat, UTPat, UTyp, URul” in TermBase.re and Term.re.
 **[My Change] If I understand correctly, “Undefined” should only be accepted as UExp, not the other four modules, so I don’t adopt the “Undefined in UPat module”. [Frankly speaking, I don’t why they put “Undefined in UPat module”.]**

(2) **For the documentation of “undefined”**, I am not sure about what should I add to ExplainThis.re, ExplainThisForm.re and TerminalExp.re in Haz3lweb/explainthis. What should I do? 
(Starting with ExplainThis.re Line 545)

(3) Could you give me some suggestions about what should I do for the following cases? [Sorry😭, I forgot the hints you gave last Friday, could you write it down here again?]**
**About Undefined behavior:
<1> Non-determinate cases: Fibonacci (-1)
    How to detect this behavior before evaluation in complex branching and conditional programs?
<2> Unexhaustive cases: match xs with … | | | not exhaustive**

I am very unfamiliar with Hazel syntax, so I just tried my best to write one good Hazel example here:
image

**Any suggestion? Feel free to list below!!!**


src/haz3lcore/dynamics/PatternMatch.re
Negabinary, line 1: Is there a reason these are DoesNotMatch instead of IndetMatch?

What's the desired behaviour of:

```
case undefined
  | 5 => 3
  | _ => 4 
end
```
?

src/haz3lweb/explainthis/data/TerminalExp.re
```diff
@@ -10,6 +10,15 @@ let triv_exp: form = {
 };
 let triv_exps: group = {id: TrivExp, forms: [triv_exp]};
 
+let undefined_exp: form = {
+  id: UndefinedExp,
+  syntactic_form: [exp("Undefined")],
+  expandable_id: None,
+  explanation: "Undefined.",
```
Negabinary: I wonder if you could explain this a little more, and maybe add an example of an incomplete case statement, or factorial only defined on positives or something?

Negabinary (changes requested):
Looks good! I've left a couple of small comments about pattern matching and documentation.

src/haz3lcore/dynamics/PatternMatch.re
Mars-Cat2023: image
Currently, it goes to the _ case. I am not quite sure about this, does it look better evaluated as "undefined"?

Negabinary: I guess I would expect it to behave the same way holes do and not evaluate at all. 

image

Mars-Cat2023: Resolved.

src/haz3lweb/explainthis/data/TerminalExp.re
Mars-Cat2023: Updated the documentation with two new examples.

cyrus- (approved):

PR #1318, opened 2024-07-01
Fix case checker on recursive types

### LINKED ISSUE #1310: case checker on recursive types
f79e50bd98dfecb56cd90018451f3df2

I expect this example to have no error warnings.

pigumar1:
Fixed by adding a specific case analysis for recursive sum types.
Before:
image

After:
image

Not sure why we are able to get away with this case analysis before though. It seems this bug was introduced after #1217 or #1274 was merged.


pigumar1:
Why would this happen?
image

cyrus-:
@pigumar1 not sure why that's happening, those files should be under `_build`?

pigumar1:
> @pigumar1 not sure why that's happening, those files should be under `_build`?

@cyrus- I reran the tasks and it worked XD

PR #1320, opened 2024-07-08
OCaml 5.2 Upgrade

disconcision:
- Upgrade OCaml from 5.0.0 to 5.2.0
- Eliminate unused dependencies
- Unpin all previous pinned versions
- Pin omd (markdown library); their latest version completely changes their interface, and the way that we use it means we have no statics assurances when updating it, so i'm going to let it lie

Upgrading dependencies has triggered two relevant changes:
1. The interface for how attributes are passed to VDom Nodes has changed
2. Json deriving for base types now requires an include (similar to Sexp deriving)

You can see plenty of examples of either change in the diff


disconcision:
Did the various README updates specified here: https://github.com/hazelgrove/hazel/blob/projectors-upgrade/docs/Updating-OCaml-Version.md

cyrus-:
@disconcision this is based on your projectors branch not `dev` - how do you want to handle merging this?

disconcision:
> @disconcision this is based on your projectors branch not `dev` - how do you want to handle merging this? 

@cyrus- do you just mean the branch name? should be against dev, I just had projectors on the brain

src/haz3lcore/tiles/Shard.re
cyrus-, line 1: delete this

cyrus-:
Closing in favor of #1333 to follow the instructions in the repo and have the expected branch name available for future cherry picking.

PR #1327, opened 2024-07-16
add hintedtest keyword

### LINKED ISSUE #694: Add description to Test form
The idea is we want to be able to attach a description to a test + display that as a hint in the exercises mode UI. 

Currently, hints are tracked separately in the exercise data structure. Instead, we can add a Hazel language feature to support this.

Until we add optional arguments to Hazel, we can define a separate keyword:

Current keyword:
```
test e end
```

Hinted version:

```
hinted_test "Hint" for e end
```

jacobziff:
In-progress Pull Request to add a hintedtest keyword to Hazel.

Current keyword:
```
test e end
```

Hinted version:
```
hint "Hint" test e end
```

TODO:

- [X] add new keyword with the same behavior as the current "test" keyword
- [X] add extra string argument to the keyword
- [X] adjust "explain this" window to reflect new keyword
- [X] interact with exercise mode to update hints with the keyword

optional for later: switch check/x icon to appear next to "test" rather than "hint"


cyrus-:
@jacobziff looks like there are merge conflicts to resolve here. 

in addition, it doesn't seem to be handling hint display correctly for the tests where there are existing hints: 
image

we should remove the existing hint field and rely on the hinted tests functionality

jacobziff:
I had some issues with src/haz3lweb/view/dec/Diag.re (the version from dev branch wouldn't compile for me - "Error: let is a reserved keyword, it cannot be used as an identifier. Try `let_` or `_let` instead"), so I had to undo the changes from the dev branch when doing merge conflicts. Other than that, I kept as much of the dev branch as possible.

cyrus-:
merged in as part of ai assistant PR

PR #1337, opened 2024-07-25
Random generation of UExps

ruiz-m:
Random generation of UExp for property-based testing.

The [QCheck](https://github.com/c-cube/qcheck?tab=readme-ov-file) library was used to perform random generation of values.

The [Alcotest](https://github.com/mirage/alcotest) library was used for unit testing.

The purely random generation of uexps will reveal some bugs.


cyrus-:
@ruiz-m could you leave instructions in the README.md file for how to run the quickcheck tests (`make test` I suppose) and how to get readable output from that? @Negabinary took a look but had trouble seeing the terms when tests failed.

ruiz-m:
Alright. I might need to work on adding some code to print the term

test/haz3ltest.re
```diff
@@ -4,6 +4,11 @@ let (suite, _) =
   run_and_report(
     ~and_exit=false,
     "Dynamics",
-    [("Elaboration", Test_Elaboration.elaboration_tests)],
+    [
+      ("Elaboration", Test_Elaboration.elaboration_tests),
+      ("Type Assignment", Test_TypeAssignment.type_assignment_tests),
+      ("Random tests", Test_TypeAssignment.random_tests),
```
7h3kk1d, line 10: We should still mark these as Type Assignment tests. What do you think about `TypeAssignment.random` or `TypeAssignment.property_tests`  for the name. We still don't have a format for doing hierarchies of tests. Dot separated matching the module structure makes sense to me.

test/Test_TypeAssignment.re
```diff
@@ -0,0 +1,603 @@
...
+        Alcotest.check(
+          Alcotest.pass,
+          "Random expression",
+          unit,
+          TypeAssignment.property_test(ty, d, m)
+            ? unit : Alcotest.fail("Term: " ++ UExp.show(u)),
+        );
+      test_case("Type assignment", `Quick, test);
+    },
+    QCheck.Gen.generate(~n=10, pure_random_uexp),
```
7h3kk1d, line 602: I think we probably want each property to be it's own test as opposed to each generated value resulting in a new test. We could then mark the tests as `Slow` instead of `Quick` so we don't slow the test runs down. It's worth looking into https://github.com/c-cube/qcheck?tab=readme-ov-file#integration-within-alcotest to see if we can go straight from the qcheck runner to alcotest.

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -319,11 +319,18 @@ let rec typ_of_dhexp =
...
 
-  switch (dhexp_typ) {
-  | None => false
-  | Some(dh_typ) => Typ.eq(dh_typ, uexp_typ)
+  switch (uexp_typ, dhexp_typ) {
+  | (None, None) => true
+  | (Some(uexp_typ), Some(dhexp_typ)) => Typ.eq(dhexp_typ, uexp_typ)
+  | _ => false
   };
 };
```
7h3kk1d, line 336: We should move this to the test file.

7h3kk1d: Up to you but the property can also be written as:


```reason
let property_test =
    (uexp_typ: option(Typ.t), dhexp: option(DHExp.t), m: Statics.Map.t)
    : bool => {
  let dhexp_typ = Option.bind(dhexp, typ_of_dhexp(Builtins.ctx_init, m))

  Option.equal(Typ.eq, uexp_typ, dhexp_typ);
};
```

test/Test_TypeAssignment.re
7h3kk1d: It's also worth doing something on the assertion to be closer to the equality we want to test so the error messages are better. We can probably do better but this is something:

```reason
let testable_typ = testable(Fmt.using(Typ.pretty_print, Fmt.string), Typ.eq);

//Generate purely random uexp test cases for alcotest
let random_tests =
  List.map(
    (u: UExp.t) => {
      print_endline("Term: " ++ UExp.show(u));
      let m: Statics.Map.t = Test_Elaboration.mk_map(u);
      let d: option(DHExp.t) = Elaborator.dhexp_of_uexp(m, u, false);
      let ty: option(Typ.t) = Elaborator.fixed_exp_typ(m, u);
      let dhexp_typ =
        Option.bind(d, TypeAssignment.typ_of_dhexp(Builtins.ctx_init, m));
      let test = () =>
        Alcotest.check(
          Alcotest.option(testable_typ),
          "Random expression: " ++ UExp.show(u),
          dhexp_typ,
          ty,
        );
      test_case("Type assignment", `Quick, test);
    },
    QCheck.Gen.generate(~n=10, pure_random_uexp),
  );
```

7h3kk1d: Qcheck has first class support for generating long/short tests https://github.com/c-cube/qcheck?tab=readme-ov-file#long-tests

test/Test_TypeAssignment.re
```diff
@@ -0,0 +1,603 @@
...
+        frequency([
+          (1, self(n / 2)),
+          (1, map(upat_list, list_size(int_range(1, 2), self(n / 2)))),
+        ])
+      }
+    ),
+  );
+
+//Generates a UExp with no regard to the typing rules
+let pure_random_uexp: QCheck.Gen.t(UExp.t) =
```
7h3kk1d, line 511: We should try to build a full arbitrary instance for the types so we can add printers/shrinking functions in the future like https://github.com/c-cube/qcheck?tab=readme-ov-file#mirrors-and-trees

7h3kk1d: I feel like we should be able to use the ppx https://github.com/c-cube/qcheck/tree/main/src/ppx_deriving_qcheck for at least the simple variants.

7h3kk1d:
We need to make sure that we're outputting the random seed for any failing tests before merging so that we can reproduce test failures locally. It would be terrible if the tests fail on CI and we can never reproduce them. I would assume using the native integration between qcheck and alcotest would do this but we need to confirm.

ruiz-m:
@7h3kk1d Thank you for the feedback. It might take me some time to work on these fixes, but I will do them

7h3kk1d:
> @7h3kk1d Thank you for the feedback. It might take me some time to work on these fixes, but I will do them

Feel free to let me know if there's any particular way I can help.

test/Test_TypeAssignment.re
7h3kk1d: @ruiz-m just an fyi that I hit this bug when doing some of my own testing https://github.com/c-cube/qcheck/issues/269#issuecomment-2347051999. So be wary if you're trying to use deriving for recursive types.

PR #1338, opened 2024-07-25
Exercise description editor

facundoy:
Make the description of the excercises editable.


facundoy:
Decided it would be a better idea to base my branch from dev rather than Russell's title-editor. I'll close this PR and create a new one with the new branch.

PR #1341, opened 2024-07-25
Use setup-ocaml github action

7h3kk1d:
Switches to https://github.com/ocaml/setup-ocaml for the ocaml installation step. Adds a `hazel.opam.locked` file to keep track of exact dependency versions that we depend on. Also adds a `dune-project` file to manage project dependencies and it generates the `hazel.opam` file on `dune build`. So we can keep track of intentional vs unintentional dependency locking which should be useful for upgrading. Also intentionally caching the opam switch separately from the ocaml compiler itself which seems more deterministic as far as caches go.

The last cached build took about 3m https://github.com/hazelgrove/hazel/actions/runs/10115441982/job/27976209701


cyrus- (changes requested):
- add instructions for updating dependencies
- update instructions for updating ocaml version if necessary
- add ocaml lsp server so make deps is sufficient to get someone started

dune-project
```diff
@@ -1,2 +1,43 @@
...
+  (omd (>= 2.0.0~alpha4))
+  ezjs_idb
+  virtual_dom
+  ppx_deriving
+  ptmap
+  uuidm
+  unionFind
+  ocamlformat
+  (junit_alcotest :with-test)
+  ocaml-lsp-server)) ; After upgrading to opam 2.2 use with-dev https://opam.ocaml.org/blog/opam-2-2-0/
```
7h3kk1d, line 42: We'll separate dev dependencies when we go to opam 2.2.0

cyrus-: Is there a reason not to go to opam 2.2 now?

7h3kk1d: I just didn't want to mix this with any upgrades in case something breaks so we can individually revert

cyrus-: Okay, let's go ahead and do that in a separate PR after this one then.

7h3kk1d: Working on that PR now. Taking a second look we all install opam system-wide from the OS so we probably want to wait until 2.2 is widely distributed by distributions https://repology.org/project/opam/versions. Ubuntu/Nix are still on 2.1. Homebrew is on 2.2.

7h3kk1d: The dune project also needs to require higher versions if we want the lock file to explicitly require 2.2 https://github.com/ocaml/dune/blob/d5809e9de4480a904ca341cbe062880648d8fa5e/src/dune_rules/opam_create.ml#L242

cyrus-: @7h3kk1d can you make a tracking issue for this

hazel.opam.locked
```diff
@@ -0,0 +1,198 @@
...
+  "ezjs_idb" {= "0.1.1"}
+  "ezjs_min" {= "0.3.0"}
+  "fiber" {= "3.7.0"}
+  "fieldslib" {= "v0.16.0"}
+  "fix" {= "20230505"}
+  "fmt" {= "0.9.0" & with-test}
+  "fpath" {= "0.7.3"}
+  "gen" {= "1.1"}
+  "gen_js_api" {= "1.1.2"}
+  "host-arch-x86_64" {= "1"}
```
cyrus-: this is causing problems on my mac -- I think it works fine to remove this and update the `make change-deps` target with the sed command I had in there before to remove it

cyrus- (changes requested):

hazel.opam.locked
7h3kk1d: I'll add the sed command and add another issue for tracking https://github.com/ocaml/opam/issues/6042

@cyrus- do you want me to see if we can have CI run for osxarm64 as well so I don't accidentally do this again?

cyrus-: @7h3kk1d seems a bit wasteful to do that on every build since we don't actually have any native code being generated in Hazel, maybe just for pushes to `dev` so we can quickly correct if it happens again?

dune-project
7h3kk1d: https://github.com/hazelgrove/hazel/issues/1365

hazel.opam.locked
7h3kk1d: @cyrus- I have updated the makefile [here](https://github.com/hazelgrove/hazel/pull/1341/commits/72d8ae302bdeb2ecbb19f5e08cecd4b18daed200#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R18). Unfortunately I had to modify your sed command because of [portability issues](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux) between osx sed and linux sed. I haven't tested it on OS X yet since I need to find a mac. If you could run make change-deps that would be useful.

7h3kk1d: https://github.com/hazelgrove/hazel/issues/1366

src/haz3lweb/view/dec/Diag.re
```diff
@@ -18,24 +18,22 @@ let tr_bl =
       (),
     ) =>
   SvgUtil.Path.(
-    {
```
7h3kk1d, line 21: This is because reason is updated from 3.11 -> 3.12. I think that's a good thing.

Makefile
```diff
@@ -8,11 +8,14 @@ all: dev
...
 	opam update
-	opam switch import opam.export
+	opam install ./hazel.opam.locked --deps-only --with-test --with-doc
 
 change-deps:
-	opam switch export opam.export
-	sed -i '' '/host-/d' opam.export # remove host- lines which are arch-specific
+	opam update
+	dune build hazel.opam
+	opam install ./hazel.opam --deps-only --with-test --with-doc
```
7h3kk1d, line 16: @cyrus- the `--with-doc` should fix the warning we saw on your computer

cyrus- (approved):

cyrus-:
Merge conflict

PR #1344, opened 2024-07-31
Try integrating brr as a online repl

7h3kk1d:
https://erratique.ch/software/brr

Add software extension from ~/opam/hazel/share/brr/console/ after installing packages

More instructions here https://erratique.ch/software/brr/doc/ocaml_console

Screenshot From 2025-07-03 10-48-15


7h3kk1d:
We need a way to only link this in dev

PR #1355, opened 2024-08-05
Exercise Description Editor

facundoy:
Make the description/prompt of the exercises editable.


facundoy:
Found a bug where the pencil icon does not disappear even though we are not in instructor mode (TODO)

PR #1356, opened 2024-08-06
Fix source maps

7h3kk1d:
image

Sets up source maps in dev so you can get actual line numbers/debugging in stack traces


src/haz3lcore/dune
```diff
@@ -4,8 +4,7 @@
  (name haz3lcore)
  (libraries util re sexplib unionFind uuidm yojson core)
  (js_of_ocaml
-  (flags
-   (:include js-of-ocaml-flags-%{profile})))
+  (flags :standard --debuginfo --noinline --dynlink --linkall --sourcemap))
```
7h3kk1d: Try going back to the profiles

7h3kk1d: The rules didn't work so I set it on the env

cyrus- (changes requested):
Seems to work for me! Haven't robustly tested but breakpoints and exceptions seem to correctly work, which is huge.

There is a section on Source Maps in `README.md` in the root directory that needs to be updated. 

Can you also include these flags in the `dune` files in the other top-level libraries (`pretty`, `haz3lschool`)?

disconcision (approved):
works for me!

cyrus- (approved):

PR #1368, opened 2024-08-10
Code Coverage

7h3kk1d:
Code coverage screenshot of Elaborator module

- [x] Add documentation
- [x] Add `make coverage` to make file 
- [ ] Integrate with CI somehow. (Maybe https://about.codecov.io/)


.gitignore
```diff
@@ -56,3 +56,63 @@ setup.log
 
 # Backup of opam lock file
 hazel.opam.locked.old
+_coverage/coverage.css
```
7h3kk1d: Why would vscode do this

7h3kk1d:
Marking this ready for review as it's done but can't go in before https://github.com/hazelgrove/hazel/pull/1341

cyrus- (commented):
check to make sure this isn't instrumenting when not running the test build target, otherwise looks gtg

7h3kk1d:
> check to make sure this isn't instrumenting when not running the test build target, otherwise looks gtg

@cyrus- Yeah, based on my reading it only gets enabled when instrumentation is enabled in the dune command which I'm only doing in make coverage

cyrus- (approved):

PR #1376, opened 2024-08-19
Theorem keyword

thomasporter522:
Adds:
- `theorem ... proof ... in ...` expression (synonym for let)
- `forall ... -> ...` type (binds expressions, not types)
- `{ ...  = ... }` type (equality between expressions)

Renames old `forall` to `type`.

The two new type constructs behave well syntactically and statically, but have no introduction or elimination forms. They are meant to connect to the stepper.


cyrus-:
@thomasporter522 needs to be merged with `dev`

thomasporter522:
There is currently a tylr-related bug. Typing `{}` and moving the cursor inside freezes the program.

Negabinary:
Closed in favor of #1308

Issue #1104, opened 2023-09-06
fix Ezjs_idb build warning issue (once upstream has been fixed)


cyrus-:
introduced with the merge of #1103 

upstream issue was filed here: https://github.com/OCamlPro/ezjs_idb/issues/1

we've silenced the warning for now, so if this ever gets fixed we can remove the code that silences it

PR #1378, opened 2024-08-20
Mark warnings as errors in release and warnings in dev

7h3kk1d:
I just set it in all the dune files for now. I'm waiting on a response to https://discuss.ocaml.org/t/default-flags-for-the-project/3217/3 to see if there's a way to configure it in dune-project/dune-workspace or if we should have a top level dune file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,8 @@ setup-instructor:
 setup-student: 
 	cp src/haz3lweb/ExerciseSettings_student.re src/haz3lweb/ExerciseSettings.re
 
-dev-helper: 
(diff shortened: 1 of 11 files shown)


7h3kk1d:
Release failing due to https://github.com/hazelgrove/hazel/issues/1104

7h3kk1d:
I got a response to use dune-workspace https://discuss.ocaml.org/t/default-flags-for-the-project/3217/3

Doing that now.

cyrus- (approved):

PR #1379, opened 2024-08-21
Run dune fmt before dev/ test without failure and add make watch-test

7h3kk1d:
- dune fmt returns a non-zero exit code if autoformatting promotion happens
- So this will run the dune fmt auto-promote and ignore the exit code then run the standard build in dev so you don't have to run make twice
- Release is left unchanged since we want the build to fail on invalid formatting
- `make watch` already seems to work
- Added `make watch-test`

### TLDR
Running `make`/`make test` will reformat then continue on with build/test


green726 (approved):
Tested this with the elaboration test branch as well and this worked. Seems good to me. Thank you!

PR #1387, opened 2024-09-06
Shaped Projectors

### LINKED ISSUE #1396: projectors break application/subtraction remolding
Screenshot 2024-09-19 at 5 58 19 PM
Screenshot 2024-09-19 at 5 58 06 PM

if you project the left unidelmited child of an application/subtraction form, additional edits within the segment (e.g. whitespace insertion) will cause the application to remold to parens, and the subtraction to remold to unary negative

disconcision:
Fixes #1396.

This PR enables projectors to replace arbitrary (full) terms instead of just terms represented by convex tiles. For example, you can fold let and fun expressions, or infix operators and their children. More experimentally, you can fold/typefold differently shaped things, including comments,  case rules, let tiles (leaving their child out).

**Features:**

- [x] Feature: Allow un/projection of arbitrary full terms
- [x] Feature: Allow folding of comments
- [x] Feature: Better logic to decide where the caret ends up after unprojection (this is currently slightly regressed relative to dev for the convex case)
- [ ] Feature: If there's a selection, try to project that selection. If can't, no-op. Only if there is no selection use projector-specified strategy to form a selection based on the indicated piece. This means that some projectors (like fold) might support multiple selection shapes
- [ ] Decide on API for allowing a projector to leave one or both unidelimited children uncovered (for example, only cover type ascription pseudo-operator and the rightwards type child while leaving the leftwards child uncovered). rn this is handled generically by term selection logic, but i guess each projector kind should decide this itself. can't quite decide what the logic should be based on though

**Bugs:**

- [x] Bug: Can't trigger projection on non-leading delimiters of non-convex tiles
- [x] Bug: exception when trying to project convex mono from right side
- [x] Bug: nested infix ops (or nested prefix and infix) sometimes throw aba exns
- [x] Bug: regrouting issue for subtraction, application when first child is projected 
- [x] Bug: cant change existing projection without unprojecting
- [x] Bug: folding case rules causes bug when moving through to the right (Not_found, maybe due to inclusion of trailing linebreak?)
- [ ] Bug: nested infix folds makes inner fold a hole in live value display
- [ ] Bug: Secondary: prevent folding of whitespace
- [ ] bug: Secondary: postfix comments still get prefix shape
- [ ] bug: Secondary:  unfolding from front of comment triggers buffer overflow (since attempted move to orig id)

**Cleanup:**
- [x] Remove param from RemoveIndicated
- [ ] TODOs

Optional:
- [ ] stretch: even better logic for caret placement?
- [ ] edge case, not necessarily undesired: for n-tuples, unfolding moves to same comma as was originally folded on


disconcision:
This has been substantively replaced by #1420, which allows projection of infix/postfix/prefix-rooted terms. That doesn't support true non-convex projections (e.g. of case rules and secondary), but those don't totally work here and need some conceptual work.

PR #1395, opened 2024-09-17
Strudelize Hazel

disconcision:
This wraps some [Strudel](https://strudel.cc/learn/getting-started/) functions in order to create audio in Hazel. Right now, it wraps only the Note function, which takes a string corresponding to a Strudel/TidalCycles cycle. This is an eDSL string which can take a space-separated list of notes (a-g), and a variety of other notations. Everything in the string is played over a second, so the more notes the faster it is.

Should be functional now on the build server; see docs/sounds:

Screenshot 2024-09-16 at 10 55 29 PM

The plan is to wrap a few more functions, defunctionalizing to create a simple ADT corresponding to Strudel code. This could be used as a basis for a future more elaborate eDSL using projectors/livelits.

Right now, audio plays whenever the program evaluates to a Note. It can be stopped via a button, and the playing audio restarts every time a valid Strudel program is returned, which is different than the last valid Strudel program for which playing began


src/util/Strudel.re
```diff
@@ -0,0 +1,67 @@
...
+    let n = note(pattern);
+    let j = jux(n, rev);
+    play(j);
+  };
+
+/* Example usage function */
+let exampleUse: unit => unit =
+  () => {
+    initStrudel();
+    playNote("<c a f e>(3,8)");
```
7h3kk1d: :coffee:

disconcision:
@7h3kk1d did you look into the below style of binding js fns when doing the ninjakeys thing? claude suggested it but i couldn't get it to work properly. interestingly, different errors on dev and release. in dev, it was trying to access the fns on an object called `runtime`, eg erroring on calling `runtime.play`. in release, it had a more specific error about the method not being implemented

```
/* First, we need to declare the external JavaScript functions */
[@bs.val] external initStrudel: unit => unit = "initStrudel";
[@bs.val] external hush: unit => unit = "hush";

/* For the `note` function, we'll need to create a wrapper */
[@bs.val] external noteRaw: string => 'a = "note";

/* For `jux` and `rev`, we'll create bindings */
[@bs.send] external jux: ('a, 'a => 'a) => 'a = "jux";
[@bs.val] external rev: 'a => 'a = "rev";

/* For `play`, we'll create a binding */
[@bs.send] external play: 'a => unit = "play";
```

7h3kk1d:
> @7h3kk1d did you look into the below style of binding js fns when doing the ninjakeys thing? claude suggested it but i couldn't get it to work properly. interestingly, different errors on dev and release. in dev, it was trying to access the fns on an object called `runtime`, eg erroring on calling `runtime.play`. in release, it had a more specific error about the method not being implemented
> 
> ```
> /* First, we need to declare the external JavaScript functions */
> [@bs.val] external initStrudel: unit => unit = "initStrudel";
> [@bs.val] external hush: unit => unit = "hush";
> 
> /* For the `note` function, we'll need to create a wrapper */
> [@bs.val] external noteRaw: string => 'a = "note";
> 
> /* For `jux` and `rev`, we'll create bindings */
> [@bs.send] external jux: ('a, 'a => 'a) => 'a = "jux";
> [@bs.val] external rev: 'a => 'a = "rev";
> 
> /* For `play`, we'll create a binding */
> [@bs.send] external play: 'a => unit = "play";
> ```

I don't think I tried the bucklescript annotation. There was another syntax that I couldn't use because my function was called `open` which is a keyword in ocaml. Let me find it.

src/util/Strudel.re
```diff
@@ -0,0 +1,67 @@
...
+  () => {
+    let addEventListenerFn = Js.Unsafe.js_expr("window.addEventListener");
+    Js.Unsafe.fun_call(
+      addEventListenerFn,
+      [|
+        Js.Unsafe.inject(Js.string("DOMContentLoaded")),
+        Js.Unsafe.inject(Js.wrap_callback(_ => initStrudel())),
+      |],
+    );
+  };
```
7h3kk1d:
```suggestion
let initStrudel: unit => unit =
  () => Js.Unsafe.coerce(Dom_html.window)##initStrudel();

let hush: unit => unit =
  () => {
    Js.Unsafe.coerce(Dom_html.window)##hush();
  };

let note: string => Js.Unsafe.any =
  pattern => {
    Js.Unsafe.coerce(Dom_html.window)##note(Js.string(pattern));
  };

let rev: Js.Unsafe.any => Js.Unsafe.any =
  pattern => {
    Js.Unsafe.coerce(Dom_html.window)##rev(pattern);
  };

let jux: (Js.Unsafe.any, Js.Unsafe.any => Js.Unsafe.any) => Js.Unsafe.any =
  (pattern, f) => {
    Js.Unsafe.coerce(pattern)##jux(f);
  };

let play: Js.Unsafe.any => unit =
  pattern => {
    Js.Unsafe.coerce(pattern)##play();
  };

/* Wrapper function to chain methods */
let playNote: string => unit =
  pattern => {
    let n = note(pattern);
    let j = jux(n, rev);
    play(j);
  };

/* Example usage function */
let exampleUse: unit => unit =
  () => {
    initStrudel();
    playNote("<c a f e>(3,8)");
  };

/* Function to stop the music */
let stopMusic: unit => unit = () => hush();

/* Function to initialize Strudel when the DOM is loaded */
let initOnLoad: unit => Dom_events.listener =
  () => {
    Dom_events.listen(
      Dom_html.window,
      Dom_events.Typ.domContentLoaded,
      (_: Js.t(Dom_html.window), _: Js.t(Dom_html.event)) => {
        initStrudel();
        false; // I don't think this does anything. 
      },
    );
  };
```

7h3kk1d: The `false` being returned on the `Dom_events.listen` looks like it would be used for stop propagation but I looked through the js_of_ocaml implementation and I don't see where this is happening.

7h3kk1d: I take it back https://github.com/ocsigen/js_of_ocaml/blob/ea51f5e6e1c54a13eb917d081c51175814b8b58b/lib/js_of_ocaml/dom_events.ml#L32 the newest version calls full_handler which does prevent_default so probably set that to `true`

cyrus-:
## Licensing plan for Strudel integration

Strudel is **AGPL-3.0-or-later** with no linking exception, so we need to be deliberate about how it's wired up. Summarizing where we landed.

### The constraint

As written, this branch loads `@strudel/web` from unpkg and calls `initStrudel`/`setPattern` through `Js.Unsafe` — Hazel's code calling Strudel's functions and passing its objects inside one JS execution context. That's a combined work, and lazy-loading the script tag doesn't change it: per the [GPL FAQ](https://www.gnu.org/licenses/gpl-faq.en.html), dynamically linked plug-ins that "make function calls to each other and share data structures" form a single program. So the main dev build would have to ship as AGPL-3.0 — which we don't want, since Hazel's permissive license is load-bearing for downstream research use and forks.

### The plan

1. **Frame-side adapter, AGPL-3.0.** A small module wrapping `@strudel/repl` (bump 1.0.3 → 1.3.0) that owns all Strudel API contact: scheduler, pattern eval, sample loading. Lives in this repo under its own `LICENSE`.
2. **Loaded as a separate document from the same build.** `<iframe src="strudel/index.html">`, same origin, same output directory — one `make serve`, no second process, works from a fresh clone with the network unplugged.
3. **Hazel-side client, MIT.** Communicates by `postMessage` only (`{type:'eval', code}` / `{type:'stop'}`), never by reaching into the frame. Enforced with a lint rule banning `contentWindow`/`contentDocument` from Hazel-side code, so nothing couples through it later by accident.

Music examples ship in the main dev build; only the adapter is AGPL.

### Why this keeps Hazel MIT

AGPL §5's aggregate clause exempts "separate and independent works... not combined with [the covered work] such as to form a larger program," and the FSF's test is the intimacy of communication — "pipes, sockets and command-line arguments are communication mechanisms normally used between two separate programs." A separate browser document exchanging serialized messages is the direct analog: we pass pattern code across, never shared objects or function calls. Hazel's MIT files stay MIT, and no contributor re-licensing is needed either way, since MIT already permits sublicensing.

### Obligations

Both cheap, since we publish sources anyway. Preserve Strudel's notices and `LICENSE`, and don't let the bundler strip headers or tree-shake Strudel into a shared chunk with the js_of_ocaml output. And provide Corresponding Source for the copy we serve — §6(d) allows hosting it elsewhere given "clear directions next to the object code," so a link to codeberg.org/uzu/strudel **pinned to the exact version**, plus our build config, is enough.

### Open issue: samples

The bigger practical problem, and not a copyleft one. `prebake()` hardcodes sample banks on `raw.githubusercontent.com`, and the PWA caching that makes strudel.cc work offline lives in the site, not the npm package — so we need our own prebake registering local banks.

A network-free clone means committing audio, and full VCSL/Dirt-Samples is far too large for Hazel's history. Proposal: commit a curated few-MB CC0 subset so the doc examples make sound offline, with `make setup-samples` fetching the full banks on demand. That also settles licensing diligence, since sample licenses are mixed and independent of AGPL (VCSL is CC0, Salamander piano CC-BY, mridangam CC-BY-SA, Dirt-Samples unstated in `dough-samples`) and a small curated set is one we can actually verify.

PR #1400, opened 2024-09-27
Switch to Bonsai

Negabinary:
Switches from Incr_Dom to Bonsai


hazel.opam.locked
```diff
@@ -1,7 +1,7 @@
 
 opam-version: "2.0"
 name: "hazel"
-version: "~dev"
+version: "dev"
```
7h3kk1d, line 4: I'm curious what's happening here. I don't think you did anything to change it.

hazel.opam.locked
7h3kk1d, line 1: I can't tell which of these dependencies have changed because you had them installed in your opam switch and which ones are necessary. I'll probably pull this down and try starting with a clean switch just to test.

src/haz3lweb/Main.re
```diff
@@ -48,7 +31,7 @@ let apply = (model, action, state, ~schedule_action): Model.t => {
   last_edit_action := JsUtil.timestamp();
   switch (
     try({
-      let new_model = Update.apply(model, action, state, ~schedule_action);
+      let new_model = Update.apply(model, action, (), ~schedule_action);
```
7h3kk1d: Should we just remove state from `Update.apply`?

Negabinary: yeah why not

hazel.opam.locked
7h3kk1d: Has to do with minor opam version differences.

7h3kk1d (approved):

cyrus- (approved):

PR #1402, opened 2024-10-01
Fix deferrals

7h3kk1d:
Before
Screenshot from 2024-10-03 09-52-57
After
Screenshot from 2024-10-03 09-53-12

### DIFF
--- a/src/haz3lcore/dynamics/Elaborator.re
+++ b/src/haz3lcore/dynamics/Elaborator.re
@@ -350,6 +350,11 @@ let rec elaborate = (m: Statics.Map.t, uexp: UExp.t): (DHExp.t, Typ.t) => {
           ((arg, _)) => Exp.is_deferral(arg),
           List.combine(args, ty_fargs),
         );
+      // TODO Should this be a product in the singleton case or not
+      // let remaining_arg_ty =
+      //   List.length(remaining_args) == 1
+      //     ? snd(List.hd(remaining_args))
+      //     : Prod(List.map(snd, remaining_args)) |> Typ.temp;
       let remaining_arg_ty = Prod(List.map(snd, remaining_args)) |> Typ.temp;
       DeferredAp(f'', args'')
       |> rewrap
--- a/src/haz3lcore/dynamics/Elaborator.re
+++ b/src/haz3lcore/dynamics/Elaborator.re
@@ -350,12 +350,10 @@ let rec elaborate = (m: Statics.Map.t, uexp: UExp.t): (DHExp.t, Typ.t) => {
           ((arg, _)) => Exp.is_deferral(arg),
           List.combine(args, ty_fargs),
         );
-      // TODO Should this be a product in the singleton case or not
-      // let remaining_arg_ty =
-      //   List.length(remaining_args) == 1
-      //     ? snd(List.hd(remaining_args))
-      //     : Prod(List.map(snd, remaining_args)) |> Typ.temp;
-      let remaining_arg_ty = Prod(List.map(snd, remaining_args)) |> Typ.temp;
+      let remaining_arg_ty =
+        List.length(remaining_args) == 1
+          ? snd(List.hd(remaining_args))
+          : Prod(List.map(snd, remaining_args)) |> Typ.temp;
       DeferredAp(f'', args'')
       |> rewrap
       |> cast_from(Arrow(remaining_arg_ty, tyf2) |> Typ.temp);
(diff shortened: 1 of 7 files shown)


src/haz3lcore/dynamics/Elaborator.re
```diff
@@ -350,7 +356,11 @@ let rec elaborate = (m: Statics.Map.t, uexp: UExp.t): (DHExp.t, Typ.t) => {
           ((arg, _)) => Exp.is_deferral(arg),
           List.combine(args, ty_fargs),
         );
-      let remaining_arg_ty = Prod(List.map(snd, remaining_args)) |> Typ.temp;
+
+      let remaining_arg_ty =
+        List.length(remaining_args) == 1
+          ? snd(List.hd(remaining_args))
+          : Prod(List.map(snd, remaining_args)) |> Typ.temp;
       DeferredAp(f'', args'')
```
7h3kk1d: This seems right to me. I don't think we always want the remaining function to be from a product.

src/haz3lcore/dynamics/Transition.re
```diff
@@ -333,15 +336,18 @@ module Transition = (EV: EV_MODE) => {
         );
       Constructor;
     | Ap(dir, d1, d2) =>
+      // TODO I don't know why this needs to be final
       let. _ = otherwise(env, (d1, (d2, _)) => Ap(dir, d1, d2) |> rewrap)
       and. d1' =
-        req_value(req(state, env), d1 => Ap1(dir, d1, d2) |> wrap_ctx, d1)
+        req_final(req(state, env), d1 => Ap1(dir, d1, d2) |> wrap_ctx, d1)
```
7h3kk1d: I still don't have a strong sense as to why this needs to be req_final but it remains indet otherwise.

src/haz3lcore/dynamics/Transition.re
```diff
@@ -394,16 +399,30 @@ module Transition = (EV: EV_MODE) => {
         }
       /* This case isn't currently used because deferrals are elaborated away */
```
7h3kk1d: This comment seems to be out of date.

src/haz3lcore/dynamics/Transition.re
```diff
@@ -394,16 +399,30 @@ module Transition = (EV: EV_MODE) => {
...
+        let-unbox args =
+          if (n_args == 1) {
+            (
+              Tuple(n_args),
+              Tuple([d2]) |> fresh // TODO Should we not be going to a tuple?
+            );
+          } else {
+            (Tuple(n_args), d2);
+          };
+        print_endline("\targs: " ++ [%derive.show: list(Exp.t)](args));
```
7h3kk1d: I don't know if there's a better way to be handling this unboxing for the singleton case than wrapping it in a tuple.

src/haz3lcore/dynamics/Elaborator.re
7h3kk1d: The code still works without this in my test though. So TODO

src/haz3lcore/dynamics/Transition.re
cyrus-: @Negabinary not sure I understand what req_value was doing here since we still need to handle casts? are casts considered values?

In any case, deferredap should be treated as if it were a lambda, so it should be a value.

Let's chat briefly about this with @7h3kk1d at the beginning of our meeting tomorrow.

src/haz3lcore/dynamics/Transition.re
```diff
@@ -396,14 +397,22 @@ module Transition = (EV: EV_MODE) => {
...
-        let-unbox args = (Tuple(n_args), d2);
+        let-unbox args =
+          if (n_args == 1) {
+            (
+              Tuple(n_args),
+              Tuple([d2]) |> fresh // TODO Should we not be going to a tuple?
+            );
+          } else {
+            (Tuple(n_args), d2);
+          };
```
7h3kk1d, line 414: This one seems to be necessary

src/haz3lcore/dynamics/Transition.re
Negabinary: @cyrus-  Oh I see, yeah I guess if you still want to do the function cast rules when the function is indet then you do need that to be req_final

7h3kk1d: I got all of the cases I can think of working. I'm struggling to find a case that req_final catches that req_value doesn't though. There might be something wrong with `Constructor`

PR #1371, opened 2024-08-13
Try making all of the ids type parameters

7h3kk1d:
This is still just an idea. Not sure if it's a good one.

PR #1409, opened 2024-10-07
Add deriving eq so we can test the editor/zippers

7h3kk1d:
I'm deriving eq for a bunch of types so we can add editor tests. Mostly having them ignore ids. I figure we can use structural equality if we want id checks for now.

I pulled out the parser tests (https://github.com/hazelgrove/hazel/pull/1411) since we don't need this to do them.

### DIFF
--- a/src/haz3lcore/dune
+++ b/src/haz3lcore/dune
@@ -12,7 +12,8 @@
    js_of_ocaml-ppx
    ppx_let
    ppx_sexp_conv
-   ppx_deriving.show)))
+   ppx_deriving.show
+   ppx_deriving.eq)))
 
 (env
  (dev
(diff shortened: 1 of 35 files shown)


7h3kk1d:
Having some issues integrating this into the new recursive types in TermBase.  I don't want to override `eq` on `IdTagged` globally but all of the `_t` types are just type aliases as opposed to new types so there's no way to overwrite the equality for just them. ClosureEnvironments also have `Id`s so it's unclear how those should work with equality.

I'm also adding on `VarBstMap` and other types so it works.

7h3kk1d:
I don't want to add even more work to consuming the `TermBase` types, but maybe we can wrap all of the `_t` types in a single constructor variant so we can customize equality without breaking `IdTagged`. Otherwise we either need to keep `IdTagged` as ignoring ids for equality or finish https://github.com/hazelgrove/hazel/pull/1371

PR #1411, opened 2024-10-13
Make term tests

7h3kk1d:


test/Test_MakeTerm.re
```diff
@@ -0,0 +1,69 @@
+/**
+ * This file contains tests to validate the `MakeTerm` module's ability to convert
+ * zippers into expressions.
+ */
+open Alcotest;
+open Haz3lcore;
+
+let exp_typ = testable(Fmt.using(Exp.show, Fmt.string), Exp.fast_equal);
```
7h3kk1d, line 8: Exp.fast_equal does not differentiate parens :(.

I can readd the eq derivings if we think that makes more sense.

cyrus- (approved):

PR #1412, opened 2024-10-18
Exp to segment

Negabinary:
Adds a function that converts expressions back into segments, also added comments to the precedence code.


disconcision:
@Negabinary what's the status of this? any known issues? i looked over the code briefly and it seems fine... not really thinking about complex cases though. anyway i want this, so i can do a proper review and merge if it's g2g

Negabinary:
I think it should be g2g

cyrus-:
Merged as part of #1297

PR #1416, opened 2024-11-03
Added introduce action

7h3kk1d:
### Introduce: Fill Empty Holes with Canonical Forms for Expressions and patterns

This PR adds Introduce, a feature that fills an empty hole expression under the cursor with a canonical form based on its expected type. This helps scaffold expressions quickly without manually constructing them.

Supported Expressions:

-  Arrow types `T1 -> T2 `:  `fun ? -> ?`
-  Tuple types `(T1, T2, ..., Tn)`:  `(? , ? , ... , ?)`
    -  Note: Labels are not inserted, as they are inferred and it's unclear whether they should be included in the syntax.
- Singleton variants: Replaced with the corresponding constructor
  - Nullary constructor `+A`: `A`
  - Non-nullary constructor `+A(T)`: `A(?)`
- Forall types `forall ? -> ?` → Replaced with typfun ? -> ?

The Introduce command is available in the command palette and is bound to Ctrl/Cmd + I for quick access.

[Screencast From 2025-03-16 13-34-59.webm](https://github.com/user-attachments/assets/795942e6-70ca-4a9e-8ce1-0c822aeeb627)

### Future Work

- ~Support for patterns~ Bassic patterns were added
- Add a Destruct (enumerate cases) action
- Consider adding a recursive introduce to scaffold deep structures


src/haz3lcore/zipper/action/Perform.re
```diff
@@ -30,6 +30,32 @@ let go_z =
...
+          _,
+        }),
+      ) =>
+      let z =
+        Printer.zipper_of_string(
+          ~zipper_init=z,
+          switch (Typ.term_of(Typ.weak_head_normalize(ctx, ana))) {
+          | Arrow(_, _) => "fun -> "
+          | Prod(tys) =>
+            "(" ++ StringUtil.repeat(List.length(tys) - 1, ", ") ++ ") "
```
7h3kk1d: This ends up adding a trailing space but that also seems to happen in the paste case as well.

7h3kk1d: We don't want extra parens if we're inside of a function application but I don't know how much it matters.

cyrus-:
- let's call it Introduce instead of Refine
- could also support single-constructor sum types

cyrus-:
- also add typfun for polymorphic functions

cyrus-:
- don't add extra parens in ap position

7h3kk1d:
Add case patterns

src/haz3lweb/app/input/Shortcut.re
```diff
@@ -251,6 +251,13 @@ let shortcuts = (sys: Util.Key.sys): list(t) =>
       "Run Benchmark",
       Benchmark(Start),
     ),
+    mk_shortcut(
+      ~mdIcon="bolt",
+      ~section="Refactoring",
+      ~hotkey=Keyboard.meta(sys) ++ "+i",
```
7h3kk1d, line 257: This is still doing ctrl on my mac

7h3kk1d: Fixed in https://github.com/hazelgrove/hazel/pull/1416/commits/0eabe893237b64e29a108ac0c5691c277c79c496. Was broken for all shortcuts.

src/haz3lweb/Main.re
```diff
@@ -138,13 +138,13 @@ let start = {
...
   let on_startup = (schedule_action, ()): unit => {
-    NinjaKeys.initialize(Shortcut.options(schedule_action));
-    JsUtil.focus_clipboard_shim();
     Os.is_mac :=
       Dom_html.window##.navigator##.platform##toUpperCase##indexOf(
         Js.string("MAC"),
       )
       >= 0;
+    NinjaKeys.initialize(Shortcut.options(schedule_action));
+    JsUtil.focus_clipboard_shim();
```
7h3kk1d, line 147: @disconcision any concern with focusing clipboard after setting OS?

7h3kk1d:
Add labels to tuples. Add toggle lablels action. Set cursor to first hole generated.

7h3kk1d:
Somehow deleting expression on f( when f has type (Int, Int) -> Int

7h3kk1d:
> Somehow deleting expression on f( when f has type (Int, Int) -> Int

I have resolve this issue by blocking the introduce action when the cursor would select something that isn't a hole. The issue was that if you do a select term on a function application with no right paren the cursor inspector is left on the hole even though the selection isn't a hole:

Screenshot From 2025-03-28 10-55-55

It's possible that we should be selecting an adjacent hole rather than just the current expression under the cursor but I don't want to break expectations with the cursor inspector yet.

> Add labels to tuples, add empty string, empty list

These are done.

src/haz3lcore/zipper/action/Move.re
7h3kk1d, line 1: @dm0n3y @disconcision I would love your thoughts on how I should be handling this. I'm using `Move.primary` in `Introduce` and it doesn't have any dependency on the `Measured` or `TermRanges` coming in from `Move.S`. So I just pulled all of those functions out of the `Make` functor. Alternatively some of these can be pushed into the `Zipper` module but overall it's annoying to have the required dependencies when the functions don't require them.

cyrus- (approved):

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -397,7 +397,8 @@ and parenthesize_pat =
   | Float(_)
   | String(_)
   | EmptyHole
-  | Constructor(_) => pat
+  | Constructor(_)
+  | Tuple([]) => pat
```
7h3kk1d, line 401: This has unit patterns rendered as () as opposed to (())

cyrus- (approved):

PR #1418, opened 2024-11-18
Cursor inspector now uses typ equality as opposed to full structural equality

### LINKED ISSUE #1520: Type consistency broken for forall types
Image

Sidenote: I'm not sure why the error is on the Pat as opposed to the expression

7h3kk1d:
Old version
Screenshot from 2024-11-18 14-22-40
New version
Screenshot from 2024-11-18 14-23-22

### DIFF
--- a/src/haz3lcore/lang/term/TPat.re
+++ b/src/haz3lcore/lang/term/TPat.re
@@ -29,3 +29,5 @@ let show_cls: cls => string =
   | MultiHole => "Broken type alias"
   | EmptyHole => "Empty type alias hole"
   | Var => "Type alias";
+
+let temp: term => t = term => {term, ids: [Id.invalid], copied: false};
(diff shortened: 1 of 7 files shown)


cyrus-:
what happens with alpha equivalent types (e.g. `forall x -> (x, x)` and `forall a -> (a, a)`)?

7h3kk1d:
> what happens with alpha equivalent types (e.g. `forall x -> (x, x)` and `forall a -> (a, a)`)?

@cyrus- According to the docstring it should do alpha equivalence but not normalization. https://github.com/hazelgrove/hazel/blob/dev/src/haz3lcore/statics/TermBase.re#L693

If we don't want it for alpha equivalent types I can introduce a 3rd form of equality that just ignores IDs.

7h3kk1d:
This may be a bug somewhere else as well though

Screenshot from 2024-11-18 14-44-16

cyrus- (approved):

PR #1420, opened 2024-11-20
Live Projectors

### LINKED ISSUE #702: Add a Probe form for debugging
Backend:
- Value/Type-wise: probe(x) = x. behaves like the identity function.
- Side effect: Like a hole, it gathers closures
- Side-effect: Like a tests in, records its argument syntax and values per-closure

Frontend:
dockable view shows list of probe occurrences, portrayed visually as the result of evaluating the indicated expression given the stored closure. hovering peeks into list of closures, clicking opens closure subeditor to re-call expression with different values… stretch: you get a repl into the environment(s) of the probe

### LINKED ISSUE #829: Live inspector

### LINKED ISSUE #879: Shift+Click for selections

### LINKED ISSUE #880: Scroll when dragging a selection down/up

### LINKED ISSUE #890: Can't scroll using mouse while selecting
When mouse is down for making a selection, my mouse scrollwheel doesn't work. Haven't tried trackpad scroll on mac.

### LINKED ISSUE #933: Bind Ctrl/Cmd+Click to Go To Definition action

### LINKED ISSUE #1147: Select All does not work at position zero

### LINKED ISSUE #1300: Structural copy/paste
If code is copied from the editor and pasted in the editor, maybe we should retain the original zipper structure rather than reparsing from text. This would be both faster, and it could retain any metadata not expressed in the text source, like projections.

(May implement this as part of #1218)

Considerations / Challenges:
- Cut/paste could reuse same ids; copy/paste would require replacing them.
- Making sure this doesn't get out of sync with system clipboard

### LINKED ISSUE #1351: Replace mousedown overlay with pointer capture
Simplifies our DOM structure and should actually make selection dragging work while your mouse pointer is outside the browser window, a current frustration.

https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

### LINKED ISSUE #1359: Projectors are not preserved under copy/paste
See discussion on #1218

### LINKED ISSUE #1396: projectors break application/subtraction remolding

### LINKED ISSUE #1421: Clean up CSS filters for drop shadows
It looks like we can implement drop shadows using just css filters instead of needing to add the svg shadow in four different places, might be worth cleaning up this code at some point. 

image

### LINKED ISSUE #1436: Selected projectors highlight in unselected editors
image

### LINKED ISSUE #1448: If arrow key is pressed when there's a selection, it breaks the selection on the wrong side
e.g. if i make a selection with the caret on the right, pressing either left/right arrow results in movement as if the previous caret position was on the left of the selection. this is a somewhat recent regression

### LINKED ISSUE #1471: make grout deco render inline with text instead of on separate layer
This is kind of a special case in the rendering system right now, ultimately easier I think to switch

### LINKED ISSUE #1489: If a text projector is directly adjacent to a token on the left, moving left from that token skips entering the text area
This is because it's using what piece is indicated to decide whether to try and give the projector focus, but since a projector is convex, if the piece directly to the left is concave, it will be indicated instead. Need to check actual sibling instead.

Image

### LINKED ISSUE #1528: padding on captioned slides has vanished
Image

### LINKED ISSUE #1561: Horizontal scroll-to-caret also scrolls vertically
ie pressing right arrow below will cause a vertical jump as well as a horizontal one.

Image

disconcision:
Closes #702, closes #829.
Projector fixes: closes #1359, closes #1396, closes #1436, closes #1489
Hazel selection fixes: closes #879, closes #880, closes #890, closes #1147, closes #1351, closes #1448.
General Hazel fixes: closes #933, closes #1300, closes #1421, closes #1471, closes #1561.

For an introduction, see the `Probes` slide here: https://hazel.org/build/projectors-live/.

# Feature: Probe Projectors

- Use the projector panel or alt/option-v to apply a probe projector to any expression or pattern
- When a term is probed, live values will show in `cells` to the right of the line
- Cells in function bodies are grouped by function call
- If the dynamics are still calculating, or if the expression is never executed, the probe icon will spin
- If closures are found for that expression, the number of closures are indicated in the probe superscript
- By default, live cells based on all gathered closures are shown; double-clicking on any cell restricts that probe to show only that one cell.
- Hovering over a cell of the currently indicated probe which is not a pattern of variable reference shows the (co-contextual) environment (i.e. the values of variables in the expression)
- Clicking on a live cell sets a global (across-probe) `closure cursor` powered by that closure's call stack. Other cells (across all probes) will get decorated based on their 'relative positions' within the call graph.
- To expand/contract a cell, hold shift and drag left/right to progressively un/summarize the value. This also un/summarizes values in the corresponding environment view.
- Double-clicking on a probe on a function application 'pins' that application; for applications with more than one execution (call), the currently selected call will be pinned
- If there are more than 30 ish cells, the others will be cut off; this is indicated by ellipses to the right. When such a probe is indicated, small arrows will appear to the left of its cells to allow moving the window of cells shown.
- Probes have limited keyboard controls. When a cell is clicked, the group of cells gets keyboard focus. Left/right arrows move the cursor, left/right+shift resizes the cursor cell, and space toggle between show-one and show-all cells.

## Probe caveats:
- Adding two probes on the same line will currently result in their cells overlapping; you can manually add a linebreak to work around this

# Feature: Card Projectors

- A new livelit-class projector for playing cards. Single cards (pairs of rank/suit ADTs) can be projected, as well as lists of cards. Single cards can be flipped by clicking on them (a view change no semantic consequences). Shift-clicking a card lets you change it to a different card. See the `Cards` slide for more details.

### Hazel Features:
- FLIP animation system: Caret movement is now animated, and a general framework has been added to make it easier to add position/scale animations to interactions. See `Animation.re` for details, or read more about the general approach here: https://aerotwist.com/blog/flip-your-animations/. Animations which use `Animation.re` can be disabled globally via a toggle in the settings menu.
- Probe system. Instrumenting a `term` by wrapping it in as `Wrap(term, Probe(options))` will make the evaluator collect information about the term's evaluations. This currently includes a value, a (filtered) environment, and a call stack. The options on Probe are intended to be added during statics, and can refine what is collected. Currently, they are used to filter the collected environment to the values occurring in the expression's co-context.  
- Clipboard cache: When you copy or cut a selection corresponding to a complete segment (a selection that contains no shards with corresponding backpack items), the underlying segment is statefully retained. If text is pasted which is identical to the text serialization of that stored segment, the segment is used instead. This speeds up pasting, and also retains projectors.
- Selection improvements: Click-drag now persists when the mouse cursor is outside the browser window. Various other selection fixes and features (see closed issues list)

## Projector System Features:

- The projector API has changed. There are new entries, described below, and some of the method arguments have been changed to provide new information to projectors.
- Projectors can get dynamic info in their infopack. This collection is gated by API flag (dynamics=true), for performance reasons. This information consists of a list of 'closures', each of which has a value, a (co-contextual) environment, and a call stack corresponding to one evaluation of the probed term or pattern.
- Can now project all terms and patterns, not just operand-shaped ones. When un-projecting non-operand-shaped projectors, the previously projected syntax will be left selected, as this now enable back-and-forth toggleability thanks to:
- Can now project based on selection (not just indication). Any selection which corresponds to a full term can be projected. Caveats: Term must not be whitespace-padded on the outside. It also suffices that the term be a term in isolation, even if it's not a subterm of the program, e.g. `1+2` in `1+2*3` can be projected. If this makes people too angry I'll consider changing it.
- Projectors are retained on cut/copy/paste (see `Clipboard cache` below)
- The can_project method and init constant have been merged into a single method (called init). This method must decide, based on a provided term, whether or not the projector applies, and if so, returns the initial model. So the initial model can now be based on the initial syntax.
- Projectors can now specify optional additional views. The offside view renders at the end of the line where the projector is located (the probe projector uses this for value displays). The overlay view renders above all code text, code decoration, and base projector views (the probe projector uses this for closure count and function pin indicators).
- All projectors now have keyboard shortcuts. Alt/option-v for probes, -t for type, and -l for livelit-style projectors (chooses the first that matches the indicated syntax)
- Projector keyboard shortcuts now show when you hover over the projector panel when the relevant projector is active
- Projectors are now provided (through info.utility) with some helper functions for piece/segment/term conversion. These are included to avoid cyclic dependencies, and will hopefully be deprecated by the upcoming Splice Projectors PR.
- Projector view functions are now provided with a view_seg callback to render non-interactive view of syntax. This is currently used for probe projector cell value displays. This will be deprecated by the upcoming Splice Projectors PR.

## Updates to existing projectors:
- Fold projector now show their hidden syntax on-hover
- Type projectors now displays at the end of the line like probes. New keyboard shortcut option/alt-t
- All livelit-style projectors now have combined keyboard shortcut option/alt-l
- All Projector keyboard shortcuts have been added to the command palette

## Hazel Internal changes:
- The code completion buffer is now implemented as a comment inside a selection (instead of a tile). This should be noticeable but should be less brittle wrt code changes as remolding/erasure will now never have to care about its shape.

### Future work (will be creating issues for these post-merge):

- Feature: Projectors: ExplainThis Override, including a Probes control panel for probes. This will show a legend / configuration panel which can be used to select which decoration overlays are active.
- Feature: Probes: Inter-probe system, including keyboard control for moving between different probe lines with arrows, and adding new probes with arrows_modifiers.
- Feature: Resize all probe cells at once with double modifier
- Feature: Make prove panel retain last used projector after unprojection for toggleability. Will need to consider when this gets reset.
- Feature: Highlight function names corresponding to pinned/indicated calls associate caller/callee more closely
- Feature: Grey out unevaluated probes (current deco arch makes this awkward)
- Feature: Adding a new probe on the same line as another removes the first or induces a linebreak or displays on the next line
- Feature: Projectors: Overlay projector option: Don't replace the syntax, but augment it over top. Could be used for probe, test. For probes: Enable editing of probed expressions (requires splices or overlays)
- Feature: Projectors: Projectors should somehow indicate strictly contained static errors. Might want to wait on a more general statics mechanism to associate errors which impact a term with that term. 
- Feature: Improve & Extend value/syntax abbreviation logic and add multiline value display (drag down). Ideally this will use (generalized) fold projectors
- Feature: Refine probe window navigation arrows UI. Integrate arrows with ellipses. Drag ellipses to rescale space available for cells (ideally snapping to cell, not just cut-off)
- Feature: Probes: 'Step' display option showing environment substituted into expression (but not evaluated)


disconcision:
@Negabinary this is only tenuously out of draft but i'd appreciate a look-over when you get the chance, especially regarding the dynamics bits. Currently I'm doing some things a bit hackily and trying to decide how bad they actually are.

There are two dynamics things in particular I'm wondering about:

1. Regarding exercise mode dynamics work. I was able to hook up to it in a basic sense, but I'm confused about dynamics across cells. Currently, probes in the correct implementation reflect values from the tests, but the user implementation probes don't reflect values from the user tests, only for executions in the user implementation editor itself. Changing the ExercisesMode piping so that the user implementation cell gets the user test dynamics also doesn't seem to make this work. (As a larger issue, it's not 100% clear to me what should get what. Also, I think I/we should consider, possibly as part of this PR, changing the way dynamics output works to make it slightly more in-line with statics. in particular, maybe dynamics should output a map, and there is just implictly a probe on the syntax root. i.e, the previous dynamics results is now 'just' an entry in the map. This clarifies my interfaces a bit but maybe fucks with stepper stuff?)
2. Another thing I don't quite understand is that in the current implementation, i'm not actually really accumulating a stack trace... it seems to be more of a lexical equivalent. i.e. I would have expected the current implementation to, if i instrument a variable x inside a function bound to f, defined at the top level, which is called inside a function g, also defined at the top level, and then g is called at the top level, to have entries corresponding to both f and g  for variable x. but it only has the immediate call. this doesn't actually immediately matter but I want to understand why my understanding of the code is wrong.

 P.S. I did some dead code removal in ClosureEnvironment as I was finding the interface was drifting a bit

src/haz3lcore/FontMetrics.re
Negabinary, line 1: FontMetrics!?? in core!!????

I'm sure it hurts you as much as it hurts me to see more and more files being slowly dragged into core. Maybe in the new year we can have a conversation about whether we can recover some semblance of modules (along with maybe renaming the 3 out of em). 

Part of me wonders whether Editor.re/Perform.re/etc really belong in web, and if we do that whether we can get the core distinction back.

src/haz3lcore/statics/TermBase.re
```diff
@@ -89,7 +89,7 @@ and exp_term =
   | Test(exp_t)
   | Filter(stepper_filter_kind_t, exp_t)
   | Closure([@show.opaque] closure_environment_t, exp_t)
-  | Parens(exp_t) // (
+  | Parens(exp_t, Probe.tag)
```
Negabinary: I think I'd vote for making an explicit Probe( ... ) form instead of using Parens. Parens to me feels like there should be no guarantee that they're preserved all the way through execution, and I can easily imagine someone writing a helper function that removes parens and creating unexpected bugs.

src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -41,6 +41,9 @@ module EvaluatorEVMode: {
   let update_test = (state, id, v) =>
     state := EvaluatorState.add_test(state^, id, v);
 
+  let update_probe = (state, id, v) =>
```
Negabinary: This feels like the right way to do it.

src/haz3lcore/Measured.re
Negabinary, line 1: Measurements in the stepper appear to have gone awry
image

(This comment could be in the wrong place.)

src/haz3lcore/dynamics/Transition.re
Negabinary, line 1: Kinda surprises me that you've had to change Fun but not Fix, Let, or Case

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -49,6 +49,9 @@ let should_add_space = (s1, s2) =>
         && !Form.is_keyword(s1)
         && String.starts_with(s2, ~prefix="(") =>
     false
+  | _ when String.ends_with(s1, ~suffix="…") =>
+    /* Hack case for probe projector abbreviations */
```
Negabinary: They're all hack cases 😅

src/haz3lcore/statics/MakeTerm.re
Negabinary, line 1: I guess this hack is a good argument for projectors being able to run their own MakeTerm.

src/haz3lcore/dynamics/Transition.re
disconcision: This comment seems to have dissociated from line number? is this about the code at (new) line 343? if so, that's not the fun case, it's the fun case within app. If not then I'm not sure what this is regarding

src/haz3lcore/statics/TermBase.re
disconcision: renamed Parens to Wrap as per convo

src/haz3lcore/FontMetrics.re
disconcision: FontMetrics no longer in core

disconcision:
@cyrus- all issues discussed yesterday have been addressed except as discussed below. 

Should be fixed:
- Selection bug: Click, then quickly move and click, and then drag
- Selection bug: Doubleclick, then quickly move and doubleclick (same cause as above)
- Selection bug: Selection dragging isn't triggered by wheel events
- Probe bug: After toggling one vs many cells shown, sometimes less cells are shown than expected
- Probe tweak: Make patterns and expression closure cells the same color
- Probe tweak: Make one vs many cells shown toggle global for all probes
- Internal tweak: Separate Parens and Probe constructors in Exp
- Probe data leak: Probe data is now removed from the `test validation` editor. (What was the specific correct implementation exfiltration you found with probes here? i can't immediately recreate one.)

Won't fix but will file issues:
- Need to juggle the mouse when dragging out-of-bounds: looks like it would require a new stateful subsystem.
- Make selections include linebreaks to improve selection behavior around end of lines (a lot of work and careful testing)

test/Test_Evaluator.re
```diff
@@ -643,7 +643,61 @@ in fn("hello")|},
...
+                          term:
+                            TupLabel(
+                              {term: Label("l"), annotation: ()},
+                              {term: String("a"), annotation: ()},
+                            ),
+                          annotation: (),
+                        },
+                      ]),
+                    ],
+                  ),
```
7h3kk1d, line 683: Yeah, this looks right. This is asserting that the tuple shows up in the probe. Test should be a lot clearer after https://github.com/hazelgrove/hazel/pull/1530/files

disconcision:
@cyrus- merge bugs resolved

disconcision:
@cyrus- the below are fixed:

- [x] use triple equals for probe+env divider; add space padding to env divider
- [x] playing card caret not visible when card is flipped
- [x] probe cells with floats shrink when maximum size is hit
- [x] unexpected extra parens around constructors
- [x] dont show functions in env
- [x] rm scratch mode behind-cell background color 
- [x] dblclick on probe doesn't collapse
- [x] purple mode --no callstack tracking
- [x] display projector keyboard shortcuts more prominently
Screenshot 2025-03-13 at 9 57 05 PM
Screenshot 2025-03-13 at 9 56 50 PM

I'm leaving the resize-all-cells and on-the-same-line mitigation to future work.

disconcision:
@Negabinary @7h3kk1d either of you interested in doing a functional review? or reading any more code? or just approving this. probably going to merge this in next several days

7h3kk1d:
> @Negabinary @7h3kk1d either of you interested in doing a functional review? or reading any more code? or just approving this. probably going to merge this in next several days

@disconcision I'll do a quick code review. I've played around with it functionally a bit. If there's anything in particular to try lmk.

7h3kk1d:
The projection toggle seems to be stuck selected for me. Tested under chrome and firefox:

[Screencast From 2025-03-14 09-32-45.webm](https://github.com/user-attachments/assets/48cce15b-aee0-4a96-a5bb-b8a16c510afd)


Related: toggling off probe switches dropdown to fold. So if you toggle twice you don't get the same state.

7h3kk1d:
Double-clicking on a probe cell seems to restrict all the probes to show only a single value. I would have expected it to show only the values under that computation

Screenshot From 2025-03-14 10-05-42

In the above example I focused into `34`. I still expected the probes in fib to show all the closures used in the calculation of fib(8). This may just be too complicated for now.

src/haz3lweb/Keyboard.re
```diff
@@ -97,10 +97,25 @@ let handle_key_event = (k: Key.t): option(Action.t) => {
...
+  | {key: D("t"), sys: PC, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
+    Some(Project(SetIndicated(Specific(Info))))
+  | {key: D("†"), sys: Mac, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
+    /* † is what holding option turns t into on Mac */
+    Some(Project(SetIndicated(Specific(Info))))
+  | {key: D("l"), sys: PC, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
+    Some(Project(SetIndicated(ChooseLivelit)))
+  | {key: D("¬"), sys: Mac, shift: Up, meta: Up, ctrl: Up, alt: Down} =>
+    /* † is what holding option turns t into on Mac */
+    Some(Project(SetIndicated(ChooseLivelit)))
```
7h3kk1d, line 118: Would you be open to putting some of these in https://github.com/hazelgrove/hazel/blob/dev/src/haz3lweb/app/input/Shortcut.re so that way we can trigger them from the command palette as well?

src/haz3lcore/Animation.re
```diff
@@ -0,0 +1,232 @@
...
+    width: float,
+  };
+
+  let box_of = (elem: Js.t(Dom_html.element)): box => {
+    let container_rect = elem##getBoundingClientRect;
+    {
+      top: container_rect##.top,
+      left: container_rect##.left,
+      height: Js.Optdef.get(container_rect##.height, _ => (-1.0)),
+      width: Js.Optdef.get(container_rect##.width, _ => (-1.0)),
```
7h3kk1d: What's the deal with the defaults being -1.0 for these and 0.0 for the ones below?

src/haz3lcore/Animation.re
```diff
@@ -0,0 +1,232 @@
...
+    };
+};
+
+open Js;
+
+/* If an element is new, report its new metrics.
+ * Otherwise, report both new & old metrics */
+type change =
+  | New(box)
+  //| Removed(box)
```
7h3kk1d: Do we want to delete this constructor?

src/haz3lcore/ClipboardCache.re
```diff
@@ -0,0 +1,33 @@
...
+let set = (seg: option(Segment.t), str: string): unit =>
+  switch (seg) {
+  | Some(seg) when Segment.deep_tile_complete(seg) =>
+    /* This check makes sure we won't create backpack orphans */
+    cache := Some((str, seg))
+  | _ => ()
+  };
+
+let intersection = (ids1: list(Id.t), ids2: list(Id.t)): list(Id.t) =>
+  List.filter(id => List.mem(id, ids2), ids1);
```
7h3kk1d: I don't think this is used and it can probably be in ListUtil

src/haz3lcore/Measured.re
```diff
@@ -275,14 +266,18 @@ let is_indented_map = (seg: Segment.t) => {
...
 };
 
-let last_of_token = (token: string, origin: Point.t): Point.t =>
-  /* Supports multi-line tokens e.g. projector placeholders */
-  Point.{
-    col: origin.col + StringUtil.max_line_width(token),
-    row: origin.row + StringUtil.num_linebreaks(token),
-  };
+/* Tab projectors add linebreaks after the end of their line */
+let deferred_linebreaks: ref(list(int)) = ref([]);
```
7h3kk1d: I haven't read this in depth but it seems like only the max of deferred_linebreaks is used. Do we need to keep a list or can we just track the max.

src/haz3lcore/lang/term/TermBase.re
```diff
@@ -280,6 +293,10 @@ and Exp: {
     | (Parens(x), _) => fast_equal(x, e2)
     | (_, DynamicErrorHole(x, _))
     | (_, Parens(x)) => fast_equal(e1, x)
+    /* Below is kind of a hack to make EvalResult.calculate go after adding a projector.
+     * We should clarify syntactic/semantic equality here */
+    | (Probe(x1, _), Probe(x2, _)) => fast_equal(x1, x2)
+    | (Probe(_, _), _) => false
```
7h3kk1d: I just added https://github.com/hazelgrove/hazel/issues/1563. Feel free to add comments on any probe specifics.

src/haz3lcore/lang/term/TermBase.re
```diff
@@ -891,121 +913,107 @@ and ClosureEnvironment: {
...
+        env_to_extend: t,
+      )
+      : t => {
+    {
+      id: Id.mk(),
+      env: Environment.union(new_bindings, map_of(env_to_extend)),
+      call_stack: Option.to_list(ap_id) @ call_stack,
+    };
+  };
+  //TODO(andrew): cleanup below
```
7h3kk1d: Is the below code still useful?

src/haz3lcore/projectors/implementations/CheckboxProj.re
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+  let toggle = (info): Base.segment =>
+    switch (
+      info.utility.lift_syntax(
+        fun
+        | Exp({term: Bool(b), _} as t) => Exp({...t, term: Bool(!b)})
+        | _ => failwith("Checkbox: Toggle: not boolean literal"),
+        info.syntax,
+      )
```
7h3kk1d, line 38: I know we have some functions that operate on the segment structure as it were an expression now but I wonder if this is a sign that the checkbox projector should have a boolean model and just set the segment to whatever the model is. Then toggle the model.

src/haz3lcore/pretty/Abbreviate.re
```diff
@@ -0,0 +1,841 @@
+/* Abbreviate a term for display, specifically for the live
+ * value probe projector. This is currently specialized for
+ * expressions which are (at least partially) values.
+ *
+ * This is an ugly rough approach, and should be rewritten when
+ * we have projectors (in particular, fold) within value displays.
+ *
+ * This approach ends up duplicating way too much info
+ * with ExpToSeg. This should probably be rewritten to
+ * use that somehow. */
```
7h3kk1d, line 10: An approach on the segment structure with abbreviations would probably work. If we want to keep this returning the `Exp` structure and use `Invalid` for abbreviations I'm not sure how it would work.

src/haz3lcore/projectors/implementations/CheckboxProj.re
disconcision: The projectors PR matt and I are working on now will hopefully eliminate the distinction between a projector model and the underlying syntax in favor of a model with a make_term method. Until then I think this is necessarily awkward. But it's not obvious to me what adding a dual source of truth would add at the moment

disconcision:
@7h3kk1d:
re: "Double-clicking on a probe cell": double clicking on a probe cell toggle one versus many values displayed. double-clicking on the probe itself (in the syntax), if that syntax is a function application, toggles restriction to values under that call in the callstack.  (this could be generalized from aps to arbitrary expressions in principle). well, specifically it restricts to values that are under the particular dynamic instance of the application whose cell is currently indicated. this is a bit subtle but the subtlety can be ignored for top-level applications which i think will be the majority initial use case. i have ideas how to refine this UI but haven't come up with a concept which doesn't involve menus, which is too much to bite off for this PR i think. but open to simple ideas to make this more straightforward.

src/haz3lcore/pretty/Abbreviate.re
disconcision: yeah i'm not happy at all with this module; i think doing this properly is a job in itself (in principle i want this to be more flexible and not just use a 'left-to-right greedy' strategy for abbreviation; i want to be able to do things like allocate half the size budget to each side of a pair for example). my tentative goal is to redo this using projectors at some point, using a more-flexible version of the fold projector to do what I'm doing with Invalids right now. but currently this would cross abstraction boundaries in an awkward way

src/haz3lcore/projectors/implementations/CheckboxProj.re
7h3kk1d: Yeah, we can probably wait for a more general fix. I still think it's less of a second source of truth and just making the model the source of truth in this example.

src/haz3lcore/pretty/Abbreviate.re
7h3kk1d: That makes sense. If there's an intermediate state that operating on segments would be useful I could look at it but it's not clear if that helps get towards the projector folding version.

7h3kk1d:
> @7h3kk1d: re: "Double-clicking on a probe cell": double clicking on a probe cell toggle one versus many values displayed. double-clicking on the probe itself (in the syntax), if that syntax is a function application, toggles restriction to values under that call in the callstack. (this could be generalized from aps to arbitrary expressions in principle). well, specifically it restricts to values that are under the particular dynamic instance of the application whose cell is currently indicated. this is a bit subtle but the subtlety can be ignored for top-level applications which i think will be the majority initial use case. i have ideas how to refine this UI but haven't come up with a concept which doesn't involve menus, which is too much to bite off for this PR i think. but open to simple ideas to make this more straightforward.

I don't have any better suggestions for the short term. It feels like pinning is a more useful operation than the single view.

src/haz3lcore/projectors/implementations/CheckboxProj.re
disconcision: but regardless of what we put in the model there's still going to be a (syntactic) boolean in the underlying syntax

7h3kk1d: Agreed. It just seems simpler to me to not have any error handling outside of the model initialization instead of on every transformation.

src/haz3lweb/Keyboard.re
disconcision: done

src/haz3lcore/projectors/ProjectorInfo.re
```diff
@@ -0,0 +1,50 @@
+/* Projector data which is dependent on semantics,
+ * separated out for dependency reasons */
+
+/* Gather utility functions/values to be sspaed to the projector.
```
gcrois, line 4: Is there a reason this needs to be associated with a specific projector rather than as a general utility? + is sspaed a typo?

disconcision: it is intended as a generic utility; it's the same for all projectors. the only reason it is included here at all is to avoid cyclic dependency issues that arise if you try to just call them directly in a projector implementation.

sspaed == passed lmao

7h3kk1d: Remember to sspae your functions and values.

PR #1425, opened 2024-12-02
Fix multiplication/division precedence

7h3kk1d:
I added a new encoding so we don't have to manually track precedence numbers. I didn't realize precedence was being stored on the tiles in the persisted programs. I'm not sure how we want to handle rolling this out with that or if we want to go back to the older number encoding so we can babysit them more. 

I'm also open to other ideas on how to avoid breaking it for everyone.

Before
Screenshot from 2024-12-02 16-47-25
After
Screenshot from 2024-12-02 16-47-31

### DIFF
--- a/src/haz3lcore/lang/Precedence.re
+++ b/src/haz3lcore/lang/Precedence.re
@@ -6,38 +6,101 @@ open Util;
 [@deriving (show({with_path: false}), sexp, yojson)]
 type t = int;
 
+[@deriving show({with_path: false})]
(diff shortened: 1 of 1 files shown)


cyrus- (changes requested):
regenerate Init.ml

cyrus-:
@7h3kk1d there are instructions in `INSTRUCTORS.md` on how to update the exercises after a breaking change to the internal representation

PR #1426, opened 2024-12-04
Remove req_value from evaluator

Negabinary:
I noticed we don't really use req_value any more in the evaluator, because we mostly rely on unboxing which gives us finer-grained control.

e.g. `[?]` is not a value, but that doesn't mean we can't do list operations on it, so evaluation shouldn't be waiting for sub-expressions to be values, they should only check whether they can unbox the desired kind of value.

ValueChecker is staying because we still use it in $v stepper filter patterns.


Negabinary:
closed in favor of #1433

PR #1434, opened 2024-12-05
Trampoline Evaluator

Negabinary:
This pr prevents stack overflows by using a system of "trampolines" in the evaluator. (Basically, we keep track of Hazel's stack ourselves, instead of making OCaml do it.)


cyrus-:
included in #1455

PR #1435, opened 2024-12-05
Closure dynamics

### LINKED ISSUE #1431: Closures around indet expressions disappear
image

```
let f = 1 in
if  then f else f
```

This should evaluate to `Closure(f → 1, if ? then f else f)`, and then we can't get rid of the closure at the end because there's still an `f`.

Problem exists in dev, but we only spotted it in #1297 because #1297 highlights the `f`s in the result red.

image

Negabinary:
Fixes #1431 

We create some new evaluation rules around closures:

- [ ] We only remove closures if the expression inside has an empty co-context.
- [ ] Since closures can remain in final expressions, we augment unboxing rules and cast calculus rules to deal with these

While we're at it we also

 - [ ] remove the function closure argument in favor of just having functions wrapped in a closure.

### DIFF
--- a/src/haz3lcore/dynamics/Builtins.re
+++ b/src/haz3lcore/dynamics/Builtins.re
@@ -1,3 +1,4 @@
+open Util;
 open DHExp;
 
 /*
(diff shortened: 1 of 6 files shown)


Negabinary:
Closed in favour of #1453

PR #1447, opened 2025-01-05
Hazatro

disconcision:
This depends on live-projectors.

Changes:

1. This adds a new livelit-style projector applying to an ADT representing playing cards. This projector can be applied to literals of the ADT, lists of those literals, patterns (which can include wildcards for either the suit or the rank) of that ADT, and lists of such patterns. The ADT doesn't need to be defined; the projector considers only the syntax.
2. The editor caret now has position transition animations on editor actions.

Under the hood:

1. New projector shape: Tab. This differs then blocks in that any trailing linebreaks are deferred until the end of the line.
2. Animation.re: A flexible mini-framework for element transition animations using the FLIP technique and the Web Animations API. FLIP animations are gated under the settings.core.flip_animation option, available in the nut menu.


cyrus- (changes requested):
Can you include a documentation page with card examples on it? Currently completely undiscoverable.

disconcision:
@cyrus- I can do it after live projectors gets merged, stacked init.ml diffs are untenable

PR #1450, opened 2025-01-07
Fix case statements in exptosegment

7h3kk1d:
Fixes the ExpToSegment functionality on match statements when inline is true. Currently this just skips all rules.

Sample output when printed
```
case x | A => 1| B => 2 end
```


Negabinary (approved):
lgtm

7h3kk1d:
Fixed in https://github.com/hazelgrove/hazel/pull/1220

PR #1393, opened 2024-09-16
Track unbound variables in expressions

Negabinary:
Fixes #1389

PR #1453, opened 2025-01-08
Fix Generalized Closures

### LINKED ISSUE #1389: can't use unbound variables - evaluator and stepper time out
image

### LINKED ISSUE #1431: Closures around indet expressions disappear

### LINKED ISSUE #1432: Elaboration of functions doesn't preserve types
```
let f : Int -> Int = fun x -> x in
f
```

becomes

```
fun x -> x
```

which has type `? -> ?`

(Exists on dev, but spotted when inspecting types on #1297)

Negabinary:
Replaces #1435 and #1393, with a simpler approach that more closely matches @jlam55555's generalized closures.

Fixes #1431 and also fixes #1389.

Also fixes #1432

### DIFF
--- a/src/haz3lcore/dynamics/DHExp.re
+++ b/src/haz3lcore/dynamics/DHExp.re
@@ -106,7 +106,7 @@ let rec strip_casts =
 let assign_name_if_none = (t, name) => {
   let (term, rewrap) = unwrap(t);
   switch (term) {
-  | Fun(arg, ty, body, None) => Fun(arg, ty, body, name) |> rewrap
+  | Fun(arg, body, None) => Fun(arg, body, name) |> rewrap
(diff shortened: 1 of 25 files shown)


cyrus- (changes requested):
Can we make it so the type annotation on functions post-elaboration is displayed (when show functions is on)?

cyrus-:
@Negabinary there is a conflict in `ExpToSegment`

cyrus- (approved):

PR #1465, opened 2025-01-17
Livelits

gcrois:
This merge brings a implementation of Livelits, proposed by https://hazel.org/papers/livelits-pldi2021.pdf, with some built-in examples.

Examples:
1. Emotion
2. ~~Timestamp~~ (Can't keep up)
3. Slider
4. JS

Issues:
- [x] Look up livelit in the context 
- [x] Show the livelits in scope in the context inspector
- [x] Livelit names are showing errors
- [x] When you press space after a livelit name, it should initialize the model argument
   - [x] and turn on the livelit projector immediately
- [x] Only displaying correct view in smiley face projector every other update (may also affect other livelits?)
- [x] ExplainThis sidebar should show information about livelit applications when relevant
- [x] Create a documentation slide showing how livelits work
- [x] Separated browser tests and URL encoding (currently they are already merged in here)
- [x] Backspace should break projection, remove default model
- [x] Fix livelit ap term -- consider modifying MakeTerm (similar to deferred Ap)
- [x] Unbound livelit error state
- [x] Documentation slide
- [x] Statics off error formatting
- [x] Remove parenthesis requirement for invocation
- [x] add Livelit.md file to docs to explain to someone what they need to do to add a new livelit

Limitations of projectors, to be fixed in separate PR:
- Livelit invocation must be directly surrounded by a parenthesis

Future PRs:
- User-defined livelits (pending modules)
- Splices (pending splices in projectors)
- Parameters (pending splices)
- Pattern livelits
- LivelitAp: Switch mode of Ap to LivelitAp when applied to a LivelitName

1. ~~Cannot define Livelits during runtime, Livelits don't have access to context -- this needs to be revealed to projectors~~
2. Cannot implement splices, requires context, merge with Live Projectors ? @disconcision 
3. ~~Livelit invocation must be directly surrounded by a parenthesis in order to be grouped in InfoExp; example: `(^emotion(100))` -- this is a current limitation of projectors~~


cyrus-:
One issue right now is that type errors on livelit projections aren't displayed, because the type error is located on the livelit ap underneath the parentheses. @disconcision any thoughts on what we should be doing here? I think in general I'd support shifting the indicators for type errors at the root of a parentheses to the parentheses (while still showing the error message when your cursor is on the parenthesized term as well). Maybe that general change would address this problem too?

disconcision:
@cyrus- i've been thinking about this one, or more generally the issue of what to do when there is a 'deep' type error that will be hidden by a projector. i think it's worth considering a more general solution, where if there is an error under a projector, meaning an error strictly inside the underlying syntax which is not inside a splice, then we place a new kind of error/warning on the projector itself. but I'm not sure about how this should be surfaced in the cursor inspector. also i think this might be worth considering in tandem with a broadened notion of error provenance/attribution, which provides at least some kind of indication that an expression contains / is downstream of some kind of error or incompleteness. this doesn't have to be as aggressive as a red outline; perhaps a tag in the cursor inspector.

i've also thought about the parentheses thing as an interim solution. as you point out there are subtleties (also wanting the error on the contained term), meaning it's not as simple as just changing parentheses statics. so yeah certainly possible to do something here but i think it's worth whiteboarding some options or at least dedicating a lunch to the short and mid-term future of error attribution

cyrus-:
yeah let's discuss this (+ rewriting) at lunch on Wednesday

gcrois:
Two things came up when trying to finish this off:

`hazel/src/haz3lmenhir/Conversion.re` -> What should we do here w.r.t. of_core (line 511)? @7h3kk1d Maybe you have insight?

You can backspace and space to unproject / reproject the livelit, but if you change the state, you can't do navigate using arrow keys, or insert characters until you click somewhere -- I suspect this is a problem with how I remove projectors in `hazel/src/haz3lmenhir/Conversion.re` line 277, maybe I am supposed to change the left and right somehow?

gcrois:
It looks like Projectors update is going to require some non-trivial reworking -- @disconcision if you have time, would be great to get some quick clarifications on things!

gcrois:
We're officially compatible with the latest changes in `dev`!

A few things to consider:
1. We no longer have a `LivelitAp` term -- this was cannibalized in a prior merge
2. The focusability of the livelit projector cannot depend upon the livelit -- this is a current limitation of Projectors
3. I fixed the `js` livelit, but haven't given it a clean interface. It is in `Livelit.re` currently as an example to `Livelit` developers, but is not in the doc slide.
      - It's not obvious to me how to use the Projector `update` function to rerender on model change
5. ~~Would love some advice from @7h3kk1d on what sort of tests would be helpful here!~~ Done :)

gcrois:
Known bug: cannot apply when caret is to the left of the LivelitName if full term is not selected, also crashes if more than Name and Ap are selected -- @disconcision maybe we need to reject projection if layout is not what is expected?

gcrois:
Cyrus: projector API should give a Term and a function to update the term; model default should be an exp; in `Insert.re`, do `exp -> segment`. `projector` member rename to -> `view`; `expansion_f` -> `expand`. Remove dynamic `explainThis`.

cyrus-:
Proposal for API when approaching user-defined livelits

```
type RawLivelit = {
  name: string,
  model_t: TermBase.Typ.t,
  model_default: string,
  expansion_t: TermBase.Typ.t,
  expansion_f: TermBase.Exp.t => TermBase.Exp.t,
  action_t: TermBase.Typ.t,
  update : (TermBase.Exp.t (* action *), TermBase.Exp.t (* old model *)) -> TermBase.Exp.t (* new model *)
  projector:
    (list(model_piece), Base.piece => Ui_effect.t(), Id.t) =>
    node_or_list,
  size: ProjectorCore.Shape.t,
  explain_this: list(string),
};

module type BuiltinLivelit = {
  name: string,
  type model_t
  let model_t : TermBase.Typ.t
  let model_to_hazel : model_t -> TermBase.Exp.t
  let model_from_hazel : TermBase.Exp.t -> model_t

  let expand : model_t -> TermBase.Exp.T

  type action_t
  let action_t : TermBase.Typ.t
  let action_to_hazel : action_t -> TermBase.Exp.t
  let action_from_hazel : TermBase.Exp.t -> action_t

  let update : (action_t, model_t) -> model_t

  let view : model_t -> node
  let size : ProjectCore.Shape.t
}

module CreateBuiltin(B : BuiltinLivelit): RawLivelit 
```

gcrois:
My understanding is that a projector's local actions currently can only modify a projector's internal model, making it impossible for a livelit to change its syntax on an update (outside of the view function) -- @disconcision is there a way to circumvent this so that the update function can change syntax? This allows a livelit to take its update functions from the context.

This is `LivelitProj`'s update function:
```
let update = (_, info, action: LivelitCtx.action_exp) => {
    print_endline("LivelitProj.update " ++ Exp.show(action));

    let ctx =
      switch (info.statics) {
      | Some(InfoExp(exp)) => exp.ctx
      | _ => []
      };

    switch (get(info)) {
    | Some((llname, model)) =>
      let ll = Ctx.lookup_livelit(ctx, llname);
      switch (ll) {
      | Some(ll) =>
        let new_model = ll.update(action, model);
        let seg = info.utility.term_to_seg(Exp(new_model));
        print_endline("new segment update: " ++ Segment.show(seg));
        ();
      | None =>
        print_endline("Warning - LivelitProj.update: not found in context")
      };
    | None => print_endline("Warning - LivelitProj.update: get is empty")
    };
    ();
  };
```

seg is what the model segment should become, ideally

===

Update: I circumvented this by creating a callback the forwards the parent action handler, rather than the local handler -- it occurs to me that the Built-in Livelit API is strikingly similar to the Projectors API. It would be a good use of our time deciding what should be a projector vs. Built-In livelit :)

gcrois:
https://github.com/hazelgrove/hazel/tree/livelit-lift-syntax

I attempted to replace [my model replacement function](https://vscode.dev/github/hazelgrove/hazel/blob/livelit/src/haz3lcore/projectors/implementations/LivelitProj.re#L55) with the existing lift-syntax, and ran into some errors when replacing the model -- namely, a nebulous error called "Not_found"... I had these in earlier versions of my replace model function, eventually finding stability when I replaced the labels of the existing syntax without modifying the shape or ids of the syntax.

Example: 
[Livelit Branch](https://hazel.org/build/livelit/?share=%255Eslider(73)) vs. [Livelit with lift_syntax](https://hazel.org/build/livelit-lift-syntax/?share=%255Eslider(73))
When the slider livelit is projected and interacted with, model changes as expected in the livelit branch, but a fatal error occurs with lift_syntax when unprojected and selected.

Screenshot 2025-04-11 at 4 43 10 PM

@disconcision Any ideas here?

gcrois:
We now do a full replacement of the syntax (including the livelit name). What cause the previous crash in [livelit-lift-syntax](https://github.com/hazelgrove/hazel/tree/livelit-lift-syntax) is still unknown, but is resolved for now. Thanks @disconcision!

gcrois:
~Create Self.re common error type for bad livelit model, needs exp_t~

For built-in livelits, we need a function to determine if an exp is a value, lives in Statics (static value checker)

gcrois:
@cyrus- Made the changes we talked about earlier today, and a couple more for stability + error reporting

cyrus- (approved):

PR #1468, opened 2025-01-20
Final tutorial sys

reevafaisal:


facundoy:
Known bug: evaluation of the YourImplementation editor is incorrect. 
I think that there may some issues with the stitching but not 100% sure. 
Tutorial edits not being saved + unable to highlight multiple lines on editor

NOTES:
- Need to specify tutorial module names when stitching in order to set wrapper to true or false since we will only need the wrapper for specific exercises

reevafaisal:
So in my last commit when I said local, I meant global.

BUG: 
Also I noticed there is a bug in the boolean exercise. Appears that whenever you type, based on the movements of the cursor the table appears to dissapear and reappear. Since we just added the new exercises in, we need to make sure it works with the widget.

reevafaisal:
Based on the print statements, it seems like there is a constant recursion over label in the normalize function. Im not sure where exactly the bug lies but it seems like it is not able to detect that the expression is incomplete.

cyrus-:
- [x]  previous button on all slides
- [x]  get rid of checkmark animation, make it smaller and think about styling of the three buttons at the bottom (same size maybe?)
- [x]  fix crash on clicking previous button or next button when that goes out of bounds
- [x]  newlines in slide text
- [x]  make the hint an option rather than required
- [x]  hover message over the info button

cyrus-:
**Instructions for creating new tutorials**

1. Create an exercise of type Tutorial.spec inside haz3l/app/explainthis/exercises/examples
2. Ensure to set the wrapper to either true or false based on the construction of your hidden_tests. If you require the user to define the variable, you may set it as false otherwise set it as true.
3. If you would like to add in a hint cell, type your hint into the display_hint variable, otherwise leave it as an empty string and the hint cell will not be displayed.
4. eds.version is the integer version of eds.id. It is used as an index (starting at 1) solely for the next/prev button functionality. Make sure to correctly assign eds.version and edit the total number of test cases in the block that generates the next button in TutorialMode.re.

reevafaisal:
- removed the cross icon entirely due to dissapearing/reappearing issue as a user types their implementation.
- checkmark functionality is fixed
- prev button only allows user to view exercises that user has previously completed

reevafaisal:
- [x] fix breakage on first arrow / remove
- [x] remove looping on last arrow and add a "done"
- [x] fix arrow css
- [x] slight issue with stitching

src/haz3lcore/lang/term/TermBase.re
```diff
@@ -118,7 +118,11 @@ module rec Any: {
         ~f_typ=continue,
         ~f_tpat=continue,
         ~f_rul=continue,
-        ~f_any=continue,
+        // ~f_any=continue,
+        ~f_any=(rec_call, x) => {
+                 print_endline("Breaking recursion at ~f_any");
```
cyrus-, line 123: debug print left here

src/haz3lweb/Export.re
```diff
@@ -40,7 +45,7 @@ let export_all = (~settings, ~instructor_mode, ~log) => {
   mk_all(~core_settings=settings, ~instructor_mode, ~log) |> yojson_of_all;
 };
 
-let import_all = (~import_log: string => unit, data, ~specs) => {
+let import_all = (~import_log: string => unit, data, ~specs, ~tutorial_specs) => {
```
cyrus-, line 48: rename `specs` to `exercise_specs` so it is more clear

src/haz3lweb/Export.re
```diff
@@ -49,6 +54,7 @@ let import_all = (~import_log: string => unit, data, ~specs) => {
         settings: all_f22.settings,
         scratch: all_f22.scratch,
         documentation: "",
+        tutorial: all_f22.tutorial,
```
cyrus-, line 57: let's rename `all_f22` to `all_public` or something like that, since it isn't about F22 semester anymore :)

src/haz3lweb/Store.re
```diff
@@ -18,6 +20,8 @@ let key_to_string =
   | Mode => "MODE"
   | Scratch => "SAVE_SCRATCH"
   | Documentation => "SAVE_DOCUMENTATION"
+  | Tutorial(id) => Haz3lcore.Id.to_string(id)
```
cyrus-, line 23: prefix with "TUTORIAL" so it is clear when debugging what this entry is, not just a random ID

src/haz3lweb/app/editors/Editors.re
```diff
@@ -99,6 +111,15 @@ module Update = {
           m,
         );
       Model.Documentation(scratch);
+    | (Tutorial(action), Tutorial(m)) =>
+      let* exercises =
```
cyrus-, line 115: call it `m'` rather than `exercises`

src/haz3lweb/app/explainthis/ExplainThis.re
```diff
@@ -178,7 +178,9 @@ let mk_translation =
     List.fold_left(
       ((msg, mapping), elem) => {
         switch (elem) {
-        | Omd.Paragraph(_, d) => translate_inline(d, msg, mapping, ~inject)
+        | Omd.Paragraph(_, d) =>
+          let (n, mapping) = translate_inline(d, [], mapping, ~inject);
```
cyrus-, line 182: why did you need to modify `ExplainThis`?

src/haz3lweb/exercises/Tutorial.re
cyrus-, line 1: let's move all the tutorial related files over to `haz3lweb/tutorials` rather than putting them under `exercises`, so we aren't confusing exercises mode and tutorials mode

src/haz3lweb/exercises/settings/TutoialSettings_student.re
cyrus-, line 1: this file name has a typo -- should be `Tutorial` not `Tutoial`

cyrus-, line 1: it doesn't look like the build process has been updated to copy in this file when building in student mode, like we do with the Exercises mode equivalent

src/haz3lweb/exercises/settings/TutorialSettings_base.re
cyrus-, line 1: let's include a .md file in docs with instructions on how to create and update tutorials, like we have for exercises

src/haz3lweb/view/Page.re
```diff
@@ -143,7 +143,12 @@ module Update = {
       model |> return_quiet;
     | FinishImportAll(None) => model |> return_quiet
     | FinishImportAll(Some(data)) =>
-      Export.import_all(~import_log, data, ~specs=ExerciseSettings.exercises);
+      Export.import_all(
+        ~import_log,
+        data,
+        ~specs=ExerciseSettings.exercises,
+        ~tutorial_specs=TutorialSettings.exercises,
```
cyrus-, line 150: let's call Tutorial pages "lessons" to distinguish them from Exercise mode's "exercises"

src/haz3lweb/exercises/Tutorial.re
```diff
@@ -0,0 +1,751 @@
...
+  p.id;
+};
+
+let find_id_opt = (id, specs: list(p('code))) => {
+  specs |> Util.ListUtil.findi_opt(spec => id_of(spec) == id);
+};
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type pos =
+  //   | Prelude
```
cyrus-, line 78: you can delete these

src/haz3lweb/exercises/Tutorial.re
```diff
@@ -0,0 +1,751 @@
...
+
+let map = (p: p('a), f: 'a => 'b, f_hidden: 'a => 'b): p('b) => {
+  {
+    id: p.id,
+    title: p.title,
+    // description: p.description,
+    version: p.version,
+    module_name: p.module_name,
+    prompt: p.prompt,
+    // point_distribution: p.point_distribution,
```
cyrus-, line 100: you can delete the commented out things

src/haz3lweb/exercises/Tutorial.re
```diff
@@ -0,0 +1,751 @@
...
+  };
+};
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type eds = p(Editor.t);
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type state = {eds};
+
+// [@deriving (show({with_path: false}), sexp, yojson)]
```
cyrus-, line 131: delete

src/haz3lweb/exercises/TutorialGrading.re
cyrus-, line 1: delete commented out things

cyrus- (changes requested):
in addition to the code comments:
- with this PR, let's make Tutorial Mode the default mode that opens up when someone loads Hazel (rather than the Documentation slide)
- let's make Tutorials mode (and Exercises mode, while you're at it) open up in Student mode by default rather than Instructor mode
- the tutorial arrows appear above everything else, e.g. the hazelnut menu. they should be at the same layer as the other stuff in the tutorial.

PR #1475, opened 2025-01-23
Fix #1424

### LINKED ISSUE #1424: Multi-argument builtin functions can't handle casts around tuple
```
let x : ( , ) = ("Oh","no") in
string_compare(x)
```

image

Negabinary:


test/Test_Evaluator.re
```diff
@@ -160,6 +160,41 @@ let tet_ap_of_hole_deferral = () =>
     |> Exp.fresh,
   );
 
+let test_multi_arg_builtin_cast = () =>
+  evaluation_test(
+    "string_compare((\"Hello\", \"World\"):(?, ?))",
```
7h3kk1d, line 165: @Negabinary fwiw I've found it a lot easier to write these strings with multiline syntax like

```reason
{|string_compare(("Hello", "World"):(?, ?))|}
```

Negabinary: ooh nice, I'll use that in future

PR #1491, opened 2025-02-01
faster pattern coverage checking

### LINKED ISSUE #607: Pattern matching exhaustiveness and redundancy
See https://github.com/hazelgrove/hazel/pull/585 for Hazel 2 effort.

### LINKED ISSUE #1319: ADT Dynamics example is broken (Exhaustiveness/Redundancy)
It claims that cases are redundant and inexhaustive where they're not.

Also separately I'm not sure about how we're localizing the inexhaustive error for lets... check out the other documentation editors, they are hard to read. Putting it on the pattern makes more sense to me in this case.

### LINKED ISSUE #1423: Redundancy check runs forever when given inconsistent types
To reproduce:

```
case ( , )
  | ([],  ) => ?
  | (x::
```

### LINKED ISSUE #1473: Statics is exponential in the number of case branches on pairs
I had to tragically abandon my card sorting aspirations as this slows to a crawl:

Image

I presume this is an exhaustiveness checking issue... could likely mitigate by putting this behind a flag

cyrus-:
# Problem

Our current implementation of pattern coverage checking (i.e. exhaustiveness and redundancy checking) uses the "incon" formulation from our OOPSLA 2023 paper defining the Peanut calculus (Yuan et al, 2023). This works but is exponential complexity (in the width of the patterns). Andrew noticed this in #1473. 

The current implementation also handles unknowns in scrutinee types in an ad hoc way.

# Solution

This PR reformulates the coverage checker to use Maranget's pattern matrix formulation (see Maranget, 2007 - "Warnings for Pattern Matching"). This is essentially the same formulation used in the OCaml compiler.

To handle holes, we use the guidance from the Peanut paper which is essentially that holes act like wildcards for exhaustiveness checking but as "bottom" for redundancy checking.

To handle unknown types in scrutinees, we treat them as having an unknown constructor set.

# Additional Contributions in this PR

- fixes some bugs with type checking for algebraic datatypes
- adds better testing infrastructure, including:
  - Fresh modules in Typ, Exp, Pat, TPat, and Term (including all of the above) that provide cleaner ways to manually construct terms.
  - Adding indications, written `{{{_}}}`, to the Menhir parser that causes parsing to emit a list of unique IDs for all terms indicated. This is used in `Test_Coverage` to make testing exactly where there are errors clean. This hasn't been backported to `Test_Statics` or other locations yet.
  - minor cleanup when noticed

# TODOs
- [x] basic exhaustiveness checker proof of concept
- [x] add support for numeric, float, and string patterns to exhaustiveness checker
- [x] redundancy checking
- [x] handle void types
- [x] handle unknowns anywhere in scrutinee type
- [x] refactor to use constructor sets
- [x] pattern holes
- [x] hook up let and fun cases
- [x] change file name / location
- [x] resolve merge conflicts
- [x] tests
- [x] code cleanup

# Spin-Off PRs?
- [ ] add %derive.show instructions to README.md for future reference

# Future PRs
This PR does not attempt to provide examples in the case of an inexhaustive match, though the infrastructure is now all there and this should be straightforward.


disconcision:
@cyrus- i found a non-pairs example of significant slowdown in redundancy checking:

Screenshot 2025-02-07 at 12 38 58 PM

here the call to is_inconsistent is taking up most of the (substantial) time spent in statics on this example:

Screenshot 2025-02-07 at 12 38 19 PM

anyway this particular example is less egregious as it is a somewhat unusual program, but thought i'd log it

cyrus-:
Okay I'll check that with the new impl. Should be okay.

disconcision:
@cyrus- may want to look at #1423 as well

src/haz3lcore/dynamics/TypeAssignment.re
```diff
@@ -79,21 +80,23 @@ let dhpat_extend_ctx = (dhpat: DHPat.t, ty: Typ.t, ctx: Ctx.t): option(Ctx.t) =>
       // TODO: make this stricter
       let* ctrs = Typ.get_sum_constructors(ctx, ty);
       let* typ = ConstructorMap.get_entry(name, ctrs);
-      let* (ty1, ty2) = Typ.matched_arrow_strict(ctx, typ);
-      Typ.eq(ty2, ty) ? dhpat_var_entry(dhp, ty1) : None;
+      switch (typ) {
```
disconcision: why does this no longer do a matched arrow? also could be a let*

src/haz3lcore/statics/ConstructorMap.re
```diff
@@ -194,7 +202,8 @@ let map = (f: option('a) => option('b), m: t('a)): t('b) => {
 let get_entry = (ctr, m) =>
   List.find_map(
     fun
-    | Variant(ctr', _, value) when Constructor.equal(ctr, ctr') => value
+    | Variant(ctr', _, value) when Constructor.equal(ctr, ctr') =>
```
disconcision, line 206: we might want to make an issue to turn at least one of these options into a custom type... strictly speaking this is an improvement but a double option is likely to cause confusion

src/haz3lcore/statics/Info.re
```diff
@@ -665,6 +666,20 @@ let is_error = (ci: t): bool => {
   };
 };
 
+let error = (ci: t): option(error) => {
```
disconcision: this function already exists (error_of, in this file)

src/haz3lcore/statics/Info.re
```diff
@@ -723,23 +738,23 @@ let fixed_typ_pat = (ctx, mode: Mode.t, self: Self.pat): Typ.t => {
...
-    )
-    : Constraint.t =>
-  switch (upat.term) {
-  | Cast(_) => constraint_
-  | _ =>
-    switch (fixed_typ_pat(ctx, mode, self) |> Typ.term_of) {
-    | Unknown(_) => Constraint.Hole
-    | _ => constraint_
-    }
-  };
```
disconcision, line 742: are we retaining this for a reason?

src/haz3lcore/statics/Info.re
```diff
@@ -799,8 +814,8 @@ let derived_pat =
     : pat => {
   let cls = Cls.Pat(Pat.cls_of_term(upat.term));
   let status = status_pat(ctx, mode, self);
+  // print_endline("Status: " ++ show_status_pat(status));
```
disconcision: cleanup

src/haz3lcore/statics/Coverage.re
```diff
@@ -0,0 +1,504 @@
...
+    };
+  };
+};
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type redundant_rows = list(int);
+
+module Matrix = {
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type row = {
```
disconcision, line 148: internalizing indices into this type feels non-obvious; not to imply i feel that strongly about it

src/haz3lcore/statics/Coverage.re
```diff
@@ -0,0 +1,504 @@
...
+    | List(elt_ty) =>
+      Finite(
+        Map.of_list([
+          (nil_ctr, []),
+          (cons_ctr, [Typ.Fresh.tprod([elt_ty, ty])]),
+        ]),
+      )
+    | Bool => Finite(Map.of_list([(true_ctr, []), (false_ctr, [])]))
+    | Unknown(_) => Unknown
+    | Int
```
disconcision: looks like this is going reject a case that actually handles all integers as non-exhaustive? literally unusable

src/haz3lcore/statics/Mode.re
```diff
@@ -132,14 +132,16 @@ let ctr_ana_typ = (ctx: Ctx.t, mode: t, ctr: Constructor.t): option(Typ.t) => {
      a sum type having that ctr as a variant, we consider the
      ctr's type to be determined by the sum type */
   switch (mode) {
+  | Ana({term: Arrow(_, ty_out), _} as ty) =>
```
disconcision, line 135: is it intentional that this no longer normalizes?

src/haz3lcore/statics/Self.re
```diff
@@ -119,22 +119,27 @@ let of_exp_var = (ctx: Ctx.t, name: Var.t): exp =>
...
-      name,
-      syn_ty:
-        switch (Ctx.lookup_ctr(ctx, name)) {
-        | None => None
-        | Some({typ, _}) => Some(typ)
-        },
-    })
-  | _ => IsConstructor({name, syn_ty: Some(ty)})
-  };
+let of_ctr =
```
disconcision, line 122: introducing a Mode dependency into Self feels weird, but I guess it makes the breakdown of the Self concept for anonymous sum types explicit. In any case this function has gotten pretty weird and awkward; probably deserves a comment

test/Test_Coverage.re
```diff
@@ -0,0 +1,748 @@
...
+  no_errors(
+    "Loooong List: Exhaustive",
+    {|let x : [Int] = ? in
+case x
+| [] => 1
+| [_] => 2
+| [_, _] => 3
+| [_, _, _] => 4
+| [_, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _] => 5
+| [_, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _] => 6
```
disconcision, line 187: shades of BigHand

disconcision (approved):
Functionally seems solid. Performant on the card comparison that brought the checker to its knees before. New checker logic feels reasonable. Some minor style stuff.

src/haz3lcore/dynamics/TypeAssignment.re
cyrus-: I changed `ConstructorMap.get_entry` to return an `option(option(Typ.t))` rather than an arrow type for constructors due to some things elsewhere

src/haz3lcore/statics/ConstructorMap.re
cyrus-: left a comment for someone looking at this code in the future

src/haz3lcore/statics/Coverage.re
cyrus-: we retain the row index from the original matrix in submatrices for redundancy checking, so its not just redundant information. left a clarifying comment.

src/haz3lcore/statics/Coverage.re
cyrus-: ha yes, I'll leave a comment noting this serious correctness issue

src/haz3lcore/statics/Mode.re
cyrus-: it didn't normalize before either. The situations I would expect to work even when the type annotation is not syntactically an arrow are still working for me. I'd like to eventually distinguish normalized and non-normalized types, it's quite complex to suss out what the invariants are.

cyrus-: it didn't normalize before either. The situations I would expect to work even when the type annotation is not syntactically an arrow are still working for me. I'd like to eventually distinguish normalized and non-normalized types, it's quite complex to suss out what the invariants are.

test/Test_Coverage.re
cyrus-: it was indeed based on bighand :)

PR #1494, opened 2025-02-02
bug fix for filter for Cast & TypFun

tonyfettes:
This PR fix the behavior of filter pattern matching inside Cast/TypFun construct.

For the TypFun specifically, we can test for the alpha-equivalence of two typfun.

Screenshot From 2025-02-02 17-50-13


Negabinary (approved):
lgtm

PR #1499, opened 2025-02-06
Make type join symmetric

### LINKED ISSUE #1370: ListLit of Sum Type with Hole report Inconsistency in Static Checker
Expect both lists pass static checker. We get 1 failed and the other passed.

Tested on `dev` [f1274b6](https://github.com/hazelgrove/hazel/commit/f1274b6c74b9da98a27222a03738024eecf0a4a5)

image

Plain Code:
```typescript
let x : +X(Bool) = in
let y : +X( ) = in
[y,x];
[x,y];
```

Negabinary:
Fixes #1370 by trying @disconcision 's suggestion:

> I worry there might be two reasons, my recollection is also fuzzy. In any case I think any reasons for the asymmetry are specific to the use site for joining the self type and expected type to determine the error status. unless things have drifted this should be a single call site where the asymmetry is load bearing. it's also possible that Matthews redo of elaboration has removed the need, which I (low confidence) recall as manifesting due to casting bugs. probably easiest just to make a branch with the asymmetry removed and see what happens

I've removed the asymmetry (removing ~fix entirely) and I can't find any problems it's caused. There are no cast failures on casts documentation and all tests still pass.


Negabinary:
@disconcision any more ideas for things to try and see if they go wrong?

disconcision:
@Negabinary have you checked the casting doc slide for console garbage?

Negabinary:
image

@disconcision  yeah nothing in console

disconcision (approved):
I played around a bit and was unable to break it.

cyrus- (approved):

PR #1502, opened 2025-02-07
Fix #1501

### LINKED ISSUE #1501: Unnecessary statics calculation on caret movement
This is a regression.

caret down action with statics disabled:

Image

caret down action with statics enabled but completion disabled:

Image

caret down action with statics and completion enabled:

Image

Statics should not be recalculated on caret movement (unless there is a new demand for this?) and caret movement should not trigger completion generation, so these profiles should be approximately the same

Negabinary:


cyrus-:
@disconcision can you confirm that this fixes the issue you reported #1501

disconcision (approved):
This works! It still does a small amount of unnecessary work when completion is enabled, but that's probably my bad, I'll file another issue for it.

cyrus- (approved):

cyrus-:
@disconcision go ahead and make an issue for the problem you're referring to

disconcision:
@cyrus- already did

PR #1506, opened 2025-02-10
Fix typo in initial Hazel example code

nskh:
Hazel initial code examples had a typo: `artihmetic` not `arithmetic`. Fixed here.


cyrus-:
There is a conflict now -- need to regenerate off of `dev`. Once you've done that, we can merge immediately to avoid this.

PR #1517, opened 2025-02-13
Init.ml formatting

Negabinary:
@cyrus- I think we didn't run the autoformatter on Init.ml when we updated it yesterday


disconcision:
there goes your negative linecount gains...

PR #1518, opened 2025-02-14
Make ID annotation on grammar a type parameter

7h3kk1d:
Currently used to make Statics tests easier to write. Should be used going forward to write more tests and have more accurate type signatures for program transformations.

I'm still looking for better encodings of the grammar but I don't want to hold up progress on that.

Moves some of the grammar/term files into the term directory and out of statics.

Another demonstration of the flexibility can be found here in probe tests: https://github.com/hazelgrove/hazel/blob/6c929c0b60da7e5a91a7455c84fa4be6e2b005a1/test/Test_Evaluator.re#L452-L517


src/haz3lcore/statics/ConstructorMap.re
```diff
@@ -191,6 +191,17 @@ let map = (f: option('a) => option('b), m: t('a)): t('b) => {
...
+  List.map(
+    fun
+    | Variant(ctr, args, Some(value)) =>
+      Variant(ctr, args, Some(f(value)))
+    | Variant(ctr, args, None) => Variant(ctr, args, None)
+    | BadEntry(value) => BadEntry(f(value)),
+    m,
+  );
+};
+
```
7h3kk1d, line 212: This function is being introduced because the existing map function can't change the type of the return value

src/haz3lcore/statics/Grammar.re
```diff
@@ -0,0 +1,415 @@
+open Util;
```
7h3kk1d: I pulled the Grammar out of `TermBase` so that all of the `TermBase` types can stay the same.

cyrus-: while we're moving things around, can we move the Term/Grammar files out of the statics directory and into term?

src/haz3lcore/statics/Grammar.re
```diff
@@ -0,0 +1,415 @@
...
+        | Cast(e, t1, t2) =>
+          Cast(
+            map_exp_annotation(f, e),
+            map_typ_annotation(f, t1),
+            map_typ_annotation(f, t2),
+          )
+        };
+      {term, annotation: new_annotation};
+    }:
+      exp_t(b) // TODO
```
cyrus-: what is TODO?

cyrus- (changes requested):

test/Test_Grammar.re
```diff
@@ -0,0 +1,26 @@
+open Alcotest;
+open Haz3lcore;
+open Grammar;
+
+let qcheck_map_annotation_test =
```
7h3kk1d, line 3: Coverage guilting me into writing property tests. Future PR I'll move the generator to the Grammar types instead of the menhir ones.

src/haz3lcore/lang/term/Grammar.re
```diff
@@ -0,0 +1,417 @@
...
+  | SynSwitch
+  | Hole(type_hole('a))
+  | Internal
+and filter('a) = {
+  pat: exp_t('a),
+  act: FilterAction.t,
+};
+
+
+let rec map_exp_annotation: type a b. (a => b, exp_t(a)) => exp_t(b) =
```
Negabinary, line 141: I feel like we should be able to combine map_?_annotation and map_term. Might be nice to save us from making another pattern match on expressions.

Negabinary (commented):
I like it; doesn't look like it's created any mess either

src/haz3lcore/lang/term/Grammar.re
7h3kk1d: All the types in `TermBase` are roughly the same right now. I didn't want to change the signatures of everything while I was introducing this. I think it definitely makes sense as a second pass to bring some of those things up and make them more generic.

For the mapping in particular. How do you see it working exactly? Use cases I see:
- Mapping an annotation with no transformation to the syntax tree
- Structure preserving transformations where the constructors don't change but the data inside them does
- Structure modifying transformations where a mapped node might be a completely different subtree
  - I found this a little difficult to figure out in [strip_casts](https://github.com/hazelgrove/hazel/blob/dev/src/haz3lcore/dynamics/DHExp.re#L57-L93) where I had to do some mutual recursion on top of the map_term function. 

As long as the interface for mapping annotations afterwards is just as easy I'm all about it.

Negabinary: I guess I was more thinking from an implementation perspective; I'm happy if we make both these mapping functions with separate signatures and have them implemented in terms of one map function. I just don't want people adding/modifying a language feature to have to modify it in two different map functions.

I agree map_term is annoyingly difficult to use for things that might modify structure so I'm definitely open to alternatives - I also wish it could do things like check to see if a certain id is in this expression, while it currently can't return anything other than an expression.

7h3kk1d: Yeah, that sounds good. I wonder if a fold on the tree structure actually can do some of this better than a map. Then implement everything in terms of the fold. If we can do it in a clean way I'm 100% in if the representation is super confusing I'd rather just have separate functions. Seems like it _should_ be doable clean though.

7h3kk1d: Took conversation off of the thread but recap:

We'll address potentially merging this with the TermBase.map_term in a later PR when more stuff from TermBase is made more generic. There's potentially more value in creating a fold but we'd need the right interface to do it.

cyrus- (approved):

PR #1525, opened 2025-02-17
Prover selection

Negabinary:
* Factor out MissingStep logic to make StepperBase slightly smaller
* Make substitutions only show on selection
* Add ugly buttons that let you use rewrites

### DIFF
--- a/src/haz3lcore/proof/ProofCtx.re
+++ b/src/haz3lcore/proof/ProofCtx.re
@@ -32,10 +32,13 @@ let rec get_rewrites = (ctx: t, exp: Exp.t) =>
     ] =>
     let bindings = get_empty_bindings(bindings);
     switch (MatchExp.match_exp([], bindings, a, exp)) {
-    | Some(_) => [b, ...get_rewrites(rs, exp)]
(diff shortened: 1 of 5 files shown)


src/haz3lcore/proof/ProofCtx.re
```diff
@@ -32,10 +32,13 @@ let rec get_rewrites = (ctx: t, exp: Exp.t) =>
     ] =>
     let bindings = get_empty_bindings(bindings);
     switch (MatchExp.match_exp([], bindings, a, exp)) {
-    | Some(_) => [b, ...get_rewrites(rs, exp)]
+    | Some(m) => [b |> MatchExp.substitute_exp(m), ...get_rewrites(rs, exp)]
```
nskh, line 35: What was happening before this? Were we not substituting in expressions at all?

nskh (approved):
Looks good to me

src/haz3lcore/proof/ProofCtx.re
Negabinary: yeah I hadn't implemented substitution yet; I think you're expecting too much from this code atm ahaha

PR #1527, opened 2025-02-18
Prover tidy

Negabinary:


src/haz3lcore/proof/Axioms.re
```diff
@@ -16,6 +16,22 @@ let v: ProofCtx.t =
        )
        |> Exp.fresh,
      )
+  |> ProofCtx.add_entry(
```
nskh, line 19: Technically stand-alone, right? Or does this depend on something added in this PR?

nskh (commented):
Looks good but can you explain the main changes so I understand a little better? 

To me, seems like:
- two modes were added for evaluation (`Substitution` and `Environment`)
- some restructuring was done for `Substitution.re` (not sure how this connects to the mode above, but I assume these two are dependent)
- One `Axiom` was added, which feels standalone?
- Big changes to `StepperBase.re` to add a `ForallStep` where we prove properties inside of a function.

Just to understand, are there dependencies between all of these additions, or could this technically go in as three standalone PRs?

src/haz3lcore/proof/Axioms.re
Negabinary: yup, technically stand alone, just lets me test more things

Negabinary:
Two modes were added to evaluation, yes, this was to make sure that we don't end up with any `Closure` s in the prover; `Closure` s are a evaluation speed optimization that just gets in our way in the prover. Substitution stuff is upstream of that and not important for the prover.

I re-adjusted the Forall datatype so that I can create a variable context in the prover.

PR #1530, opened 2025-02-20
Syntax factory

7h3kk1d:
# Add factory functions for AST

Adds a bunch of helper functions for building up the AST with default annotations.

Replaces the constructors in tests with the new helper functions to remove fresh calls. Also moved builtins.re and transition.re to use the new functions.


src/haz3lcore/lang/term/Grammar.re
```diff
@@ -415,3 +415,385 @@ and map_type_hole_annotation:
...
+    term: Match(e, l),
+    annotation: Option.value(~default=DefaultAnnotation.default_value(), ann),
+  };
+  let cast = (~ann=?, e, t1, t2): exp_t(DefaultAnnotation.t) => {
+    term: Cast(e, t1, t2),
+    annotation: Option.value(~default=DefaultAnnotation.default_value(), ann),
+  };
+
+  // pat
+  let pat_invalid = (~ann=?, s): pat_t(DefaultAnnotation.t) => {
```
7h3kk1d: We can also nest modules inside here instead of prepending sort. This works with local open so you could do:

```reason
FTemp.(Exp.(list([int(1), int(2), cast(int(3), Typ.int(), Typ.unknown(Internal)))))
```

test/Test_Statics.re
```diff
@@ -382,173 +366,127 @@ let tests = (
...
+            ty_prod([
+              ty_tup_label(
+                ty_label("b"),
+                ty_prod([
+                  ty_tup_label(ty_label("c"), ty_unknown(Hole(EmptyHole))),
+                ]),
+              ),
+            ]),
+          ),
+        ]),
```
7h3kk1d: Example of the fresh calls being removed.

test/Test_Statics.re
```diff
@@ -889,214 +792,172 @@ let tests = (
...
+                    );
+                  },
+                  tup_label(label("a"), string("hello")),
+                ],
               ),
             ),
-            no_error_exp(Bool(true)),
-          ),
+            bool(true),
+          )
```
7h3kk1d: The test is significantly smaller and only 30 out of the remaining ~90 lines are the AST. The rest are the annotations.

src/haz3lcore/lang/term/Grammar.re
```diff
@@ -417,3 +417,394 @@ and map_type_hole_annotation:
...
+  type exp = exp_t(DefaultAnnotation.t);
+  type pat = pat_t(DefaultAnnotation.t);
+  type typ = typ_t(DefaultAnnotation.t);
+  type tpat = tpat_t(DefaultAnnotation.t);
+  type typ_provenance = type_provenance(DefaultAnnotation.t);
+
+  let default_annotation = ann =>
+    Option.value(~default=DefaultAnnotation.default_value(), ann);
+  module Exp = {
+    let invalid = (~ann=?, s): exp_t(DefaultAnnotation.t) => {
```
7h3kk1d, line 436: Figure out if we want to have a pattern match from the original one to make sure we don't miss one of these.

test/Test_Grammar.re
```diff
@@ -18,7 +23,227 @@ let qcheck_map_annotation_test =
...
+      }
+    )
+  );
+};
+
 let tests = (
   "Grammar",
-  [QCheck_alcotest.to_alcotest(qcheck_map_annotation_test)],
+  [
+    test_case(
```
7h3kk1d, line 166: I added these test cases to walk over all of the expressions, types, patterns, ...  to make sure that we can build one for each `cls` and the class is in 1-1 correspondence with the term types. This turns out not to be true for types for Constructor.


The tests help ensure cls are in order and that we have the factory functions. I think that's good enough for now.

7h3kk1d: Maybe I'll find a way to derive both in the future.

src/haz3lcore/lang/term/Grammar.re
```diff
@@ -417,3 +417,400 @@ and map_type_hole_annotation:
     | MultiHole(l) => MultiHole(List.map(x => map_any_annotation(f, x), l))
     };
   };
+
+module type DefaultAnnotation = {
+  type t;
+  let default_value: unit => t;
+};
```
7h3kk1d, line 424: You could have 2 types here `t` for the type of the annotation and another type `s` that could be used in the factory methods below and require a `default_value : option(s) -> t` as part of the builder. I haven't bothered but for optional annotations it allows the people calling the builder to use a different type like not provide `Some` constructors.

cyrus-: leave as a comment in the file

cyrus- (approved):

PR #1533, opened 2025-02-21
Proof Steps

Negabinary:
## Overview

This branch adds new kinds of steps to the stepper 

In particular it adds steps for:
* Cases (e.g. case on whether a boolean is true/false)
* Axioms (e.g. `x * 0 = 0`)
* Rewrites (write your own axiom and check it with a SAT solver)

We are also porting over the coq export from https://github.com/hazelgrove/hazel/pull/1193 for basic axiom steps.

## TODOS:

- [x] Skel-based selection
- [x] Fix hidden steps
- [x] substitute variables in axioms
- [x] fix closure issues
- [x] add context to proofs
- [x] Fix duplicated ids in substitution
- [x] Prevent variable escape at the end of forall (last_step in forall)
- [x] Put rewrites in a box

## TODOs for this pr:

- [x] toggle proof actions button
- [x] fix turn off history
- [x] fix turn off stepper
- [x] Add experimental warning to Algebrite thing
- [x] Replace star button in axioms
- [x] Add back the styling for hidden steps
- [x] print out expression that's going to be replaced in algebrite thing

## After this pr:

- [ ] inexhaustive error message
- [ ] add an error message if cases don't end at the same value
- [ ] Coq export for axioms
- [ ] Make selection select subtrees
- [ ] add inductive hypotheses to cases
- [ ] Fix capture in substitution
- [ ] Fix capture in evaluation etc
- [ ] Add conditions to axioms
- [ ] Make evaluate work
- [ ] Require filling out under-specified rewrites
- [ ] Make the exp substitution more principled
- [ ] Cleanup: combine Match Exp and Filter match

 - Add a Theorem keyword and put the proofs in splices
 - Make the patterns in case actually patterns and run exhaustivity, redundancy, ...
 - Pattern unpacking


package-lock.json
```diff
@@ -821,6 +831,15 @@
         "node": ">= 8"
       }
     },
+    "node_modules/big-integer": {
```
cyrus-, line 834: @nskh is this something you are using with Algebrite? I believe the Hazel big integers are using an OCaml library not this one, so if you aren't using it we can remove this.

src/haz3lcore/dynamics/TypeAssignment.re
cyrus-, line 1: move to ProofHacks

src/haz3lcore/proof/MatchExp.re
```diff
@@ -0,0 +1,426 @@
...
+          switch (List.assoc(x, sub)) {
+          | None => exp
+          | Some(e) => e
+          }
+        | _ => cont(exp)
+        },
+    exp,
+  );
+
+// [@deriving (show({with_path: false}), sexp, yojson)]
```
cyrus-: deleted

cyrus- (approved):

PR #1541, opened 2025-02-24
fix: remove filter vars in context inspector

### LINKED ISSUE #1535: stepper filter variables should not be shown in context inspector
Image

tonyfettes:
This PR adds an extra filter to statics/Ctx.re, so that we can filter out stepper filter variables when rendering the `ContextInpsector`.

For now, every stepper filter variable starts will a `$` sign, therefore we utilize this to find out all the stepper filter variables and drops them in the `filter_stepper_filter_variables`.

@Negabinary Is there any other part of Hazel uses `$` for identifiers/variables? If so I guess we need other ways to differentiate stepper filter variables out.

Fixes #1535

### DIFF
--- a/src/haz3lcore/statics/Ctx.re
+++ b/src/haz3lcore/statics/Ctx.re
@@ -175,5 +175,24 @@ let filter_duplicates = (ctx: t): t =>
      )
   |> (((ctx, _, _)) => List.rev(ctx));
 
+let filter_stepper_filter_variables = (ctx: t): t =>
+  ctx
+  |> List.fold_left(
+       (ctx, entry) => {
+         switch (entry) {
+         | VarEntry({name, _})
+         | ConstructorEntry({name, _})
+         | TVarEntry({name, _}) =>
+           if (String.starts_with(~prefix="$", name)) {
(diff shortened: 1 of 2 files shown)


Negabinary (approved):
@tonyfettes I'm pretty sure filters is still the only way to introduce a variable with a `$` sign, so this will work.

src/haz3lcore/statics/Ctx.re
```diff
@@ -175,5 +175,25 @@ let filter_duplicates = (ctx: t): t =>
...
+let filter_stepper_filter_variables = (ctx: t): t =>
+  ctx
+  |> List.fold_left(
+       (ctx, entry) => {
+         switch (entry) {
+         | VarEntry({name, _})
+         | ConstructorEntry({name, _})
+         | TVarEntry({name, _}) =>
+           if (String.starts_with(~prefix="$", name)) {
+             Printf.printf("filtering stepper filter variables: %s\n", name);
```
cyrus-: remove print before merge

cyrus- (approved):

PR #1542, opened 2025-02-25
Create selection-based buttons for prover

Negabinary:


nskh (approved):

src/haz3lweb/app/editors/stepper/StepperBase.re
```diff
@@ -426,22 +431,23 @@ module Update = {
       ) => {
     switch (step_kind) {
     | SingleStep(m) =>
-      calculate_single_step(~settings, expr, state, m, hidden, editor)
+      calculate_single_step(~settings, expr, ctx, state, m, hidden, editor)
```
nskh, line 434: Why do we need to send in `ctx` here and elsewhere? Unless I'm missing it, it doesn't seem to get used.

src/haz3lweb/app/editors/stepper/MissingStep.re
```diff
@@ -80,21 +141,312 @@ module Update = {
...
+                            Widgets.button(Icons.star, _ =>
+                              signal(
+                                AddAxiomStep(
+                                  unboxed_selected_exp,
+                                  unboxed_cached_exp,
+                                ),
+                              )
+                            ),
+                          ]
+                        | Some(false) => [Web.Node.text("Invalid")]
```
nskh, line 436: ok, so here is where we're using whether the rewrite is valid or not - nice

PR #1554, opened 2025-03-07
Use keyword

Negabinary:
Todo:

- [x] Rename to Atom
- [x] Make use ? in not a static error
- [x] Add float to integers & add bigint / sint
- [x] Add `use SInt in`
- [x] Add ExplainThis
- [x] Add monus builtin
- [x] Make monus work

Fixes:
- [x] crashes when writing case analysis
- [x] remove nat_of_nat
- [x] change max_int and min_int to say sint and return those types
- [x] explainthis for integer literals should change to number literals
- [x] add mod for nats and sints and floats

Wontfixes:
- [x] leave issue for BigInt performance -- maybe we can see in the future if JS native BigInt is faster than the OCaml bignum library
- [x] make issue for adding suffixed versions of literals, e.g. 1n and 1i. Need to work out how this works in dynamics for display
- [ ] refactor to add a contextual version of the operators for use in surface programs


cyrus- (commented):
- crashes when writing case analysis
- remove `nat_of_nat`
- leave issue for BigInt performance -- maybe we can see in the future if JS native BigInt is faster than the OCaml bignum library
- change `max_int` and `min_int` to say `sint` and return those types
- explainthis for integer literals should change to number literals
- add `mod` for nats and sints and floats
- refactor to add a contextual version of the operators for use in surface programs
- make issue for adding suffixed versions of literals, e.g. `1n` and `1i`. Need to work out how this works in dynamics for display.
-

PR #1565, opened 2025-03-16
Stop parenthesizing non-nullary constructors in ExpToSegment

7h3kk1d:
Stops adding parentheses around variant constructors when they're being applied to arguments in exptosegment.

Before
Screenshot From 2025-03-16 06-35-11
After
Screenshot From 2025-03-16 06-34-45

### DIFF
--- a/src/haz3lcore/pretty/ExpToSegment.re
+++ b/src/haz3lcore/pretty/ExpToSegment.re
@@ -44,7 +44,8 @@ let rec external_precedence = (exp: Exp.t): Precedence.t => {
   | BuiltinFun(_)
   | Undefined
   | Label(_)
-  | TupLabel(_) => Precedence.max
+  | TupLabel(_)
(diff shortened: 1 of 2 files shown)


src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -54,7 +55,6 @@ let rec external_precedence = (exp: Exp.t): Precedence.t => {
 
   // Other forms
   | UnOp(Meta(Unquote), _) => Precedence.unquote
-  | Constructor(_) // Constructor is here because we currently always add a type annotation to constructors
```
7h3kk1d, line 57: @Negabinary this comment seems outdated as far as I can tell we don't add the type annotations onto the constructors

7h3kk1d:
Fixed in https://github.com/hazelgrove/hazel/pull/1420

PR #1570, opened 2025-03-20
Fix operation explanation typos

nskh:
I happened to notice the following:
Screenshot 2025-03-20 at 12 17 40 AM

Fixes typos where ExplainThis said things like "multiplied be" instead of "multiplied by" and "evalute" instead of "evaluate".


disconcision (approved):
lgtm

PR #1573, opened 2025-03-21
Treat List Cons as constructor when it has indet tail

### LINKED ISSUE #1572: List cons not treated as a constructor
Instead, Hazel represents list values as literals. So, only list literals are treated as boxed values.

If the tail part of a list cons is indeterminate, but not a list literal, e.g. ? or [1] @ ?, it cannot be represented as a list literal:

Image

We should also treat these specific cons operations as 'boxed values', allowing unboxing on a cons request (but not unboxing on a list literal request). Allowing the following match statement to evaluate to true:

Image

MaxCarroll0:
Closes #1572 

List Cons with indeterminate tail, e.g. `1::?` is now treated as a constructor and can be unboxed with the Cons unbox request.

A cons explicitly does NOT match against a list literal with smaller length than the number of conses in the expression. But, indet matches against list literals with larger length. e.g.
- 1 :: ? Indet matches with [1]
- 1 :: ? does NOT match with []

Additionally:
- Renames List unboxing request to ListLit to make clear that you cannot unbox a cons with indet tail into a finite list literal


src/haz3lcore/dynamics/Transition.re
```diff
@@ -695,13 +695,17 @@ module Transition = (EV: EV_MODE) => {
...
+      switch (Unboxing.unbox(ListLit, d2')) {
+      | Matches(ds) =>
+        Step({
+          expr: list_lit([d1', ...ds]),
+          state_update,
+          kind: ListCons,
+          is_value: true,
+        })
+      | DoesNotMatch => Indet
+      | IndetMatch => Constructor // Treat list cons with indet tail as constructors
```
Negabinary: I agree this doesn't currently have any effect on semantics

src/haz3lcore/dynamics/Unboxing.re
```diff
@@ -3,13 +3,19 @@ open Util;
 /* What is unboxing?
 
    When you have an expression of type list, and it's finished evaluating,
-   is it a list? Sadly not necessarily, it might be:
+   is it a list literal? Sadly not necessarily, it might be:
 
-    - indeterminate, e.g. it has a hole in it
-    - a list with some casts wrapped around it
+    - an indeterminate list cons: e.g. 1 :: ?
```
Negabinary: it's still possible for things like Cast(EmptyHole, EmptyHole, List(EmptyHole)) to have a list type too so it would be helpful to keep the old first bullet point / have a third bullet point to cover that case.

Negabinary (changes requested):
Looks good to me! I just had one comment on a comment.

Would you also be able to add a couple tests to make sure we don't ever accidentally delete this functionality in the future?

Also, for future, it would be helpful if you make branches on main repo instead of on a fork, because then we can automatically run tests on it and try it out in the browser at https://hazel.org/build/<branch name>

src/haz3lcore/dynamics/Unboxing.re
MaxCarroll0: Yep, that would make sense to keep as a third point.

Negabinary (approved):

cyrus-:
@MaxCarroll0 the diff here appears to have a bunch of formatting regressions, probably because you need to update to the latest dependencies. please update to a clean diff and then we can merge.

Negabinary:
@MaxCarroll0  (feel free to message me or @7h3kk1d if you need help with updating your deps to fix the formatting issues)

MaxCarroll0:
@Negabinary diff is now clean and mergeable!

cyrus- (approved):

cyrus-:
Thanks for the fix @MaxCarroll0!

PR #1574, opened 2025-03-21
Bundle npm dependencies using esbuild

### LINKED ISSUE #1557: Add a bundler/build step to package js dependencies
Flakiness on unpkg is causing some page load issues to load the ninja keys js from the cdn. We should figure out a bundler or at least vendor this dependency https://github.com/hazelgrove/hazel/blob/6df1774d64832ebaf8397fed1527b0b48431707c/src/haz3lweb/www/ninja_module.js#L1

7h3kk1d:
### ESBuild Bundling Dependencies

This is being used to bundle our npm dependencies at build time so we're not reliant on unpkg or esm.sh. This is currently being used for ninja-keys and it's transitive dependencies but will also be used for any other npm dependencies we add such as in https://github.com/hazelgrove/hazel/pull/1568.

This does unfortunately add another package manager `npm` for managing our JS dependencies and an esbuild step to our dune rules to create a bundled.js file that has the dependencies. I'm interested in finding better solutions but this seemed to work so far. 

#### Alternative approaches

- https://github.com/ahrefs/opam-check-npm-deps
  - I couldn't figure out how to get this to work after trying for a bit. Complicated more by having dune between us and opam.
- https://github.com/tmattio/opam-esbuild#7738af5912cc9f554d31da81d2334e88052102fb OCaml.org seems to use this in some places but I haven't tried figuring out how
- https://melange.re/v5.0.0/ seems to have first class npm support but it's a newer compiler so I haven't bothered investigating.

#### TODO

- [x] Add a make step for installing the js dependencies
- [x] Add readme information for installing npm and dependencies

### DIFF
--- a/.github/workflows/deploy_branches.yml
+++ b/.github/workflows/deploy_branches.yml
@@ -6,7 +6,7 @@ jobs:
   Deploy:
     runs-on: ubuntu-latest
     steps:
-      # NOTE: position the below lines in the code between two steps
+      # NOTE: position the below lines in the code between two stepse
       # and uncomment them to open an ssh connection at that point:
       #- name: Debugging with ssh
       #  uses: lhotari/action-upterm@v1
@@ -16,6 +16,13 @@ jobs:
           path: source
       - name: Add the name of the current branch to the environment as BRANCH_NAME
         uses: nelonoel/branch-name@v1.0.1
+      - name: Install node/npm
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'latest'
+      - name: Install NPM dependencies
+        run: npm install
+        working-directory: ./source
       - name: Set-up OCaml
         uses: ocaml/setup-ocaml@v3
         with:
--- a/.github/workflows/deploy_branches.yml
+++ b/.github/workflows/deploy_branches.yml
@@ -6,7 +6,7 @@ jobs:
   Deploy:
     runs-on: ubuntu-latest
     steps:
-      # NOTE: position the below lines in the code between two stepse
+      # NOTE: position the below lines in the code between two steps
       # and uncomment them to open an ssh connection at that point:
       #- name: Debugging with ssh
       #  uses: lhotari/action-upterm@v1
--- a/package.json
+++ b/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "hazel",
+  "repository": {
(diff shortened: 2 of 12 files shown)


gcrois:
@7h3kk1d and I just went through what using `esy` might look like (https://esy.sh/docs/what-why)

We would have to maintain the package.json so that it includes the dependencies referenced in dune-project; for us, this command: 
```
esy add ocaml@">=5.2.0" \
  @opam/menhir@">=2.0" \
  @opam/yojson \
  @opam/reason@">=3.12.0" \
  @opam/ppx_yojson_conv_lib \
  @opam/ppx_yojson_conv \
  @opam/incr_dom \
  @opam/bisect_ppx \
  @opam/omd@">=2.0.0~alpha4" \
  @opam/ezjs_idb \
  @opam/bonsai \
  @opam/ppx_deriving \
  @opam/ptmap \
  @opam/uuidm@=0.9.8 \
  @opam/unionFind \
  @opam/ocamlformat \
  @opam/junit_alcotest@">=2.1.0" \
  @opam/ocaml-lsp-server \
  @opam/ocaml-index \
  @opam/qcheck \
  @opam/qcheck-alcotest \
  @opam/ppx_deriving_qcheck
```

Ultimately, not resulting in much benefit (except making it slightly easier to run for the first time).

I am in favor of adding esbuild, but not esy unless we restructure the project to run using melange.

cyrus- (changes requested):
add docs/external_dependencies.md with instructions on how to add new js dependencies

.github/workflows/deploy_branches.yml
```diff
@@ -16,6 +16,13 @@ jobs:
           path: source
       - name: Add the name of the current branch to the environment as BRANCH_NAME
         uses: nelonoel/branch-name@v1.0.1
+      - name: Install node/npm
```
gcrois, line 19: My larger projects all use pnpm rather than npm -- almost identical interface-wise, but much faster. I like npm for now, as it's more common, but if js building becomes a bottleneck, we should consider it

package.json
gcrois, line 1: If we ever export components of hazel as npm-accessible packages, we would indicate that here.

gcrois (approved):

cyrus- (approved):

PR #1575, opened 2025-03-22
LLM Hole fillings and Assistant Sidebar

russell-rozenbaum:
_**To-Do**_

**Before Initial Merge**
- [x] fix or move structured edit actions in Perform that aren't working
- [x] remove system prompt spam
- [x] make sure hinted test feature works (or move to another branch)
- [x] make sure nothing outside of the assistant code is commented out or instrumented with prints
- [x] go through Andrew's comments and see if there are any other items in this category
- [x] make each `??` invocation open up a new chat automatically
- [x] fix issues with resizing the sidebar
- [x] clean up Perform.re and Indicated.re and anything else impacted by structure edit actions (mainly navigation actions)

🔴 **Most Important** 🔴
- [x] **Make it possible for LLM to respond with a chain of tool calls**, which will in turn make the task completions much more cost-effective.
- [x] **Allow LLM to navigate to and edit function bodies if need-be**. This is most critical at the end of a program sketch, where the tail of the program sketch is inherently the body of a _let_, thus we should allow for the LLM to somehow navigate here. How is the main question... we could allow it to _only_ access the body of the final let statement, or we could allow it to freely choose to perhaps "goto_body" of any let statement. 
- [x] **BUG: User needs to manually click into editor before LLM responds with a tool call, otherwise there will be no cursor placed and nothing will happen**
- Update on the above: It seems like this was fixed by fixing the whitespace bug (they were seemingly the same thing actually, and not a problem with the cursor)
- [x] **Server becomes very slow as chats get long**. This is likely due to the unpacking of code messages... mainly since it updates with every mouse action. 

🟡 **Less Important** 🟡
- [ ] **Implement functionality for more tool calls**. We still need to implement scrolling up/down and show_references.
- [ ] **Improve prompting**. This is quite broad, but some focusses could be to improve the automated context retrieval and showing the LLM where its cursor currently is/what it has selected/what it just updated. This is likely to be an ongoing process throughout the implementation of other features, such as tool calls or historical context. 
- [x] **Restyle/Rename chat parties**. This is mainly for increased intuition, but we would supposedly want the "user" message to be what the user actually typed in the text box, the "Assistant" to be the LLM response, the "System" to be automated prompts/contexts/etc, and an "Error" output (maybe instead of "System", as errors can stem from external API calls) to report any failures. 
- [x] **Make collected chats (for historical context) much shorter/make more cost-efficient**. This could be done by summarizing collected chats (each time they are collected) into a predefined number of words, disposing of unnecessary information/context (eg. old sketches from prior messages), and/or possibly even using an LLM or ML mechanism itself to determine whether or not a given user message actually requires historical context (eg. Feed the msg "Why did you implement it like that?" into an LLM asking "Does this require historical context: Yes or No?", expecting Yes. eg. example of a "No" response would be feeding in something like "please make this function recursive") 

**🐞 Bugs / 🕳️ Pitfalls / 🔭 Things to Look Into** 
- [ ] **Cannot properly select "test" defs and bodies.** There are likely other language server navigation errors like this. 
- [ ] **Sketch can quickly become corrupt.** When improper syntax is used, the language server fails to navigate using purely the "structure" of the code. It also seems that it might cause the LLM to be confused on how to navigate using purely the structure of the code. Take, for example, the following sketch: ```let case in fun```. This program isn't syntactically correct or semantically meaningful, so is there a way to navigate on it in a "structured" manner?
- [ ] **Comments above a definition cannot be edited/changed** as of right now. Also, comments aren't properly inserted into sketch in general. Either they aren't inserted/parsed out of a response properly, or the LLM doesn't use proper syntax when making comments.
- [x] **Refine Few-Shot Examples**. Try to cover all cases, syntax, and overall complexity of task completion in Hazel.
- [x] **Prompt Generation doesn't work when the cursor has selected whitespace in the program sketch**. Getting the cursor inspector information fails, which causes prompt generation to fail, and no message is sent.
- [ ] **Can't copy from code blocks**
- [x] **Prompt display and history menu can interfere**. This should be a simple fix, just close the other when one is opened.
- [x] **Suggestion mode UI message displays need to be updated to match new architecture.** Applies to the frontend and backend of things.
- [ ] Editor sent to update function changes when the user changes the editor. So in chains/loops of agent calls, the editor is not persisted.


disconcision:
I made a variety of style and formatting tweaks... let me know if you object to any. I have some other ideas, mostly for simplifying things, but I'll talk about them when we meet next. here are some of the things I changed:

1. Minor updates to ExplainThis for consistency; sidebar resizing works with explainthis now, and I got rid of the now mostly empty header in favor of a floating toggle for highlighting:

Screenshot 2025-06-10 at 7 27 15 PM

2. Consistency updates for the settings UI. A variety of changes here, mostly minor. The css/dom structure is a bit messy here; I started folding some things into reusable components but didn't go all the way. Right now things like font sizes, border radii etc are defined in many different places, so there were lots of different values. I partially consolidated these, and manually homogenized some others. I also commented out the manual model entry field; not sure we need that given the dropdown?

Screenshot 2025-06-10 at 7 27 31 PM

Some things still to do:
1. We still need to find a way to specify recommended model(s) here.
2. Minor: I think what I've now called the New and Current API Key components could be combined into a single component that just displays whatever the current stored key is, or shows the message to enter one if there's none. This could also do some kind of validation on_change. Not sure the update button is strictly necessary either; it could just update on_change. Unless update actually validated the key with the server, but it doesn't do that right now.
3. Minor: Similar for the current model ID.

One thing I'm noticing now is that when you try to do a ?? hole filling, the Suggest panel fills up with a lot of system messages:

Screenshot 2025-06-10 at 7 27 48 PM

russell-rozenbaum:
> the manual model entry field; not sure we need that given the dropdown?

The current list filters out models that don't support tool calls. In general, if we ever filter the list, we might want to let the user add their own model that may have been filtered out from that list.

> We still need to find a way to specify recommended model(s) here.

We could implement a simple regex filter based on our belief of what the top models are. I currently have something like this in OpenRouter.re with the `is_top_model` function, but I've excluded it from being used anywhere for now. This would require maintenance a few times a year—updating recommendations with better models and removing out-of-date ones. 

Everything sounds good. And yea, I've yet to get around to concatenating the suggestion mode system prompt, so '??' will send many detached prompts to the model.

src/web/view/ScratchMode.re
```diff
@@ -117,9 +116,68 @@ module Update = {
     JsUtil.QueryParams.set_param("share", StringUtil.compress(c));
   };
```
disconcision: why is this calculate function copy pasted? there are two versions in this file now

src/web/app/Page.re
```diff
@@ -263,6 +437,7 @@ module Update = {
     | Globals(action) => Globals.Update.can_undo(action)
     | Editors(action) => Editors.Update.can_undo(action)
     | ExplainThis(action) => ExplainThisUpdate.can_undo(action)
+    | Assistant(_) => false // TODO: ??
```
disconcision: should any assistant actions be undoable?

src/web/app/helpful-assistant/AssistantUtil.re
```diff
@@ -0,0 +1,25 @@
+open Util;
```
disconcision: this is more generally applicable, move to something like Util/TimeUtil

src/haz3lcore/zipper/action/Action.re
```diff
@@ -35,13 +35,19 @@ type rel =
   | Current
   | Id([@equal (_, _) => true] Id.t, Direction.t);
 
+[@deriving (show({with_path: false}), sexp, yojson, eq)]
```
disconcision: don't totally get what these actions are supposed to be. definition of what? body of what?

src/haz3lcore/zipper/action/Indicated.re
```diff
@@ -141,3 +141,25 @@ let ci_of =
       ctx: Language.Statics.Info.ctx_of(ci),
     });
   };
+
```
disconcision: not sure why this is here?

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -78,7 +96,26 @@ let go_z =
       }
     };
 
-  let smart_select = (n, z): option(Zipper.t) => {
```
disconcision, line 81: is this code just for testing?

src/language/dynamics/state/TestMap.re
```diff
@@ -2,16 +2,31 @@ open Util;
 
 /* FIXME: Make more obvious names. */
 [@deriving (show({with_path: false}), sexp, yojson)]
-type instance_report = (DHExp.t, TestStatus.t);
+type instance_report = {
+  exp: DHExp.t,
+  status: TestStatus.t,
+  hint: string,
+};
+// type instance_report = (DHExp.t, TestStatus.t);
```
disconcision: commented code

src/language/dynamics/transition/Transition.re
```diff
@@ -338,7 +338,46 @@ module Transition = (EV: EV_MODE) => {
...
+          update_test(
+            state,
+            DHExp.rep_id(d),
+            {
+              exp: d,
+              status: result,
+              hint: "No hint available.",
+            },
+          ),
+        // update_test(state, DHExp.rep_id(d), (d', result)),
```
disconcision, line 327: commented code

src/web/Settings.re
```diff
@@ -239,6 +265,71 @@ module Update = {
...
+            },
+          }
+        | ToggleHistory => {
+            ...settings,
+            assistant: {
+              ...settings.assistant,
+              show_history: !settings.assistant.show_history,
+            },
+          }
+        | SetLLM(llm_id) =>
```
disconcision: we should find another way to do this... this module should really just be about a functional updating of the settings, we shouldn't be doing effects here. these should be actions in a separate assistant component

src/web/Settings.re
```diff
@@ -239,6 +265,71 @@ module Update = {
...
+          };
+          settings;
+        | ToggleAPIKeyVisibility => {
+            ...settings,
+            assistant: {
+              ...settings.assistant,
+              show_api_key: !settings.assistant.show_api_key,
+            },
+          }
+        | SetAPIKey(api_key) =>
```
disconcision: same as above; this isn't just a settings toggle; we shouldn't be doing schedule_action here

src/web/app/Page.re
```diff
@@ -196,6 +230,59 @@ module Update = {
           editors,
         };
       };
+    //TODO: Figure out how below is impacted by new global undo/redo
```
disconcision: the below changes were on code that's no longer extant due to matt's new global undo/redo system... maybe ask him if any of the assistant logic beneath is necessary?

src/web/app/Page.re
```diff
@@ -236,6 +328,88 @@ module Update = {
         ...model,
         explain_this,
       };
+    | Assistant(action) =>
```
disconcision, line 270: there's a lot of logic below. this should be either somewhere else or broken out into its own file or possibly some helpers. i'm having a hard time following exactly what's happening here

src/web/app/editors/code/CodeEditable.re
```diff
@@ -247,6 +248,8 @@ module View = {
           ]);
         | 2 => inject(Perform(Select(Smart(2))))
         | 3 => inject(Perform(Select(Smart(3))))
+        | 4 => inject(Perform(Select(Smart(4))))
```
disconcision: what are these additional states?

src/web/app/explainthis/data/HintedTestExp.re
```diff
@@ -0,0 +1,40 @@
+open Haz3lcore;
+open Example;
+open ExplainThisForm;
+let hinted_test_true_ex = {
+  sub_id: HintedTestTrue,
+  term: mk_example("hint \"Always true\"\n test true end"),
+  message: "This is reported as a passing test because the body of the test is true.",
+};
+// TODO are these really the correct messages/explanations; maybe include something about the result being triv
```
disconcision: TODO

src/web/app/explainthis/data/HintedTestExp.re
```diff
@@ -0,0 +1,40 @@
...
+let hinted_test_false_ex = {
+  sub_id: HintedTestFalse,
+  term: mk_example("hint \"Always false\"\n test 3 < 1 end"),
+  message: "This is reported as a failing test because the body of the test is 3 < 1 which evaluates to false.",
+};
+let _exp_body = exp("e");
```
disconcision, line 15: rm?

src/web/view/ScratchMode.re
```diff
@@ -246,50 +315,60 @@ module Selection = {
   open Cursor;
 
   [@deriving (show({with_path: false}), sexp, yojson)]
-  type t = CellEditor.Selection.t;
```
disconcision, line 336: i don't get why we need to reify TextBox selection here? why do we need to change this type and add the resulting logic?

src/web/app/Page.re
```diff
@@ -62,11 +66,27 @@ module Update = {
...
     | Editors(Editors.Update.t)
     | ExplainThis(ExplainThisUpdate.update)
+    | Assistant(AssistantUpdate.t)
     | MakeActive(selection)
     | Benchmark(benchmark_action)
     | Start
     | Save;
 
+  // Helper function to check for insertion of '??' or '?a' and trigger assistant updates
+  let send_assistant_insertion_info =
```
disconcision: this fn seems to be a fairly thin wrapper around check_req? is it worth defining?

src/web/app/helpful-assistant/AssistantUpdate.re
```diff
@@ -0,0 +1,1252 @@
...
+        editor,
+        chat_id,
+      ),
+    );
+  };
+
+  // Check if user just typed ??
+  switch (caret, Zipper.neighbor_monotiles(siblings)) {
+  | (Outer, (_, Some(_))) =>
+    switch (Zipper.right_neighbor_monotile(siblings)) {
```
disconcision: we already should have this from the above call, no need to do another switch

src/web/app/helpful-assistant/AssistantUpdate.re
```diff
@@ -0,0 +1,1252 @@
...
+        let advanced_reasoning = true;
+        send_message(tileId, advanced_reasoning);
+      | _ => ()
+      }
+    | _ => ()
+    }
+  | (Outer, (_, None)) =>
+    switch (Zipper.left_neighbor_monotile(siblings)) {
+    | Some(c) =>
+      switch (c) {
```
disconcision: this seems duplicated from above, let's find a way to abstract this

src/web/view/ScratchMode.re
```diff
@@ -131,10 +189,21 @@ module Update = {
...
         scratchpads: new_sp,
       };
+      switch (a) {
+      // Check for assistant hole completion triggers
+      | MainEditor(Perform(Insert(char))) =>
+        // todo: This might hinder runtime performance since we'll ultimately
+        //       calculat the info map twice. A hacky idea of fixing this is
+        //       to set a flag if '??' or '?a' was inserted and only send the
+        //       completion request after scheduled info calculation is complete.
+        send_assistant_insertion_info(~char, ~editor=new_ed.editor)
```
disconcision: is this calculating the info_map? this should be a very simple syntactic check (ie it should be O(1)).

disconcision (changes requested):
See questions. If we're going to try and get this merged as cyrus suggests, let's focus for a bit on cleanup, especially around the editor-centric parts. the assistant-specific code can stay messy but let's try to get the parts where it interactions with the editor+language a bit tidier

disconcision:
I resolved the merge conflicts after the language PR merge and reinstated the model free text field as per russell's comment above

> And yea, I've yet to get around to concatenating the suggestion mode system prompt, so '??' will send many detached prompts to the model.
@russell-rozenbaum what does it mean exactly? how many prompts are we sending when you enter `??`? this sounds slightly worrying

src/web/Settings.re
russell-rozenbaum: I moved these out of AssistantModel and into AssistantSettings because my thinking was that they are setting parameters in the Assistant Architecture, so, by definition, they are updating the settings, not performing any effects. I sort of see them as analogous to something like toggling instructor mode in exercise mode; they "set" things within the assistant architecture.

src/web/app/Page.re
russell-rozenbaum: These are here to avoid circular dependencies. We could look into breaking them out.

src/web/app/editors/code/CodeEditable.re
russell-rozenbaum: Testing/debugging

src/web/app/explainthis/data/HintedTestExp.re
russell-rozenbaum: This is a remnant from a merge. Not sure what this todo entails

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -11,12 +11,30 @@ let buffer_clear = (z: t): t =>
...
   | Some(z) => z
   };
 
+let set_llm_buffer = (z: t, response: string): t =>
+  switch (TyDi.set_llm_buffer(z, response)) {
+  | None => z
+  | Some(z) => z
+  };
+
+let paste = (z: Zipper.t, str: string): option(Zipper.t) => {
```
disconcision, line 36: is this copied from somewhere?

src/language/dynamics/state/TestMap.re
disconcision: also should this hint be an option string? what happens for a regular test?

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -243,14 +243,20 @@ and exp_term: unsorted => (Exp.term, list(Id.t)) = {
         | term => ret(ListLit([term]))
         }
       | (["test", "end"], [Exp(test)]) => ret(Test(test))
+      | (["hint", "test", "end"], [Exp(hint), Exp(test)]) =>
```
disconcision, line 249: this seems to indicate that the parsing of a hint form fails entirely if the hint is not a string? seems like this could have negative consequences in-editor. i can't test it though as hints don't seem to work in-editor? tried to enter "hint" and it doesn't expand

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -243,14 +243,20 @@ and exp_term: unsorted => (Exp.term, list(Id.t)) = {
...
       | (
           ["case", "end"],
           [Rul({term: Rules(scrut, rules), annotation: {ids, _}})],
         ) => (
           Match(scrut, rules),
           ids,
         )
-      | ([t], []) when t != " " && !Form.is_explicit_hole(t) =>
+      | ([t], [])
+          when t != " " && !Form.is_explicit_hole(t) && !Form.is_llm_hole(t) =>
```
disconcision: this case is starting to get messy and too implicit. for this and the similar cases for other sorts, let's create a separate positive condition case, ie above the invalid case, add a case which returns a hole when the negation of this condition is met, i.e. when it's an explicit or llm hole. not sure what the whitespace case is here, seems like it should be an illegal state; maybe try to remove it and see if anything breaks?

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -89,10 +126,14 @@ let go_z =
       Piece.is_term(p)
         ? Select.parent_of_indicated(z, statics.info_map)
         : Select.current_term(~defs_exclude_bodies=true, ~case_rules=true, z);
+    | 4 => select_definition(z)
+    | 5 => select_body(z)
     | _ => None
     };
   };
 
+  //let go_to = (z): option(Zipper.t) => {};
```
disconcision: commented code

src/web/Settings.re
disconcision: this one is writing to disk though, and the other is issuing an API request. the other settings actions just update the model. this is not necessarily a dealbreaker but definitely a smell

src/web/app/Page.re
disconcision: dependency injection is possibly a way to go... which is the problematic reference?

src/web/view/ScratchMode.re
disconcision: also, we might want to generalize this function a bit. maybe call it something like editor_effect, which takes an action and an editor, and document it, saying its a callback to perform an effect on editor actions. then you can internalize this switch into the implementation. the purpose here being to get a cleaner API between the editor and the assistant that is not overly specific.

disconcision (changes requested):
A few more comments/replies. We can talk about it this afternoon but I think to merge we want to consider cherry picking some of the currently non-functional or incomplete parts into a new PR. I think it's fine to leave incomplete stuff in the assistant code, but I'm hesitant to leave incomplete code which integrates with Language or Editor as it complicates other people's work in those areas.

src/web/app/helpful-assistant/ChatLSP.re
```diff
@@ -209,6 +209,10 @@ module Composition = {
     | Definition
     | All;
 
+  type kind_of_goto =
```
disconcision: both are kinds variables; perhaps Value and Type?

disconcision:
I looked fairly extensively at the hint crash issue (repro instructions: enter a linebreak, press up arrow, then type `hint"`). It seems to be a fairly subtle regrouting related issue. I've pushed a workaround which hopefully doesn't break anything else; let me know if it seems okay on your end

russell-rozenbaum:
> I looked fairly extensively at the hint crash issue (repro instructions: enter a linebreak, press up arrow, then type `hint"`). It seems to be a fairly subtle regrouting related issue. I've pushed a workaround which hopefully doesn't break anything else; let me know if it seems okay on your end

I was messing around with this a bit more, and found a bug that could be related or separate...

Eg. (although, probably applies to any fully defined hinted test)
`hint "here" test x == 3 end;`

Attempting to delete any character from the test keyword results in the Nonconvex_segment.

disconcision:
@russell-rozenbaum this might actually an existing dev issue; trying to delete the `=` in a let expression also throws a convex segment error. out of curiosity, does this error still occur when typechecking is turned off in the settings?

russell-rozenbaum:
> @russell-rozenbaum this might actually an existing dev issue; trying to delete the `=` in a let expression also throws a convex segment error. out of curiosity, does this error still occur when typechecking is turned off in the settings?

Right, I see—this doesn't happen when type checking is switched off

russell-rozenbaum:
I notice that when turning off types, the bottom bar, along with its associated child components (context and projectors) disappears; Is this expected behavior? 

Other than this, all of the requests have been addressed, and the PR should be ready for a final review now.

disconcision:
@russell-rozenbaum: yes, that's expected, the bottom bar depends on types. i guess the projector panel could stick around in principle though... the deletion crash issue should be fixed, or at least mitigated

src/web/Main.re
```diff
@@ -148,6 +148,9 @@ let start = {
       >= 0;
     NinjaKeys.initialize(Shortcut.options(schedule_action));
     JsUtil.focus_clipboard_shim();
+    schedule_action(
+      Assistant(AssistantUpdate.ChatAction(FilterLoadingMessages)),
+    );
```
disconcision, line 153: Why is it necessary to run an effect here at startup?

src/web/app/helpful-assistant/prompts/InitPrompts.re
```diff
@@ -0,0 +1,93 @@
+open Util;
+
+let get_documentation_as_text = () => {
+  let (_, slides) = ScratchMode.StoreDocumentation.load();
+  let documentation =
```
disconcision, line 5: feels a bit weird to do a load from disk here. can't we get this from the main model? nbd tho

disconcision (changes requested):
Code looking good. Did some cleanup myself and a few style tweaks. There'a couple things I'm wondering about.

Let's change the PR title before merge to more reflect what's actually usable at this time.

Functionally there are a couple issues that should be fixed before merge, and some others that we should either fix now or create issues for after merge; see followup comments.

Should be able to merge before the meeting this week if these can be adressed.

disconcision:
A few functional things:

Breaks existing functionality:
1. The editor startup time has regressed by 3-5x (try loading https://hazel.org/build/task-completion-assistant/ vs https://hazel.org/build/dev/). Not sure why but one hint I can find is in the console. The line "Warning - LivelitProj.get: Not an InfoExp term" repeats 6x in this branch on startup as opposed to 2x on dev, indicating to me that documentation editor loading is happening 3 times.   

Major feature functionality:
1. Whenever I get a completion (tried several models), it seems to insert a ton of whitespace after; see here both in the editor (knocking the code off the screen) and in the console:
Screenshot 2025-06-29 at 1 32 48 AM

2. Submitting any completion request results in dozens of these in the console:
Screenshot 2025-06-29 at 1 33 05 AM

3. There's leading space getting consistently inserted before completions... either figure out if that's coming from somewhere internal, or probably actually just trim leading whitespace before inserting completions...

4. Completions seems to be displacing error holes somehow? Like note below the the three line completion seems to be displacing following error holes by three lines... mysterious.
Screenshot 2025-06-29 at 3 03 45 AM

(also, above is an example of trailing whitespace insertion... curious if this is coming from somewhere internal... if so, trim. if it's coming from the model, it might be because we have a bunch of whitespace in a prompt or something, since it seems so consistent)


Minor feature functionality (don't have to fix this time but should create issues for if merged):
1. Picking any Anthropic model triggers some kind of API error about tool support or something. Not a huge deal but we should fix at some point ( I miss Claude)
2. Entering the second `?` in `??` cause a half-second lag in editor responsiveness... curious why this is. Should dispatch message async. Not a huge deal but we should figure out where this perf hit is coming from.
3. Weren't we using code editors to display code in the sidebar as opposed to just text? I might be misremembering:
Screenshot 2025-06-29 at 1 37 04 AM

4. `??` typed in Exercises mode still gets same decoration even though it's (intentionally?) non functional? Perhaps should have some kind of easter egg...

russell-rozenbaum:
> 2\. Entering the second `?` in `??` cause a half-second lag in editor responsiveness... curious why this is. Should dispatch message async. Not a huge deal but we should figure out where this perf hit is coming from.

This comes from creating a new chat. You can see this performance bottleneck when creating a new chat in any of the three modes. 

In particular, following the execution path, when a new chat is made we call _init\_chat_  which calls _mk\_message\_display_ which calls _parse\_blocks_, and it's here in parse\_blocks where performance takes a hit. It is using regexp to parse out each code block. This doesn't occur in normal back and forth chatting, but due to the quantity of example code snippets in our prompts, this does occur for initial prompt parsing. 

A simple fix would probably be storing a serialized version of these parsed initial messages somehow. Although let me know if there are other ideas.

russell-rozenbaum:
> 3\. Weren't we using code editors to display code in the sidebar as opposed to just text? I might be misremembering:

Looks like we might not be properly prompting the agent for '??' completions. The "content" field in the json payload shows the agent isn't wrapping their response in triple backticks.

russell-rozenbaum:
> 1. Whenever I get a completion (tried several models), it seems to insert a ton of whitespace after; see here both in the editor (knocking the code off the screen) and in the console

Yea, we talked about this at one of the meetings, and we're not sure as to what's going on here. I think our best guess was that the opaque suggestion is provided by the backpack doesn't format properly, though I'm still understanding the framework for editors, zippers, terms, tiles, etc, so I'm not entirely sure what's happening here.

russell-rozenbaum:
Todos
- [ ] Allow for copying from code blocks
- [ ] Implement streaming and/or reintroduce loading dots. O.w. user does not know if LLM is responding.
- [x] Move suggestion prompt creation from ChatLSP to AssistantModel (into init_prompt_data)
- [ ] (Optional) Allow for copying from API key field

disconcision:
> Yea, we talked about this at one of the meetings, and we're not sure as to what's going on here. I think our best guess was that the opaque suggestion is provided by the backpack doesn't format properly, though I'm still understanding the framework for editors, zippers, terms, tiles, etc, so I'm not entirely sure what's happening here.

is the whitespace in the received completion from the API, or is it something the editor is doing? in any case, let's just trim leading and trailing whitespace from completions.

disconcision:
> A simple fix would probably be storing a serialized version of these parsed initial messages somehow. Although let me know if there are other ideas.

not sure about best approach, would need to hear a bit more about the internals which i haven't read that closely. in principle i feel it's not optimal to have code completion speed depend on stuff going on in the chat, as for regular completions the chat will often be closed. so if there's a way to take chat UI stuff off the critical path for completions that's probably the way to go, but pre-serialization also sounds reasonable i think? should be at compile time though

russell-rozenbaum:
> > Yea, we talked about this at one of the meetings, and we're not sure as to what's going on here. I think our best guess was that the opaque suggestion is provided by the backpack doesn't format properly, though I'm still understanding the framework for editors, zippers, terms, tiles, etc, so I'm not entirely sure what's happening here.
> 
> is the whitespace in the received completion from the API, or is it something the editor is doing? in any case, let's just trim leading and trailing whitespace from completions.

It comes from the editor. The LLM will insert appropriate line breaks and spaces for nice formatting. You can see that everything is formatted well when accepting those chaotic suggestions.

disconcision:
I confirmed that the slow startup time is coming from the same parse_blocks cause.

dev startup graph:

Screenshot 2025-06-29 at 3 43 33 PM

this branch startup graph:

Screenshot 2025-06-29 at 3 43 50 PM

I think that regardless of whether we pre-serialize we ideally want to find a way to make sure these kinds of calls don't happen if the chat isn't open to avoid the possibility of these kinds of regressions

russell-rozenbaum:
> I confirmed that the slow startup time is coming from the same parse_blocks cause.
> 
> dev startup graph:
> 
> Screenshot 2025-06-29 at 3 43 33 PM
> this branch startup graph:
> 
> Screenshot 2025-06-29 at 3 43 50 PM
> I think that regardless of whether we pre-serialize we ideally want to find a way to make sure these kinds of calls don't happen if the chat isn't open to avoid the possibility of these kinds of regressions

Is this suggesting we don't want to store them on initialization of the model as is done here https://github.com/hazelgrove/hazel/pull/1575/commits/6826547b7ae769c4fd2dff8ef9b6f0cbc5becb29 ?

We will inevitably need to parse blocks at least once at some point, currently we do it on initial startup, before we did it each time a new chat was created. We could do it instead on the first use case of a user typing '??'/'?a' or opening tutor mode/composition mode. But this again comes with the consequence of being slow when the user actually wants to use it.

We could also look into improving the runtime of parse blocks itself, I'd need to look into the current regexp to see how feasible this is.

disconcision:
Doing it when the chat sidebar is opened for the first time rather than editor startup is my first-order suggestion. In general the idea here is that things the user might not use at all shouldn't slow down initial load if possible. Additionally, is this process on the critical path for generating completions as opposed to using the chat sidebar? it's not obvious to me that getting completions should require this parsing step.

> We could also look into improving the runtime of parse blocks itself, I'd need to look into the current regexp to see how feasible this is.

I see you're using Str.Regexp; use the js regexp instead, see StringUtil, it can be an order of magnitude faster

russell-rozenbaum:
> Additionally, is this process on the critical path for generating completions as opposed to using the chat sidebar? it's not obvious to me that getting completions should require this parsing step.

Right, this step is only for the display present in the sidebar. We send a raw string message to the model. However, for the sake of guaranteeing consistency between what is sent to the model and what is displayed to the user, the creation of the string messages to be sent and the creation of the viewable UI message displays are always done jointly. I think moving this process off the critical path loses this guarantee.

disconcision:
I looked into the extra whitespace issue and I think what is happening is that the layout system assumes the buffer has no linebreaks, so the caret is placed where it would be if that is true. the same issue is probably causing the error holes to be displaced. i though this wasn't an issue before though so i'm wondering what might have changed?

> Right, this step is only for the display present in the sidebar. We send a raw string message to the model. However, for the sake of guaranteeing consistency between what is sent to the model and what is displayed to the user, the creation of the string messages to be sent and the creation of the viewable UI message displays are always done jointly. I think moving this process off the critical path loses this guarantee.

don't know enough about the current implementation to know what to advise but i think the bottom line here is that processes which are only relevant to the chat bar view shouldn't slow things down when the chat bar isn't being used. if the current architecture makes that difficult i think we need to revisit it

russell-rozenbaum:
> don't know enough about the current implementation to know what to advise, but I think the bottom line here is that processes which are only relevant to the chat bar view shouldn't slow things down when the chat bar isn't being used. if the current architecture makes that difficult i think we need to revisit it

I've imagined generating completions non-separable from using the sidebar. 

There are two versions of a message: 
1. The openrouter one, which gets sent to the model (this is absolutely necessary for completions) 
2. The displayed version, which is what parse blocks parse code blocks out of

When a user presses enter or the send message button -> the string is created into **both** versions of a message -> the OpenRouter message is sent to the LLM _and_ the display version is concatenated to a list of other display messages to be displayed in the chat interface.


We can set flags up for when the user initially opens the assistant sidebar. They will need to open it before requesting completions, as they'll need to set an API key. Also, hopefully that suggestion of using js regexp and StringUtil rather than Str regexp will help, though I'll need to study how js regexp works and how I can copy the implementation over.
Other than this, we could probably have the displays only be parsed when a chat is opened, but this would be a more involved refactoring.

disconcision:
@russell-rozenbaum @CyrusD123 
I've pushed a modified version of the completion buffer which solves the decoration issues above, and some other buffer aesthetic issues, like wonky formatting and the trailing hole. This new completion style works more akin to a regular selection; it doesn't contain unparsed text inside, but a fully parsed hazel code segment.

It looks like this now:

Screenshot 2025-06-29 at 5 46 55 PM

(Not married to the style; both the tokens contained within the buffer and the buffer backing can be styled; here, I'm using a modified version of the selection deco for the backing, and a 50% opacity filter for the tokens)

This is a pretty different approach to the buffer though so I'd appreciate it if you guys could play with it and see if you find any bugs or other undesirable behaviors. One way in which this might be different is that insertion will fail silently if the completion is too badly borked (see set_llm_buffer; it first tries to parse the completion individually, and then it checks if the resulting backpack is empty (so no missing delimiters); if either of these fails, it fails silently. You guys might be doing similar checks upstream already, I'm not sure.

disconcision:
After recent changes the completion system prompt now seems to be missing in the chat and in the api requests

russell-rozenbaum:
> This is a pretty different approach to the buffer though so I'd appreciate it if you guys could play with it and see if you find any bugs or other undesirable behaviors.

I notice we can't undo to get back suggestions anymore, where previously we could. Not sure how important this is, though.

russell-rozenbaum:
> Doing it when the chat sidebar is opened for the first time rather than editor startup is my first-order suggestion.

https://github.com/hazelgrove/hazel/pull/1575/commits/03ad9522f3fa4e0bcd7da42da53fba5bc0f7e775 should offer somewhat of a fix for this.

disconcision:
@russell-rozenbaum I'm still getting that slowdown after entering the second ?. in the below flame graph, the initial small spike is the first ?, and the second is the second ?, which takes over a second. as can be seen below, almost this entire time is spent in parse_blocks. not an absolute dealbreaker for merging but we should definitely prioritize fixing this

Screenshot 2025-07-01 at 8 41 36 PM

russell-rozenbaum:
> @russell-rozenbaum I'm still getting that slowdown after entering the second ?. in the below flame graph, the initial small spike is the first ?, and the second is the second ?, which takes over a second. as can be seen below, almost this entire time is spent in parse_blocks. not an absolute dealbreaker for merging but we should definitely prioritize fixing this
> 
> 
> 
> Screenshot 2025-07-01 at 8 41 36 PM
> 
> 

Interesting. Have you hard reset Hazel? Does that have any information as to what function call might causing that?

disconcision:
I have hard reset, yes. So there's no latency when pressing the second ? on your machine?

Unrelatedly: I'm a bit confused about the console output during a completion:

Screenshot 2025-07-01 at 9 31 06 PM

First, minor issues: i'm confused about what the regexp one is saying? and also why there's an error round one, as there was no error round? at least i don't think? relatedly and most importantly, what's that second API response at the end? is there another round happening here? there always seems to be a second response.

Perhaps relatedly, what is the stuff called 'resuggest' in the code?

disconcision:
re what's being called, it's AssistantUpdate.update calling mk_message_display calling parse_blocks:

Screenshot 2025-07-01 at 9 38 41 PM

it may be other things too but that's the first call i think

disconcision:
@russell-rozenbaum I investigated a bit. The slow part is the zipper_of_string call... this is unfortunately a very expensive call, we need to be careful where we use this. Right now, on every completion request, it is calling this on the program sketch, which is currently the whole program. This can be very slow. Ideally we could memoize it but it looks like it's running on the sketch after the llm hole has been replaced by the numbered version, so it's different every time. Seeing as this is for view formatting purposes only, for now let's just skip rendering these things nicely, or just take the zipper/segment of the editor and use it without reparsing it from its string version.

For the sake of expidiency for this PR, do you mind if I just bypass parse_blocks for now? i.e.:
Screenshot 2025-07-01 at 9 59 04 PM

I'm not sure if that has any unintended consequences; seems to work fine for chat. I pushed that change for now

russell-rozenbaum:
> I have hard reset, yes. So there's no latency when pressing the second ? on your machine?

Yea, I don't get much latency, if any, when using the task-completion-assistant branch. It's maybe like .05 seconds. Given you found it's from the zipper_of_string though, this might be cause I've been using fairly small sketches.

russell-rozenbaum:
> relatedly and most importantly, what's that second API response at the end? is there another round happening here? there always seems to be a second response.

This the API request sent to request a summarization title for the chat. See create_chat_descriptor in AssistantUpdate.re. We only do this after the first few chat exchanges... given we make a new suggestion mode chat every time now, this does make things doubly expensive.

russell-rozenbaum:
> i'm confused about what the regexp one is saying?

I believe this is just cause we expect the agent's response to be wrapped in triple backticks, but it seems like we don't prompt it with that. I think _maybe_ if you use '?a' the regex match should succeed since we almost certainly have it wrap it in triple backticks for that prompt... not sure what happened to the '??' prompt, I thought we prompted it to do that.

russell-rozenbaum:
> and also why there's an error round one, as there was no error round?

We schedule an error round action by default. If there were no errors in the response, then we don't actually send an API request to the LLM, instead we stop looping and return the result.

russell-rozenbaum:
> Perhaps relatedly, what is the stuff called 'resuggest' in the code?

Resuggest was an idea I had for allowing a user to get back the suggestion at a specific tile. This felt wonky to use, and only worked if the user didn't change the tile. It's since been removed from actually being used for the most part... could probably clean up that artifact.

disconcision:
> Yea, I don't get much latency, if any, when using the task-completion-assistant branch. It's maybe like .05 seconds. Given you found it's from the zipper_of_string though, this might be cause I've been using fairly small sketches.

ah yes that's definitely it. I was using a two-page program (the basic reference)

> Resuggest was an idea I had for allowing a user to get back the suggestion at a specific tile. This felt wonky to use, and only worked if the user didn't change the tile. It's since been removed from actually being used for the most part... could probably clean up that artifact.

might not be a bad idea to do a cruft pass if you have the time before tomorrow late afternoon... otherwise we'll fix it in post

russell-rozenbaum:
> @russell-rozenbaum I investigated a bit. The slow part is the zipper_of_string call... this is unfortunately a very expensive call, we need to be careful where we use this. Right now, on every completion request, it is calling this on the program sketch, which is currently the whole program. This can be very slow. Ideally we could memoize it but it looks like it's running on the sketch after the llm hole has been replaced by the numbered version, so it's different every time. Seeing as this is for view formatting purposes only, for now let's just skip rendering these things nicely, or just take the zipper/segment of the editor and use it without reparsing it from its string version.
> 
> For the sake of expidiency for this PR, do you mind if I just bypass parse_blocks for now? i.e.: Screenshot 2025-07-01 at 9 59 04 PM
> 
> I'm not sure if that has any unintended consequences; seems to work fine for chat. I pushed that change for now


Do you have any ideas of a better approach than zipper_of_string? 

I'm okay with this temporary change for now... shouldn't have any consequences as these are only for the message display blocks. I'm wondering, since this only happens for larger programs (and the lag is almost unnoticeable for smaller programs, <250 for certain, from what I've tested), what if we did something like the following?

Screenshot 2025-07-01 at 11 14 34 PM

cyrus- (approved):

cyrus- (approved):

disconcision (approved):
lgtm

test/Test_Editing.re
```diff
@@ -80,7 +80,7 @@ let mv_r_token = (n: int): list(Action.t) =>
   List.init(n, _ => Action.Move(Local(Right(ByToken))));
 
 let mk = (init: string): list(Action.t) => {
-  /* This harness uses a ¦ to represent caret position.
+  /* This harness uses a  to represent caret position.
```
7h3kk1d, line 83: @disconcision is this an accidental change?

PR #1578, opened 2025-03-24
Improvements to stepper rewrite checker

nskh:
improvements to rewrite checker:

- [x] Don't handle strings
- [x] Unique identifiers for unknown terms
- [x] Handle `Casts` naively
- [ ] Think more about bugs arising from blindly printing the `exp` of a `Cast`
- [ ] Handle `Nat`s, once added
- [x] `UnOp` handling
- [x] Rewrite `**` to `^` for `Power` in Algebrite
- [ ] Type-check incoming `Exp`s a little bit when the rewrite checker is invoked
- [ ] Pass around `Some(Bool)` in case we ever can't handle some sort of expression

### DIFF
--- a/src/haz3lweb/app/editors/stepper/RewriteChecker.re
+++ b/src/haz3lweb/app/editors/stepper/RewriteChecker.re
@@ -4,9 +4,9 @@ open Js_of_ocaml;
 let rec print_exp_for_algebrite = (exp: Exp.t): string =>
   switch (exp.term) {
   | Int(value) => string_of_int(value)
+  // TODO Nats?
   | Float(value) => string_of_float(value)
   | Bool(value) => string_of_bool(value)
-  | String(value) => value
   | BinOp(op, exp_left, exp_right) =>
     "("
     ++ print_exp_for_algebrite(exp_left)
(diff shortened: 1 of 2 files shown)


src/haz3lweb/app/editors/stepper/RewriteChecker.re
```diff
@@ -15,14 +23,23 @@ let rec print_exp_for_algebrite = (exp: Exp.t): string =>
...
+  // TODO: think harder about weird corner cases where we'd want to ensure the types in Cast are valid
+  | Cast(exp, _, _) => print_exp_for_algebrite(exp)
+  // If we don't know how to print the expression, we can use a hash to create a unique string
+  // Modulo keeps it short
+  | _ => "unknown_" ++ string_of_int(Hashtbl.hash(exp) mod 100000)
   };
 
 let checkEquality = (expr1, expr2): bool => {
   let algebrite = Js.Unsafe.global##.Algebrite;
+  print_endline(expr1);
```
Negabinary: Do we still need these prints?

Negabinary (changes requested):

src/haz3lweb/app/editors/stepper/RewriteChecker.re
nskh: removed

PR #1582, opened 2025-03-28
[cleanup] Create interfaces for evaluator and evaluator step

Negabinary:
This pr creates some simpler interfaces for the evaluate and stepping functions in dynamics, to better contain some of the performance hacks (e.g. making sure duplicate ids don't escape the evaluator), along with cleaning up the directory structure of dynamics a little


src/haz3lcore/dynamics/Evaluator.re
```diff
@@ -110,36 +108,38 @@ let rec evaluate = (~in_closure=?, state, env, d) => {
...
-      print_endline("EXN:" ++ Printexc.to_string(exn));
-      ResultFail(UnknownException(Printexc.to_string(exn)));
-    | (state, result) =>
-      ResultOk({
-        result,
-        state,
-      })
-    }
-  };
+// let evaluate =
```
cyrus-: delete

cyrus- (approved):

PR #1590, opened 2025-04-02
Hot Module Refresh via Vite

gcrois:
Using `Vite`, this PR adds the ability to automatically reload a development window when .css (hot reload!) or compiled .js (not-quite-as-hot reload...) as is visible in the video below (if you squint)

https://github.com/user-attachments/assets/16501d1d-8055-41ed-8a9f-72fa86707353

To use it, run `npx vite` in a terminal other than `make watch`.


gcrois:
@7h3kk1d lmk if you can figure out why this crashes the test? Maybe because I changed `package.json` to `"type": "module"`

7h3kk1d:
> @7h3kk1d lmk if you can figure out why this crashes the test? Maybe because I changed `package.json` to `"type": "module"`

I'm surprised if that breaks it since that's just being used for the esbuild stuff and I wouldn't expect it to be used for tests. Are we moving the js files or just copying them with vite?

7h3kk1d:
@gcrois I looked into this locally. We're not changing what js is generated but you are right that it's the module configuration in package.json that's breaking it as the js_of_ocaml code being generated isn't ECMAScript Module compatible. I think this is the related issue https://github.com/ocsigen/js_of_ocaml/issues/551 on the js of ocaml side. I'm not sure if there's another way to configure vite to be compatible without modules or we can look into other hot-reloading/hot-swapping solutions.

gcrois:
Fixed by renaming `vite.config.ts` to `vite.config.mts` 🥴🥴🥴

PR #1594, opened 2025-04-08
add forall explicit and succint syntax to test parser

joseemds:


cyrus-:
We would need to add this to the main tylr parser as well to justify including it in the test parser

PR #1600, opened 2025-04-10
Fix menhir ast generator sizes

7h3kk1d:
I made mistakes with renaming the size param in sub-patterns and accidentally referencing the outer one. So this is a bunch of fixes so our expressions are actually the right size.

### DIFF
--- a/src/haz3lmenhir/AST.re
+++ b/src/haz3lmenhir/AST.re
@@ -304,6 +304,7 @@ let rec gen_exp_sized = (n: int): QCheck.Gen.t(exp) =>
         pure(EmptyHole),
         pure(TupleExp([])),
         pure(ListExp([])),
+        map(x => Constructor(x, None), gen_constructor_ident),
       ]);
     fix(
       (self: int => t(exp), n) => {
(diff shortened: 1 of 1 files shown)


src/haz3lmenhir/AST.re
```diff
@@ -304,6 +304,7 @@ let rec gen_exp_sized = (n: int): QCheck.Gen.t(exp) =>
         pure(EmptyHole),
         pure(TupleExp([])),
         pure(ListExp([])),
+        map(x => Constructor(x, None), gen_constructor_ident),
```
7h3kk1d, line 307: Constructors can be leaf nodes of size 1.

src/haz3lmenhir/AST.re
```diff
@@ -314,23 +315,22 @@ let rec gen_exp_sized = (n: int): QCheck.Gen.t(exp) =>
...
               let* sizes = gen_non_singleton_array(n);
               let+ exps =
                 flatten_a(
                   Array.map(
-                    (size: int) =>
+                    (n: int) =>
                       oneof([
                         {
                           let* l = gen_label;
                           let+ e = self(n - 1);
```
7h3kk1d, line 330: This was the bug where self was using n but the param was called size.

PR #1602, opened 2025-04-10
Fix CI fatal warnings

### LINKED ISSUE #1601: CI isn't failing for unused variable errors
The dune-workspace flags for the release profile seem broken and it isn't catching unused variables anymore. It's still being caught by the tests but doesn't fail the build https://github.com/hazelgrove/hazel/actions/runs/14364361051/job/40273552148#step:15:57

7h3kk1d:
Resolves https://github.com/hazelgrove/hazel/issues/1601

I just inlined all of the warnings we have configured in development (the dune default). By default dune has [minimal warnings](https://github.com/ocaml/dune/blob/897bd784ba2877388001f5e88bd873449d9e83e1/src/dune_rules/ocaml_flags.ml#L50) enabled for release. This should at least make it to where anything failing locally fails on release. It may also be worth it to us to go through and pick an intentional set of warnings/errors that work for us.

P.S. I can also suppress the warning 58 (no cmx file) that happens on every build if we want.

### DIFF
--- a/src/haz3lcore/lang/term/Typ.re
+++ b/src/haz3lcore/lang/term/Typ.re
@@ -562,7 +562,7 @@ let rec get_sum_constructors = (ctx: Ctx.t, ty: t): option(sum_map) => {
        the below code will be incorrect! */
     let ty =
       switch (ty |> term_of) {
-      | Rec({term: Var(x), _}, ty_body) =>
+      | Rec({term: Var(x), _}, _ty_body) =>
         switch (Ctx.lookup_alias(ctx, x)) {
         | None => unroll(ty)
         | Some(_) => unroll(ty)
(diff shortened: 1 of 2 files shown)


src/haz3lcore/lang/term/Typ.re
```diff
@@ -562,7 +562,7 @@ let rec get_sum_constructors = (ctx: Ctx.t, ty: t): option(sum_map) => {
        the below code will be incorrect! */
     let ty =
       switch (ty |> term_of) {
-      | Rec({term: Var(x), _}, ty_body) =>
+      | Rec({term: Var(x), _}, _ty_body) =>
```
7h3kk1d, line 565: @Negabinary this is the case that I caught it for. Lmk if this is actually a bug.

PR #1638, opened 2025-04-29
HTTPS link for Algebrite

nskh:
Rewrite checker was broken due to loading an unencrypted HTTP link; I updated the algebrite link to an HTTPS one.


7h3kk1d:
@nksh you can add algebrite to https://github.com/hazelgrove/hazel/blob/dev/package.json and https://github.com/hazelgrove/hazel/blob/dev/src/haz3lweb/www/prebundle.js and it will get bundled into the bundle.js file. Then you don't need to load it from the cdn. lmk if you need any help.

nskh:
@7h3kk1d Oh that's way better - I'll do that instead.

cyrus-:
@nskh lots of conflicts, I think primarily due to your refmt version being different from the latest deps?

nskh:
@cyrus- looks like these changes actually were already made, I'll close this PR.

PR #1650, opened 2025-05-13
TDOS

### LINKED ISSUE #1588: Evaluator unboxing errors
As part of https://github.com/hazelgrove/hazel/tree/cat-resiliency I've mostly found unboxing bugs

# Cosntructor bugs
Image

Image

Image

Image

# Unquote bugs

Image

Image

# List bugs
"":: (type x = + A in A) 

Image

### LINKED ISSUE #1625: Type application in dynamics error happening in elaboration and dynamics
`let [(A: (Bool(Bool))), (_: (String))] = 0 in (())` breaks elaboration

`(fix(A: ((+A) -> Bool(+A))) ->  )(?)` breaks dynamics

7h3kk1d:
# Type-Directed Operational Semantics

This PR removes cast insertion from Elaboration and instead gives a direct operational semantics to the type ascriptions that are present in the syntax. Historic unboxing failures, which resulted from incorrect cast insertions, are now indeterminate. Expressions are also assumed to reduce to final forms before surrounding ascriptions are passed through using the existing `Cast.transition` function. In a future PR the `Cast` constructor will be replaced with a `Ascription` constructor that only has 1 type. This has been left off to make the refactors easier to pass through in 2 stages.

## Diff
Line numbers in the PR are somewhat disingenuous from the test splitup.
`src` difference:  21 files changed, 489 insertions(+), 780 deletions(-)
`test` difference:  22 files changed, 1611 insertions(+), 1134 deletions(-)

# Differences from dev
### Constructor patterns
```hazel
let match :? = fun x->
  case x
    | (A : +A) => true
    | _ => false    
end in

(match(A :(+A +B +C)), match(A: (+A +B)))
```

This is stuck on the tdos branch. On dev it doesn't consider the type when matching patterns. I think this is better. 

```hazel
case Var("yo")
  | Var(n) => 1  
end
```

For unknown constructors dev behaves differently depending on if there's a payload or not. I'm just having both evaluate as a match for now. I think it would be better for this to evaluate as indet but I'm open to suggestions.


src/haz3lcore/dynamics/transition/Unboxing.re
```diff
@@ -93,27 +88,12 @@ let rec unbox: type a. (unbox_request(a), DHExp.t) => unboxed(a) =
...
-      | TupLabel(_, e) => unbox(request, Cast(e, e1, e2) |> DHExp.fresh)
-      | _ => unbox(request, Cast(e, e1, e2) |> DHExp.fresh)
-      }
-
     /* Base types are always already unboxed because of the ITCastID rule*/
     | (Atom(r), Atom(x)) =>
       switch (Atom.unbox(r, x)) {
       | Some(x) => Matches(x)
-      | None => DoesNotMatch
+      | None => IndetMatch
```
7h3kk1d, line 94: @Negabinary This seems more correct to me even absent the tdos approach. Not sure if you have thoughts

src/haz3lcore/lang/term/Grammar.re
```diff
@@ -646,6 +646,16 @@ module Factory = (DefaultAnnotation: DefaultAnnotation) => {
...
+    let asc = (~ann=?, e, t): exp_t(DefaultAnnotation.t) =>
+      cast(
+        ~ann?,
+        e,
+        t,
+        {
+          term: Unknown(Internal),
+          annotation: default_annotation(ann),
+        },
+      );
```
7h3kk1d: In a future pr we'll replace cast with ascription.

test/Test_Evaluator.re
7h3kk1d, line 1: Split up into multiple files

src/haz3lcore/dynamics/transition/Unboxing.re
Negabinary: yeah, fair enough

src/haz3lcore/dynamics/transition/Casts.re
```diff
@@ -1,181 +1,244 @@
...
-    }
-  | Sum(sm) =>
-    sm |> ConstructorMap.is_ground(is_hole)
-      ? Ground : NotGroundOrHole(Sum(grounded_Sum()) |> Typ.temp)
-  | Arrow(_, _) => grounded_Arrow
-  | Forall(_) => grounded_Forall
-  | List(_) => grounded_List
-  | Ap(_) => failwith("type application in dynamics")
-  };
-};
```
7h3kk1d, line 89: These are unused in Casts so have been moved to TypeAssignment.re

7h3kk1d:
Check on reverse application

disconcision:
thumbs_b_c_25db01058f7397bcd39013c4604a6054

PR #1659, opened 2025-05-21
Global undo

Negabinary:
This is a simpler implementation of global undo that only works for a single Bonsai component (the way Hazel is currently structured.)


Negabinary:
todo: fix the bug where it requires an extra ctrl+z

src/haz3lweb/app/editors/mode/ExercisesMode.re
```diff
@@ -188,6 +188,16 @@ module Update = {
     | ExportTransitionary
     | ExportGrading;
 
+  let can_undo = (action: t) => {
+    switch (action) {
+    | SwitchExercise(_) => true
```
cyrus-: probably false here

src/haz3lweb/view/ScratchMode.re
```diff
@@ -89,6 +89,18 @@ module Update = {
     | Export
     | Encode;
 
+  let can_undo = (action: t) => {
+    switch (action) {
+    | CellAction(action) => CellEditor.Update.can_undo(action)
+    | SwitchSlide(_) => true
```
cyrus-: probably false

cyrus- (changes requested):

PR #1660, opened 2025-05-22
Splices

Negabinary:
This branch is a major rewrite of projectors, it also implicates most of the files in tylr.

**High-level approach:**

We begin conceptually separating tylr code and projector code, tylr code is now parameterized over projectors, and projector code is now parameterized over tylr. Instead of projectors simply covering up a piece, they can now have whatever model they want, the only obligation is that they must be able to make a term.

**Todos:**

- Core functionality
  - [x] Get cursor info working again for splices
  - [x] Fix projector serialization
  - [x] scrolls when you use the arrow keys
  - [x] cursor getting stuck in between terms (not checking for ancestors)

 - Serialization
   - [x] (A) Come up with a backup text for projectors

- Project/unproject
  - [ ] (T) Generalized project
  - [ ] (T) Unproject method
     - [ ] (T) ISSUE: unprojecting pair proj loses projectors in splices
     - [ ] (T) ISSUE: unprojecting (say) a fold or probe uses mk_term and hence loses formatting
  - [ ] (A) Smarter logic to reproject using the panel (see ProjectorCursor; use cached term?)
  - [ ] (A) Reinstate livelit autoprojection (see Insert.re and possibly elsewhere?)
  - [ ] (A) Unproject all projectors on backspace

- Restore Existing Projector functionality
  - [ ] (A) BUG: Resizing cells doesn't work in probes (by keyboard or by dragging)
  - [ ] (A) BUG: Style for error holes inside probe/info/fold projectors
  - [ ] (A) BUG: projector error outlines don't show up (for non text-area projectors)
  - [ ] (A) Probe/Type: Abbreviate syntax view
  - [ ] (M) Probes don't have dynamics on first load

- Splices
  - [ ] (M) Selection bug
  - [ ] (M) BUG: can't double click to select in splices
  - [ ] (A) Add test checkmarks in splices
  - [ ] (A) Unprojecting pair has weird behavior (try to arrow through the result). Probably due to UUID duplication

 - Helper Functions
   - [ ] (M) Create an editor manager component that can be reused in stepper (and possibly restore stepper statics)
   - [ ] (M) Package editor methods etc in meaningful bundles for projector methods

- Cleanup 
  - [ ] Bundle projector methods, especially those passed to Perform.go
  - [ ] Rename Haz3lcorep to something that makes more sense
  - [x] Add editor length method, possibly replacing editor string method
  - [ ] (A) EditorView.Focus.applicable_projectors
  - [ ] (A) Different action category or something for unproject (see ProjectorPanel)

- After this PR
   - Fixed size splices e.g. single-line
   - Make it easier to thread arguments to cursor_info. use case: route focus method to cursor_info so that can make the projector panel toggle return the focus to the editor pane
   - Figure out where offside views in splices should go (currently they overlap the parent)
   - Restore ClipboardCache (Currently broken in dev so not a priority)

- Done (Andrew)
  - [x] Reinstate info arg to view including sort
  - [x] BUG: Probe: Immediate unhandled exception (Util__Calc.PendingValue)
  - [x] Projector keyboard shortcuts
  - [x] Projector keyboard shortcut displays on hover
  - [x] BUG: Cards: Doesn't work for patterns 
  - [x] Fold: Exp hardcoded. Need sort in views.
  - [x] comment in ProjectorCursor about "different sorts"
  - [x] Add term to view info 
  - [x] Probe/Info: syntax view css is broken 
  - [x] BUG: arrow key movement gets stuck in bidelimited contexts sometimes
  - [x] BUG: Textarea linebreaks don't work
  - [x] BUG: Textarea arrow keys scroll
  - [x] BUG: For non-focussing projectors, clicks should move parent splice caret to projector
  - [x] BUG: editor caret shows up when textarea caret is active
  - [x] BUG: projector panel options don't show up on non-convex tile positions
  - [x] Make focus decorations nicer. Both the outlines indicating what is in focus, and whether the editor caret is shown
  - [x] Probe: Maketerm and several view things need to be redone
  - [x] BUG: No keyboard control for probes

- Handoffs
    - [x] out of a projector
    - [x] into a projector
    - [x] out of a splice
    - [x] into a splice


gcrois:
@Negabinary and I merged livelits in! Right now, livelits follow the new projector model, but are not yet able to incorporate their own splices.

We need to add back the editor shortcuts, automatically projecting a livelit with its default value after writing its name, and unprojecting it on backspace.

https://github.com/hazelgrove/hazel/blob/cba7fd955a869427dad7e571583f80671c78e83e/src/haz3lcore/zipper/action/Insert.re#L227C1-L227C32

https://github.com/hazelgrove/hazel/blob/cba7fd955a869427dad7e571583f80671c78e83e/src/haz3lcore/zipper/Zipper.re#L392

We are discussing how a runtime-defined livelit might designate that it needs slices. It is untenable to express the full editor state as a hazel term.

I proposed that we might have a `ref` livelit (think of a symbol in js https://developer.mozilla.org/en-US/docs/Glossary/Symbol).

Matt proposed an atomic datatype that represents a hazel editor state.

I think these ideas are compatible!

cyrus-:
@gcrois take another look at the livelits paper for a refresher on how we did splices there (using refs in a monad)

src/haz3lcore/dune
```diff
@@ -1,7 +1,7 @@
 (include_subdirs unqualified)
 
 (library
- (name haz3lcore)
+ (name haz3lcorep)
```
7h3kk1d, line 4: typo?

Negabinary: I wish

Hopefully we'll blow away this name scheme when we move files at the end though.

PR #1668, opened 2025-05-28
Type Preservation Property Based Tests and Fixes

7h3kk1d:
### Summary

This PR introduces a number of fixes to ensure **type preservation** across evaluation.  
It also includes a **disabled property-based test** for preservation, which will be enabled once remaining bugs are resolved.

---

### 🔒 Type Preservation

Implements a modified version of type preservation:

> If `Γ ⊢ e ⇒ t` and `e → e'`, then `Γ ⊢ e' ⇒ t'` where `t' ⊑ t`,  
> with `⊑` being the precision relation on gradual types,  
> and therefore `t' ~ t`.

---

### 🐛 Outstanding Bugs (Blocking the test)

The following bugs still violate preservation and prevent the test from being enabled:

- **[List concatenation with bad constructors (#1670)](https://github.com/hazelgrove/hazel/issues/1670)**  

- **[Nested ascriptions in fixpoint patterns (#1672)](https://github.com/hazelgrove/hazel/issues/1672)**  
  - e.g. `fix a : Int : Bool -> true` → `true`  
  - We're currently fixing preservation by just adding a cast to Bool on the resulting expression but the nested type annotations are ignored in dynamics

- **[Inexhaustive `let` pattern falls back to unknown type (#1671)](https://github.com/hazelgrove/hazel/issues/1671)**  
  - Example:
    ```hazel
    let [A, false] = if false then false else "" in (()) ⇒ ()
    ```
    After evaluation:
    ```hazel
    let [A, false] = "" in () ⇒ ?
    ```

---

### ✅ Fixes and Improvements

- 🛠️ **Short-circuiting boolean operators now preserve type**  
  - Before: `false || 7` → `7`  
  - Now: → `7 : Bool`

- 🛠️ **Ascriptions now properly propagate through:**  
  - Short-circuiting operators  
  - `test` expressions  
  - List literals like `[1, 2, "": ?]` now yield `[1, 2, "": Int]`
- 🛠️ **Fixed list unboxing for fixed-size `cons`**

- ✅ **Improved list `cons` typing**  
  - Before: `1 :: ["hello"]` ⇒ `[Int]`, ignoring the type of `"hello"`  
  - Now: → `[1, "hello" : Int]`  
  - 🚧 Future improvement: consider analyzing both elements to infer a more consistent element type

- ✅ **Fixpoint behavior clarified**  
  - `fix () -> []` now reduces correctly  
  - `fix () : ? -> []` is stuck, reflecting correct behavior when ascriptions are in the function position


src/haz3lcore/dynamics/transition/Ascriptions.re
```diff
@@ -139,7 +140,6 @@ let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
     | (Constructor(_, Some(Some(t))), t')
         when Typ.is_consistent(Ctx.empty, Typ.unroll(t), t' |> Typ.temp) =>
       Some(e)
-    | (Test(_), Prod([])) => Some(d)
```
7h3kk1d: This could just be Some(e)

src/haz3lcore/dynamics/transition/Transition.re
```diff
@@ -271,61 +271,49 @@ module Transition = (EV: EV_MODE) => {
...
+    | FixF({term: Asc(p, t), _}, d1, env) =>
+      // TODO Decide if this is the right approach
+      let. _ =
+        otherwise(env |> Option.value(~default=ClosureEnvironment.empty), d);
+      Step({
+        expr: FixF(p, Asc(d1, t) |> DHExp.fresh, env) |> rewrap,
+        state_update,
+        kind: FixUnwrap,
+        is_value: false,
+      });
```
7h3kk1d: I don't think this is the right approach alone. It's contingent on how we do statics for fixpoints.

If we get the type for a fixpoint from the pattern then `fix a: Int : String -> ?` -> `fix a : Int -> ? : String`. Which means we have something that was a string is now an int violating preservation. So we either want to use the expression inside the  fixpoint to get the type or we need some form of eager pattern matching that results in something of a consistent type. Since ascriptions come off of patterns in reverse the type that it fundamentally has is the most centered one.

src/haz3lcore/dynamics/transition/Transition.re
```diff
@@ -271,61 +271,49 @@ module Transition = (EV: EV_MODE) => {
...
-              (env |> Option.value(~default=ClosureEnvironment.empty)).
-                call_stack,
-            Environment.singleton((f, FixF(dp, d1, env) |> rewrap)),
-            env |> Option.value(~default=ClosureEnvironment.empty),
+          VarBstMap.Ordered.mapo(
+            ((p, exp)) =>
+              FixF(
+                Var(p) |> Pat.fresh,
+                exp,
+                Some(ClosureEnvironment.of_environment(env')),
```
7h3kk1d: I need to figure out what the correct closure environment here is.

src/haz3lcore/dynamics/transition/Unboxing.re
```diff
@@ -97,8 +97,10 @@ let rec unbox: type a. (unbox_request(a), DHExp.t) => unboxed(a) =
     | (ListLitn(_), ListLit(_)) => DoesNotMatch
     /* A cons final form is always indet, so either does NOT match or indet matches with a listliteral*/
     | (ListLitn(0), Cons(_)) => DoesNotMatch // Cons is not an empty list
-    | (ListLitn(n), Cons(_, xs)) => unbox(ListLitn(n - 1), xs)
-    | (ListLit, Cons(_)) => IndetMatch // WIthout length of ListLit we cannot know
+    | (ListLitn(n), Cons(x, xs)) =>
+      let* tail = unbox(ListLitn(n - 1), xs);
+      Matches([x, ...tail]);
+    | (ListLit, Cons(_)) => IndetMatch // Without length of ListLit we cannot know
```
7h3kk1d: I think this was just a bug. @MaxCarroll0 / @Negabinary can you confirm?

test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -175,11 +175,64 @@ let qcheck_pattern_equivalence_test =
...
+          let statics =
+            Statics.mk(CoreSettings.on, Builtins.ctx_init(Some(Int)), next);
+          let info: option(Info.t) =
+            Statics.Map.lookup(next.annotation.ids |> List.hd, statics);
+
+          info;
+        }
+      ) {
+      | Some(InfoExp({ty, _})) =>
+        let ret = Typ.is_consistent(Ctx.empty, ty, orig_ty);
```
7h3kk1d: Needs to also check that it's more precise.

src/haz3lcore/dynamics/transition/Transition.re
7h3kk1d: I'm leaning towards not modifying evaluation much and changing statics to mark the fixpoint as unknown with inconsistent pattern/expression like:
Screenshot From 2025-05-28 15-59-08

src/haz3lcore/dynamics/transition/Unboxing.re
MaxCarroll0: Yep this was just fixing a bug.

MaxCarroll0: But, when a cons expression has finished evaluating it is necessarily indeterminate (other it would be a list literal term?). Therefore it cannot match a listlit request, hence the original definition.

Did you find a situation that this reasoning was false?

MaxCarroll0: If not, we should probably keep the original to avoid accidentally matching invalid values (if the transition semantics gets bugged).

7h3kk1d: I had some case in this branch where it was causing an issue because the term inside the fixpoint wasn't fully reduced and then wasn't pulling the head element out with ListN because of
`     | (ListLitn(n), Cons(_, xs)) => unbox(ListLitn(n - 1), xs)`

I'll try to see if I can reproduce it but it might not be possible on this branch anymore.  But the above case is always wrong right since the request is returning n-1. If we don't unbox the full list we should just do indet.

7h3kk1d:
On the typejoin side I'm currently adding ascriptions for expressions that are joining types to the more precise one.

e.g. `[1,2, "hello" : ?]` currently becomes `[1, 2, "hello"]` but typechecks to `[Int]

So this will become `[1,2, "hello" : Int]`.

I wonder if it's better to just do typechecking like ts for lists/conditionals and give the least precise type rather than the most precise.

Screenshot From 2025-05-29 10-05-30

src/haz3lcore/dynamics/transition/Unboxing.re
MaxCarroll0: The reason it doesn't just return indet is that we still might be able to determine that it does not match concretely. For example, consider that `1::?` does not match `ListLitn(0)`

src/haz3lmenhir/Parser.mly
```diff
@@ -324,7 +321,8 @@ exp:
     | c = CONSTRUCTOR_IDENT { Constructor(c, None)}
     | c = CONSTRUCTOR_IDENT; SLASH_TILDE; { Constructor(c, Some(None)) } 
     | c = CONSTRUCTOR_IDENT; TILDE; t = typ;  { Constructor(c, Some(Some(t))) }
-    | c = CONSTRUCTOR_IDENT; COLON; t = typ;  { Asc(Constructor(c, None), t) }
+    // | c = CONSTRUCTOR_IDENT; COLON; t = typ;  { Asc(Constructor(c, None), t) }
```
7h3kk1d: Remove line

test/evaluator/Test_Evaluator_Fixpoint.re
```diff
@@ -0,0 +1,48 @@
...
+    test_case("Fixpoint pattern ascription", `Quick, () =>
+      parse_and_evaluate_test(
+        "() : Bool",
+        {|fix (_ : Bool) -> test false end|},
+      )
+    ),
+    test_case("Fixpoint with type ascription", `Quick, () =>
+      parse_and_evaluate_test("fix 0 -> () : Int", {|fix 0 -> ()|})
+    ),
+    test_case("Wild statics fixpoint list cons", `Quick, () =>
```
7h3kk1d: Rename

cyrus- (approved):

PR #1669, opened 2025-05-28
Language

Negabinary:
Replaces #1619 - needed to cherry-pick commits.

This is  breaking language out of core.

Language contains what you'd expect, I think. Borderline cases are things like Operators which are maybe more aware of the surface syntax than is strictly seemly.

Haz3lcore is a bit more fraught. Currently it contains:

tylrCore proper (Editor.re and dependencies)
maketerm + expToSeg (Interface between tylr and semantics)
TyDi (most logic could live in Semantics, but not sure if appropriate. I feel @7h3kk1d's Introduce is in a similar boat)
projectors


cyrus-:
@disconcision @Negabinary where are we at with this? ready to merge?

PR #1708, opened 2025-06-22
Add the ability to pass focus to an editor from the projector panel

disconcision:
@Negabinary to look at when you're back, probably... my thus far failed attempt to complete this threading... I keep on getting confused. Right now trying to deal with a bunch of errors in the various stepper files, increasingly unsure I'm on the right track with this....


Negabinary:
So I think you're on the right track - as currently implemented this will require a lot of work to thread all the arguments through the stepper. 

The threading is exactly the same as the threading required in the view function, which is making me wonder whether there's a way to combine them - like maybe the view function should just return a Node + a list of contextual actions?

PR #1712, opened 2025-06-23
Unprojection method (Non-compiling draft)

disconcision:
Some questions are coming up here.

The exact signature of this method is unclear to me. I think we basically need to return an editor, even though what we really want out is a segment. Not sure what it needs as input, besides the model. From working through the projectors, it seems like it needs sort as well, for at least the pair projector, unless we pass the sort through the pair projector model.

For the pair projector, it's not totally clear how to abstract the parts of unproject which are common with maketerm. It's not obvious to me how to just use maketerm in the unproject implementation, given the Calc abstraction? but presumably that's possible?

But if we wanted to do a real implementation in this case, I think we maybe need another Editor method, the opposite of Editor.split which I implemented to extract sub-editors. But the API in the other direction is slightly unclear. Morally we want to do something like: take a term-with-holes-with-ids, and a map of ids to editors, and make an editor by basically doing an ExpToSegment on the holey term, except when we hit subterms with the given ids, we substitute those with the corresponding things in the map of editors/segments. I guess we could just adapt ExpToSegment to support this functionality, but it all feels a bit loose to me. Like we can use our EmptyHoles for the 'holes' in the term-with-holes, but that feels slightly arbitrary.

### DIFF
--- a/src/haz3lcore/interface/ProjectorInterface.re
+++ b/src/haz3lcore/interface/ProjectorInterface.re
@@ -106,6 +106,8 @@ module type PROJECTOR = {
       model'
     ) =>
     View.t;
+
+  let unproject: model' => editor_model;
 };
 
 module Defaults = {
--- a/src/haz3lcore/projectors/Projector.re
+++ b/src/haz3lcore/projectors/Projector.re
@@ -316,6 +316,41 @@ let init =
   };
 };
 
+let unproject =
+    (
+      type ed,
+      type ed_a,
+      type ed_f,
+      ~editor_module,
+      model: model(ed, ed_a, ed_f),
+    )
+    : ed => {
+  let.gadt W(kind_gadt) = kind;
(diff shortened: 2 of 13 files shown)


Negabinary:
> I think we basically need to return an editor, even though what we really want out is a segment.

Agreed, I think we have to do this because segment is an implementation detail of editor that we're trying to hide.

> Not sure what it needs as input, besides the model.

I think model as input makes sense. We could also give it the cached maketerm, which I think maketerm should be guaranteed to have run before unproject so we can just give it an `Exp.t` instead of a `Calc.saved(Exp.t)`. This would include the sort as well.

> I think we maybe need another Editor method, the opposite of Editor.split

Yeah I agree with this. Since the term has already been calculated, I think we can just use the full term instead of any term-with-holes as input? There might be performance implications of this. The un-split function then would take the full term, and then a mapping from Ids to Editor.ts, and return a new editor.

PR #1714, opened 2025-06-24
fix list concatenation statics for free constructor

### LINKED ISSUE #1670: List concatenation synthesizes wrong type when arguments are bad constructors
Image

Right now

`A @ A` ⇒ `+A +?`. This should be synthesizing `[?] at least because list concatenation should return a list.

7h3kk1d:
fix #1670 

This was caused because regardless of mode free constructors return a syn type. So when analyzed against lists the join is a sum type. I resolved it by using the ana type unless it's in syn mode. I think this points to a larger issue about multiple errors where this should be inconsistent and free constructor.

This also reorganizes the statics tests to be similar to the evaluator tests


src/language/statics/Info.re
```diff
@@ -728,44 +728,54 @@ let fixed_typ_ok: ok_pat => Typ.t =
...
+    | NoType(FreeConstructor(c)) =>
+      if (Typ.is_syn_plus(ana)) {
+        Sum([
+          ConstructorMap.Variant(c, [Id.invalid], None),
+          ConstructorMap.BadEntry(Unknown(Internal) |> Typ.temp),
+        ])
+        |> Typ.temp;
+      } else {
+        ana;
+      }
```
7h3kk1d, line 738: This is the fix.

7h3kk1d: @disconcision or @Negabinary not sure if either of you have a better idea of how we should handle these errors and the ana type. My assumption is that the type on the info should always be consistent with the ana type.

7h3kk1d: For example this is still broken for free variables. `a @ a` => ? instead of [?]

src/language/statics/Info.re
```diff
@@ -739,45 +739,60 @@ let fixed_typ_ok: ok_pat => Typ.t =
...
+    switch (err) {
+    | NoType(FreeConstructor(c)) =>
+      if (Typ.is_syn_plus(ana)) {
+        Sum([
+          ConstructorMap.Variant(c, [Id.invalid], None),
+          ConstructorMap.BadEntry(Unknown(Internal) |> Typ.temp),
+        ])
+        |> Typ.temp;
+      } else {
+        ana;
```
7h3kk1d, line 737: I guess this could also be the join of the ana/sum if they're consistent :shrug:

cyrus- (approved):

PR #1722, opened 2025-06-26
Editing tests, Printer/Parser consolidate, nonconvex exception fix

### LINKED ISSUE #953: expansion fails on "if then|else"
This triggers a concavity exception, but I'm not sure why... not sure why it has to be well-formed before the standard remold/regrout pass. I had a hacky fix for this in, regrouting after doing expand/barfing the left neighbor, but I realized that causes barfing to fail in (more common) cases like "if a|then"; the then doesn't get barfed because a grout is inserted so its no longer the left neighbor. Not sure about the correct way to approach this.

### LINKED ISSUE #1088: Deleting the middle delimiter of `let = in` or `if then else` triggers Nonconvex Segment exception
`(Exception "Haz3lcore.Skel.Nonconvex_segment")`

This only seems to occur:
1. when I try to delete (only a) "=" or "then" starting from complete syntax; note that if I delete a selection also containing any other non-whitespace token (part of a different form) to the left or the right of the =/then, it works fine.
2.  when try to delete a selection containing both the first and second delimiter, e.g. "let ...=" or "if ... then"

I feel like there's one or more issues for this but I can't immediately find any.

@dm0n3y any feeling for how hard it would be to fix?

disconcision:
Closes #1088, closes #953.

This:
1. Adds a testing harness for editing tests, including some starter tests for Insert/Delete, which can be fleshed out more going forward. Tests are written entirely in concrete syntax, incorporating special characters for the caret, and for both kinds of holes.
3. Cleans up a longstanding mess of printer and parser functions. From now on, use `Printer.of_segment`, `Printer.of_zipper`, `Parser.to_segment`, and `Parser.to_zipper` for all your printing and parsing needs. This change was made here to support the new testing harness without creating yet another variant.
4. The original purpose of this branch: fixing for a very longstanding syntax bug. In particular, the following two situations should no longer trigger editor-crashing exceptions, and indeed should behave as expected:

1. Pressing delete on `let x =| 1 in x`
5. Pressing space on `if true then|else`

This is a complicated issue that touches a bunch of hacks, so I'd appreciate some functional review. As an example of some of the things that could have been impacted, and at times were during testing:

1. typing `let x = 1 in x` and `if true then 1 else 2` left-to-right should behave stable with respect to token movement. that is, although concave grout will appear during entry, it should be transmuted to/from whitespace when appropriate to maintain caret position
2. typing space on `if true|then` should work as expected 
6. typing space on `if|then` should work as expected
7. typing space on `if|if` should work as expected
8.  typing space on `[|]` and `(|)` and `case|end` and`use|in` should work as expected (only a hole inside, no addn whitespace)

For all of the above, entering an operator or symbolic delimiter instead of space should also do something reasonable (no exceptions, no weird space insertion). There are now tests for all of these and more


7h3kk1d:
I'm still seeing `split_kids: index out of bounds` errors but those might be unrelated to this fix.

https://github.com/hazelgrove/hazel/issues/1724

7h3kk1d:
This is a pretty weird way to edit and I'm not sure if it's related but it doesn't remold. 
[Screencast From 2025-06-27 10-28-20.webm](https://github.com/user-attachments/assets/fb52b215-cfbc-43ad-a5f8-d7d4593327dd)

src/haz3lcore/zipper/Printer.re
```diff
@@ -3,35 +3,39 @@ open Util.OptUtil.Syntax;
...
+  seg |> List.map(of_piece(~holes, ~concave_holes)) |> String.concat("")
+and of_piece = (~holes, ~concave_holes, p: Piece.t): string =>
   switch (p) {
-  | Tile(t) => of_tile(~holes, t)
+  | Tile(t) => of_tile(~holes, ~concave_holes, t)
+  | Grout({shape: Concave, _}) when concave_holes != None =>
+    Option.get(concave_holes)
   | Grout({shape: Concave, _}) => " "
   | Grout({shape: Convex, _}) when holes != None => Option.get(holes)
   | Grout({shape: Convex, _}) => " "
```
7h3kk1d: Is there a reason we make these options as opposed to just defaulting `concave_hole`/`hole` to `" "`

src/haz3lcore/zipper/action/Insert.re
```diff
@@ -93,11 +93,10 @@ let expand_neighbors_and_make_new_tile = (char: Token.t, state: t): option(t) =>
      The order here could be revisited if barfing was more sophisticated.
      */
   let* z = expand_or_barf_left_neighbor(state);
-  //let (z) = regrout(Left, z);
-  /* Note to david: I'm not sure why the above regrout is necessary.
-     Without it, there is a Nonconvex segment error thrown in exactly
-     one case, the double barf case: insert space on "if then|else" */
   let+ z = expand_or_barf_right_neighbor(z);
+  /* If contemplating changing regrouting behavior here, try these
+   * two cases: pressing space on `if then|else` and `if true|then` */
```
7h3kk1d: I don't understand what this comment is saying. Both cases seem to work afaict

test/Test_Editing.re
```diff
@@ -0,0 +1,237 @@
...
+  ),
+  test(
+    ~name="Insert char inside token",
+    ~acts=mk("fi¦me") @ [Insert("x")],
+    ~goal="fix¦me",
+  ),
+  test(
+    ~name="Inserting string quote inserts closing quote as well",
+    ~acts=mk("¦") @ [Insert("\"")],
+    ~goal="\"¦\"",
```
7h3kk1d: fwiw, I find theses a lot easier to read with the `{|"¦"|}` syntax

7h3kk1d (approved):
Left some questions but tests look great

src/haz3lcore/zipper/action/Insert.re
disconcision: this comment is... mostly for me. it's saying check those cases still work if you change this code, based on things ive broken before. but probably not necessary now that we have tests, ill remove

src/haz3lcore/zipper/Printer.re
disconcision: This question accidentally caused a refactor

cyrus- (approved):

PR #1725, opened 2025-06-28
Wrap Current Selection in Comment

russell-rozenbaum:
Note: Ctrl on PC, Cmd on Mac...

**Adds a shortcut to comment in and out segments of code. Select line(s) of code you'd like to comment out and press _Ctrl+/_ to trigger**. This is based on a common shortcut convention in several languages. 

Additionally, the user can **comment out the lowest-enclosing type/value binding's pattern+definition based on the code at the location of the cursor by pressing _shift+ctrl+/_**. Under the hood, this comments out in the same way as the line-based approach, except it does an additional preliminary step in selecting the pattern+definition before proceeding.

Buffer(Set(TyDi)) used to use Ctrl+/; switched it to use Ctrl+b.

If any line in a selection does not begin with a comment delimiter, then all lines are wrapped in comments. O.w, if all lines do begin with a comment delimiter, then all lines are unwrapped from comments. 

**This does not retain preexisting comments that exist within a selected segment or on a line. This is the only pitfall that I cant think of that this comes with.**


7h3kk1d:
I thought about adding this with ctrl+/ but didn't because of the ability to turn it back into code. How would you feel about just having it delete the comment and paste the text? So a reinsertion

disconcision:
cool! It does actually seem to work to be to uncomment? perhaps accidentally so? (if you press # with a comment selected) with minor artefacts

disconcision:
https://github.com/user-attachments/assets/f4f41ae5-6e33-4c69-8b56-df5fc13e87a6

russell-rozenbaum:
> cool! It does actually seem to work to be to uncomment? perhaps accidentally so? (if you press # with a comment selected) with minor artefacts

Yea, just implemented this (and just updated description)

russell-rozenbaum:
> I thought about adding this with ctrl+/ but didn't because of the ability to turn it back into code. How would you feel about just having it delete the comment and paste the text? So a reinsertion

Yea, that's essentially what I ended up doing. 

eg. selecting:
```
#let x = 3 in#
let y = 4 in
```
and pressing ctrl+/ will turn it into 

```
let x = 3 in
#let y = 4 in#
```

So it essentially inverts comments/code, a pitfall being that it does so by checking the first character of a line.
Other languages like Python will just insert a '#' no matter what. Because Hazel is more so a structure editor, comments are encapsulated, so that approach doesn't seem to work too well. It would be ideal to preserve comments, such as this case:
`let y = 4 in # y is 4 # let u = 6 in`

Which I think just requires nifty regexp

7h3kk1d:
Yeah, we need a way to escape comments/strings if we're going to do the nested version which I imagine is ideal.

cyrus-:
@russell-rozenbaum can you fix the merge conflicts here? happy to get it merged soon.

cyrus-:
we have basic line comments in Hazel now I think from another PR. this might have some things that we want still, but if so please open another PR

PR #1728, opened 2025-06-30
Handle Deletion of Last Remaining ScratchPad

russell-rozenbaum:
Would fail silently when the user deleted a scratchpad from a list of one scratchpad.

Now defaults to creating a new, empty scratchpad.

### DIFF
--- a/src/web/view/ScratchMode.re
+++ b/src/web/view/ScratchMode.re
@@ -116,6 +116,38 @@ module Update = {
     JsUtil.QueryParams.set_param("name", name);
   };
 
+  let mk_new_scratchpad =
+      (model: Model.t, is_documentation: bool)
+      : list((string, CellEditor.Model.t)) => {
+    let new_key =
+      switch (is_documentation) {
+      | false =>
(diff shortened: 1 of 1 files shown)


src/web/view/ScratchMode.re
```diff
@@ -116,6 +116,38 @@ module Update = {
...
+
+        "Scratchpad " ++ string_of_int(new_number);
+      | true =>
+        JsUtil.prompt("Enter new buffer name:", "New Buffer Name")
+        |> Option.get
+      };
+    let new_sp =
+      model.scratchpads
+      @ [(new_key, CellEditor.Model.mk(Editor.Model.mk(Zipper.init())))];
+    new_sp;
```
7h3kk1d: If we don't have the type annotation we can just leave off the variable.
```suggestion
model.scratchpads
      @ [(new_key, CellEditor.Model.mk(Editor.Model.mk(Zipper.init())))]
```

7h3kk1d (approved):

cyrus- (approved):

PR #1729, opened 2025-07-01
Fix fixpoint env substitution

7h3kk1d:
In #1668 I had broken substitution for the fixpoint closure. This fixes it tests give a couple examples.

### DIFF
--- a/src/language/dynamics/transition/Transition.re
+++ b/src/language/dynamics/transition/Transition.re
@@ -281,12 +281,11 @@ module Transition = (EV: EV_MODE) => {
         let env'' =
           VarBstMap.Ordered.mapo(
             ((p, exp)) =>
-              FixF(
-                Var(p) |> Pat.fresh,
-                exp,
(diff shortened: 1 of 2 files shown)


disconcision:
wow this branch number is expressible as the sum of two positive cubes in two different ways

7h3kk1d:
> wow this branch number is expressible as the sum of two positive cubes in two different ways

I'll take this as a LGTM

src/language/dynamics/transition/Transition.re
```diff
@@ -297,6 +296,7 @@ module Transition = (EV: EV_MODE) => {
             env'',
             env |> Option.value(~default=ClosureEnvironment.empty),
           );
+
```
7h3kk1d:
```suggestion
```

disconcision (approved):
LGTM. fixes the mergesort eval issue on demo-hzl-builtins-moar

cyrus- (approved):

Issue #1563, opened 2025-03-14
Replace Grammar equals and TermBase fast_equals functions with a more flexible equality function


7h3kk1d:
We want one equality function for each sort that has flags for alpha equivalence, paren normalization, etc.

7h3kk1d:
@Negabinary It seems like currently typ equality is alpha equivalent and expression equality is not. Given these reference each other are there situations we need different behavior in a single call site?

Negabinary:
Not that I'm aware of, I suspect it was just for convenience / what we usually use in each sort.

Both equality of types within expressions and especially equality of expressions within types are somewhat afterthoughts

7h3kk1d:
Sounds good. I just don't want to have to add `alpha_equivalent_type`, `alpha_equivalent_exp` params.

disconcision:
There is (as of #1420) a special case for probes in fast_equals. This is because EvalResult.calculate checks equality on the syntax to decide whether to recalculate dynamics; we want recalculation to happen if we add a probe, even though it doesn't effect the dynamic semantics of the probed code. This is a bit of an awkward case, as in principle the calculate check should be gated on semantic as opposed to syntactic equality; although probes are debatably neither, they are closer to the latter than the former. So likely some sort of hack will still be required, unless we decided to fully side-channel information about what is to be probed.

Negabinary:
One more design consideration, adding a form to the current fast-equal is a trap that at least three people (including myself) have fallen into, because if you define equality wrong, the evaluator continuously restarts the webworker, thinking elab has changed when it hasn't. (https://github.com/hazelgrove/hazel/issues/1545)

So perhaps we could add a comment for people implementing new terms explaining that you can't use (==) to define this, and I'd also favour doing a sort of alternation between correct cases and fallback cases, instead of having all the correct cases first.

i.e.
```
| (Seq(e1, e2), Seq(e1', e2'))
| (Seq(_), _) 
| (If(e1, e2, e3), If(e1', e2', e3'))
| (If(_), _)
```

instead of

```
| (Seq(e1, e2), Seq(e1', e2'))
| (If(e1, e2, e3), If(e1', e2', e3'))
| (Seq(_), _) 
| (If(_), _)
```
which I think was originally my bad

7h3kk1d:
> So perhaps we could add a comment for people implementing new terms explaining that you can't use (==) to define this, 

[...]

PR #1734, opened 2025-07-02
Add all of the builtins

disconcision:
1. What: This PR add a 3 types and 74 functions or function alises to the builtins, and reorganizes the existing builtins to streamline the process of adding new ones. The functions that could be standard library functions are implemented as Hazel code, with accompanying concrete syntax.
2. Why: This greatly expanded selection is intended as a temporary measure to support LLM, MVU, dataframes, and advent of code initiatives, and is to later be replaced by a Hazel standard library.
3. Non-total functions return 'undefined' as the dynamic error constructor doesn't seem to show up in the value display. All added functions have dynamic tests.

 Types:
 - Option = None + Some(?) (used for several list functions)
 - Either = Left(?) + Right(?) (used for partition_map)
 - Ord = Lt + Eq + Gt (used for sort)

 String functions (some use regexps provided as plain strings)
 - string_escaped
 - string_uppercase
 - string_lowercase
 - string_capitalize
 - string_uncapitalize
 - string_match
 - string_replace
 - string_search

 Pair functions:
 - fst
 - snd

 Option functions:
 - option_map
 - option_bind
 - option_to_list

 List functions:
 - cons
 - hd
 - tl
 - is_empty
 - length
 - map
 - filter
 - fold_left
 - fold_right
 - flat_map
 - zip
 - unzip
 - reverse
 - take
 - drop
 - take_while
 - drop_while
 - slice
 - init
 - range
 - enumerate
 - any
 - all
 - intersperse
 - nth
 - append
 - concat
 - mapi
 - filteri
 - find
 - mem
 - partition
 - rev_append
 - fold_left2
 - fold_right2
 - map2
 - all2
 - any2

 List functions involving option:
 - hd_opt
 - tl_opt
 - filter_map
 - nth_opt
 - find_opt
 - find_index
 - find_map
 - find_mapi

 List functions involving either:
 - partition_map

 Association list functions:
 - assoc
 - assoc_opt
 - mem_assoc
 - remove_assoc

 List function aliases:
 - contains
 - flatten
 - rev
 - exists
 - for_all
 - exists2
 - for_all2
 - concat_map
 - split
 - combine


test/Test_Menhir.re
```diff
@@ -14,11 +14,12 @@ let strip_Wrap_and_add_builtins =
         | Parens(e)
         | Probe(e, _) => cont(e)
         | Var(x) =>
-          let builtin = Util.VarMap.lookup(Builtins.Pervasives.builtins, x);
+          let builtin = Util.VarMap.lookup(Builtins.builtin_map, x);
```
7h3kk1d: So it looks like some of the tests are breaking since the documentation buffers are shadowing the builtins (map).  We do the builtins based on the context which is shadowable and not in the parser ordinarily.

I think we should just stop parsing builtins as builtins in menhir. The original use case to mostly use it for elaboration tests has faded somewhat and the only test that fails if we parse them as vars is `Built-in function (menhir).`. If we end up needing a literal syntax for the builtins (which I hope we don't) we can add it back to menhir with special syntax.

7h3kk1d: Let me know if you want me to make this change

src/language/builtins/BuiltinsADT.re
```diff
@@ -0,0 +1,192 @@
...
+let create_type_alias = (name: string, typ: Typ.t): Ctx.entry =>
+  Ctx.TVarEntry({
+    name,
+    id: Id.invalid,
+    kind: Ctx.Singleton(typ),
+  });
+
+// Convert type aliases to context entries
+let types: list(Ctx.entry) =
+  List.map(((name, typ)) => create_type_alias(name, typ), type_aliases);
```
7h3kk1d, line 173: It kills me that this doesn't uncurry automatically to make it point-free

src/language/builtins/BuiltinsBase.re
```diff
@@ -0,0 +1,470 @@
...
+      Some(Exp.string(String.trim(s)));
+    },
+  },
+  {
+    name: "string_escaped",
+    arg: Atom(String),
+    ret: Atom(String),
+    imp: d => {
+      let-unbox s = (Atom(String), d);
+      Some(Exp.string(String.escaped(s)));
```
7h3kk1d, line 301: Do we want unescaped?

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+            None,
+          )
+        )
+      );
+    },
+  },
+  {
+    str: {|fix filter -> fun xs f -> case xs
+             | [] => []
+             | [x] => if f(x) then [x] else []
```
7h3kk1d: This middle case is unnecessary. Can't tell if it's here as an optimization.

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+            ),
+            None,
+          )
+        )
+      );
+    },
+  },
+  {
+    str: {|fix flat_map -> fun (xs, f) -> fold_left(xs, fun (acc,x) -> acc @ f(x), []))([1,2],fun x ->[x,x])|},
+    name: "flat_map", //TODO: weird string content above
```
7h3kk1d: maybe todo should be todone

src/language/builtins/BuiltinsList.re
7h3kk1d: fold_left does the same thing

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+          )
+        )
+      );
+    },
+  },
+  {
+    str: {|fix reverse -> fun xs -> case xs
+             | [] => []
+             | x :: xs => reverse(xs) @ [x]
+           end|},
```
7h3kk1d, line 340: I expected this to be slow and was going to suggest

```hazel
let reverse = let go = fun (xs, acc) -> case xs 
| [] => acc
| (x :: xs) => go(xs, x::acc) 
end in
go(_, []) 
```

But since list concatenation is implemented in ocaml the existing implementation is way faster. I still suspect it's quadratic but thousands of times faster.

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+        )
+      );
+    },
+  },
+  {
+    str: {|fix take -> fun (xs, n) -> case xs
+             | [] => []
+             | _ => if n <= 0 then [] else case xs
+             | x :: xs => x :: take(xs, n - 1)
+           end|},
```
7h3kk1d: For take and drop if you put the if on the outside you don't need the double case statement which also would make the matches exhaustive.

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+        )
+      );
+    },
+  },
+  {
+    str: {|fix enumerate -> fun xs -> enumerate_helper(xs, 0)
+           fix enumerate_helper -> fun (xs, index) -> case xs
+             | [] => []
+             | x :: xs => (index, x) :: enumerate_helper(xs, index + 1)
+           end|},
```
7h3kk1d: Fwiw, this string isn't accurate.

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+          )
+        )
+      );
+    },
+  },
+  {
+    str: {|fix mem -> fun (xs, x) -> case xs
+             | [] => false
+             | y :: xs => if x == y then true else mem(xs, x)
+           end|},
```
7h3kk1d: Where we can I'd probably start using the stdlib

```hazel
fun xs, x -> any(xs, fun y -> x == y)
```

src/language/builtins/BuiltinsList.re
```diff
@@ -0,0 +1,2844 @@
...
+            ),
+            None,
+            None,
+          )
+        )
+      );
+    },
+  },
+  {
+    str: {|fix slice -> fun (start, len, xs) -> take(drop(xs, start),len)|},
```
7h3kk1d: We should consider if we want end or length on slice. Python/js do end index but I could see it either way.

src/language/term/TermBase.re
```diff
@@ -177,8 +177,8 @@ and Exp: {
     ) =>
     t;
 
-  let fast_equal: (t, t) => bool;
-  let equal: (t, t) => bool;
+  let fast_equal: (~ignore_constructor_types: bool=?, t, t) => bool;
```
7h3kk1d, line 180: Self reminder to do https://github.com/hazelgrove/hazel/issues/1563

test/evaluator/Test_Evaluator_Prelude.re
```diff
@@ -3,7 +3,11 @@ open Language;
 
 module UG = Grammar.UnitGrammar;
 
-let dhexp_typ = testable(Fmt.using(Exp.show, Fmt.string), DHExp.fast_equal);
+let dhexp_typ =
+  testable(
+    Fmt.using(Exp.show, Fmt.string),
+    DHExp.fast_equal(~ignore_constructor_types=true) // This is a bad idea to just blanket ignore the type
```
7h3kk1d: Are we good just keeping this for now? I'm not sure where the constructor types are used in dynamics and when it would matter.

disconcision: i didn't know we had types on constructors sooo

test/Test_Menhir.re
disconcision: @7h3kk1d yeah i was going to ask you why the parser was handling builtins specially but i forgot. i think we should get rid of that if we can. if you could do it that would be great

src/language/builtins/BuiltinsADT.re
disconcision: sometimes i make functions take pairs for this reason, usually makes me feel somewhat dirty

test/evaluator/Test_Evaluator_Prelude.re
7h3kk1d: types on types on types

src/language/builtins/BuiltinsBase.re
disconcision: huh. i'll add it. missed it as it's not in the String lib (its in Scanf)

src/language/builtins/BuiltinsList.re
disconcision: changed it

test/evaluator/Test_Evaluator_Prelude.re
disconcision: is this actually necessary? didn't we change it so that the tests do have the types on them?

test/Test_Menhir.re
disconcision: wait, what are the failing tests? are these ones that don't run with `make test-quick`?

test/evaluator/Test_Evaluator_Prelude.re
7h3kk1d: These tests will fail:

https://github.com/hazelgrove/hazel/pull/1734/files#diff-a5b6e71606758826f0a3f4c4052b3fa0223d021509a0238b4103574a0127d484R209

But we can rewrite those not to use the parser for just the ones that have constructors.

test/Test_Menhir.re
disconcision: oh. i see. i though test-quick just did the same tests but... faster?

7h3kk1d: If only. It skips the pbt and the longer tylr ones

test/evaluator/Test_Evaluator_Prelude.re
disconcision: This is the only remaining issue; are you okay just leaving this in place for now?

cyrus- (approved):

PR #1752, opened 2025-07-03
Duplicate variable checking in patterns

joshqwang:
Causes Hazel to give an error message on patterns that contain duplicate variables, including labels in tuples.
Screenshot 2025-07-03 at 5 18 05 PM

Added a new get_duplicate_bindings function in Term.re.
Changed Duplicate type in Self.re to DuplicateLabel so it isn't confused with the new DuplicateVar type.


Negabinary:
I'm getting a crash when I try typing

`let (x,x`

joshqwang:
Should be fixed now.

Negabinary:
I'm a little confused how the changes in statics.re work - specifically I don't understand how you managed to reuse the duplicate-label list to also do duplicate variables. Even though I don't understand it somehow almost every example I use to try and break it works including:

`let (x, x=?) = ? in  `  No error
`let (x=(x, y)) = ? in  `   No error

The one weird behaviour I could come up with is 

`let (x, x=(x, y)) = ? in  `

image

which puts the errors on the first and second `x` instead of the first and third.

Aside from that one issue in Statics.re I'm happy with the rest of this pr; I appreciate the bonus cleanup in elaborator/pattern!

joshqwang:
I intended initially to have duplicate variable bindings and duplicate label bindings share the same list of duplicates. 

In an earlier commit, `let (x, x=?) = ? in ` produced a DuplicateVar error _and_ a DuplicateLabel error.

But, when I did the code cleanup in Term.re, I ended up breaking that functionality, since the function I deleted and the Term.bindings function had more differences than I realized.

I'm a little unsure what correct behavior should be here. I have a version on my machine that fixed that functionality.

Screenshot 2025-07-29 at 4 16 44 PM

But, in these case, should the labels have no error on them?

MaxCarroll0:
> I'm a little unsure what correct behavior should be here. I have a version on my machine that fixed that functionality.
> ...
>But, in these case, should the labels have no error on them?

The `x = e` pattern doesn't actually bind `x` to anything, so the left hand side (`x`) should never be in a duplicate binding error (so, as you say, it should not be a duplicate label in any of your examples: there are **not** two labels with name `x`, rather a *label* & a *variable*). 

And, if the right hand side (`e`) has duplicate variables they should be errors (e.g. the 3rd `x` in the 3rd example is correctly detected as an error). BUT, the 2nd `x` in the 2nd example should **not** be a duplicate variable error (only one variable `x` is used, the other `x` is a label, not a variable).

It seems like you do indeed need to distinguish duplicate labels and duplicate variables?

joshqwang:
Hopefully should be working now.
Screenshot 2025-07-31 at 1 20 24 PM

MaxCarroll0:
Looks good to me

test/statics/Test_Statics_Sums.re
cyrus-, line 1: - can you also add a test verifying that it works for fun
- also add test(s) for labeled tuple patterns

cyrus- (changes requested):
lgtm other than needing a few more tests

dm0n3y:
@cyrus- ready for review

dm0n3y:
@cyrus- resolved merge conflicts, checked that josh's example code above works, should be good to merge

src/language/statics/Statics.re
```diff
@@ -210,7 +211,9 @@ and uexp_to_info_map =
         go(~ana, ~duplicates, e, m) |> (((e, m)) => (es @ [e], m)),
       ([], m),
     );
-  let go_pat = upat_to_info_map(~ctx, ~ancestors, ~duplicates);
+  // TODO: probably broken
```
disconcision: is it?

src/language/term/Pat.re
```diff
@@ -302,3 +302,11 @@ let bound_var_ids = (ctx, pat): list(Binding.t) =>
          }
        }
      );
+
+let get_duplicate_bindings = (pat: t) => {
+  let bindings = bound_vars(pat);
+  List.filter(
```
disconcision, line 308: should be a fn in listutil that does this

src/language/statics/Statics.re
dm0n3y: no, at least not wrt dev, but there was a benign dev bug that was passing in duplicate expression-level labels into the pattern duplicates argument. benign bc the pattern duplicates arg would then later be overwritten. fixed that here while reviewing.

src/language/term/Pat.re
dm0n3y: i think you mean the `dedup` fn? that returns entire list deduplicated, this returns just the duplicates.

cyrus- (approved):

PR #1754, opened 2025-07-04
Syntax fixes

### LINKED ISSUE #1115: Unexpected move behavior on incomplete poly tokens
Description: When my cursor is at this position and press ←:
{3A294E62-5B5B-45b7-B039-3E0A4B6CFC68}
The cursor dosn't appear where it's expected
{596D28BA-8496-4e3d-BAB5-2069A51BBC38}

This bug seems to occur on incomplete combined keyword.

### LINKED ISSUE #1253: Pasting a string into an empty tuple or list does not have desired effect
Screenshot 2024-03-19 at 4 41 27 PM

This screenshot shows the results of pasting the `"yo"` into `()`. Considering strings will still be special-cased in the new syntax model this may still be an issue.

### LINKED ISSUE #1259: Certain strings fail to paste invisibly
The simplest example is `([)(`; it will print a messages to the console "WARN: zipper_of_string: Invalid_argument("option is None")" but there is no other feedback. This is a minified version of the following, which arose naturalistically during LLM completion experiments:

```
fun (playList, action) ->
  case action
    | PlaySong(id) =>
      let (songs, _) = playList in (songs, Playing(id))
    | PauseCurrentSong =>
      let (songs, state) = playList in
      case state
        | Playing(id) => (songs, PausedOn(id))
        | _ => playList
      end
    | RemoveSong(id) =>
      let (songs, state) = playList in
      (List.remove(id, songs),
       case state
         | Playing(id') if id == id' => NoSongSelected
         | PausedOn(id') if id == id' => NoSongSelected
         | _ => state
       end)
    | AddSong(id) =>
      let (songs, state) = playList in (List.append(songs, [id]), state)
  end
```

(the problematic part is the 'if' that the llm is trying to use as a 'when' clause).

Likely a wontfix, pending the new syntax model, but I'm logging it.

### LINKED ISSUE #1354: elaboration returns none when entering a case rule
prior to pressing space here and expanding the rule, we get this:

Screenshot 2024-08-04 at 10 40 51 PM

### LINKED ISSUE #1362: Any selection touching a hole on the left gets convex tip even if should be concave
Screenshot 2024-08-08 at 5 34 24 PM

### LINKED ISSUE #1470: Adding parentheses starting with closing doesn't remold
If you insert a closing parenthesis, and then an opening one around an existing term, the parentheses and closed term are treated as expressions regardless of surrounding sort context.

Not sure if this is a regression or longstanding.

Image

### LINKED ISSUE #1564: Go to definition and jump to id places cursor to the left of expression giving wrong cursor inspector status
https://github.com/user-attachments/assets/dceaeb45-e74c-4f7d-94d3-bad522db01af

If you ctrl+click or go to defn of a variable binding the cursor jumps to the left of the binding. If there's no space to the left of the variable the cursor inspector gives the status of whatever is to the left of it as opposed to the binding. In the above example the cursor inspector shows the let expression as a tuple instead of the int x

### LINKED ISSUE #1596: Floating point operator completion inserts extra star
Image

Only happens when the overall operation is in analytic position. Maybe we should disable that logic now that we have `use` coming?

### LINKED ISSUE #1702: ExplainThis for fn lit taking pair overflows
Image

just add overflow-x: scroll to example cell i think

### LINKED ISSUE #1724: Cant_insert if you delete the beginning of misfix form then try to edit the middle
If you delete `let` or `if` and retain the rest of the compound form you can't insert characters in the middle.

E.g.
`let = in` then you delete the let and try to modify between the = and in

https://github.com/user-attachments/assets/755e3dc3-086b-46c2-8e68-8b4818ab3629

### LINKED ISSUE #1727: Console warnings from Livelits: Molds.get and LivelitProj.get
Image

### LINKED ISSUE #1753: Exception on deleting certain leading delimiters
E.g. pressing backspace here throws an exception. This is a regression due to my last PR, fix incoming.

Image

### LINKED ISSUE #1755: string_match, string_replace, string_search only use literal regexps
As in they can only match exact strings, not general patterns.

### LINKED ISSUE #1759: Trying to merge two prefix + operators into infix ++ in type sort results in wrong caret position
Here's an analogous but less impactful example with `!`:

given: (caret is to left of rightwards `!`)

Image

pressing backspace yields:

Image

which is not only the wrong caret position visually, it's also broken internally (note that the nib shape is straight ie wrong)

### LINKED ISSUE #1767: Not_found exception being thrown after adding let expression before negation
https://github.com/user-attachments/assets/ee6de85c-9d12-4741-b80c-9e661a3ff82c

If you type `-?` then move to the beginning of the line and type `let x= ` it adds a hole to the left of the minus. Then when you press right it throws a Not_found exception. The exception seems to be thrown from Deco.term_ranges here https://github.com/hazelgrove/hazel/blob/dev/src/web/app/editors/decoration/Deco.re#L280

I think this is new from https://github.com/hazelgrove/hazel/pull/1722

disconcision:
This PR fixes all outstanding syntax system issues which I've added or been assigned. Or at least the ones which are strictly bugs as opposed to design flaws.

Core Editing:
Closes #1753: Recently introduced bug which hard crashes on deleting any leading delimiter.
Closes #1724: Ancient and mysterious many-headed exception-throwing syntax bug dating back to pre haz3l tylr code. This should also fix the issue with the hinted tests form
Closes #1115: Longstanding movement edge case bug
Closes #1470: Remold bug
Closes #1354: 'Elaboration returns none' for incomplete case
Closes #1253: Bug when pasting string literal splits tokens
Closes #1259: Certain delimiters cannot be entered
Closes #1362: Selection decoration gets wrong shape on spaces near grout
Closes #1596: No longer suggest floating point exponentiation operator at the expense of multiplication
Closes #1759: Crashes on deleting/adding leading operands/operators in prefixed sum types
Closes #1767

Misc:
Closes #1755: String regexp builtins were broken
Closes #1727: Resolves the mold ambiguity warning in the console, and quiets a livelit warning
Closes #1702: ExplainThis embedded editors get horizontal scrolls if necessary
Closes #1564: Jump to id edge case fix

Others:
- Alphabetizes builtins (just functions, not types or constructors)
- Related to #1354 I made some adjustments to sort conflict styles aimed to reduce visual noise, mostly around entering case expressions. The angry red tile decoration when on unrecognized operators is reduced (it's annoying for freshly inserted rule bars in cases especially), and inconsistent-sorted tokens now just use the same color as expressions. This is less informative but less shouty, a good balance for now I think until we have a better experience around deferring yelling at the user during entry.
- Renames incorrectly-named Rul constructor Hole to MultiHole
- Editing movement tests

### DIFF
--- a/src/haz3lcore/zipper/Zipper.re
+++ b/src/haz3lcore/zipper/Zipper.re
@@ -69,6 +69,18 @@ let unzip = (seg: Segment.t): t => {
 let pop_backpack = (z: t) =>
   Backpack.pop(Relatives.local_incomplete_tiles(z.relatives), z.backpack);
 
+let will_barf = (t: Token.t, z: t): bool =>
+  switch (pop_backpack(z)) {
+  | Some((_, {content: [p], _}, _)) =>
+    switch (p) {
+    | Tile({shards: [i], label, _}) =>
+      assert(i < List.length(label));
+      List.nth(label, i) == t;
+    | _ => false
+    }
+  | _ => false
+  };
+
 let left_neighbor_monotile: Siblings.t => option(Token.t) =
   s => s |> Siblings.left_neighbor |> OptUtil.and_then(Piece.monotile);
 
--- a/src/haz3lcore/zipper/Zipper.re
+++ b/src/haz3lcore/zipper/Zipper.re
@@ -282,7 +282,7 @@ let remold_regrout_prev = (z: t): t =>
   switch (move(Left, z)) {
   | None => z
   | Some(z_left) =>
-    let z_left = z_left |> regrout(Right);
+    let z_left = z_left |> remold |> regrout(Right);
     switch (move(Right, z_left)) {
     | None => failwith("Zipper.put_down: move fail")
     | Some(z_right) => z_right
--- a/src/haz3lcore/zipper/Zipper.re
+++ b/src/haz3lcore/zipper/Zipper.re
@@ -26,7 +26,6 @@ let next_blank = _ => Id.mk();
 [@deriving (show({with_path: false}), sexp, yojson, eq)]
 type chunkiness =
   | ByChar
-  | MonoByChar
   | ByToken;
 
 [@deriving (show({with_path: false}), sexp, yojson, eq)]
@@ -195,6 +194,9 @@ let directional_unselect = (d: Direction.t, z: t): t => {
     selection,
   });
 };
+
+let unselect = (z: t): t =>
+  z.selection.content == [] ? z : directional_unselect(z.selection.focus, z);
 
 let move = (d: Direction.t, z: t): option(t) =>
   if (Selection.is_empty(z.selection)) {
@@ -563,7 +565,9 @@ let try_to_dump_backpack = (zipper: t) => {
 
 let smart_seg = (~dump_backpack: bool, ~erase_buffer: bool, z: t) => {
   let z = erase_buffer ? clear_unparsed_buffer(z) : z;
+  // print_endline("z before: " ++ Siblings.show(z.relatives.siblings));
   let z = dump_backpack ? try_to_dump_backpack(z) : z;
+  // print_endline("z after: " ++ Siblings.show(z.relatives.siblings));
   unselect_and_zip(~erase_buffer, z);
 };
 
--- a/src/haz3lcore/zipper/Zipper.re
+++ b/src/haz3lcore/zipper/Zipper.re
@@ -81,11 +81,10 @@ let neighbor_monotiles: Siblings.t => (option(Token.t), option(Token.t)) =
 let regrout = (d: Direction.t, z: t): t => {
   assert(Selection.is_empty(z.selection));
   let relatives = Relatives.regrout(d, z.relatives);
-  let z = {
+  {
     ...z,
     relatives,
   };
-  z;
 };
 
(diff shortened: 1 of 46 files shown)


src/haz3lcore/zipper/Zipper.re
```diff
@@ -280,7 +282,7 @@ let remold_regrout_prev = (z: t): t =>
   switch (move(Left, z)) {
   | None => z
   | Some(z_left) =>
-    let z_left = z_left |> regrout(Right) |> remold;
+    let z_left = z_left |> regrout(Right);
```
disconcision, line 285: This seems to fix #1753 but i'm not quite sure why...

src/web/app/editors/code/Code.re
```diff
@@ -157,7 +157,15 @@ module Text =
...
       );
-    let is_consistent = Sort.consistent(t.mold.out, expected_sort);
+    let consistent = (s: Sort.t, s': Sort.t) =>
+      switch (s, s') {
+      | (Any, _)
+      | (_, Any) => true
+      | (Rul, Exp) => true
+      | (Exp, Rul) => true
+      | _ => s == s'
+      };
```
disconcision, line 167: sort consistency isn't really a concept, besides a somewhat ad hoc style heuristic. moving here to make that clearer

PR #1761, opened 2025-07-06
Indentation

### LINKED ISSUE #882: Fix indentation for case rules

### LINKED ISSUE #887: Indentation of case rules when previous rule's expression is not bidelimited is incorrect

### LINKED ISSUE #1087: Indentation triage

disconcision:
Replaces #1099.

Closes https://github.com/hazelgrove/hazel/issues/882
Closes https://github.com/hazelgrove/hazel/issues/887
Closes https://github.com/hazelgrove/hazel/issues/1087

This makes indentation work more normally. After a bunch of false starts I'm fairly happy with this conceptual approach:
1. Lots of tests
2. Trying to write a formatter for incomplete syntax is fraught and very non-orthogonal.
3. Therefore we should write a formatter for complete syntax, and then try to best-effort complete incomplete syntax.
4. The main goal for incomplete syntax is to make the normal left-to-right entry experience non-janky. There are two potential sources of jank here: (a) The level of indentation at the caret changing unexpectedly during entry, and (b) The level of indentation of the rest of the code changing during entry.
5. To reduce both, we want to complete the syntax in a way the user would expect, either wrapping or not wrapping following forms. To accomplish this we use the following heuristic: In a given segment, for each incomplete tile that is missing one or more trailing delimiters, insert all such trailing delimiters after a blank line which is after the incomplete tiles, or failing that, at the end of the segment. There are a couple ways this can go wrong, but in practice it seems to work pretty well. Often people will automatically insert a blank line in such a situation, or there will already be one there. If they don't, it's not a hard thing to teach them to do.
6. Previously I had considered using the caret position as the place to drop remaining delimiters. This is probably worth revisiting in the future, but it proved complex to make this efficient, i.e. avoiding totally recalculating Measured for every movement action with a non-empty backpack (although this should be doable in principle). It's also a more dramatic approach, as code magically re-indents as you move. I generally like the effect, but it should be carefully considered from a UX perspective.

Also fixes:
- Indentation bug for tuples (as seen in the `even` example if basic references)
- Most re-indentation jank when entering incomplete forms, provided you leave at least one blank line between where you're entering new code and the rest of the program


disconcision:
@7h3kk1d if you could kick the tires functionally it would be appreciated. The design goals are basically that: (a) Indentation of complete programs is what you'd expect, and (b) Indentation is what you'd expect when doing left-to-right entry, provided that you've left a blank line between where you're entering and the rest of the program, and modulo indentation shifts like when a variable `i` turns into an `in` delimiter etc.

PR #1768, opened 2025-07-09
Add failing test for inserting let binding before prefix negation

### LINKED ISSUE #1767: Not_found exception being thrown after adding let expression before negation

7h3kk1d:
Just adding the failing test here to track the issue. I'm not sure how to address it yet.

### DIFF
--- a/test/Test_Editing.re
+++ b/test/Test_Editing.re
@@ -198,6 +198,22 @@ let insertion_tests = [
     ~acts=mk({|if the¦else|}) @ [Insert("n"), Insert(" ")],
     ~goal={|if? then?¦else?|},
   ),
+    test(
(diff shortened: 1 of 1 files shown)


disconcision:
Now fixed in #1754. not a good fix but a fix

PR #1769, opened 2025-07-09
Assistant Actions v2

russell-rozenbaum:
#### Code Cleanup
- [ ]  Namely, AssistantUpdate.re file and CompositionModel.re file. Do so via: Organize into Modules -> Further Separate Concerns into Submodules -> Split into Files/Folders if necessary -> Supply "Public" Modules (I find this helps with clarity on what is local/helper and what is external/publicly used)

#### Task Nodules and Before/After Views 
- [ ]  Store tool calls with each subtask. This can be done via tracking when a tool is called and which task was marked as active.
- [ ]  Store a before segment and after segment. Simply, store what the segment was when the task/todo list was created, then, store what it is when it is completed. NOTE: Need to add a task completion action. Currently, only archiving a task exists.

#### Add Tool Calls for Agent to Test Code it Writes
- [ ]  ^

#### Write Robust Test Cases
- [ ]  Composition Action Tests
- [ ]  Assistant Model/Update Tests

#### Improve Agent View Logic
- [ ] We might want to default to expanding the definitions of recently implemented code

#### General Agent Toolkit Improvements
- [ ] Rename UpdateBody to something else, leaning towards UpdateScope. The agent seems to get confused sometimes as to what exactly the body is. Understandably so tho, eg. "function bodies" could lead to some confusion, or the fact that updating the ultimate, final body of a program is just updating the last defined variable's body... oftentimes it tries to call "UpdateBody" on a variable called "root".


src/haz3lcore/AssistantCore/AssistantTreeHelper.re
```diff
@@ -273,6 +273,7 @@ let build_curr_node_info =
           | FixF(_, e, _)
           | Use(_, e)
           | TypAp(e, _)
+          | TypFun(_, e, _)
```
russell-rozenbaum, line 276: this was a bug fix

src/haz3lcore/AssistantCore/CompositionTools.re
```diff
@@ -135,6 +144,15 @@ let string_of = (action: action) => {
...
     ++ ")"
+  | Nav(GoToSibling(Stepwise(d))) =>
+    "go_to_sibling("
+    ++ (
+      switch (d) {
+      | Left => "Prev"
+      | Right => "Next"
+      }
+    )
+    ++ ")"
```
russell-rozenbaum, line 155: this was just for reorganizing

src/haz3lcore/AssistantCore/CompositionTools.re
```diff
@@ -146,16 +164,19 @@ let string_of = (action: action) => {
...
-      | Right => "Next"
+      switch (index) {
+      | Some(index) => ", " ++ string_of_int(index)
+      | None => ""
       }
     )
     ++ ")"
-  | Read(ViewDefinition) => "view_definition"
+  | Read(ViewEntireDefintion) => "view_entire_definition"
+  | Read(ShowUseSites) => "show_use_sites"
```
russell-rozenbaum: stubbing these tools here

russell-rozenbaum: still have yet to implement

cyrus-:
see #2048

PR #1781, opened 2025-07-10
Adjusting some selection movement behavior

7h3kk1d:
Make movement out of selection work like other editors.

https://github.com/user-attachments/assets/e8139ce6-59cc-4ee1-ac8c-3dd25a7a7e9a

### DIFF
--- a/src/haz3lcore/zipper/action/Move.re
+++ b/src/haz3lcore/zipper/action/Move.re
@@ -425,23 +425,13 @@ module Make = (M: S) => {
        * even if we don't actually move */
       switch (d) {
       // By char just unselects
-      | Local(Left(ByChar)) => Some(Zipper.directional_unselect(Left, z))
-      | Local(Right(ByChar)) => Some(Zipper.directional_unselect(Right, z))
-      | Local(Left(_))
-      | Extreme(Left(_) | Up)
-      | Local(Up) =>
-        // Go to beginning of selection and apply move
-        let z = Zipper.directional_unselect(Left, z);
-        switch (move_dispatch(d, z)) {
-        | Some(z) => Some(z)
-        | None => Some(z)
-        };
-      | Extreme(Right(_))
-      | Local(Down)
-      | Local(Right(ByToken))
-      | Extreme(Down) =>
-        // Go to end of selection and apply move
-        let z = Zipper.directional_unselect(Right, z);
+      | Local(Left(ByChar) as planar)
+      | Local(Right(ByChar) as planar) =>
+        Some(Zipper.directional_unselect(Zipper.from_plane(planar), z))
+      | Local(planar)
+      | Extreme(planar) =>
+        // Escape selection and apply move
+        let z = Zipper.directional_unselect(Zipper.from_plane(planar), z);
         switch (move_dispatch(d, z)) {
         | Some(z) => Some(z)
         | None => Some(z)
--- a/src/haz3lcore/zipper/action/Move.re
+++ b/src/haz3lcore/zipper/action/Move.re
@@ -423,10 +423,27 @@ module Make = (M: S) => {
     } else {
       /* Always empty selection on move action,
        * even if we don't actually move */
-      let z = Zipper.directional_unselect(z.selection.focus, z);
-      switch (move_dispatch(d, z)) {
-      | Some(z) => Some(z)
-      | None => Some(z)
+      switch (d) {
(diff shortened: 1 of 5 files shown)


src/haz3lcore/zipper/action/Move.re
```diff
@@ -423,10 +423,27 @@ module Make = (M: S) => {
...
+        | Some(z) => Some(z)
+        | None => Some(z)
+        };
+      | _ =>
+        let z = Zipper.directional_unselect(z.selection.focus, z);
+        switch (move_dispatch(d, z)) {
+        | Some(z) => Some(z)
+        | None => Some(z)
+        };
       };
```
7h3kk1d: This aligns with my understanding of how vscode and other editors do it. Left/Right just cancel the selection in that direction and by token cancels in that direction then takes the step.

src/haz3lcore/zipper/action/Move.re
```diff
@@ -423,10 +423,23 @@ module Make = (M: S) => {
...
       /* Always empty selection on move action,
        * even if we don't actually move */
-      let z = Zipper.directional_unselect(z.selection.focus, z);
-      switch (move_dispatch(d, z)) {
-      | Some(z) => Some(z)
-      | None => Some(z)
+      let z =
+        switch (d) {
+        | Local(planar)
+        | Extreme(planar) =>
```
disconcision: why the change for the Extreme case here?

test/Test_Editing.re
```diff
@@ -16,14 +16,52 @@ module Fresh = Language.IdTagged.FreshGrammar;
...
+    let selection_pos = caret_index + size;
+    String.sub(serialized_with_caret, 0, selection_pos)
+    ++ selection_char
+    ++ String.sub(
+         serialized_with_caret,
+         selection_pos,
+         String.length(serialized_with_caret) - selection_pos,
+       );
+  };
+};
```
disconcision, line 30: is there a reason why you didn't bake the selection printing logic into Printer? it seems generically useful

src/haz3lcore/zipper/action/Move.re
7h3kk1d: I had presumed there's a performance improvement by exiting the direction of the extreme movement rather than from the selection focus. I mostly thought about it for extreme up/down. Looks like this diverges from vs code in the multiline case for home/end. This intuitively makes more sense to me but I can move the extreme case to be with the goal one. @disconcision do you have any intuition for how it should work?

test/Test_Editing.re
7h3kk1d: Mostly a skill issue and not wanting to figure out how the buffers work in selection.

7h3kk1d: Would you calculate it the way I am and then add selection_point to Zipper like caret_point? Mostly I'm afraid of making something wrong in the actual code vs the tests.

src/haz3lcore/zipper/action/Move.re
7h3kk1d: Just tested with prosemirror/vscode/some online text inputs. Home/End, Extreme(Left | Right), seem to always break on the caret focus direction. It doesn't make a ton of intuitive sense to me but I can change it to that.

Interestingly move left/right by token seems to just directionally cancel on prosemirror and textinput boxes, but it does move in vs code. I'm planning to stay compliant with vs code.

test/Test_Editing.re
7h3kk1d: I spent about an hour trying to get this to work by rewriting the printer to operate over the zipper structure with the caret/selection handled throughout. It seems possible but requires a bunch of rewriting. I couldn't figure out the right way to do the selection anchor in terms of Measured so if I do this in the Printer it would just be doing this string offset trick in the of_zipper function. Not sure that's worth it.

disconcision: i know i said to do it the way youre doing, which is still fine by me, it but to do it via measured, i think you just do: take the first and last pieces in the selection and look up their first/last in the measurements. use the focus to decide which is the caret. use this logic only when the selection is non-empty, still caret_point otherwise

7h3kk1d: @disconcision I think I did it but would like your review on https://github.com/hazelgrove/hazel/pull/1781/commits/e50db3b88b39746bdaa0614f5441e62ef27af4c8

disconcision (approved):
lgtm

PR #1785, opened 2025-07-11
Fix type of sort function and make string_compare compatible

7h3kk1d:
Fixes the type of the builtin sort to have a comparator that returns Ord.t as opposed to int to be compatible with it's dynamics. Also changed string_compare to return Ord.t so you can do:

```hazel
sort(string_compare,["ab", "", "ca","Al","T","c"])
```
Without converting between int and Ord.t


disconcision (approved):
lgtm, seems to work fine

cyrus- (approved):

PR #1788, opened 2025-07-14
Modules

gcrois:
- [x] [Dynamics](https://github.com/hazelgrove/hazel/blob/f7e26bd432a78146f409ff6da8c0ff6c8d1d5d5c/test/evaluator/Test_Evaluator_Modules.re)
- [ ] Analysis case in statics
- [ ] Fix todos ([especially type hygiene](https://github.com/hazelgrove/hazel/blob/643a4a4aef6d327d92a96d788f2beac81da926f8/src/language/term/Typ.re))
- [ ] Use real syntax -- sorry triple semicolons :(
- [ ] Fix highlighting 

- [ ] Add keyword + form for submodule definition (`module`)
- [ ] Automatic typealias

Discussion:
https://github.com/hazelgrove/hazel/discussions/1558

Related PR:
https://github.com/hazelgrove/hazel/pull/1020/files

image


gcrois:
Thank you @7h3kk1d and @Negabinary for your help!!

We officially have basic dynamics working for modules, with a few problems (that you might have suggestions on!)

@Negabinary:
I think we are reversing somewhere we shouldn't be - Statics seems to get the bindings in their intended order (red highlight when using a binding after its defined), but dynamics can only access _future_ findings? [I tried swapping the pattern match](https://github.com/hazelgrove/hazel/blob/775c07f29a8dd51f22fb4424579e7d2722c64798/src/language/dynamics/transition/Transition.re#L878) but OCaml did _not_ like that. Any advice off the top of your head?

@7h3kk1d:
I want to get the Dot operator working -- I took a look at [what we do for tuples](https://github.com/hazelgrove/hazel/blob/775c07f29a8dd51f22fb4424579e7d2722c64798/src/language/dynamics/transition/Transition.re#L635), but I feel like I shouldn't just search through the entries and find a `ValBinding`. My instincts tell me we should have a context or something we can directly look through here?

To do ExpToSegment, I want to follow the pattern we have elsewhere where we use the [`Form.re`](https://github.com/hazelgrove/hazel/blob/0a15c06eda2c0813a83c2838f81bc6342bb26fa1/src/haz3lcore/lang/Form.re#L498) definition, but I kept getting a nebulous error about not matching the number of `_in` variables?

gcrois:
image

[Statics + Dynamics work for dot!](https://hazel.org/build/gc-modules/?name=Scratchpad%202&share=let%2520before_before%2520%253D%2520%2522before%2522%2520in%250Alet%2520module%2520%253D%2520%2520%257B%250Aval%2520before%2520%253D%2520before_before%2520%252B%252B%2520%2522...%2520%2522%2520%253B%253B%250Aval%2520after%2520%253D%2520before%2520%252B%252B%2520%2522after%2522%250A%257D%2520in%250Alet%2520after_after%2520%253D%2520module.after%2520%252B%252B%2520%2522...%2522%2520in%2520after_after)

gcrois:
I added some statics test cases we were missing! I'm able to get most of these if I use the [factored-out def binding code](https://github.com/hazelgrove/hazel/blob/c4944c4b942014d3c16f0676bf7b3a8556f0b817/src/language/statics/Statics.re#L353), but this seems to break evaluation? @Negabinary any thoughts [here](https://github.com/hazelgrove/hazel/blob/c4944c4b942014d3c16f0676bf7b3a8556f0b817/src/language/statics/Statics.re#L1228)?

@7h3kk1d can you help me make [these tests prettier](https://github.com/hazelgrove/hazel/blob/c4944c4b942014d3c16f0676bf7b3a8556f0b817/test/statics/Test_Statics_Modules.re#L25)? I wasn't sure the best way to do the ones I commented out, and I'm not convinced the last few are actually testing what we want to test. Also, I was getting some pretty rough crashes when editing modules because of holes -- [I went ahead and stripped them from `MakeTerm`](https://github.com/hazelgrove/hazel/blob/cc92fc04d9b5757a406ad6047e2acea3ba60dad8/src/haz3lcore/lang/MakeTerm.re#L307) -- this is probably a blunt solution. Can you think of something better?

For now, I'd like to avoid adding `Ctx` to the syntax of `ModuleSignature`, simply because of some truly frightening cyclic dependencies. We can pretty easily emulate this by [constructing a ctx when needed from the entries](https://github.com/hazelgrove/hazel/blob/f0c4d0eba7c9b466986f8e9213ffd5c592ef44c8/src/language/term/Term.re#L961).

PR #1791, opened 2025-07-15
Stepper lists variable names in substitution

joshqwang:
Added which variables are being substituted in justification for substitution in let expressions.

Screenshot 2025-07-15 at 1 12 19 PM


Negabinary (approved):
looks good to me!

PR #1793, opened 2025-07-16
Use more beginner friendly `git clone` command

haydenflinner:
My session trying to clone hazel after seeing the super impressive talk and web demo:

```
~/code/world
% git clone git@github.com:hazelgrove/hazel.git
Cloning into 'hazel'...
The authenticity of host 'github.com (140.82.112.3)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

~/code/world took 3s
128 % cd hazel
cd: no such file or directory: hazel

~/code/world
1 % ls

~/code/world
% git clone https://github.com/hazelgrove/hazel.git
Cloning into 'hazel'...
remote: Enumerating objects: 186865, done.
remote: Counting objects: 100% (3286/3286), done.
remote: Compressing objects: 100% (1031/1031), done.
remote: Total 186865 (delta 2783), reused 2351 (delta 2251), pack-reused 183579 (from 3)
Receiving objects: 100% (186865/186865), 178.48 MiB | 12.04 MiB/s, done.
Resolving deltas: 100% (140802/140802), done.
```

Cloning from the public URL means a beginner user doesn't have to have git ssh keys set up, though it may require `gh auth` to be able to push back seamlessly? I'm not sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ can also be accessed at:
 If you already have `ocaml` version 5.2.0, at least version 2.0 of `opam`, and `npm`
 installed, you can build Hazel by running the following commands.
 
-- `git clone git@github.com:hazelgrove/hazel.git`
+- `git clone https://github.com/hazelgrove/hazel.git`
 - `cd hazel`
 - `make deps`
 - `make dev`
@@ -147,7 +147,7 @@ To obtain an clean build, you may need to:
   enter the project root of your cloned Hazel project.
 
   ```sh
-  git clone git@github.com:hazelgrove/hazel.git
+  git clone https://github.com/hazelgrove/hazel.git
   cd hazel
   ```
 


cyrus-:
@7h3kk1d opinions on this? I don't know all the ins and outs of how github auth works. I'd like folks newly added to the org to not have to jump through any hoops to be able to push to branches, but also hear the point Hayden is making about folks that want to clone without having ssh keys set up. We could list both options if there isn't a seamless option.

7h3kk1d:
I don't have any strong opinions. It's probably worth thinking about who we're optimizing for in terms of github experience. The HTTPS origin does make cloning simpler but then you need to use a personal access token to push which I find more confusing, since it asks for "password", than just using ssh keys.

I've never used the `gh` cli  as opposed to just git. If most people are eventually going to contribute I'd probably favor the ssh url with a small comment about getting keys configured. We could also list both :shrug:.

haydenflinner:
The flow I've bumbled through with prior GItHub repos (IIRC) is that it will ask for a password, then reject whatever you give (since GitHub doesn't accept plain passwords anymore) along with printing a link for setting up the `gh` cli. I believe that's as simple as installing `gh` and running `gh auth login`, which opens your browser and uses that login. So, not perfect on a remote server, but works well enough locally, though ssh is at least one less tool to install (on Unixes)

7h3kk1d:
> The flow I've bumbled through with prior GItHub repos (IIRC) is that it will ask for a password, then reject whatever you give (since GitHub doesn't accept plain passwords anymore)

Yeah, the password that you have to give it is the personal access token which is very confusing because it looks like an http simple authentication password.

cyrus-:
I believe this was handled with the changes in #1797. Thanks for the contribution!

PR #1794, opened 2025-07-16
Fix INSTALL.md 'til next dep update

haydenflinner:
Necessary to avoid this err with current directions at least on Mac.

```
% make deps
opam update

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><>  🐫
[default] no changes from https://opam.ocaml.org
opam install ./hazel.opam.locked --deps-only --with-test --with-doc
[ERROR] Package conflict!
  * Missing dependency:
    - deps-of-hazel → ppx_deriving_qcheck < 0.7
    no matching version

No solution found, exiting
make: *** [deps] Error 20
```

Ref:
https://github.com/ocaml/opam-repository/issues/28065#issuecomment-2995585882

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -133,6 +133,11 @@ follow these instructions instead of the shorter instructions in the
   eval $(opam env --switch=5.2.0)
   ```
 
+- Add the opam archive (TODO Remove this at next release version bump)
+- ```sh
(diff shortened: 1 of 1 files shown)


cyrus-:
This was handled in #1797 as well. Thanks!

PR #1804, opened 2025-07-18
Structural Insertion and Delete

disconcision:


disconcision:
This branch adds an option to make forms fully expand, advancing the caret to the first contained hole, limited to certain syntactic contexts e.g. with nothing after then on the line. This isn't straightforwardly adaptable to the way that dev works now, in that keyword expansion doesn't wait for space to be pressed anymore, so this structural insertion precludes e.g. straightforwardly entering a variable beginning with 'let'. This would need to be rethought at a fundamental level

PR #1805, opened 2025-07-18
Backpack II

### LINKED ISSUE #752: cannot enter variable name when keyword in backpack is the prefix
Can't type out a variable prefixed by a keyword if the keyword is in the backpack.
For example, I can't type out a variable called `incr` when the `in` keyword is in the backpack.
 
Screen Shot 2022-09-01 at 22 58 49

Screen Shot 2022-09-01 at 22 59 00

disconcision:
### Abstract
This PR removes the backpack as a persistent data structure, instead re-deriving it when necessary from the rest of the zipper. We materialize a `local backpack` from analyzing just the local segment to figure out whether to match a newly-completed keyword tile to an existing incomplete one, and a `global backpack` from the whole zipper when calculating the backpack display. 

### Details

The global backpack simply consists of the missing shards of all incomplete tiles in the zipper/segment, not counting those in the selection. The local backpack consists exactly of the rightwards-missing shards of incomplete tiles to the left of the caret, and vice-versa in the other direction. All local backpack shards can be dropped in any order, except for `ambiguous polymorphs` like `=` and `=>`, which each occur as delimiters in at least two different forms, one of which is a monotile and the other a polytile. These are gated under the delimiters above them (corresponding to rightwards-more incomplete tiles) in order to permit entering the monotile form after a parenthesis, as otherwise it would not be possible to enter such forms during left-to-right text entry. 

### Supporting changes
- The backpack now shows over the top bar
- The backpack flagpole now has dynamic height when the caret is near the top of the screen, and the flagpole pixel alignment is fixed
- Remove legacy backpack actions: including: `pick_up`, `move to backup target`, `rotate backpack` (last already not surfaced in dev)
- Remove display of backpack targets

### Secondary changes to expansion mechanics
- Expansion of trailing keywords now requires inserting a space or other token-breaking character to close #752.
- Introduces `amorphous keyword prefix tiles` as a replacement for whitespace/grout transmutation. Any prefix of a keyword can now get a concave mold as well as the existing convex option, which prevents grout insertion jank while completing a non-leading keyword. This may sometimes cause non-optimal molding when a variable which is a proper prefix of a keyword is next to another operand with no intervening operator.
- Case rules now don't expand from `|` when outside a case

### Semantics changes supporting incomplete states
- Expression multi-holes are now treated semantically as seq (semicolon) for both statics and dynamics

## TODO:
- [x] half char flagpole height
- [x] border top instead of passing top for #main
- [x] put tydi form list in forms
- [x] generalize the instant closing expansion hack in Insert.re
- [x] failing tests
- [x] cases/tests in ExplainThis
- [x] delayed expansion for `>` issue on polymorphism slide
- [x] bug: the (global) backpack doesn't show some shards from incompletes in siblings of parent
- [x] don't show incompletes during selection (or otherwise prevent false incompletes owing to selection)
- [x] bug: backpack shows over nut menu
- [x] check if commas have a similar problem to ambiguous polymorphs (they do not)
- [x] ambiguous polymorph delimiters like `=` and `->`
- [x] reinstate dump backpack
- [x] bug: failure hd on entering `case` with statics on
- [x] add eg `end` from case to amorphous keywords
- [x] require space for trailing expansion
- [x] now TAB at end of complete keyword must trigger expansion

## FUTURE PRs:
- Local whitespace formatting on putdown
- Unify dump backback with indentation completion
- Match keywords on move-away actions
- Segment-wide rematching, including non-keyword delimiters (eg `( )` with `) (` in the backpack)
- Investigate segment-breaking rematching


disconcision:
Ready for functional review. Note there are at least two outstanding bugs, indicated above. I'm looking for feedback related mostly to:
1. the core change in the dropping mechanic. Any unexpected behavior here?
2. the change how trailing delimiters (like `=` and `in` for a let) now need the insertion of space or other token-breaking character to expand. this feels a bit awkward to me and i'm trying to decide if there's actually an issue or i'm just use to the old behavior.

cyrus-:
image
I'm not satisfied with this `in` behavior. The decorations here make it look like an infix operator. We should probably do something more specialized to the situation where pressing space would mold the token as an infix operator.

disconcision:
@cyrus- can you clarify? if you mean the arm decorations, i think changing those substantively is beyond my realistic capabilities atm. but could drop them entirely for such situations. could also drop the shard backing deco as well.

(i agree that the decorations here are bad; they are however honest wrt to the underlying parse for that state, the `in` is indeed treated as an infix operator. i'm down in principle to mitigate this by hiding the reality of the situation from the user though)

anyway i tentatively dropped the deco in these situations; feels better to me, lmk what you think

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -25,25 +25,16 @@ let set_llm_buffer = (z: t, response: string): t =>
       open OptUtil.Syntax;
       //TODO: Error feedback on below
       let* content = Parser.to_zipper(response);
-      let+ _ = [] == content.backpack ? Some() : None;
+      let+ _ = [] == Zipper.local_backpack(content) ? Some() : None;
```
cyrus-: convoluted

PR #1808, opened 2025-07-18
Retain/restore projector placements in concrete syntax

### LINKED ISSUE #1787: Lightweight projector placement serialization to retain projectors in backup text
This would appear in backup and externally pasted text as `$<projector_name>(<init_syntax>)`, e.g. `$slider(23)`. In principle, this new 'serialization ap form' is not intended as persistent in editor syntax; its entry should provoke replacement with a projector. Initially though we'll probably only trigger replacement in the specific circumstances sufficing to support deserialization for performance/complexity reasons. Note that this RFC covers projector _placement_ serialization, not projector serialization as such, in that it is based on un/projected syntax rather than the projector model. This could be changed in the future though.

Comparison to livelit invocation: This will behave analogously to livelit invocation in that finishing entering the serialization will trigger projection. Unlike livelit invocation, it includes entire ap, not just the leading token, and the serialization application is not included in the initializing syntax.

The new editing logic will be something like:

- Condition: A closing parenthesis character is entered, resulting in an exp/pat application to the immediate left of the caret, whose function expression (don't need to parse, can just check the token) is a token beginning in $ followed by a valid projector name.
- In this case, select the application, then overwrite the selection with just the part in parentheses
- Then, call the init method of the applicable projector
- Then, unselect, leaving the caret to the right

The hope is that this logic suffices to permit restoration of projector placements from backup text.

This will require an analogous system in serialization to replace projectors with a serialization application, using their unproject method. This can hopefully just be the default behavior in the segment printer.

It is not a goal at this time to serialize projector state, but the hope is that this could be achieved through essentially analogous means; instead of having the projector's init/unprojected syntax in the app, it can be syntax which is specified by a projector serialize_to_hazel method, or a generic method which serializes projector models to an sexp then to a hazel string.

disconcision:
Closes #1787.
Necessary for #1660.

- This introduces a lightweight concrete syntax to communicate to the editor that a projector should be placed on a term. 
- The primary goal is to provide a mechanism to prevent loss of projector placements when editors are reparsed from backup text, a complication which has been slowing development.
- Secondary goals include the ability to retain projectors on copy/paste, whether or not the internal segment cache is used, and when using the 'reparse editor' menu item. Reparsing any existing editor should now retain all projectors.
- It is NOT a goal of this PR to provide a user-friendly form for textual projector invocation. It's not actually that bad for this, but making it nice would require making it a new term type, requiring 50+ of change across dozens of files.

The syntax for invocation is `^^<projector_name>(<init_syntax>)` e.g. `^^probe(1 + 1)`. The double-caret was chosen instead of the previously-discussed `$` as the existing quote operator complicates the parsing story there, and I wanted to avoid burning another character. Invocation (replacement of concrete invocation syntax with projected syntax) happens whenever there is an insertion action which ends with the caret directly to the left of an invocation, i.e. the state after an insertion is  `... ^^<projector_name>(<init_syntax>)| ...`. This condition was chosen as it is the simplest to implement which satisfies the primary and secondary goals.


disconcision:
@7h3kk1d not sure how to satisfy the property tester for the parser here... commit [e50f9dd](https://github.com/hazelgrove/hazel/pull/1808/commits/e50f9dd2d7c1f6b5ce0de6a2d07ac54514714205) is my attempt. I just tried to make it to treat these things is invalids the same way maketerm does. not sure why it's even bugging me about this? livelits don't seem to be handled afaict and it doesn't complain about them? how does it know...

cyrus- (approved):

PR #1826, opened 2025-07-24
clean up evalresult

Negabinary:
Tidies up the eval result code and fixes a couple perf issues.

* Delete dead code
* Stop showing elaboration while running code for perf reasons
* Stop recalculating whole editor with dynamic data (we still redraw after dynamic data)
* Always run evaluator in the background, even when stepper is open

Later:

- [ ] Abbreviate results

### DIFF
--- a/src/web/view/ExerciseMode.re
+++ b/src/web/view/ExerciseMode.re
@@ -378,44 +378,40 @@ module Update = {
         model.cells,
       );
 
-    if (List.length(worker_request^) > 0) {
-      WorkerClient.request(
-        worker_request^,
-        ~handler=
-          List.iter(((pos, result)) => {
-            let pos' = Exercise.pos_of_key(pos);
-            let result':
-              Language.ProgramResult.t(Language.ProgramResult.inner) =
-              switch (result) {
-              | Ok((r, s)) =>
-                ResultOk({
-                  result: r,
-                  state: s,
-                })
-              | Error(e) => ResultFail(e)
-              };
-            schedule_action(
-              Editor(pos', ResultAction(UpdateResult(result'))),
-            );
-          }),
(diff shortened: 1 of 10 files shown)


disconcision:
Feels great on an initial pass. I'm wondering if we should also suppress automatic display of indeterminate forms, or at least large ones:

Screenshot 2025-07-24 182904

we could use the Abbreviate by default, like the probe displays do, with more available on-toggle. perhaps this toggle could be combined with the stepper somehow, ie if regular eval gets stuck you can try to keep indeterministically stepping it? this isn't fully baked. but i do think we should suppress large result displays by default even if the method is rough

(post splices Abbreviated displays will use folds)

disconcision (approved):
lgtm, other than the proliferation of multiple apostrophe ending function names, but i cant cast the first stone there.

pumped to have the calculating anim / greying out back.

Negabinary:
> lgtm, other than the proliferation of multiple apostrophe ending function names, but i cant cast the first stone there.

Thanks for reminding me to delete one of them.

src/web/view/ExerciseMode.re
```diff
@@ -377,39 +377,45 @@ module Update = {
...
-      ~timeout=_ => {
-        let _ =
-          Exercise.map_stitched(
-            (pos, _) =>
-              schedule_action(
-                Editor(
-                  pos,
-                  ResultAction(UpdateResult(ResultFail(Timeout))),
+
+    if (List.length(worker_request^) > 0) {
```
cyrus-: move check into WorkerClient.request

cyrus- (changes requested):
can be merged after minor change

cyrus- (approved):

PR #1831, opened 2025-07-25
Dynamic witnesses

MaxCarroll0:
Automated generation of dynamic witnesses for static type errors and also searching for runtime errors in dynamic code. Based upon [Seidel et al](https://arxiv.org/pdf/1606.07557).

Can be combined with type slicing #1830 to produce potentially more concise error highlighting for static errors.

This more generally implements a form of non-deterministic evaluation of programs with holes with further applications to property testing among others.

Further details on functionality and motivation found in [my dissertation](https://github.com/MaxCarroll0/bachelors-dissertation/blob/main/Type-Error-Debugging-in-Hazel.pdf) (sections 3.5-3.6).

Also implements a simple form of hole substitution (without hole closures).

## Checklist
- [x] Indeterminate evaluation & generation of dynamic witnesses
- [ ] Remove type slicing code
- [ ] Update to latest Hazel
- [ ] Toggle holes to be indeterminately evaluated
- [ ] More intuitive UI for retrieving dynamic witnesses of static errors
- [ ] Associate static errors with cast errors more directly
- [ ] UI demonstrating instantiations & custom instantiation direction

PR #1839, opened 2025-07-29
fix stepper value classification

### LINKED ISSUE #1632: Stepper classifies Indet terms as Values
This differs from the results from the value checker (which is correct).

For example, the result of the following is a failed cast classified as a value by the stepper but is an indet (obviously and confirmed by the value checker):
```
let sumList : [Int] -> [Int] = fun xs -> case xs 
  | [] => []
  | h1 :: h2 :: t => h1 + h2(sumList)(t) 
end in sumList(?::?::?)
= (? + (? : ? -> ?)(<sumList>)(?)) : [?]
```

MaxCarroll0:
Fixes #1632

### DIFF
--- a/src/language/dynamics/stepper/EvaluatorStep.re
+++ b/src/language/dynamics/stepper/EvaluatorStep.re
@@ -265,7 +265,7 @@ module Decompose = {
     let req_final = (cont, wr, d) => {
       (
         switch (cont(d)) {
-        | Result.Indet => Result.BoxedValue
+        | Result.Indet => Result.Indet
         | Result.BoxedValue => Result.BoxedValue
         | Result.Step(objs) =>
           Result.Step(List.map(EvalObj.wrap(wr), objs))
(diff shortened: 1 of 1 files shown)


Negabinary (approved):
changes look good to me

Out of interest what was your application where you needed to tell whether something was a value/indet? (I've been struggling to work out what the difference philosophically represents and was wondering if your application might provide clarity)

MaxCarroll0:
I was writing a non-deterministic version of the stepper as part of #1831 which would look at indeterminate terms and instantiate holes within them to continue evaluation. So effectively it would continue evaluation if the result was either a Step or an Indet term (but finish if it is a value).

The stepper would classify some of it's results as boxed values when they were actually indeterminate (and hence this search/evaluation being terminated early).

MaxCarroll0:
Basically, it makes sense to have consistent behaviour between the value checker and stepper: 
If the value checker returns `Indet` then so does the Stepper (as does `Value` <-> `BoxedValue`) and finally if the value checker returns `Expr` then the Stepper returns a list of Steps.

PR #1858, opened 2025-08-03
Add dynamic mode to type projector

7h3kk1d:
# Add dynamic mode to type projector

Adds a **Dynamic** mode to the type projector, alongside Expected and Self. It infers the type of the values that actually flowed through an expression at runtime, and colours the parts of that type runtime supplied rather than statics.

Screenshot From 2025-11-19 13-11-50

## How it works

**Inferring the type.** `DynamicTypInfer` runs statics over each of a probe's sample values in the program's context and meets the results. It returns an `option`, so a set of samples with inconsistent types is distinguishable from success. ~~It has a `.rei`: `type_of_sample` and the `or_unknown` fallback are steps of the other two, not API.~~

It has a `.rei` exposing that and `segment_and_dynamic_ids`, which prints an inferred type against the static one; `type_of_sample` is a step of the first and stays internal. Neither decides what to show when nothing could be inferred — that is the projector's call, below.

**Deciding what turns green.** Only the components runtime supplied go green, not the whole type. Given `(fun x -> ^^statics((1, x)))(true)`, statics knows `(Int, ?)` and runtime says `(Int, Bool)`, so `Bool` is green and `Int` stays the ordinary type colour. `Typ.diff` decides this by walking the two types together and returning the ids in the dynamic one that the static one does not account for.

Two things make that harder than it sounds:

- *The comparison has to run on the types as they will be laid out, not as they arrive.* Rendering desugars `Sig`, inserts defensive parens as real nodes, and pads out id slots — and those added nodes carry tokens of their own. Diffing the types as the caller passed them cannot name their ids, which is why parentheses inside a wholly runtime-derived type used to render in the static colour.

  ~~Both types now go through `ExpToSegment.normalize_typ` first. It is not idempotent, so the normalized type is rendered once and handed back together with the ids to colour, rather than re-rendered later.~~

  ~~That state is a type now: `ExpToSegment.PreparedTyp` is abstract, buildable only by `prepare`, and carries the settings it was prepared for — so "prepare once, render under those settings" is held by the type rather than by a comment. It never leaves `ExpToSegment`; projectors get a single `utility.typ_to_seg_with_diff_ids`, which prepares both sides, diffs them, and returns the one render those ids describe.~~

  That state is a type now, and it lives in `TypToSegment`: the prepared type paired with the settings it was prepared under, built only by `prepare` and named nowhere in the `.rei`, so "prepare once, render under those settings" is held by the module rather than by a comment. Projectors get a single `utility.typ_to_seg_with_diff_ids`, which prepares both sides, diffs them, and returns the one render those ids describe.
- *Statics builds types with `Typ.temp`,* which stamps every node with the same `Id.invalid` sentinel. Left in, the whole set of ids collapses onto whichever tile the renderer emits first — which on a partially-known type means turning a component statics supplied green instead. The sentinels are replaced where a type's ids become the ids of rendered tokens.

`Typ.diff` also expands type aliases through `~ctx` to decide whether they differ, but returns the alias token's own id rather than the expansion's, since an alias renders as a single token. It terminates on cyclic alias chains. Sums get fine-grained diffing via `ConstructorMap.venn_regions`, falling back to colouring the whole dynamic Sum when the static type has constructors it lacks.

**What you see.** The tiles named by that id set get a `dynamic` class, threaded through `ProjectorView` and `Code.re` as a `~classes` callback, and `proj-type.css` renders those green. When statics are absent — switched off, or briefly mid-edit — the projector says so instead of showing `?`, which would claim the type is unknown when what is unknown is whether we looked.

The arrow, its tooltip and the content are one decision, so one cascade in `TypeProj` settles all three. The dynamic reading carries the type runtime supplied beside the static type it is coloured against, and `None` where runtime supplied nothing: a probe that has sampled nothing is not a probe that observed `?`. The renderer is what resolves that case, showing the static type — which diffs against itself, so nothing goes green.

**Probes under nested ascriptions.** In `3 : Int : Int` the two ascriptions are consistent, so the transition meets them and rewrites to a single `3 : Int` — and the inner node is gone. A probe sitting on it goes too, and reports nothing. That matters for a feature whose only input is probe samples.

~~Ascription transitions now collect the value through a writer monad in `Ascriptions.re` and report it as a `RecordAscriptionProbe` side effect, so `^^probe(3 : Int) : Int` records `3`. Pattern matching deliberately opts out with `Sample.no_targets`: the scrutinee is already final there, so its probes fired at the stack evaluation actually passed through, and collecting again would double-count.~~

Replaced by one line in `Ascriptions.re`: the collapsed ascription keeps the inner node's id, so the observation span the evaluator already opens for that id closes on the final value, and every level of a chain gets its sample. The writer-monad plumbing and `RecordAscriptionProbe` are gone.

## Testing

- `Test_DynamicTypInfer.re` — inference from probe samples (basic types, meet behaviour, inconsistent samples, user-defined ADTs, unevaluated branches), and colouring driven through the same composition the projector runs.
- ~~`Test_DynamicTypIds.re` — properties tying the green ids to the render: every dynamic id is actually emitted, a wholly runtime-derived type has every tile green, a type runtime merely confirmed has none, and a prepared type already carries every id its rendering consumes.~~
- ~~`Test_DynamicTyp.re` — end-to-end dynamic type rendering.~~
- `Test_TypToSegment.re` — took over both: which tokens of a printed type came from runtime (tuples, arrows, sums, aliases), and the properties tying the green ids to the render — every dynamic id is actually emitted, a wholly runtime-derived type has every tile green, a type runtime merely confirmed has none, and a prepared type already carries every id its rendering consumes.
- `Test_Typ.re` — `Typ.diff` unit cases and the identity property.
- `Test_Evaluator_Probes.re` — probe attribution under recursion and down ascription chains: one sample per call, calls told apart by their call stack, every level of a chain sampling the value.

Driven in the browser too: green on a partial diff, and the unavailable state with Types switched off.

## Notes

The projector is not offered on type positions: `Info.typ` carries no expected or self type and types are not evaluated, so every mode would render `?`.

The mode arrows are pinned to the bundled `--code-font` — left to a fallback they rendered differently per browser. That font has no bidirectional double arrow, so the types-agree mode shows `↔` rather than `⇔`.


src/language/term/Typ.re
```diff
@@ -561,16 +576,34 @@ let rec match_synswitch = (t1: t, t2: t) => {
...
+ */
+let consistent_join = (ctx: Ctx.t, tys: list(t)): t => {
+  join_all(
+    ~inconsistent=Unknown(Inconsistent) |> temp,
+    ~empty=Unknown(SynSwitch) |> temp,
+    ctx,
+    tys,
+  )
+  |> Option.value(~default=Unknown(SynSwitch) |> temp);
+};
```
7h3kk1d: This could be used for all the places we're using join to synthesize a type from inconsistent internal errors.

7h3kk1d:
Update documentation slides.

cyrus- (changes requested):
update projectors doc slide with new behavior for type probes

src/haz3lcore/projectors/ProjectorPerform.re
```diff
@@ -195,18 +195,14 @@ let go =
...
-              map,
+            ListUtil.update_nth(idx, map, ((id, entry: Refractors.entry)) =>
+              (
+                id,
+                Refractors.{
+                  kind: entry.kind,
+                  model: new_model,
+                },
+              )
             ),
```
7h3kk1d: Fix for https://github.com/hazelgrove/hazel/issues/2097

7h3kk1d:
"Add Statics" -> "Add Type probe"

7h3kk1d:
The hover text on mode is added. I started doing the resizing and abbreviate isn't implemented on types. Plus the sample lengths are a bit ad-hoc for the length state. I can duplicate all the mouse handling but if we're not going to add some sort of a resizable component this is going to be a maintenance burden. @disconcision if you want to see if our 🤖 overlords can make the type probe resizable that would be great.

cyrus- (dismissed):
1. 

```
let x : ? = 2 * 3 in 
^^probe(x)
```
is not showing the probed value in this branch (works on dev)

2. arrow fonts are inconsistent (down arrow looks weird on Chrome, on Firefox they are all small and display differently - pick a fixed font for those that renders the same)

7h3kk1d:
Converted to draft as this is blocked until #2363

src/language/dynamics/transition/Transition.re
```diff
@@ -1236,36 +1236,45 @@ module Transition = (EV: EV_MODE) => {
       let.wrap_closure _ = (env, d);
       Indet;
     | Asc(d', t) =>
-      switch (Ascriptions.transition(d)) {
+      let (samples, stepped) = Ascriptions.transition(~targets, d);
```
7h3kk1d: **Ascription samples are collected before `req_final`, so a probe can report unevaluated syntax as a runtime value.**

This first `Ascriptions.transition` runs on the raw `d`; only the `None` branch goes on to `req_final`. When it returns `Some`, the samples it emitted hold the pre-evaluation operand (env-substituted, but not reduced), so `DynamicTypInfer.type_of_sample` ends up running statics over *syntax* and the "dynamic" type collapses to the static result type for that sample.

Repro:

```
let f = fun x -> x + 1 in
if true then ^^probe(f(3) : ?) else 2
```

```
$ ./hazel probe --many b.hz
if true then ⟦f(3) : ?⟧ else 2     ≡ ^^fold((fun x -> x + 1))(3)
```

Drop the `: ?` and the same probe reads `≡ 4`. Only one sample is recorded, so the bogus one is what the UI shows. A `case` rule triggers it identically; a `let` def does not, because it gets no outer ascription — the trigger is an ascription nested inside something `fresh_ascription` wraps (`if` branches, `case` rules, list elements).

Fix: collect only on the post-`req_final` path.

src/language/dynamics/state/EvaluatorState.re
```diff
@@ -252,6 +253,26 @@ let update =
           call_stack,
           add_test(state, instance_report),
         )
+      | RecordAscriptionProbe((id, capture_spec, ascribed_exp)) =>
+        let step = state.step_count;
+        /* Substitute env so a Var body resolves to its runtime value. */
+        let ascribed_exp = Substitution.in_exp(env, ascribed_exp);
```
7h3kk1d: **This is the only `Sample.mk` site that substitutes the environment into the sample value.**

The other three (`RecordPrint` just below, `RecordPatMatch`, `ObsTrace.SpanClose`) pass the value as-is and let `Sample.Env.elide` substitute lazily, per captured binding. Substituting here also breaks `Substitution.re`'s documented PRECONDITION ("the environment should list all free variables (each free variable in the env should just be v |-> v)") — a raw mid-evaluation env is not of that shape — and costs a deep `Exp.replace_all_ids` copy plus a recursive re-substitution for every bound variable it hits, over an env that includes all of `Builtins.env_init`, per ascription sample, inside evaluation.

Two visible consequences, both against a control probe on the same value:

```
let n = 41 in
^^probe((fun y -> y + n) : ?)   ->  fun y -> y + 41     # this path: env inlined
^^probe(fun y -> y + n)         ->  fun y -> y + n      # every other probe
^^probe((fun n -> n + 1) : ?)   ->  fun n' -> n' + 1    # binder alpha-renamed
```

The last one is the one a user would report: the probe shows a function whose parameter has been renamed, purely because the term sits under a nested ascription.

Fix: drop the substitution and let Sample's env elision do it.

src/haz3lcore/projectors/DynamicTypInfer.re
```diff
@@ -0,0 +1,75 @@
+open Language;
+
+/* Infer a type from a single sample value by running statics on it.
+   Uses the provided context so user-defined types are visible. */
+let type_of_sample = (~ctx: Ctx.t, sample: Sample.t): option(Typ.t) => {
+  let (info_map, _elab) = Statics.mk(CoreSettings.on, ctx, sample.value);
```
7h3kk1d, line 10: **`type_of_sample` runs statics on an already-elaborated value using the source context's `use_mode`, so the sample gets re-kinded a second time.**

`ctx` here comes from `Info.ctx_of(info.statics)`, i.e. the ambient source mode. `Ctx.re:45` documents the field as `use_mode: option(Operators.mode) // None if elaboration has already occurred`, and `CachedStatics.re:101` passes `Builtins.ctx_init(is_dynamic_term ? None : Some(Int))` for exactly this reason. With `use_mode = Some(Nat)`, `Operators.replace_literal`'s `(Int(n), _, Some(Nat)) => Nat(n)` arm retypes an `Int` sample as `Nat`.

Repro:

```
let f : Int -> Int = fun x -> x + 1 in
use Nat in
^^statics(f(3))
```

`f` is `Int -> Int` and the value is `4`, so both static modes show `Int` — but Dynamic mode shows **`Nat`**, in green as runtime-derived. (Confirmed in the browser.)

Fix: `Ctx.set_use_mode(ctx, None)` before `Statics.mk`. `TypeProj.re:34`'s `Builtins.ctx_init(Some(Int))` fallback has the same problem.

7h3kk1d (commented):
Three bugs from a review pass, each with a reproduction. Inline below.

src/language/dynamics/transition/Transition.re
7h3kk1d: Fixed in 1382274722: the collapsed ascription keeps the inner ascription's id, so the evaluator's observation span closes on the final value and the collapse-time sample is gone. Pinned by the five ascription-chain tests in the same commit.

src/language/dynamics/state/EvaluatorState.re
7h3kk1d: Made unreachable by 1382274722 (no ascription samples are minted any more, so this arm never runs). The RecordAscriptionProbe effect, the SampleWriter threading in Ascriptions.transition, and util/WriterMonad.re are now dead code and still to be removed.

src/haz3lcore/projectors/DynamicTypInfer.re
7h3kk1d: Fixed in 5e832d1f7f: Ctx.set_use_mode(ctx, None) before Statics.mk.

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -3129,21 +3259,63 @@ let exp_to_segment =
...
+   * Exposed because a caller that wants to reason about the ids in a rendered
+   * segment has to reason about THIS type, not the one it passed in: the parens
+   * and the padded id slots are nodes that only exist after normalization, and
+   * their ids are what the renderer puts on tokens. Diffing the un-normalized
+   * type cannot name them.
+   *
+   * Not idempotent: parenthesize_typ mints a fresh id for each Parens it adds,
+   * so normalizing twice yields two different types. Normalize once, then pass
+   * the result to normalized_typ_to_segment. */
+let normalize_typ = (~settings: Settings.t, typ: Typ.t): Typ.t =>
```
7h3kk1d: We shouldn't call this normalize since it's different than type normalization.

src/haz3lcore/projectors/implementations/TypeProj.re
```diff
@@ -24,11 +24,40 @@ let totalize_ty = (expected_ty: option(Typ.t)): Typ.t =>
...
   };
 
+/* The segment to display in Dynamic mode, and the ids of its tokens that
+   came from runtime. Rendered here rather than by the caller because the
+   dynamic_ids only describe this one render -- see DynamicTypInfer. */
+let get_dynamic_segment =
+    (utility: utility, info: info): (Base.segment, Id.Set.t) => {
+  let ctx =
+    Option.map(Info.ctx_of, info.statics)
+    |> Option.value(~default=Builtins.ctx_init(Some(Int)));
```
7h3kk1d: I'm nervous about a default int ctx.

src/haz3lcore/projectors/DynamicTypInfer.re
```diff
@@ -0,0 +1,79 @@
...
+/* Like dynamic_typ_of_samples, but defaults to Unknown(Internal) on failure. */
+let dynamic_typ_of_samples_or_unknown =
+    (~ctx: Ctx.t, samples: list(Sample.t)): Typ.t =>
+  dynamic_typ_of_samples(~ctx, samples)
+  |> Option.value(~default=Typ.fresh(Unknown(Internal)));
+
+/* The segment to show in Dynamic mode, and the ids of its tokens that came
+     from runtime rather than from statics.
+   *
+   * Both types are normalized first -- ExpToSegment.normalize_typ, the same
```
7h3kk1d: This is not type normalization.

src/haz3lcore/projectors/DynamicTypInfer.re
```diff
@@ -0,0 +1,79 @@
...
+   * The normalized dynamic type is rendered once and returned: normalize_typ is
+   * not idempotent (it mints a fresh id per added Parens), so re-normalizing or
+   * re-rendering would produce a segment the dynamic_ids do not describe. */
+let displayed_segment_and_dynamic_ids =
+    /* Injected rather than called directly: ExpToSegment sits downstream of
+       the projectors, so this module cannot name it. Pass
+       utility.normalize_typ and utility.render_normalized_typ, which are
+       built from one settings value so both halves agree. */
+    (
+      ~normalize: Typ.t => Typ.t,
```
7h3kk1d: Would we benefit from a abstract type or wrapper somewhere that marks these types as having been padded and parenthesized?

src/haz3lcore/projectors/ProjectorBase.re
```diff
@@ -34,6 +34,14 @@ type utility = {
...
   /* Convert a term to a segment */
   term_to_seg: (~inline: bool, Any.t) => Base.segment,
+  /* The two halves of term_to_seg for types, split apart. A caller that
+   * needs to reason about the ids in a rendered segment must normalize
+   * first and reason about the NORMALIZED type: parens are inserted as
+   * real nodes and id slots are padded during normalization, and those are
+   * the ids the renderer puts on tokens. Normalization is not idempotent,
+   * so normalize once and render that. */
+  normalize_typ: (~inline: bool, Typ.t) => Typ.t,
+  render_normalized_typ: (~inline: bool, Typ.t) => Base.segment,
```
7h3kk1d: Not normalization

src/language/dynamics/transition/Ascriptions.re
```diff
@@ -27,11 +27,21 @@ let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
...
         // This is only necessary because sometimes we add two ascriptions and aren't marking it as a non-value
         when Typ.is_consistent(Ctx.empty, Typ.unroll(t), Typ.unroll(t')) =>
       switch (Typ.meet(Ctx.empty, Typ.unroll(t), Typ.unroll(t'))) {
-      | Some(t) => Some(recur(Asc(e, t) |> DHExp.fresh))
+      | Some(t) =>
+        /* Collapse under e's id, per ID PRESERVATION above: the evaluator
+           opens an observation span for a target id and closes it on the
+           final value (Evaluator.eval_3_record_probe_sample), so keeping the
+           id here is what gets this ascription a sample. Minting one at
+           collapse time instead would capture the operand unevaluated. */
```
7h3kk1d: Comment probably unnecessary.

test/Test_DynamicTyp.re
```diff
@@ -0,0 +1,409 @@
...
+          ),
+        );
+      check(
+        list(region),
+        "Int static, String dynamic",
+        [s("(Int,"), d("String"), s(")")],
+        result,
+      );
+    },
+  );
```
7h3kk1d: For the tests in this file that don't need special hidden info on the type (like constructors) we should use the parser to construct the types so the tests are easier to read.

test/Test_DynamicTypIds.re
```diff
@@ -0,0 +1,188 @@
...
+   DynamicTypInfer.displayed_segment_and_dynamic_ids: normalize both, diff the
+   normalized forms, render the normalized dynamic type once.
+
+   The two id sets are deliberately different. `all` includes Grout and
+   Secondary, because a runtime-derived node can legitimately render as Grout --
+   an Unknown does, with show_unknown_as_hole off -- and including it is harmless.
+   `tiles` is what Code.re actually colours, so it is the right set to
+   require full coverage of. */
+let dynamic_ids_and_rendered =
+    (~static_typ: Typ.t, ~dynamic_typ: Typ.t): (Id.Set.t, Id.Set.t, Id.Set.t) => {
```
7h3kk1d: let's return a record here

src/haz3lcore/pretty/ExpToSegment.re
7h3kk1d, line 1: A lot of this file is now padding stuff for the dynamic type inference. Think about which files would benefit from rei files for more explicit APIs in this pr and if anything should/can move.

src/haz3lcore/pretty/ExpToSegment.re
7h3kk1d: 🤖 Agreed, and it's gone rather than renamed. `normalize_typ` and `normalized_typ_to_segment` are no longer exposed at all; what a caller sees now is `typ_to_segment` and `typ_to_segment_with_diff_ids`, neither of which asks them to think about a separate pass. The three steps live behind `PreparedTyp.prepare` — see the reply on DynamicTypInfer.re:60.

src/haz3lcore/projectors/implementations/TypeProj.re
7h3kk1d: 🤖 Removed. Inference needs the expression's own context — the types a sample mentions are the ones in scope where it was sampled, so a builtin-only stand-in reports them as something else — and `get_dynamic_segment` now takes `~ctx` rather than defaulting one.

That left the question of what to show when statics are absent, which happens when statics are switched off and briefly while an edit is being checked. `typ_view` now matches on `info.statics` up front and renders `no type information`, in every mode rather than only Dynamic: `?` in Expected/Self was the same claim in a quieter form, since what's unknown there is whether we looked, not the type.

Not browser-checked yet, and the wording is a guess — happy to change it.

src/haz3lcore/projectors/DynamicTypInfer.re
7h3kk1d: 🤖 Right — fixed by removing it rather than renaming. See ExpToSegment.re:3275.

src/haz3lcore/projectors/ProjectorBase.re
7h3kk1d: 🤖 Fixed. These two fields are gone entirely; `utility` now has a single `typ_to_seg_with_diff_ids` that exchanges only plain types and segments. See DynamicTypInfer.re:60 for why it ended up as one operation rather than two renamed ones.

src/haz3lcore/projectors/DynamicTypInfer.re
7h3kk1d: 🤖 Yes — `ExpToSegment.PreparedTyp`. Abstract, built only by `prepare`, and carrying the settings it was prepared for, so the "prepare once, render with the same settings" rule is held by the type instead of by a comment. Its signature is `prepare` / `to_segment` / `diff_ids` / `ids_sufficient`, deliberately with no accessor back to `Typ.t`: a prepared type can't be unwrapped and prepared again, and `diff_ids` can only be handed two prepared sides.

It could not go in `utility`, though. `ProjectorBase` can't name `ExpToSegment` — there's a real cycle, two ways round:

```
ProjectorBase → FoldProj → ExpToSegment
ProjectorBase → ProjectorInit → Refractors → ZipperBase → Zipper → Dump → MakeTerm → ExpToSegment
```

A type declared upstream of both could be mentioned there, but up there it can't reach `parenthesize_typ` or `pad_typ_ids`, so its constructor would be callable by anyone and the wrapper would be a label rather than an enforcement.

So the two callbacks collapsed into one `typ_to_seg_with_diff_ids` instead. The prepared type never crosses the boundary, which means the window it would have had to guard doesn't exist there, and `DynamicTypInfer` no longer threads two callbacks that have to be kept in step.

One gap worth naming, which predates this thread: `ids_sufficient` compares padding against `necessary_ids`, and both sides of that comparison come from `necessary_ids` — so it can't catch `necessary_ids` drifting from what `typ_to_pretty` actually spends, which is what the old `raise_if_padding` flag caught. Co-location plus the unit count pins are still the real guard, which is why I left `necessary_ids` beside the `pad_ids` calls (see the reply on ExpToSegment.re:1). The honest fix is a property saying every id in the rendered segment appears in the prepared type; it needs the type to have distinct ids, since `uniquify_repeated_tiles` freshens duplicates, so it isn't a one-liner. Say the word and I'll do it here rather than leaving it.

src/language/dynamics/transition/Ascriptions.re
7h3kk1d: 🤖 Removed. The file header's ID PRESERVATION FOR PROBES section already states the rule, and states it for every arm rather than one.

test/Test_DynamicTyp.re
7h3kk1d: 🤖 Done — there's a `typ` helper over `Parser.to_segment(~root=Typ)` + `MakeTerm.for_projection`, so the product, arrow and alias tests now read `typ("(Int, ?)")`, `typ("Int -> ?")`, `typ("[Int]")` and so on. About a hundred lines lighter.

Kept explicit where the construction carries something the parser doesn't express: the sum tests, for the constructor annotations you called out, and the `Ctx.extend_tvar` in the alias tests — though the types inside those contexts are parsed now.

test/Test_DynamicTypIds.re
7h3kk1d: 🤖 Done — `{dynamic_ids, emitted, tiles}`. The comment explaining why `emitted` and `tiles` differ now sits on the type.

src/haz3lcore/pretty/ExpToSegment.re
7h3kk1d: 🤖 Added `DynamicTypInfer.rei`. Two of its four values — `type_of_sample` and the `or_unknown` fallback — were only ever steps of the other two, so hiding them leaves the module stating what it's for.

Nothing moved, and I want to flag that as a decision rather than an omission. The padding code is the obvious candidate, but `necessary_ids` carries a comment saying it must sit beside the `pad_ids` calls in `typ_to_pretty` that spend it, and that argument still holds: nothing currently checks the two against each other, so co-location is the only thing keeping them in sync (detail in the reply on DynamicTypInfer.re:60). Moving it to a padding module would make that drift cheaper, not dearer. Happy to move it as part of closing that gap properly, but not before.

No `ExpToSegment.rei` either. At 3331 lines with consumers across haz3lcore, web, CLI and test, writing its signature is a mechanical diff that would dominate review of this branch and touch code this PR doesn't otherwise. I'd rather that were its own branch — tell me if you'd sooner have it here.

src/haz3lcore/projectors/implementations/TypeProj.re
7h3kk1d: 🤖 Browser-checked now, so replacing the caveat above.

Turning off Types in the Semantics settings does reach this branch, and the cell reads `unavailable` with the sentence as a tooltip. The first wording, `no type information`, wrapped onto two lines and made the probe box taller than any type it stands in for, so `deafa9035c` shortened it.

Round trip checked too: Types off → `unavailable`, Types back on → the dynamic type returns with its colouring intact.

src/haz3lcore/projectors/implementations/TypeProj.re
```diff
@@ -46,45 +47,130 @@ module M: Projector = {
...
+      | (Some(self), Some(expected)) => Typ.fast_equal(self, expected)
+      | _ => false
+      }
+    );
+
+  /* What the cell shows: the model says which reading was asked for, this
+     says which it came to. Runtime carries no type -- its segment is built
+     from the samples and comes with the ids to colour. */
+  type content =
+    | FromRuntime
```
7h3kk1d: This should just include the dynamic type and the static type on it I think.

src/haz3lcore/projectors/implementations/TypeProj.re
```diff
@@ -46,45 +47,150 @@ module M: Projector = {
...
-    | Self => statics |> self_ty
-    | Expected => statics |> expected_ty
+  /* Whether statics has an expectation to show. An expression in synthetic
+     position has none, so Expected mode falls back to Self and toggling
+     skips it. */
+  let has_expected = (statics: Info.t): bool =>
+    switch (expected_ty(statics)) {
+    | None => false
+    | Some(ty) => !Typ.is_syn(ty)
     };
```
7h3kk1d: Can this be put into the expected_ty function and just return none otherwise? What triggers the arrow can just be whether it's some

7h3kk1d: 🤖 Done — `expected_ty` decides it now: `Typ.is_syn(ana) ? None : Some(ana)`, and `has_expected` is gone.

`readings_agree` drops its leading conjunct, since the `(Some, Some)` pattern carries it. The `Expected` arm of `reading` matches the option rather than guarding on a predicate and then putting the type back through `totalize_ty`. `update` binds through the statics option and asks `Option.is_some`.

No behavior change: `Option.is_some` of the new `expected_ty` is exactly the old `has_expected`, and every arm that could reach a syn expectation was already guarded by it. Build clean and `make test-quick` green; not clicked through in the browser yet.

PR #1859, opened 2025-08-05
Reflexivity

### LINKED ISSUE #1843: Some polyeq noop steps show up in stepper
Image

This step should probably be hidden

Negabinary:
image

When proof is enabled, allows the user to step any syntactically equal expression to true by clicking a purple step.

### DIFF
--- a/src/language/proof/ProofHacks.re
+++ b/src/language/proof/ProofHacks.re
@@ -315,7 +315,8 @@ let find_refls = e => {
       ~f_exp=
         (cont, exp) => {
           switch (exp |> Exp.term_of) {
-          | BinOp(Poly(Equals), e1, e2) when Exp.fast_equal(e1, e2) =>
+          | BinOp(Poly(Equals), e1, e2)
+              when MatchExp.match_exp([], [], e1, e2) |> Option.is_some =>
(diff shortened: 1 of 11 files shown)


src/language/proof/ProofHacks.re
```diff
@@ -214,3 +214,115 @@ let rec get_inductive_hypotheses = (m: Statics.Map.t, t: Typ.t, p: Pat.t) => {
...
+      ? true : CoCtx.has_any(with_coctx, bvn);
+  };
+  let replace_exp = (in_exp): Exp.t =>
+    replace_exp(replace, replace_coctx, with_exp, with_coctx, in_exp);
+  Exp.map_term(
+    ~f_exp=
+      (continue, exp) => {
+        let (term, rewrap) = Exp.unwrap(exp);
+        switch (term) {
+        | _ when Exp.fast_equal(exp, replace) =>
```
cyrus-, line 235: leave a comment about this not being alpha-equivalence

cyrus- (changes requested):
- when a term is both steppable and reflexive, show it only in green not both
- change proof step for reflexivity to say "reflexivity"

cyrus- (approved):

PR #1863, opened 2025-08-06
Fix term appearing in its own list of ancestors

### LINKED ISSUE #1862: Term appears in its own list of ancestors (in some cases)
An example case is:
```
type x = (
  a = Int, 
  b = Float
)
in
```
The ancestors of the tuple type `(a = Int,   b = Float)` has itself at index 0. This is also observed if the term is a basic type, and potentially more cases.

Likely an issue with implementations not using the helpers and using the helpers to add stuff to the info map... if they do not use the helpers, then they will be using the shadowed version of ancestors, which has the current node already included.

russell-rozenbaum:


cyrus-:
@russell-rozenbaum this seems to only have a debug command change but no actual fix, so closing for now but feel free to reopen or leave an issue on GitHub.

PR #1865, opened 2025-08-07
Per-character Insert/delete actions now effect all delimiters equally

disconcision:
Previously, insert/delete actions affecting a polytile delimiter either failed or acted on the delimiter as a whole. For example, backspacing on an `in` which is part of a `let` would remove the whole `in`, or inserting `x` given `le|t x = 4 in x` would fail, or inserting `t` on `let| x = 1 in x` would result in a separate `t` token being inserted.

Give the recent backpack changes (#1805), the monotile delimiter logic can now be generalized to cover polytiles as well. Now, insert/delete actions will result in the affected delimiter being separated from its tile.

To support this, expansion and matching (drop/put-down) behavior has been simplified. **There is no longer any concept of delayed expansion or matching**. Both are now instant, and require no special character insertion/positioning to trigger (although they still are only triggered local to caret). For example, when entering the variable `letter` there will be an expansion after the first `t` is entered, but then it will un-expand after the second `t`.

Other changes:

- The concept of duotile merging/splitting (e.g. `[|]` splitting on insertion, merging on delete, etc) has been eliminated as a special case, subsumed into the new general shard logic.

- The concepts of expansion and gloming (where a token becomes matched with an existing incomplete tile, previously called barfing) have been consolidated and moved upstream; both now happen in exactly one place, in the core Insert.insert_shard function. Calling this function will handle expansion or gloming of the inserted token.

- There is a new concept, Triggers, defined in `Triggers.re` which define a list of syntax operations to be applied automatically after insertion actions. This is a shallow and un-opinionated  abstraction, designed to abstract both projector and livelit invocation, but technically permitting arbitrary rewrites. We might want to make it more restrictive in the future

- The model for within-token caret positioning logic has been simplified. Previously, the caret field on Zippers retained which polytile delimiter the caret was on as well as an index into that delimiter (`Inner(delimiter_idx, token_idx)`). I realized (very gradually) that there is no reason to retain this, and in fact the only purpose served seemed to be creating bugs due to entirely self-imposed inconsistency. **The caret field now only tracks the internal index**.

- All string-only logic from Form.re has bee moved to Mold.re, in particular token identification predicates. String functions operating on tokens throughout the codebase have been pointed to Token to facilitate future refactorings, including using Unicode and moving to using lexed tokens in labels instead of strings. Molds.re has been integrated into Form.re as the boundary between these has always been vague.

- This makes a deep change to the way tile assembly works. Delimiter (re)matching in hazel is universally triggered by a shard being unselected. Unselection triggers a process called assembly. During assembly, one or more parts of the cursor segment are pinched off into children of the tile being assembled. Regrouting now happens internal to this process, which should prevent many of the eternally recurring `nonconvex segment` exceptions. This probably unnecessarily slows down non-editing operations involving selections; if this turns out to be the case, we can add a flag for regrouting. Regrouting of the focal segment itself though is still not handled internally, as there are still-persisting special cases around the positioning of grout in multiple Insert/Destruct sub-cases. Hopefully one day they will be unnecessary.


disconcision:
@Negabinary @7h3kk1d soliciting functional opinions and opinions on the division of Token.re and Form.re. The latter is intended to be somewhat of a transitional state; we probably want to do a future refactor which turns Token.re into a tokenize proper, perhaps driven by the Menhir lexer.

7h3kk1d:
I did a couple euler problems and it feels substantially better imo.

7h3kk1d:
One issue I'm having trouble figuring out why is this isn't inserting an if when I type `if` then a space.
[Screencast From 2025-08-11 13-55-05.webm](https://github.com/user-attachments/assets/553e2d92-f98b-4095-8b1f-fdd29c2b9c86)

disconcision:
it's a bug. looks like what's happening here is interaction between the initial grout insertion relative to the caret, and the order of expansion and regrouting.        

Edit: Fixed

cyrus- (approved):

Issue #1522, opened 2025-02-16
Display run-time errors for invalid operations


cyrus-:
Currently, invalid operations like `0/0` and invalid results from built-ins are not displayed in any way in the result.

PR #1870, opened 2025-08-08
Equality

Negabinary:
Adds a parameterized equality function that separates out syntactic and semantic equality, but also lets us specify different kinds of equality. Replaces fast_equal, and the filter matcher, could eventually be used to implement match_exp.

image

Currently failing 51 tests.


src/language/term/Equality.re
```diff
@@ -0,0 +1,634 @@
...
+    typ: typ(Alphas.empty, Alphas.empty),
+    tpat: (tp1, tp2) => tpat(tp1, tp2) |> Option.is_some,
+    rul: rul(Alphas.empty, Alphas.empty),
+    any: any(Alphas.empty, Alphas.empty),
+  };
+};
+
+// Useful defaults
+
+let syntactic =
```
7h3kk1d: Looking at the failing tests and it seems for at least some it's things that want syntactic equality _but_ they want to ignore parens. I'm wondering instead of having a bunch of optional params if we should have a `Settings` record like `ExpToSegment` and give a default settings. That way you could use `SyntacticSettings` and modify the one field you care about.

src/language/term/Term.re
```diff
@@ -338,6 +341,9 @@ module Exp = {
 
   include TermBase.Exp;
 
+  let fast_equal = Equality.syntactic.exp;
```
7h3kk1d: This used to ignore parens and now it's not. I'm looking at the test failures but I'm not sure what other implementation issues might pop up from this.

test/evaluator/Test_Evaluator_Prelude.re
```diff
@@ -3,10 +3,23 @@ open Language;
...
 
-let testable_exp = (~ignore_constructor_types=?, ()) =>
+let testable_exp = (~ignore_constructor_types=false, ()) =>
   testable(
     Fmt.using(Exp.show, Fmt.string),
-    DHExp.fast_equal(~ignore_constructor_types?),
+    Equality.equality(
+      ~type_alpha=false,
+      ~exp_alpha=false,
+      ~ignore_wrappers=false,
```
7h3kk1d: One consequence of using ignore_wrappers=false is that we can't use parens in test assertions for precedence/tuple grouping because that's expected in test output. This is related to the idea that we have ASTs that have no surface syntax representation (ExpToSegment inserts paren nodes on serialization)

src/language/term/Equality.re
```diff
@@ -0,0 +1,634 @@
...
+                  "full closure equality has not been implemented yet",
+                ),
+                c1,
+                c2,
+              )
+        )
+      | None => false
+      }
+    | (FixF(_, e, _), _) when ignore_fixpoints => exp'(e, e2)
+    | (_, FixF(_, e, _)) when ignore_fixpoints => exp'(e1, e)
```
7h3kk1d, line 244: Does this not have capture issues?

test/evaluator/Test_Evaluator_Prelude.re
```diff
@@ -3,10 +3,23 @@ open Language;
...
+      ~ignore_casts=false,
+      ~ignore_constructor_types,
+      ~ignore_function_names=false,
+      ~ignore_filters=false,
+      ~ignore_unknown_provenance=false,
```
7h3kk1d: Same issue with unknown provenance and surface representation. We're using `?` everywhere to mean unknown but not specifying a provenance.

src/language/term/Equality.re
```diff
@@ -0,0 +1,634 @@
...
+      };
+
+  let combine: (t, t) => t = (@);
+};
+
+let equality =
+    (
+      ~type_alpha: bool,
+      ~exp_alpha: bool,
+      ~ignore_wrappers: bool,
```
7h3kk1d: I think we probably want to be able to differentiate paren wrapping which is _mostly_ surface syntax for parsing (and should be reinserted on serialization and dynamic error holes. I am confused what the dynamic error holes are being used for right now though since it doesn't seem like it shows up anywhere.

7h3kk1d (commented):
I added https://github.com/hazelgrove/hazel/pull/1873 to address the tests. Still looking through the pr for any other thoughts as well.

src/language/term/Term.re
7h3kk1d: My personal preference for this pr would be to instantiate `Equality` with the params to make this the same as it used to be (ignore wrappers). Then in a separate pr go through and remove every reference to fast_equal.

src/language/term/Equality.re
```diff
@@ -0,0 +1,634 @@
...
+
+                "Equality was a mistake" - Andrew Blinn, 2025
+
+ This module aims to capture the many different versions of equality that we need in Hazel under
+ one function with many parameters. It is rare that a simple (==) check will suffice, often we need to
+ do things like:
+
+     - ignore parantheses or other wrapping forms
+     - not traverse big environments for performance reasons
+     - have the equality check look up free variables in an environment
```
7h3kk1d, line 25: We might want to mention something about id-relevance to equality.

src/language/term/Equality.re
```diff
@@ -0,0 +1,634 @@
...
+let semantic =
+  equality(
+    ~type_alpha=true,
+    ~exp_alpha=true,
+    ~ignore_wrappers=true,
+    ~ignore_casts=true,
+    ~ignore_constructor_types=true,
+    ~ignore_function_names=true,
+    ~ignore_filters=true,
+    ~ignore_unknown_provenance=true,
```
7h3kk1d: I'm not sure what semantic equality _means_ here. Since the constructor types and casts are definitely semantically relevant.

src/language/term/Equality.re
Negabinary: I like this idea, made a record.

src/language/term/Term.re
Negabinary: oh good spot, I didn't notice that was different to syntactic

test/evaluator/Test_Evaluator_Prelude.re
Negabinary: ok let's set it to true then

src/language/term/Equality.re
Negabinary: oh, it's awful, yeah. I think it's only used in stepper step-skipping logic so it's not used anywhere too critical.

test/evaluator/Test_Evaluator_Prelude.re
Negabinary: will ignore provenance in tests

src/language/term/Equality.re
Negabinary: separated them; dynamic error holes might not be currently used because we don't give any error feedback (see #1522)

src/language/term/Equality.re
Negabinary: done

src/language/term/Equality.re
Negabinary: I can get behind changing types and casts in this definition. I don't think this version of equality is used in enough places for me to have too good an idea of what "cluster" of equalities this is meant to be in the middle of.

test/evaluator/Test_Evaluator_Prelude.re
```diff
@@ -3,10 +3,20 @@ open Language;
...
+let testable_exp = (~ignore_constructor_types=false, ()) =>
   testable(
     Fmt.using(Exp.show, Fmt.string),
-    DHExp.fast_equal(~ignore_constructor_types?),
+    Equality.(
+      equality({
+        ...syntactic_settings,
+        ignore_parens: true,
+        ignore_function_names: true,
+        ignore_casts: true,
```
Negabinary: I had to add ignore_casts to get all the tests to pass again.

src/language/term/Equality.re
```diff
@@ -0,0 +1,674 @@
...
+  closures_by_id: bool, /* Currently "false" option is not implemented.
+                           compares closures by their IDs to save time
+                           traversing through massive closures */
+  ignore_filters: bool,
+  ignore_unknown_provenance: bool,
+  use_expr_wildcards: option((ClosureEnvironment.t, Exp.t) => bool),
+  ignore_fixpoints: bool, // Hideously unsound, used to hide function steps in the stepper
+  free_var_handler: option((Alphas.t, string, Exp.t) => bool), // Note[Matt]: to be used in MatchExp
+  env1: option(ClosureEnvironment.t), // The environment to look up variables on the left in
+  env2: option(ClosureEnvironment.t) // The environment to look up variables on the right in
```
7h3kk1d, line 87: Can we move this from `settings` to the equality params?

src/language/term/Typ.re
```diff
@@ -423,7 +423,8 @@ let unroll = (ty: t): t =>
 
 /* Type Equality: This coincides with alpha equivalence for normalized types.
    Other types may be equivalent but this will not detect so if they are not normalized. */
-let equal = (t1: t, t2: t): bool => fast_equal(t1, t2);
+let fast_equal = Equality.semantic.typ;
+let equal = (t1: t, t2: t): bool => Equality.syntactic.typ(t1, t2);
```
7h3kk1d, line 426: Why the two different ones here?

test/evaluator/Test_Evaluator_Prelude.re
7h3kk1d:
```suggestion
        ignore_unknown_provenance: true,
```
The ascriptions are failing because the types are wrong up to unknown provenance.

src/web/app/inspector/CursorInspector.re
```diff
@@ -237,7 +237,7 @@ let common_ok_view =
         view_type(ana),
       ]
     | (_, Ana(Consistent({ana, syn, _})))
-        when Typ.fast_equal(~alpha_equivalence=false, ana, syn) =>
+        when Equality.semantic.typ(ana, syn) =>
```
Negabinary, line 270: Looks like this needs to not use alpha equivalence

src/language/term/Equality.re
```diff
@@ -0,0 +1,677 @@
...
+    | (Deferral(_), _) => false
+    | (Atom(c1), Atom(c2)) => c1 == c2
+    | (Atom(_), _) => false
+    | (Label(l1), Label(l2)) => l1 == l2
+    | (Label(_), _) => false
+    | (LivelitName(s1), LivelitName(s2)) => s1 == s2
+    | (LivelitName(_), _) => false
+    | (MultiHole(xs1), MultiHole(xs2))
+        when List.length(xs1) == List.length(xs2) =>
+      List.equal(any', xs1, xs2)
```
7h3kk1d: Coming across this today. Is this not a form of "provenance" for the hole? We seem to ignore this at the type level but not the expression.

7h3kk1d: And whether EmptyHole should be equal to MultiHole

cyrus- (approved):
lgtm pending a few additional requested comments in our meetings, will be merged as part of #1979

PR #1873, opened 2025-08-10
Fix tests after equality

7h3kk1d:
A lot of this is relaxing the notion of equality. Some of them can be made by changing the test expected value but there's some tradeoffs worth discussing.


test/Test_Elaboration.re
```diff
@@ -3,7 +3,25 @@ open Language;
...
   an equal function (dhexp_eq) and a print function (dhexp_print) */
-let dhexp_typ = testable(Fmt.using(Exp.show, Fmt.string), DHExp.fast_equal);
+let dhexp_typ =
+  testable(
+    Fmt.using(Exp.show, Fmt.string),
+    // This is syntactic with ignore_wrappers=true
+    Equality.equality(
+      ~type_alpha=false,
+      ~exp_alpha=false,
+      ~ignore_wrappers=true,
```
7h3kk1d, line 13: This is being done because we have parens in the expectations for the post-elaboration code. We can remove parens speculatively but we'd either have to write the ast in ocaml syntax or make sure the expression doesn't need it.

test/Test_Elaboration.re
```diff
@@ -3,7 +3,25 @@ open Language;
...
+      ~ignore_casts=false,
+      ~ignore_constructor_types=false,
+      ~ignore_function_names=false,
+      ~ignore_filters=false,
+      ~ignore_unknown_provenance=true, // Backwards compatibility for tests
```
7h3kk1d, line 18: This is because of `?` being used to mark unknown types. If we care about the unknown we need to use the ocaml version of the AST and not the parser.

test/evaluator/Test_Evaluator_Prelude.re
```diff
@@ -9,12 +9,12 @@ let testable_exp = (~ignore_constructor_types=false, ()) =>
     Equality.equality(
       ~type_alpha=false,
       ~exp_alpha=false,
-      ~ignore_wrappers=false,
+      ~ignore_wrappers=true,
       ~ignore_casts=false,
       ~ignore_constructor_types,
-      ~ignore_function_names=false,
+      ~ignore_function_names=true,
```
7h3kk1d, line 15: Same issue here with the expected coming from the parsed code. https://github.com/hazelgrove/hazel/pull/1737 could be used to address some of it but it's more work.

test/statics/Test_Statics_Labeled_Tuple.re
```diff
@@ -246,11 +246,13 @@ let tests = [
                         Inconsistent(
                           Expectation({
                             ana:
-                              prod([
-                                tup_label(label("a"), int()),
-                                tup_label(label("b"), float()),
-                                string(),
-                              ]),
+                              parens(
```
7h3kk1d, line 249: For some tests you can add the parens but it's annoying when you seemingly don't care. I'd rather remove the parens from the ast but I don't think we should do that as part of this pr.

This is the type of issue that comes up though when you have irrelevant stuff strewn through the AST.

Negabinary:
ok, yeah I think I'm happy to just have most of the tests ignore parens, function names, and unknown provenence for now.

PR #1876, opened 2025-08-15
Stepper shows variable substitution dropdowns

joshqwang:
On let binds with only one variable, shows "substitution for (variable name)"
Screenshot 2025-08-15 at 5 09 58 PM

On let binds with more than one variable, function application, and case selection, shows variables substituted when hovered over.

Screenshot 2025-08-15 at 5 10 36 PM

### DIFF
--- a/src/web/www/style/dynamics.css
+++ b/src/web/www/style/dynamics.css
@@ -180,15 +180,24 @@
 
 .stepper-justification {
     width: fit-content;
+    position:relative
 }
 .stepper-justification-sub-vars {
     display: none;
-    position: relative;
-    filter: drop-shadow(3px 3px 2px #00000010);
+    position: absolute;
+    right: 0;
+    width: max-content;
+    text-align: right;
+    z-index: var(--projector-hover-z);
+}
+.stepper-justification-sub-vars span{
+  display:inline-block;
+  background-color: var(--shard-any);
+  padding: 2px;
+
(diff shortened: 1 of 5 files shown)


src/web/www/style/stepper.css
```diff
@@ -169,10 +169,10 @@
 
 
 /* STEPPER (OLD) */
-
+/*
```
Negabinary: Huh, I guess this is all duplicate code. We should probably just delete all of the rules here that also appear in dynamics.css (or delete the ones in dynamics)

src/web/www/style/dynamics.css
```diff
@@ -179,7 +179,16 @@
 /* STEPPER */
 
 .stepper-justification {
-  width: 260px;
+    width: fit-content;
+}
+.stepper-justification-sub-vars {
+    display: none;
+    position: relative;
```
Negabinary: I think it might look better if the popup hovers above the stepper instead of all the spacing changing on hover.

Also we should probably make the UI hint to the user that you can hover over it, maybe by replacing "substitution" with "substitution of ..."

Negabinary (changes requested):

joshqwang:
ok, I made some changes.
Screenshot 2025-08-23 at 12 47 37 PM
Here's what it's looking like right now.

PR #1877, opened 2025-08-19
Handle function application in rewrite checker

### LINKED ISSUE #1842: Algebrite should treat arbitrary equal expressions as equal
Image

In here, we don't need to give the rewrite checker any semantics of length, but it should be able to see that these two are the same.

nskh:
Allows us to rewrite-check things like `f(x) + 1` being equivalent to `1 + f(x)`; all we have to do is unpack function application. Even works when `f, x` have not been defined: 
image

Closes #1842.


Negabinary:
Had a play with it; these are some examples I'd prefer if they gave the opposite answer

image

image

image

image

nskh:
Fixed length (builtins) and list concat handling. Still working on the function stuff, we may need to chat about how to handle those generally.

PR #1878, opened 2025-08-19
Fix text projector line height

7h3kk1d:
The text alignment of the lines was off

Before:
Lines overlap text
After:
Lines between text


disconcision (approved):
🏆coloring within the lines award 🎖️

PR #1879, opened 2025-08-20
Probes II

disconcision:
Core features:
- [x] Refractors: A variant on projectors that do not replace the underlying syntax, but instead decorate it, while allowing a custom offside display.
- [x] Probes: Probes are now refractors, so you can edit a term even with a probe on, and you can have probes inside other probes.
- [x] Autoprobes: These automatically add probes to each line of the autoprobed expression.
- [x] Context menu: Right click on syntax to open. Replaces projector panel. Also has jump-to-definition.
- [x] Sample dropdown: Hover over a probe on a function to access contextual actions (Pin and Step-Into; see below)
- [x] Probe Sidebar: (camera lens icon) for probe extras and configuration

Core internals:
- [x] Clarify separation internal dynamics (Sample collection) versus instrumentation UI (Probes). Probe construct removed from AST in favor of a map of sample targets which is passed to the evaluator

Non-probe Bug fixes:
- [x] Fixed bug when click on code when caret is located offscreen would cause a sudden scroll to initial caret position

Non-probe supporting features:
- [x] Better emoji and general Unicode support in strings
- [x] Messages to/from worker are no longer serialized and hence much faster
- [x] Projector system: Viewport culling for projector view rendering perf (for scratch and documentation mode only)
- [x] Concave grout now the precedence that semicolons *should* (but don't) have. This is looser than any operator but tighter than things like function literals in order to get live feedback when inserting a new let expression inside an existing function literal
- [x] Scroll to caret now keeps a buffer around the caret
- [x] Underline decoration when hovering on references when command is held to indicate jump to def
- [x] Script to generate a string of all builtins for documentation and prompting purposes

Context menu:
- [x] Add context menu
- [x] Add jump to definition
- [x] Only show probe options for expressions/patterns
- [x] Don't show probe option for fake expressions (e.g. labels) 
- [x] Restrict autoprobe to expressions
- [x] Keyboard shortcut display 

Adding/remove probes:
- [x] New keyboard shortcut: Cmd/Ctrl-E to toggle a manual refractor probe on the indicated term
- [x] Add via context menu
- [x] Allow probes (refractors) to be placed on projectors
- [x] Replace Probe projector in Panel with refractor
- [x] Command to remove all probes within a selection (reuse Cmd-E)
- [x] Make manual and auto probes mutually exclusive when toggling
- [x] Remove a probe if another is added on the same line
- [x] Smart probe placement for arrow types. If you'd be probing a term of arrow type, try to probe subterms containing or contained by that term instead. We currently consider 4 cases: (a) If probing a function literal, probe parameters and body instead. (b) If probing a variable in function position of an application, probe the application instead. (c) If probing a partially applied function, in function position wrt an application (typically but not limited to a reverse application), probe the application instead. (d) (combining (b) and (c)) Variable in fun pos of partial ap in fun pos of ap (typically reverse). In a separate case, probing a let definition just probes the definition term. The following expression cases do not support probing: type alias definition (no point), deferrals in partial aps (no point + breaks parse), and labels (same). 
- [x] Probe remove keyboard toggle works when a sample is focussed

Sample display:
- [x] Only cap sample value length if there is *actually* more than one closure on that line
- [x] Style Indet samples differently
- [x] Add Sample legend in sidebar
- [x] Add time-based coloring option
- [x] Add duration-Interval-based coloring option
- [x] Hide dyncursor-unrelated samples in Single sample mode
- [x] Explanatory icons when no samples are shown; clicking these will try to show sample when possible

Internals:
- [x] State: Trigger/deserialize probe projector invocation into refractor probes
- [x] State: Serialize refractor probes into invocation
- [x] Rebuilt all doc slides
- [x] Sample collection: Eliminate Probe AST wrapping form in favor of a set of probed UUIDs
- [x] Sample collections: Now collects times and other data 

StepInto:
- [x] StepInto jumps to function definition (function parameters) and adds an autoprobe
- [x] StepInto shouldn't change probe status if already probed
- [x] Add StepInto to sample dropdown
- [x] StepInto pins the ap
- [x] StepInto updates the DynCursor to include the source ap 

Pinning:
- [x] Adding: Add pin action to sample dropdown
- [x] Adding: Add keyboard shortcut (p) when sample is focussed 
- [x] Adding: Disallow pinning constructor applications
- [x] Adding: Disallow pinning builtins
- [x] Adding: Only allow one pin at a time
- [x] Removing: When removing (manual/auto) probe with pinned sample, remove pin
- [x] Removing: When bulk-removing probes via selection, remove a pin on sample within selection
- [x] View: Move pin icon to left instead of right so visible for large values
- [x] View: When a probe would show samples, but doesn't because of pin, show an unpin affordance in lieu of samples

Autoprobes:
- [x] Add via keyboard shortcut: Cmd/Ctrl-Shift-E
- [x] Add via context menu, or convert to manual probe 
- [x] Indicate current probed state of term
- [x] Basic placement logic: Autoprobing a term probes the 'last largest term' on each line of that term.
- [x] Advanced logic: Modifying the above, favor probing useful things over less useful things.
- [x] Less useful things determinable syntactically include: function literals (no value rep).
- [x] Less useful things determinable statically include references of variables already seen through the patterns (either singly as a probe on a pattern variable, or in composite as in a tuple or list pattern, e.g. if we have a probe on the pattern `(a, b)` then we've seen and can tell the bindings a and b). UPDATE: This actually results in confusing omissions; disabled; was annoying when e.g. the ref is a branch of an a if expression
- [x] Special case: For tuples and list literals where the literal and the last element but not the second last element end on the same line to put the probe on the last element. This is perhaps an awkward description. the idea is that when we want this behavior: ` (a, // show value of a here [linebreak] b) // show value of b here, NOT the value of (a,b)` 
- [x] Special case: Technically in hazel tuples don't require parens around them, ie the parens counts as a separate term, the same as disambiguation parens. but if we have closing parens on a line following the last element's last line, we don't want to bother putting a probe on the parens (we are assuming that given that the last element will be probed, other elements of the tuple that the user wants to probe can or are being probed as well; not 100% sure about this logic, tbh)
- [x] More special cases in AutoProbe.re
- [x] Make autoprobe line change color on hover/indication to clarify which expression correspond to which value

Sidebar:
- [x] Show indicated call if any as indication in call cursor (red outline)
- [x] Show index in call cursor (brown outline), dim entries below index 
- [x] Show pinned call in call cursor
- [x] Make sure indicated call is always in call cursor? i would think it would be but maybe not
- [x] Combine manual and auto probe lists in sidebar
- [x] Nest auto probes together in sidebar
- [x] Partition probes in sidebar by top level definitions
- [x] Toggle:  Single/Many
- [x] Toggle: Step/Call
- [x] Toggles for below/above colors 
- [x] Toggles for caller/callee colors
- [x] 1 versus Many toggles for below/above/caller/callee

---------

Review:

Done:
- [x] Context menu should say "add/remove projector"
- [x] Allow both "P" and "Shift-P" for pinning
- [x] Disable debug popup which causes a focus issue
- [x] Don't show step into when application has anything other than a variable (e.g.function literals) in function position (For now I've just straight-up disabled both stepping and pinning for anything other than an ap with a variable in function position. We can revisit this later)
- [x] Remove Option/Alt-V shortcut for probes everywhere in favor of Cmd/Ctrl-E
- [x] Double click the sample to toggle back to single sample mode
- [x] Replace bottom right corner-decoration with an indicator of whether there are any statics errors remaining. This should be already calculated in the statics cache. See what other global feedback we could offer that is already calculate and we could also have a status icon for no perf cost. Probably prioritize statics errors but if none could show if there are empty holes (but only if we can get this info cheap! don't want to iterate again over the syntax JUST for this) 
- [x] Exercise mode: Style for probe counts when there is an error on the probed term and not focused
- [x] "Ad probe" typo in comments on doc slide for probe; remove option-v; general consistency pass for comments. Will need to be updated again in probes-III after sample coloring scheme changes
- [x] Generalize mechanism that prevents two probes being on a line. Probably need to run a check after certain or all code edits. Iterate through probes (after measured has been recalculated; this may be subtle!) and see if two probes have ended up on the same line. If so, get rid of one of them. Find a reasonable way of deciding which
Bugs:
- [x] Bug: Clicking the "..." next to a sample toggles you to an arbitrary (?) sample
- [x] Bug: Step into causes frankenstein state i.e. focus isn't kept in sync with browser notion of focus (Just transfers focus to parent editor for now)


7h3kk1d:
@disconcision any idea why there's some regex failures going on in tests?

TypeError: Method RegExp.prototype.test called on incompatible receiver undefined

disconcision:
@7h3kk1d no, and there are other weird test errors as well. it's on my list...

7h3kk1d:
> @7h3kk1d no, and there are other weird test errors as well. it's on my list...

I'm debugging this one now. Looks like the regex changes for Unicode are somehow not playing nicely with bisect ppx for coverage

7h3kk1d:
@disconcision temporary fix https://github.com/hazelgrove/hazel/commit/e1e74167bccf7541645e9836cfa585aad47c1918

disconcision:
@7h3kk1d do you know any more about what the deal is with those regexp errors? i don't get them locally...

7h3kk1d:
> @7h3kk1d do you know any more about what the deal is with those regexp errors? i don't get them locally...

It's an issue with the regexp code plus the bisect_ppx code. I haven't minimized it to create an issue yet.

dm0n3y (commented):
lgtm except you should probably extend the grout precedence changes to `Grout.shapes` and then just defer to `Piece.shapes` on line 83 of `Skel.re` as it was originally

cyrus- (changes requested):
most critical:
- clicking the "..." next to a sample toggles you to an arbitrary (?) sample
- cursor frankenstein state (focus isn't kept in sync with browser notion of focus)
- step into causes frankenstein state

other stuff (triage):
- Context menu should say "add/remove <name> projector"
- On bottom right of chrome, have it be an indicator of whether there are any errors/warnings/holes remaining
- "Ad probe" typo in comments on doc slide for probe
- Do general consistency pass for comments
- Disable debug popup which causes a focus issue
- allow both "P" and "Shift-P" for pinning 
- hide the "P" and "Enter" keyboard shortcut indicators when not focused 
- double click the sample to toggle back to single sample mode
- remove Opt/Alt-V shortcut in favor of Cmd/Ctrl-E (and change text in doc slides)
- style for probe counts when there is an error on the probed term and not focused
- can still have two probes on the same line if it happens via an edit (e.g. backspace)
- include icon for samples not aligned with cursor in legend in many mode
- consistency of naming of "cursor" -> should be "dynamic cursor" whenever possible
- move Dynamic Cursor box in probearium to top or some other way to make clear that the legend reference to "Cursor" is the dynamic cursor
- think about being able to toggle off probes on a sub-expression when it is inside an autoprobe (lifetime as long as the autoprobe lives)
- step into doesn't work for function literals in function position
- be able to bulk probe via selection using same logic as auto probing
- when probes show up there appears to be some sort of animation/shimmer maybe involving the depth indicators?

disconcision:
@cyrus- Addressed most of your comments; see the 'Review' section above for what is fixed here. Two things were defered to probes-III (search 'cyrus' on #2061) and four things were deferred to TBD (search 'cyrus' on #2058). This is good to merge imo

PR #1890, opened 2025-08-21
Selection snapping for associative terms

nskh:
Enable Hazel to snap to proper subterms of an expression (i.e. highlight all of `1 + 2` when user selects over just, say, `+ 2` or `1 +`). Handle left-associative binary operators for arithmetic by traversing the AST (described in a comment). 

#2307 adds a related feature: taking hidden steps to reparenthesize parts of the AST that are selected before taking Stepper actions, so we can act on proper sub-terms instead of gating that behind only the outer-most left-associative term.


cyrus-:
@nskh we just merged a PR where @disconcision rewrote some of the selection code, feel free to ping him

nskh:
@cyrus-  #1906? I'll take a look.

cyrus- (commented):
- selection snapping does not change background color of folds / otherwise trigger projector selection state (fix architecturally)

Image

- not working in this example:

Image

src/language/term/Operators.re
```diff
@@ -222,6 +222,21 @@ let show_op_bin_float: op_bin_float => string =
   | Equals => "Float Equality"
   | NotEquals => "Float Inequality";
 
+// Useful for associativity handling, selection snapping, etc
```
Negabinary: Is this still used?

src/language/term/Operators.re
```diff
@@ -222,6 +222,21 @@ let show_op_bin_float: op_bin_float => string =
...
+let is_associative_op = (op: op_bin): bool => {
+  switch (op) {
+  | Int(Plus)
+  | Int(Times)
+  | Int(Power)
```
Negabinary: not associative

src/web/app/editors/decoration/Highlight.re
```diff
@@ -548,36 +557,25 @@ let selection_expanded =
...
-      switch (seg) {
-      | None => []
-      | Some(seg) =>
-        of_segment(
+      @ of_segment(
           ~measured,
           ~shape_map,
           ~font_metrics,
           ~shape_init=Some(fst(Siblings.shapes(z.relatives.siblings))),
-          ~clss=["selected-expanded", Selection.buffer_cls(z.selection)],
```
Negabinary, line 619: Is not drawing a separate selection-expanded (lighter for expansions) deliberate? I could see it going either way.

Negabinary (commented):
Left a few minor comments, otherwise looks good!

nskh:
Cleaned up and so standard constructs fall back to the dev-style selection behavior, with special snapping only for associative cases.

Tested over the forms in the basic reference and verified against `dev`:
 - `case`, `|`, and `=>` snap to the enclosing case
 - tuple comma selection snaps over the tuple contents
 - list cons `::` snaps over the right-associated list
 - unary `-` snaps over the full negative expression
 - rewrite checker now sees `-1`, not just `1`

Tests added for these cases as well.

PR #1895, opened 2025-08-22
Fold projector un/folds on click instead of hover

### LINKED ISSUE #1756: Make fold hover less hyperactive in stepper

disconcision:
Closes #1756.

Fold projectors now require a click as opposed to hover to toggle display. It intentionally only toggles when the fold projector is indicated, i.e. if the caret is not at the fold, one click will bring it to the fold, and another will toggle the display. Double-click to remove is removed it's too easy to do this accidentally when clicking. If people want mouse-based removal I can find another mechanism but I don't think it's a big deal.

This also aggressively changes the way unfolded folds are styled. I think they blend in badly with the background in their current state. I don't think the new style is fully successful but it stands out at least. The operative metaphor here is that they are dark as they are 'underground' / 'in a cave' / 'closed over'.

As an implementation note, there are two codepaths for how this display toggle works, as chosen by the `always_display` field. For in-editor folds, this is false, meaning that the code view will not be calculate when the fold is in a collapsed state, which makes folds useful to mitigate editing perf. This doesn't work folds which are in non-persisted editors like the results view / stepper, so these have a different codepath where collapsing just hides the code. This division is awkward and should be revisited later.

Other changes:

Projector API Update: The arguments to projector view functions are now a record to prevent having to necessarily update every projector when an argument is added or removed. There is also a new argument, status, which tells the projector if it is indicated/selected/etc in the parent editor. 

- This also make some minor updates to the Projectors slide


cyrus-:
- z index issue with result editor
- when folding a type you get a style issue with the dark tip

disconcision:
@cyrus- do you recall what the z-index issue was here? i recall seeing text from the background editor over the fold, but I can't reproduce, either in the result editor or the main editor:

Screenshot 2025-09-04 at 11 59 53 PM

Screenshot 2025-09-05 at 12 00 08 AM

disconcision:
Issues addressed

cyrus- (approved):

Issue #1838, opened 2025-07-29
Buitlin functions should show name in probes


7h3kk1d:
<img width="420" height="150" alt="Image" src="https://github.com/user-attachments/assets/2ffbd342-0758-4a6b-bf46-e34d46eec545" />

The probe on builtins should at least show the function name

PR #1898, opened 2025-08-22
Builtin fixes

### LINKED ISSUE #1841: Builtins shouldn't be anon in stepper
Image

possibly related to #1838

Negabinary:
Fixes #1841 and #1838


disconcision:
not a big deal but having a separate field in the impl which is always just the main record name field with a plus appended feels redundant... can that name just be derived when the builtins are processed?

disconcision:
what happens without the fixes?

Negabinary:
> what happens without the fixes?

It's mostly just stepper optics, without the fixes:
- builtin functions show up as anonymous
- the lookup step highlights the application not the function
- it's harder to compare pre-and-post-looked up functions in the stepper

Negabinary:
> not a big deal but having a separate field in the impl which is always just the main record name field with a plus appended feels redundant... can that name just be derived when the builtins are processed?

Probably. If you think it's worth going back over the file and changing this I can change it. There's already some redundancies like having the implementation as a string and an AST though.

cyrus- (approved):

PR #1916, opened 2025-08-31
Prevent same scratchpad names

Cryplo:


cyrus-:
@Cryplo can you resolve the merge conflict and make the PR based on a branch rather than your fork of Hazel so it runs through our CI pipeline

PR #1919, opened 2025-09-02
Switch to init using segment instead of zipper

7h3kk1d:
This PR updates the initialization logic to store the segment as a s-expression, similar to our previous approach with the zipper. The main motivation for this change is that we always want the caret to be positioned at the beginning of the slide for init slides, making explicit caret storage unnecessary. As a result, the related test for caret position can be removed.

### Motivation

- For init slides, the caret should always start at the beginning, so storing caret position is redundant.

### Notes

- This approach continues to use s-expressions in init files, which are parsed at runtime. While moving away from this would be ideal, migration will require a strategy for rebuilding files when updating the zipper type. This is deferred to a future PR.


src/web/init/docs/Livelits.ml
```diff
@@ -1,154 +1,777 @@
...
-         e3c391a3-2c6c-4cd4-ae1b-dc741c9950fd)(label(\"(\"\")\"))(mold((out \
-         Exp)(in_(Exp))(nibs(((shape(Concave 23))(sort Exp))((shape \
-         Convex)(sort Exp))))))(shards(0 1))(children(((Tile((id \
-         7e7a62e9-51a3-4bbb-9555-54e8b2f2c869)(label(0))(mold((out \
-         Exp)(in_())(nibs(((shape Convex)(sort Exp))((shape Convex)(sort \
-         Exp))))))(shards(0))(children()))))))))))))))(model\"()\")))(Secondary((id \
-         44ddd530-82a8-4184-800a-21652348cfe7)(content(Whitespace\" \
-         \")))))))))(Grout((id 3534f38b-9cab-4478-9cde-bd0d16f2fae0)(shape \
-         Convex))))))(ancestors())))(caret Outer))";
+      segment =
```
7h3kk1d, line 4: @disconcision I'm conflicted between storing this as the sexp or just the ml code. The ml code doesn't have to parse but the formatted code is so much longer. Do you have any thoughts?

disconcision: how would export work to generate the ml code?

7h3kk1d: I have it locally. It just runs `Zipper.zip` to get the segment. Builds this record:
```reason
{
  segment: Segment.t,
  backup_text: string,
}
```

Then runs show on it.

7h3kk1d: Basically this code https://github.com/hazelgrove/hazel/pull/1919/files#diff-bf0419f9509267c72a8bef1b8106779b0711842cd703b69d97e330e612c995d0R73-R83

src/web/init/Init.re
```diff
@@ -21,23 +21,12 @@ let startup: PersistentData.t = {
...
-      // GUIDEComposingExpressions.out,
-      // GUIDEComputingEquationally.out,
-      // GUIDEVariables.out,
-      // GUIDECompositionality.out,
-      // GUIDEScope.out,
-      // GUIDEShadowing.out,
-      // GUIDEBoolsandTypes.out,
-      // GUIDEConditionals.out,
-      // GUIDEFunctions.out,
-      // TESTSTypesandStaticErrors.out,
```
7h3kk1d, line 34: I just removed these unused ones.

disconcision:
@7h3kk1d so one thing this is going to complicate for me is persisting refractors, which are currently a map on the zipper. it remains non-obvious to me in our current setup what should live on the zipper as opposed to the editor, so what i've been going with is that non-ephemeral stuff goes on the zipper. (This also makes refractor actions zipper actions as opposed to editor actions, keeping them in line with projector actions, but this is a side issue).

I do think it's probably better to consider cursor position / selection ephemeral here though, so i like this change directionally. I assume to adapt my stuff it doesn't really matter where the refractors are stored, I just need to replace PersistentSegment with a record that stores both the PersistentSegment and the refractors? So this isn't really a blocker. But it might be worth clarifying in terms of naming our different notions of persistence for editors. Like, we have one notion of persisting editors to localstore for session maintenance, which preserves the zipper and hence some (but not literally all eg col_target) ephemeral editor state. And we have this notion, which eschews all ephemeral state. Maybe we can call this PersistentNonEphemeral, or ideally something better, which is currently just an alias for PersistentSegment but will be updated to a record after my changes?

7h3kk1d:
> @7h3kk1d so one thing this is going to complicate for me is persisting refractors, which are currently a map on the zipper. it remains non-obvious to me in our current setup what should live on the zipper as opposed to the editor, so what i've been going with is that non-ephemeral stuff goes on the zipper. (This also makes refractor actions zipper actions as opposed to editor actions, keeping them in line with projector actions, but this is a side issue).
> 
> I do think it's probably better to consider cursor position / selection ephemeral here though, so i like this change directionally. I assume to adapt my stuff it doesn't really matter where the refractors are stored, I just need to replace PersistentSegment with a record that stores both the PersistentSegment and the refractors? So this isn't really a blocker. But it might be worth clarifying in terms of naming our different notions of persistence for editors. Like, we have one notion of persisting editors to localstore for session maintenance, which preserves the zipper and hence some (but not literally all eg col_target) ephemeral editor state. And we have this notion, which eschews all ephemeral state. Maybe we can call this PersistentNonEphemeral, or ideally something better, which is currently just an alias for PersistentSegment but will be updated to a record after my changes?

Yeah, `PersistentSegment` is currently only used for Init so you could also just add the refractors there and/or rename it to `InitEditor` if you want to remove confusion. I don't think the reasoning for what gets persisted or not is intrinsic enough to have a huge taxonomy. It's probably just worth naming them based off of use case and storing whatever is neccesary. So if we want the name  `Persistent` to literally mean what's stored in the browser we could call what is now `PersistentZipper` `PersistentEditor`. I could then rename `PersistentSegment` to be `InitEditor`. The real change is splitting up the editor persistence from the initial state. Let me know if you have thoughts on naming. I'm leaning towards merging this as is until we come to a solution there.

src/haz3lcore/zipper/Zipper.re
```diff
@@ -30,10 +30,14 @@ let delete_parent = (z: t): t => {
 let zip = (z: t): Segment.t =>
   Relatives.zip(~sel=z.selection.content, z.relatives);
 
-let unzip = (seg: Segment.t): t => {
+let unzip = (~direction: Direction.t=Right, seg: Segment.t): t => {
```
disconcision, line 33: been meaning to do this for literal years

test/Test_ReparseDocSlides.re
```diff
@@ -52,8 +52,4 @@ let tests = [
     "DocSlides.ReparseBackuptext",
     List.map(doc_slide_reparses, List.tl(doc_slides)) // Dropping the first basic reference slide to avoid the issue with whitespace shifting around convex grout
   ),
-  (
-    "DocSlides.CaretAtBeginning",
-    List.map(caret_is_at_beginning, doc_slides),
```
disconcision, line 57: is this predicate used elsewhere?

disconcision (approved):
lgtm. Export for init seems to work. Noticing that Export Scratchpad still seems to use the zipper; should we switch that one as well? Maybe not necessary in this PR but if it's advisable we might want to create an issue. Also noticing that Export Scratchpad no longer works on doc slides (in dev, not specifically here), going to make an issue for that.

test/Test_ReparseDocSlides.re
7h3kk1d: Nope, will remove.

src/haz3lcore/zipper/Zipper.re
7h3kk1d: The time has come.

7h3kk1d:
> Noticing that Export Scratchpad still seems to use the zipper; should we switch that one as well? 

I've never used it so I'm not exactly sure what it's for. I don't mind keeping the caret position in it so it shows up the same on import but I don't really know the context of use.

cyrus- (approved):

PR #1922, opened 2025-09-03
run dynamics on startup

### LINKED ISSUE #1918: Calculation not Triggered in Exercise Mode when Refreshing
Tested on [6dc60b4](https://github.com/hazelgrove/hazel/commit/6dc60b430ce370f539bfef097e03826f9910d9ee) deployment.

Reproduce as follows:
1) switch to exercise mode page 1
2) wait for auto save, and refresh
3) In section Test Validation: No results

Negabinary:
Fixes the issue where the dynamics don't run when you load Hazel.

This version still waits until after the first autosave to run dynamics, I would like to come up with a more immediate way to run it.

### DIFF
--- a/src/web/Main.re
+++ b/src/web/Main.re
@@ -57,7 +57,11 @@ let apply =
   // ---------- CALCULATE PHASE ----------
   let model' =
     updated.model
-    |> History.Update.calculate(~schedule_action, ~is_edited=updated.is_edit);
+    |> History.Update.calculate(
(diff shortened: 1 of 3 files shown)


disconcision:
@Negabinary is this just draft because you want to find a more immediate way? i'm into merging it as is as this is starting to drive me nuts

cyrus- (approved):

PR #1923, opened 2025-09-03
Remove fake Typ Ap

### LINKED ISSUE #1459: Invalid type application breaks statics throwing exception
Screenshot From 2025-01-10 10-45-49
Trying to write this expression with an invalid type `Int(7)` or `Int(Float)` throws an exception as soon as you try to assign it a value.

This triggers a statics exception (Failure "Type join of ap")
```hazel
type x = Int(Float) in let y : x =  1
```

This alone triggers an exception (Failure "type application in dynamics") in dynamics:

```hazel
type x = Int(Float) in let y : x
```

### LINKED ISSUE #1621: Typ.weak_head_normalize does not terminate in statics
If you try to run `type y = y in type ? = y in ?` you break the editor.

### LINKED ISSUE #1626: Coverage checker fails on invalid ap typ
`fun (?: (Float((+ A(Bool))))) -> ""` fails with exception:

```
"all_ctrs_of_type called with a non-normalized type: (Ap ((Atom Float),\
  \n   (Sum\
  \n      [(Variant (\"A\",\
  \n          [Option.get(Haz3lcore.Id.of_string(\"58aa635a-f9d2-43e0-a2c5-0625a89ff439\"));\
  \n            Option.get(Haz3lcore.Id.of_string(\"e07c88ea-9b7e-4282-ba89-cf46633dffa5\"))\
  \n            ],\
  \n          (Some (Atom Bool))))\
  \n        ])\
  \n   ))") [hazel.js:16924:29](https://hazel.org/build/dev/hazel.js)
```

### LINKED ISSUE #1627: Typ normalization in elaboration causes infinite recursion
`type x = rec y -> x in 0 @< x > `

Causes Typ.normalization to infinitely recurse and elaboration to non-terminate

### LINKED ISSUE #1628: Type T = T possibly creating recursive type
Image

```
type T = T in
let x : T = 
```

crashes the editor.

### LINKED ISSUE #1921: Unhandled type join of ap
I obtained `Unhandled exception (Failure "Type join of ap")` while trying to edit in the first hole shown below. I think I tried to insert a `[`.

Image

disconcision:
Closes #1921.
Closes #1459, at least in the basic sense that the exceptions are no longer thrown. We still have the issue of a weird-looking type percolating through semantics.
Closes #1628, again in the sense that that no longer crashes the editor.
Closes #1627, ibid.
Closes #1626, ibid.
Closes #1621, ibid.

1. This removes the application case in Term.Typ. This case was just scaffolding to support a maketerm parsing hack for sum type definitions; it was not intended to persist past the end of MakeTerm into Statics and beyond. It looks to have since become instantiated throughout the codebase. Fully removing the sum type definition parsing hack would require creating a new sort; I don't have time for that right now. Instead, instead of having the `Ap` case in for Typ actually materialized, the fallthrough multihole is used instead, and then is consumed internally in MakeTerm by the sum type parser. Type applications outside sum type definitions will now be parsed as Multihole and shown as 'broken types' in the cursor inspector. I've change the is_error code to ensure error holes drawn around them. These will no longer make it to statics.

2. This also generalizes the type alias definition form to add the rec constructor to any type definition that self-mentions. Non sum-type recurse types still won't work properly, but this will prevent them from infinite-loop crashing weak head normalization as they do now.

3. Semi-unrelated change: backpack dumping exposed some non-advisable parameter combinations, and was available through some ill-advised aliases which don't advertise that something funky is happening. This functionality is moved to its own module, under the hope that people will think twice before calling it. My general advice is to not call dump_backpack without a good reason.


disconcision:
This might have implications for type parameterization. not sure if there's an actually live branch there?

cyrus- (approved):

cyrus-:
@odruzgal This PR removes Ap since it was quite buggy -- you can add it back in your parameterized types PR, with proper error handling and normalization etc.

PR #1933, opened 2025-09-03
Fix #1929

### LINKED ISSUE #1929: Failure hd crash on edit pattern
to reproduce, write the following:

```
let x =
  case  
```

and then add an "s" to the "x" to make it "xs".

disconcision:
Closes #1929.
I will not be taking any questions


Negabinary (approved):
Indiscernible dark magic but seems to work

PR #1938, opened 2025-09-04
Property based test that says parsing is history independent

7h3kk1d:
Currently parsing isn't history independent so this is more an aspirational goal.

Known issues
* Grout shifts whitespace on reparse
  * The test addresses this by stripping the whitespace in the assertion
* Removing grout conjoins adjacent tokens
  * [?.?.? reparse returns ?..?](https://github.com/user-attachments/assets/a58f5f67-1fd7-4b13-b490-a50d03b79d8a)

TODO Test Improvements
* Need to add character deletion
* Prioritize inserting keywords
* Once known bugs are addressed prioritize situations that could cause problems


7h3kk1d:
@disconcision closing this for now

Issue #1386, opened 2024-09-05
Exercise Import menu item doesn't accept clicks on label


disconcision:
You have to click the Icon, unlike other menu items. Would've fixed immediately but looks like the widget needs to be refactored.

cyrus-:
Closed by #1944

Issue #980, opened 2023-02-06
Second file upload is not registered by on_input handler of file upload button


evanweissburg:
From Slack:

Evan: "I'm running into an odd bug where uploading a file twice does not trigger the `on_input` handler of the file upload button on the second time. Has anyone else seen this before/can repro?"

Cyrus: "yeah I was seeing that too. not sure. maybe look on stack overflow to see if its a general issue in the browser or specific to js_of_ocaml?"

cyrus-:
can't reproduce this anymore

PR #1944, opened 2025-09-10
fix clicking file import button in nut menu

cyrus-:


disconcision (approved):
Works. Duplicates a chunk of code from the function above but whatever. This closes #1386. Also possibly related: #980; not sure exactly what this is saying.

PR #1969, opened 2025-09-25
Probes and dyntype probes on types

7h3kk1d:
Screenshot 2025-09-25 at 9 53 25 AM

Allows you to add probes and dyntype probes on types. This works by adding a `Probe` type similar to the one for patterns and expressions. The `Type` "observes" any expression that flows through it in an ascription e.g. `"Hello" : Probe(?)` will add "Hello" to the probe.

### Known issues
* We optimize away some ascriptions
  * Consistent branches and lists both add ascriptions if the types don't match. If the type could contain a probe we want it to exist so that we get the probe on the type.
  * Sum type constructors don't add ascriptions and instead just rerun statics on the contained expression:
Screenshot From 2025-09-25 10-38-22

* Ascriptions sometimes evaluate before the scrutinee evaluates
Screenshot From 2025-09-25 10-39-05
* Ascriptions don't currently pass through probe types causing some indet issues with builtins.
* This now has dynamics set to true for the type proj in order to get the dynamic type. Unfortunately this wraps the term in a probe which has the same statics as parens and gets the wrong self type


src/haz3lcore/lang/MakeTerm.re
```diff
@@ -661,7 +661,12 @@ and typ_term: unsorted => (Typ.term, list(Id.t)) = {
         | ([t], []) when Token.is_quoted_label(t) =>
           Label(Token.sub(t, 1, Token.length(t) - 2))
         | (["(", ")"], [Typ(body)]) => Parens(body)
-        | (label, [Typ(body)]) when is_probe_wrap(label) => body.term
+        | (label, [Typ(body)]) when is_probe_wrap(label) =>
+          print_endline("Parsing probe wrap for type");
+          print_endline("Id: " ++ Id.str3(id));
```
7h3kk1d:
```suggestion
```

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -661,7 +661,12 @@ and typ_term: unsorted => (Typ.term, list(Id.t)) = {
...
+          let should = should_instrument(id);
+          print_endline("Should instrument: " ++ string_of_bool(should));
```
7h3kk1d:
```suggestion
```

src/language/dynamics/transition/Transition.re
```diff
@@ -189,6 +189,15 @@ module Transition = (EV: EV_MODE) => {
         d,
       )
       : EV.result => {
+    let update_probe' = (id, d, env, call_stack, pr) => {
+      print_endline(
+        "Updating probe: " ++ Id.str3(id) ++ " with value " ++ DHExp.show(d),
+      );
```
7h3kk1d:
```suggestion
```

src/util/Monads.rei
```diff
@@ -2,7 +2,6 @@
  * https://ocaml.janestreet.com/ocaml-core/v0.13/doc/base/Base__/Monad_intf/index.html */
 
 module type MONAD_BASIC = {
-  [@deriving sexp]
   type t('a);
```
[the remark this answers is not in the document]
7h3kk1d: The deriving sexp wasn't used.

src/language/dynamics/Dynamics.re
```diff
@@ -119,12 +119,12 @@ module Probe = {
...
-    refs: Statics.Map.refs_in(m, id),
+  let instrument_exp = (m: StaticsBase.Map.t, id: Id.t, _: Probe.t): Probe.t => {
+    refs: StaticsBase.Map.refs_in(m, id),
   };
 
-  let instrument_pat = (m: Statics.Map.t, id: Id.t, _: Probe.t): Probe.t => {
-    refs: Statics.Map.bound_in(m, id),
+  let instrument_pat = (m: StaticsBase.Map.t, id: Id.t, _: Probe.t): Probe.t => {
+    refs: StaticsBase.Map.bound_in(m, id),
   };
```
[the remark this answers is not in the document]
7h3kk1d: not a breaking change

PR #1975, opened 2025-09-30
Fix paste sometimes leaving caret on left

### LINKED ISSUE #1973: Pasting leaves caret on the left
For example, select all on any docs editor, and paste. The caret is left on the left as opposed to the right as expected. Recent regression I think.

disconcision:
Closes #1973


disconcision:
@russell-rozenbaum this should fix the issue you're seeing

cyrus- (approved):

PR #1977, opened 2025-10-01
Stop Indet forms evaluating to false with polymorphic equality

Negabinary:
This stops

image

from evaluating to false, (allowing us to use reflexivity in proofs)


cyrus- (approved):

cyrus-:
will be merged as part of #1979

PR #1974, opened 2025-09-30
Delete term

Negabinary:
Splits the functions in Term.re up into the appropriate other files (i.e. Exp.re, Pat.re, Rul.re, Any.re)

PR #1835, opened 2025-07-28
Theorems at the bottom

Negabinary:
### Main features

This pr adds some new expressions for defining theorems, notably: `theorem <p> = <e> in <e>` for defining theorems, `forall <p> -> <e>` for quantifying those theorems. (Note: the previous **forall** has been renamed to **poly**) For every theorem keyword added, a new theorem box appears at the bottom.

image

Theorems get evaluated to expressions of the form `proof_object <e> end`, these expressions have type `proof_of <e> end`.

 - [x] "search assumptions"
 - [x] get rid of proof_of in assumptions box

### Development

Uses a theorem keyword to collect theorems and add a stepper at the bottom of the screen for each "theorem".

- [x] Add `theorem ... in ...`, `proof_of ... end`, `yes ... indeed`, `forall ... -> ...` keywords
- [x] Collect theorems at the bottom of the screen
- [x] Give theorems the appropriate static context
- [x] Collect the environment for the theorem and allow substitution steps using that environment
- [x] Pull equalities out of env/ctx and put them in the axiom box
- [x] Rename "InductionStep" to "CasesStep"
- [x] Add Induction step similar to cases
- [x] Add IH to axioms

Done
- [x] #1843
- [x] #1845
- [x] #1847
- [x] Special UI treatment for reflexivity [dev]

Wednesday (08-06)
- [x] Axiom box experience [thm-cap]
   * Hook up equalities (0.5h)
   * Use AssumptionBox in axiom box (0.5h)
   * Add search functionality (2h) [0:4]
   * #1844 (0.5h)
- [x] fix up refl branch (0.25h)
- [x] equality modulo env [thm-cap]
   * substitute in all expressions before assumption enters env (currently just IH, eq) (0.5h)
   * make MatchExp take an env (0.5h)

Thursday (08-07)
- [x] Serialize proofs [#1869]
   * Add serialized form to all the proof steps (1h)
   * Make sure that ids are recoverable (2h)

Friday (08-08)
- [x] Use theorem keyword, not just proof_of [thm-cap] [#1871]
   * try out expression-forall thing (3h)
   * Fix unannotated foralls get weird types (1h)
       image

Monday (08-18)
- [x] Some sense of proof completion [thm-cap]
   * Case exhaustiveness (1h)
   * analyse proof steps against bool (0.5h)
   * Make something go green when we reach true (??)

Tuesday (08-19)
- [x] Clean up the css
   * Make theorems appear separate (2h)
   * Otherwise prettify (4h)

[...]

PR #1979, opened 2025-10-01
Theorem capture merge

Negabinary:
Combines #1835 #1870 #1974 and #1977 into one pr for git reasons


Negabinary:
Closed because we'll merge these in separately.

PR #1986, opened 2025-10-13
Line numbers

Cryplo:


7h3kk1d:
@Cryplo it might be worth looking at the cards documentation slide as it has some multi-line projectors that I assume we want to offset with the line numbers.

[Screencast From 2025-10-13 14-17-52.webm](https://github.com/user-attachments/assets/37cd1115-9842-4ff3-a84e-70dadf758ad9)

In the above screenrecording there is no line 15 so we probably want to shift it down.

Cryplo:
> @Cryplo it might be worth looking at the cards documentation slide as it has some multi-line projectors that I assume we want to offset with the line numbers.
> 
>  Screencast.From.2025-10-13.14-17-52.webm 
> 
> In the above screenrecording there is no line 15 so we probably want to shift it down.

Great point, I'll write some code to offset the line numbers for multiline projectors

Cryplo:
@7h3kk1d For the most part, line skipping for multi-line projectors is working now. However, there's a small issue. If there's a multi-line tab projector on the last line, as pictured below, the line doesn't seem to skip. Would you happen to know why this might be? 

Currently, the way my code works is by checking each row and seeing if any multiline projectors span across that row, then skipping that row. Specifically, my code checks if each row is greater than a measurement's origin row and less than a measurement's final row (which both means the projector is multi-line and the row is within this multi-line projector). I think that this error is because the last secondary measurement doesn't consume the line break that the Card projector should give it, so the secondary measurement is treated as if it doesn't have any projectors that cause deferred line breaks?

Screenshot 2025-10-19 at 11 17 48 PM

src/haz3lcore/derived/Measured.re
```diff
@@ -326,7 +326,12 @@ let of_segment =
           );
         // add seg to map and reset seg
         //TODO(andrew): decide if should actually add linebreak here
```
disconcision: get rid of my TODO, we're basically doing this now

src/haz3lcore/lang/Token.re
```diff
@@ -42,6 +42,7 @@ let bounding_box = (t: t): Point.t =>
 /* Token Recognition Predicates */
 
 /* A. Secondary Notation (Comments, Whitespace, etc.)  */
+let empty = "";
```
disconcision: make a note saying this is invalid from a view perspective; the use case is pretty much designating something is a dummy value, which should never be displayed

src/web/view/NutMenu.re
```diff
@@ -104,12 +104,17 @@ let stepper_group = (~globals: Globals.t) => {
   );
 };
 
-let dev_group = (~globals) => {
+let dev_group = (~globals: Globals.t) => {
   settings_group(
```
disconcision, line 116: let's break out a new subdivision here, called 'Code Display'. It should contain: Whitespace, Animations, and your two line numbering options.

src/web/www/style/editor.css
```diff
@@ -22,16 +22,56 @@
...
+
 /* This is a hack to compensating for default linebreak handling
 * preventing a trailing linebreak in code editors from leaving a
 * trailing blank line, as per https://stackoverflow.com/questions/43492826/.
 * The mysterious chracter below is a zero-width space, used here
 * to avoid inserting an extraneous whitespace context for inline
 * syntax views, e.g. type views in the cursor inspector */
 .code-text:after {
-  content: "​";
+  content: "";
```
disconcision: mysteries of the diffing algorithm...

disconcision (changes requested):
Looks solid; minor changes requested, then we're good to merge from my perspective. @cyrus- maybe do a quick functional review to make sure it satisfies your use case. Let's consider visually de-emphasizing line numbers a bit; below I use colors --BR1 and --BR2. that's plenty visible on my screen but if it seems too faint we could increment those numbers. I also changed the L/R passing to 0em/1em, which I think  feels a bit less cramped while not actually taking up any more space
 
Screenshot 2026-01-15 at 1 35 48 PM

src/web/www/style/editor.css
7h3kk1d: The old version had an invisible unicode zero-width character

Screenshot 2026-01-15 at 1 55 13 PM

I'm assuming it was an accident but I'm not sure.

```bash
alexanderbandukwala@Alexanders-MacBook-Pro-4 hazel % grep 'content:' src/web/www/style/editor.css | od -c
0000000            c   o   n   t   e   n   t   :       " 342 200 213   "
```

disconcision: ohhh right. i thought i left a comment about this. this is load bearing. ("loading bearing zero width character", the most beautiful phrase in the english language. this was my budget fix for the following situation (turn line numbers off; and imo these should be off by default, forgot to mention that in the review, though don't feel strongly about it):

Screenshot 2026-01-15 143738

this is what happens on this branch now if you have a trailing linebreak in your program.

Cryplo: Thanks for catching this, I added back the zero width character. It should be in commit 2da9845: https://github.com/hazelgrove/hazel/pull/1986/commits/2da9845f12b7145f3a41063433a2a6af74c948eb

disconcision (approved):
lgtm! except let's remove the opacity 0.5 from the line numbers... just --BR1 is fine but with the opacity its too light. also remove the opacity: 1 from the bolded line number style as it doesnt do anything. also need to fix the merge conflict from today's merge. 

ping me when those are done and i'll merge this

Cryplo:
Thanks for the feedback @disconcision! I just made the following changes:

1. Deleted the opacity 0.5 and opacity 1
2. Merged dev into this branch
3. Turned off line numbers by default

Please let me know if there is anything else I should change!

PR #1988, opened 2025-10-17
Live typing

7h3kk1d:
# Live typing

Adds a toggleable in-editor type feedback mode that fills in statically unknown types from what actually happened at runtime. Where a type is statically unknown, statics has nothing to check against and stays quiet; once runtime fills the unknown in, a genuine misuse becomes reportable.

Stacked on #1858, which contributes the machinery for coloring the runtime-derived parts of a type.

## Example

[Screencast From 2025-11-18 13-45-45.webm](https://github.com/user-attachments/assets/42d6e2c5-e0f2-4093-b1ef-547807dc26d0)

## Probe selection

Integrated with the `Probe` pinning system, so you can get dynamic types for one particular function application by pinning it. The type projector's Dynamic mode honors the same pin.

[Screencast From 2025-11-18 13-53-21.webm](https://github.com/user-attachments/assets/9d9bdec4-12d6-47d8-8fa3-d906989b9bc8)

## What you see

- Live-typing errors get their own decoration in the editor — a pulsing dashed outline in a color distinct from ordinary static errors — so it is clear that an error depends on runtime information.
- The cursor inspector shows the refined type with the runtime-derived components in green, and a ⚡ badge in its header when only the dynamic info reports an error.
- They also collect in the problem panel under a `Live Typing Errors` category.
- Only a *witnessed misuse* is reported (`Mark.is_live_reportable`): an observed value's refined type contradicting how the program uses it. Join failures between independently refined siblings (`case`/`if` branches, list elements) and exhaustiveness/redundancy claims recomputed from narrowed types are deliberately excluded — samples witness what occurred, not what is possible.

## Implementation

- **Probing what statics could not resolve** — with `live_typing` on, `CachedStatics.compute_targets` unions `ids_with_unknown_types(info_map)` into the probe targets, so every expression and pattern whose `elab_syn_ty` still contains an unknown gets sampled. `CachedStatics` now also records the *placed* probe ids separately: `targets` carries derived ids too, so it can no longer serve as the statics-refresh gate.
- **`LiveTyping.re`** — carries per-expression dynamic samples and per-TPat type instantiations, consumed by `Statics.mk(~dynamics=...)`. A term's samples are only consulted when its type actually has unknowns to fill. `LiveTyping.reabstract` re-generalizes a `typfun` body's live-refined type against its annotation, so the synthesized `Poly` does not quantify a variable its own body has already concretized (previously a source of spurious Poly/Poly mismatches).
- **`CodeWithStatics.re`** — filters `Dynamics.t` by the current `sample_focus` and runs statics a second time with the dynamic info, surfacing dynamic-only, live-reportable errors as `live_typing_error_ids`. During streaming evaluation the second pass runs at most once per 500ms; the settled result always runs it.
- **`Dynamics.TypeInstMap` / `EvaluatorState.RecordTypeInstantiation`** — TypAp transitions record the concrete type at each type-variable instantiation, so polymorphic code gets instantiation-aware feedback instead of a spurious error on the parameter. `EvaluatorState.append` merges these, so incremental-eval cache reuse keeps them.
- **`Typ.join` / `join_all`** — the lattice dual of `meet`, used by `LiveTyping.extend_ctx_with_instantiations` to combine the observed instantiations of a type variable. Like `meet`, it hands back the alias when the join does not refine it, deduplicates repeated tuple labels, keeps an unbound variable's name, and stops rather than looping on a cyclic alias chain.
- **Closure scoping** — sample values are *closed* before statics sees them: `Dynamics.to_live_typing_map` (and `DynamicTypInfer.type_of_sample`) substitute each closure's own environment, and strip stuck root casts so `(x : ?) : String` with `x = 1` is evidence of `Int`, not `String`. Statics discards a closure's env, so the body's free variables would otherwise resolve against the probe site's context and capture whatever same-named binder happens to be in scope there. For closures that still reach statics unclosed (the result editor), `Ctx.of_closure_env` shadows the captured names with `?`.
- **Incremental evaluation** — `EvalResult` tracks the program-derived targets separately and drops the incremental map when one appears or vanishes while the elaboration is unchanged (toggling live typing), since `IncrEval.reuse_check` cannot tell that a cached slice now needs sampling.
- **Cursor inspector** — `CursorInspector.view_type` takes `~live_typing_info` and, where the dynamic type differs from the static one, renders the segment and green ids that `DynamicTypInfer.segment_and_dynamic_ids` returns.
- **Settings** — `CoreSettings.live_typing`, off by default, toggled from the nut menu (with a performance warning) or the command palette.

## Testing

`test/evaluator/Test_Evaluator_LiveTyping.re` covers the end-to-end flow: sample collection through unknown ascriptions and pattern binders, meet across samples, closure shadowing, polymorphic instantiation (annotated and conflicting instantiations included), the reporting policy (a case-join `NoMeet` is not reported; `ExpectationMismatch` and `LabelNotFound` are), and pinning: one program is run unpinned and pinned on each of two calls, and the set of live errors, a binder's live type, and a `typfun` parameter's instantiation all follow the pin. A 500-case QCheck property checks the live pass never makes a synthesized type less precise.

`Test_Typ.re` covers `Typ.join` — equal atoms, Unknown, sums, unbound variables (name preserved), duplicate labels, alias preservation, cyclic-alias termination — plus a 10k-case property that `join` is at least as imprecise as both inputs. `Test_DynamicTypInfer.re` adds a closure-scoping case, and a check that a type built the way statics builds them colors only its runtime-derived components. `Test_Evaluator_Incremental.re` checks type instantiations survive cache reuse; `Test_Evaluator_Probes.re` pins that probes on nested ascriptions still sample.

## Notes

- `Grammar.fn` now takes `~typ=?` and `~name=?` instead of two positional options. Mechanical churn across builtins, table transforms, the menhir conversion, and tests.
- `CodeWithStatics.Model.dynamics` is now the full `Dynamics.t` (probe map plus type instantiations), and `EvalResult.Model`'s accessors return `Calc.t` values; call sites in the stepper, agent, and exercise modes changed accordingly.
- `.changetours/1988-live-typing.changetour.md` is a Change Tour walkthrough, authored at an earlier head (b3f9d69d84). It predates the reporting-policy filter, `reabstract`, and the incremental fixes.
- `editor.css` carries formatter whitespace churn (roughly a third of its diff is `>` spacing and blank lines).

## Known limitations

- `DynamicTypInfer` runs a full `Statics.mk` per sample with no memoization, and the live pass does the same per sample inside `refine_typ_with_dynamics`. Tolerable for a single projector in Dynamic mode, more of a concern here, where live typing needs dynamic types for many expressions at once.
- `Typ.join` inherits `meet`'s Poly hole: a free type variable in one operand that shares the other's binder name is captured. The precision property excludes that shape.


src/haz3lcore/projectors/implementations/TypeProj.re
```diff
@@ -46,7 +46,10 @@ let get_dynamic_typ = (info: info): Typ.t => {
            };
            let types = List.map(type_of, d) |> Util.OptUtil.sequence;
 
-           Option.map(Typ.consistent_join(Ctx.empty), types);
+           Option.bind(
+             types,
+             Typ.join_all(~empty=Unknown(Internal) |> Typ.temp, Ctx.empty),
+           );
```
7h3kk1d: This is switching to ordinary type join for the dynamic type we assign expressions. So `[1,2,3]` ⊔ `["", ""]` is `?` as opposed to `[?]`

src/language/term/Typ.re
```diff
@@ -596,106 +624,111 @@ let rec join =
...
+    | (Unknown(Inconsistent), _) => Some(ty1)
+    | (_, Unknown(Inconsistent)) => Some(ty2)
+    | (Unknown(_), _) => Some(ty2)
+    | (_, Unknown(_)) => Some(ty1)
+    | (Var(n1), Var(n2)) =>
+      if (n1 == n2) {
+        Some(ty1);
+      } else {
+        let* ty1 = Ctx.lookup_alias(ctx, n1);
+
```
7h3kk1d, line 843:
```suggestion
```

src/language/term/Typ.re
```diff
@@ -596,106 +624,111 @@ let rec join =
...
+        let* ty2 = Ctx.lookup_alias(ctx, n2);
+
```
7h3kk1d, line 945:
```suggestion
```

src/language/term/Typ.re
```diff
@@ -596,106 +624,111 @@ let rec join =
...
+        let+ ty_join = join'(ty1, ty2);
+        !resolve && equal(ty1, ty_join) ? ty1 : ty_join;
+      }
+    | (Var(name), _) =>
+      let* ty_name = Ctx.lookup_alias(ctx, name);
+
```
7h3kk1d, line 945:
```suggestion
```

src/web/util/WorkerClient.re
```diff
@@ -2,7 +2,7 @@ open Js_of_ocaml;
 open WorkerServer;
 
 let name = "worker.js"; // Worker file name
-let timeoutDuration = 20000; // Worker timeout in ms
+let timeoutDuration = 60000; // Worker timeout in ms
```
7h3kk1d:
```suggestion
let timeoutDuration = 20000; // Worker timeout in ms
```

src/language/dynamics/transition/PatternMatch.re
```diff
@@ -55,6 +55,12 @@ let rec matches =
   | Ap({term: Constructor(ctr, _), _}, p2) =>
     let* d2 = Unboxing.unbox(SumWithArg(ctr), d);
     matches(p2, d2);
+  | Ap({term: Probe({term: Constructor(ctr, _), _}, _), _}, p2) =>
+    // TODO There's probably a better way to handle the incremental pattern matching for this
+    // TODO We need to actually capture the constructor here
+    // This is addressing https://github.com/hazelgrove/hazel/issues/1987
+    let* d2 = Unboxing.unbox(SumWithArg(ctr), d);
+    matches(p2, d2);
```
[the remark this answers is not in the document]
7h3kk1d: There's gotchas here I would rather wait and solve separately.

src/web/app/inspector/CursorInspector.re
```diff
@@ -90,17 +92,44 @@ let view_any = (~globals, any: Any.t) =>
...
+  dyn_type
+  |> Option.value(~default=typ)
+  |> CodeViewable.view_typ(
+       ~globals,
+       ~settings=code_view_settings,
+       ~is_dynamic=is_dynamic_id,
+     )
   |> code_box_container;
-
+};
```
7h3kk1d: To review and cleanup.

src/web/www/style/cursor-inspector.css
```diff
@@ -17,6 +17,8 @@
   border-color: var(--R1);
 }
 
+
+
```
7h3kk1d:
```suggestion
```

7h3kk1d:
Move the lightning bolt in the error message to the right side of the cursor inspector

src/haz3lcore/derived/CachedStatics.re
```diff
@@ -45,19 +49,57 @@ let all_probeable_ids = (info_map: Statics.Map.t): Id.Map.t(unit) =>
     Id.Map.empty,
   );
 
+/* Collect IDs of expressions/patterns with partially unknown types.
+ * Used for live_typing to automatically probe terms that need dynamic type feedback.
+ * Note: We check the SELF (synthesized) type, not the fixed TY, because
```
7h3kk1d: Rename self to elab_syn_type based on dev changes

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -169,6 +169,23 @@ module SampleLength = {
   let set = (id: int, length: int): unit => Hashtbl.add(lengths, id, length);
 };
 
+/* Remove opaque values like function literals */
+let rm_opaques: list(Sample.Env.entry) => list(Sample.Env.entry) =
```
7h3kk1d: I think this is unused

src/language/dynamics/state/EvaluatorState.re
```diff
@@ -277,6 +287,37 @@ let update =
           step_count: state.step_count + 1,
         };
         (call_stack, state);
+      | RecordAscriptionProbe((id, capture_spec, ascribed_exp)) =>
+        let step = state.step_count;
+        /* Resolve free variables in the ascribed expression via the live
+           env. The ascription transition records `ascribed_exp` as the
+           inner body of a collapsed nested Asc *before* that body has
+           been reduced; if the body is a Var, the recorded sample would
+           otherwise be the symbol rather than the runtime value. */
```
7h3kk1d: Noisy comment

src/language/dynamics/state/EvaluatorState.re
```diff
@@ -277,6 +287,37 @@ let update =
...
+            id,
+            ascribed_exp,
+            env,
+            call_stack,
+            capture_spec,
+          );
+        print_endline("Adding sample");
+        let state = add_sample(state, sample);
+        /* Advance step count past ascription evaluation */
+        // TODO Check with disconcision
```
7h3kk1d: remove todo

src/language/dynamics/transition/Ascriptions.re
```diff
@@ -17,139 +17,163 @@
  either by returning `Some(e)` directly, or by using `IdTagged.fast_copy(DHExp.rep_id(e), ...)`
  when constructing a new expression structure.
  */
-let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
-  let recur = (d: DHExp.t): DHExp.t =>
+// TODO Rename to samples
```
7h3kk1d: remove todo

src/language/dynamics/transition/PatternMatch.re
```diff
@@ -69,7 +74,9 @@ let match_pattern =
   | Parens(p)
   | Projector(_, p) => recur(p, d)
   | Asc(p, t1) =>
-    recur(p, Ascriptions.transition_multiple(Asc(d, t1) |> DHExp.fresh))
+    let (_samples, d') =
```
7h3kk1d, line 77: Why is this unused?

src/language/dynamics/transition/PatternMatch.re
```diff
@@ -110,12 +117,12 @@ let rec matches_inner =
           d: DHExp.t,
         )
         : match_result => {
-  let d = Ascriptions.transition_multiple(d);
+  let (_samples, d) = Ascriptions.transition_multiple(~targets, d);
```
7h3kk1d, line 120: Why is this unused

src/language/dynamics/transition/Unboxing.re
```diff
@@ -130,7 +130,11 @@ let rec unbox: type a. (unbox_request(a), DHExp.t) => unboxed(a) =
     | (LabeledTupleEntries, Tuple(ds)) =>
       let unbox_tup_label =
           (d: Exp.t): option((option(LabeledTuple.label), Exp.t)) => {
-        switch (Ascriptions.transition_multiple(d).term) {
+        // TODO Samples
```
7h3kk1d, line 133: TODO

src/language/statics/Info.re
```diff
@@ -335,6 +335,36 @@ let projector_kind_of = (info: t): option(ProjectorKind.t) =>
...
+      let dyn_typs = OptUtil.traverse(calculate_dynamic_type, exps);
+      let dyn_typ = Option.bind(dyn_typs, Typ.meet_all(~empty=ty, ctx));
+      switch (dyn_typ) {
+      | None => ty
+      | Some(t) => t
+      };
+    };
+  } else {
+    ty;
+  };
```
7h3kk1d: Move to LiveTyping.re

7h3kk1d: unless it creates a circular dependency

src/language/statics/Statics.re
```diff
@@ -187,8 +188,20 @@ and uexp_to_info_map =
       m: Map.t,
     )
     : (Info.exp, Exp.t, Map.t) => {
+  let calculate_dynamic_type = (uexp: Exp.t) => {
+    let (ie, _, _) =
+      uexp_to_info_map(
+        ~dynamics=LiveTyping.Map.empty,
```
7h3kk1d, line 206: Should this be empty?

src/language/statics/Statics.re
```diff
@@ -1511,9 +1540,19 @@ and uexp_to_info_map =
             (module
              {
                let uexp_to_info_map =
-                   (~ctx, ~ana=?, ~is_in_filter=?, ~ancestors=?, exp, m) =>
+                   (
+                     ~dynamics as _=LiveTyping.Map.empty,
```
7h3kk1d, line 1683: Should this be empty?

src/language/statics/Statics.re
```diff
@@ -1598,9 +1637,19 @@ and uexp_to_info_map =
           (module
            {
              let uexp_to_info_map =
-                 (~ctx, ~ana=?, ~is_in_filter=?, ~ancestors=?, exp, m) =>
+                 (
+                   ~dynamics as _=LiveTyping.Map.empty,
```
7h3kk1d, line 1795: Should this be empty

src/language/statics/StaticsBase.re
```diff
@@ -43,6 +43,30 @@ module Map = {
...
+      (_: Uuidm.t, info: Info.t) =>
+        switch (Info.marks_of(info)) {
+        | [] => None
+        | ms => Some(ms)
+        },
+      map,
+    );
+
+  let has_errors = (map: t): bool =>
+    Id.Map.exists((_: Uuidm.t, info: Info.t) => Info.is_error(info), map);
```
7h3kk1d, line 77: Check if these are used

7h3kk1d:
### Review Guide

This PR adds the _live typing_ feature to hazel which supplements Hazel's static type checking with dynamic type inference provided by live probes. It works by automatically observing terms which have statically incomplete types using live probes, and incorporating that data back into the static type checker. This work is downstream of https://github.com/hazelgrove/hazel/pull/1858 which just extends the _type probe_ projector with the ability to view the _live types_ but does not extend the error marks generated from the inferred types. The feature is gated behind a settings toggle.

#### Suggested Order
- [src/language/statics/LiveTyping.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-41d8d4305d2eb767e88c7bf7ce5c5547606b59099ee9e2dedc820f5a1ac1218e) - Provides the core data structure that needs to be produced by the evaluator to be provided by live typing
  - This includes the collection of sampled values by Id as well as instantiated types for System-F style polymorphism
- [src/language/statics/Statics.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-78d299b3e7a12a38d178508ca26481e42e3b8bbb5d0478c396e4452c44df2e6c) - Extends the static type checker with the aforementioned live typing data structures
  - This should perform static type checking as ordinary while falling back to the live typing information in the case that a static type is not fully known using LiveTyping.refine_typ_with_dynamics
- [test/evaluator/Test_Evaluator_LiveTyping.re](https://github.com/hazelgrove/hazel/pull/1988/changes?w=1#diff-17155de2d8548f9de68c625059f0a70c96a9736c40391b6cf9d3b0ef41692d8a) - tests exhibiting the types of live typing errors that should show up 
- [src/language/dynamics/state/EvaluatorState.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-6f594d4d237e47256eeeb7c863d7880f44826308d910b1d52d49d3c37b66f3d4) - Extends the evaluator state with the abililty to capture type instantiations
  - The type instantiations are used to allow for live typing to be compared against the join of all instantiations in a polymorphic function 
- [src/haz3lcore/ProblemCollection.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-c6b5b3ad80ff6591a8cdc2affef4ab34348b12a8e038e77650135105f49c1b81) - adds live typing errors to the problem sidebar
- [src/web/app/inspector/CursorInspector.re](https://github.com/hazelgrove/hazel/pull/1988/changes?w=1#diff-8bed94fcc687e96fc3fb088f1b59001a30879f2f43681e5d23a48074685de07c) - Cursor inspector is changed to thread live typing errors to the bottom bar prioritizing static errors first and adding styling for live typing errors.
- [src/web/app/editors/code/CodeWithStatics.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-eb11abe6a7f6001c1b56c4cc1ec9920d32e0c018eddeb610d28b38154b756511) - The CodeWithStatics logic also has to cache/update the live typing information from dynamics now. This is also where live typing errors and code decorations are added to the state to be rendered.
- [src/web/app/editors/code/CodeViewable.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-6f594d4d237e47256eeeb7c863d7880f44826308d910b1d52d49d3c37b66f3d4) - the code viewing function has been extended to allow for a function projecting html classes for certain terms (specified by id) so that we can mark part of types as "dynamic"

#### Semi-related changes
- [src/language/term/Grammar.re](https://github.com/hazelgrove/hazel/pull/1988/changes#diff-21e4d78e8f3ba603efde8c18d0b5bad9ec8ac220f0d00d887d3d203ed4cf98e3) - moves the type and name for the `Fun` factory builder to optional arguments

PR #1998, opened 2025-11-03
Rich Probes + Table Projector

7h3kk1d:
## Summary

Adds **Rich Probes** (interactive domain-specific visualizations of probed values), the **Table Projector** (renders lists of labeled tuples as interactive tables), and **auto-label support** so the table projector works with unlabeled / out-of-order tuples whose labels come from the expected type.

Ported from `tables-study` (Rich Probes + Table Projector) plus PR #2207 (auto-labeling via elaboration), extracting only these features without pulling in unrelated changes (tutorials, pretty printer rewrite, dynamic statics, etc.).

### New features
- **RichProbe module signature** — extensible system for domain-specific probe renderers with parse/init/render/update lifecycle
- **TableRenderer** — renders probed table values with interactive column menus (drop, rename, sort, group by, filter, move, transform, add column)
- ~~**CardRenderer** — renders probed card/hand values with visual card sprites and chooser UI~~ Reverted; the card projector stays as on `dev`.
- **TableProj** — standalone projector that renders lists of labeled tuples as read-only tables
- **Auto-label support** — TableProj renders unlabeled tuples that receive labels from the expected type, and tuples with out-of-order labels, by consulting the elaborated expression at view time. When elaboration changes in a way that breaks the table (e.g. the surrounding type changes), the raw syntax is rendered inline with an error banner and dashed red border
- **Projector errors surfaced in UI** — new `Projector` problem category; errors appear in the cursor inspector and problem sidebar
- ~~**Settings toggles** — "Rich Probes" and "Tables" in Value Display settings menu~~
  **Settings toggle** — "Tables" in Value Display settings menu. (The `rich_probes` feature flag was removed in `13fe26eddd`; rich-probe renderers are now always available via the sample context menu.)
- ~~**Comparison builtins** — `int_compare`, `float_compare`, `sint_compare`, `nat_compare` for table sorting~~
  **Comparison builtins** — `int_compare`, `float_compare`, `sint_compare`, `nat_compare`, plus `invert_ord` for descending sort

### Rich-probe UX
~~Rich-probe renderers are activated by a per-row badge icon to the left of each probe value.~~
Rich-probe renderers are surfaced as action-items ("View as …") in the existing sample context menu, one per compatible renderer, next to "Pin enclosing call" / "Step into". When activated, the renderer view opens in a floating modal anchored below the sample — the active menu flips to `position: relative` and the flex row top-aligns (scoped via `:has(.modal-backdrop)`) so the modal stacks cleanly below the menu rather than overlapping it. The modal has a floating `×` close button outside its top-right corner. To keep the sample's outline/background from wrapping the now-in-flow menu, visual fills (`background-color`, `outline`, `border-radius`, `box-shadow`) moved from `.sample` to `.value`; specificity is preserved via `.sample:has(.value.X) .value` selectors, so non-rich probes render identically to `dev`.

### Infrastructure changes

*Refractor & syntax rewrite plumbing (Rich Probes):*
- `~inline` parameter on `lift_syntax`/`term_to_seg` — controls whether syntax rewrites use multi-line formatting
- `core_settings` plumbed through `ProjectorBase.View.args` → `ProjectorView` → `RefractorView`
- `SetSyntax` action gains `kind` parameter for refractor-aware syntax rewriting (select term range, replace, preserve manual/ephemeral refractor state)
- `Refractors.to_projector` now receives real syntax from `TermData` instead of dummy whitespace
- `ExpToSegment`: newline before `|>` when not inline, `project_table_if` for auto-projecting tables (with graceful fallback), reverse application parenthesization fix
- `MakeTerm.for_projection`: `Any` case now returns the underlying exp for non-tuples (enables `project_table_if`)
- `Refractors.mk_entry` and `ZipperBase.add_manual` gain `~model` parameter to preserve state across rewrites
- `ZipperBase.update_refractor` — applies updates to both `manuals` and `ephemerals` maps so model state persists for auto-probes
- `TableTransforms.apply_transforms` extracted as a public function for testability

*Elaboration & error plumbing (Auto-labeling, #2207):*
- `elaborate_syntax: bool` flag on the `Projector` module type (like the existing `dynamics` flag)
- `info.elaborated` — elaborated sub-expression looked up by inner expression ID, passed to projectors at view time
- `View.t.error: bool` flag, kept separate from `Model.status` so the upcoming `ViewCache` can key on status without view-time state leaking in
- `error: (model, info) => option(error)` function on `Projector` module type (renamed from `diagnose`) for reporting render-time errors
- `CachedSyntax.projector_errors` map computed alongside shape_map, plus `refresh_shapes` recomputing shapes and errors after statics update (gated on `statics !== old_statics`)
- `Exp.find_by_id` utility moved from `ProofHacks` to `Exp.re`
- `ContextMenu.Projectors.is_applicable` consults the elaborated form, so "Add Table" appears for auto-labeled lists; `actions_data` keeps the current projector's kind available so "Remove Table" stays reachable once elaboration breaks

*Misc:*
- `WebUtil.empty_hole_svg` — helper for rendering empty-hole indicators in table headers without labels

## Notes
- Branch was brought up to date with `dev` in `7694d63e8b`. Notable accommodations for upstream API drift: `MakeTerm.from_zip_for_sem` and `CachedStatics.init` now require `~root` (threaded through `ProjectorPerform.go` and `CompositionView` collapse helpers); `ExpToSegment.Settings` got a new `project_tables` field added to derivation/test settings records.

## Test plan
- [x] `make dev` compiles cleanly
- [x] All 2466 tests pass (`make test-quick`)
- [x] Labeled list of tuples — Table projector activates when "Tables" setting enabled
- [x] Unlabeled tuples with annotated list-of-tuples type — Table projector infers column headers
- [x] Tuples with out-of-order labels — Table projector reorders columns per the annotated type
- [x] When a type change breaks the projected table, raw syntax renders inline with error banner and dashed red border
- [x] Projector errors appear in cursor inspector and problem sidebar
- [x] "Remove Table" stays available in the context menu after elaboration breaks
- [x] ~~Place probe on table expression, enable "Rich Probes" — badge appears on probe values~~ ~~Place probe on table expression, enable "Rich Probes" — dropdown menu shows "View as table" entry~~ Place probe on table expression — sample context menu shows "View as table" entry (rich probes always on)
- [x] ~~Click badge — modal table with interactive column menus (drop, rename, sort, etc.)~~ Click "View as table" — modal table opens below the menu with interactive column menus (drop, rename, sort, etc.)
- [x] Close modal via the floating × in the top-right corner
- [x] Column operations rewrite underlying syntax correctly
- [x] ~~Card renderer works on card/hand expressions~~ (CardRenderer reverted)
- [x] ~~Settings toggles for "Tables" and "Rich Probes" work correctly~~ "Tables" settings toggle works correctly
- [x] Rich probe modal renders above result cells (z-index)
- [x] Rich probe menu entry works on auto (ephemeral) probes


src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -8,11 +8,105 @@ open Language;
...
+        ~parent,
+        (),
+      );
+    },
+    update_packed: (model_str, action_str) =>
+      R.update(deserialize_model(model_str), deserialize_action(action_str))
+      |> serialize_model,
+    badge: R.badge,
+  };
+};
```
7h3kk1d: I hate this encoding

src/haz3lcore/Editor.re
```diff
@@ -120,13 +120,15 @@ module Update = {
...
       (
+        ~print_segment: Segment.t => string,
         ~settings: Language.CoreSettings.t,
         a: Action.t,
         old_statics: CachedStatics.t,
         old_dynamics: Dynamics.Map.t,
         {state, syntax}: Model.t,
       )
       : Action.Result.t(Model.t) => {
+    // print_endline("CodeEditable.Update action: " ++ Action.show(a));
```
7h3kk1d:
```suggestion
```

src/haz3lcore/derived/TermData.re
```diff
@@ -80,9 +80,24 @@ let extreme_measures = (id: Id.t, data: t, measured: Measured.t) =>
...
+      "Termdata.segment Skel range: "
+      ++ string_of_int(l)
+      ++ ", "
+      ++ string_of_int(r),
+    );
+    print_endline(
+      "Termdata.segment: "
+      ++ [%derive.show: Segment.t](ListUtil.sublist((l, r + 1), base_seg)),
+    );
+  };
```
7h3kk1d:
```suggestion
```

src/haz3lcore/derived/TermData.re
```diff
@@ -80,9 +80,24 @@ let extreme_measures = (id: Id.t, data: t, measured: Measured.t) =>
   };
 
 /* The segment corresponding to the `id` term */
-let segment = (id: Id.t, data: t): option(Segment.t) => {
+let segment = (~debug=false, id: Id.t, data: t): option(Segment.t) => {
```
7h3kk1d:
```suggestion
let segment = (id: Id.t, data: t): option(Segment.t) => {
```

src/util/Unicode.re
```diff
@@ -1,3 +1,4 @@
+[@coverage exclude_file];
```
7h3kk1d, line 1: For some reason coverage breaks the regex in here. We should open an issue in the bisect_ppx repo

src/web/app/common/RefractorView.re
```diff
@@ -40,9 +40,25 @@ let mk_data =
...
+            |> Option.map(Segment.parenthesize),
+            ~default=
+              Base.Secondary({
+                id: Id.invalid,
+                content: Whitespace(""),
+              }),
+          ),
+          id,
+          entry,
+        );
```
7h3kk1d: @disconcision I need the syntax on refractors so I can use lift_syntax to rewrite the underlying syntax. This is how I have it working after all the recent changes.

The old way I was doing it is here: https://github.com/hazelgrove/hazel/blob/e9ebe540c1f363f60df4883ecd6be365ca44146f/src/web/app/common/ProjectorView.re#L390-L399

I was similarly pulling the  segment out of the termdata map. The old ProjectorCore took a segment and the new on takes a piece so I'm parenthesizing it. I'm also having to trim secondary to avoid issues with parsing it to a term in lift_syntax. 

I'm not sure if you have thoughts here on a better way to do this.

src/web/www/style/projectors/proj-probe.css
```diff
@@ -376,6 +376,10 @@
 
 .live-offside .value {
   cursor: pointer;
+  padding: 0 0.1em;
+  border-bottom-color: var(--bottom-border-clr);
+  border-radius: 0.05em 0.05em 0.05em 0.2em;
```
7h3kk1d, line 381: Added `padding`, `border-bottom-color`, `border-radius`, and `transition: box-shadow` onto `.value` because these moved off `.sample`. When `.sample-context-menu` flips to `position: relative` (while the rich-probe modal is open), `.sample`'s box grows to include the menu — any fill on `.sample` would then extend to wrap the menu visually. Painting on `.value` scopes the visual to just the probed value. In the default menu-absolute state `.value` was already the only visible child of `.sample`, so for every existing probe in the codebase this paints the same pixels as `dev`.

src/web/www/style/projectors/proj-probe.css
```diff
@@ -386,33 +390,37 @@
...
   --depth-shadow: oklch(0.75 0 0);
   --num-levels: 7;
   --base: 0;
-  padding: 0 0.1em;
-  border-bottom-color: var(--bottom-border-clr);
-  border-radius: 0.05em 0.05em 0.05em 0.2em;
   transition:
     scale 0.1s ease-in-out,
     transform 0.1s ease-in-out,
     translate 0.1s ease-in-out,
```
7h3kk1d, line 405: What stayed on `.sample`: layout concerns that must move the whole sample (incl. its menu) together — `position: relative` (menu anchor), CSS custom properties (they inherit into `.value`), layout transforms (`top`, `translate`, `scale`, `height` on depth variants), and the `scale/transform/translate/top` transition. Plus the `:hover` z-index and the `filter: contrast(…)` on `.Many.sample:has(.value.unrelated/.tangent-*)` (filters cascade to descendants, so applying on `.sample` still filters `.value`).

src/web/www/style/projectors/proj-probe.css
```diff
@@ -386,33 +390,37 @@
...
 
 /* Indet styling only applies to focus-aligned samples; for off-context
  * samples (blue/pink), the call-context color is more informative than
  * the indet indicator, so we let it show through. */
 .live-offside .sample:has(.value.indet.focus) {
-  background-color: var(--T2) !important;
   --depth-shadow: oklch(0.46 0 0);
   --bottom-border-clr: oklch(0.4 0 0);
 }
+.live-offside .sample:has(.value.indet.focus) .value {
```
7h3kk1d, line 416: Note the selector pattern: `.sample:has(.value.X) .value` rather than the simpler `.value.X`. `:has()` contributes the specificity of its argument, so this is 0,5,0 instead of 0,3,0. Without the extra class from `:has()`, `.live-offside.Single .sample .value` (0,4,0) wins over state-specific rules on equal specificity — and we hit that regression during development (Single's green fallback overrode tangent-before / focus / indet bgs). Same pattern used for all value-bound state rules below.

src/web/www/style/projectors/proj-probe.css
```diff
@@ -535,28 +569,25 @@
     border-bottom-color: transparent;
   }
 }
-.projector.probe:not(.indicated) .sample:hover {
-  /* outline-color: var(--exp-indicated); */
-  /* outline-width: 0;
-  outline-style: none; */
+.projector.probe:not(.indicated) .sample:hover .value {
```
7h3kk1d, line 567: Selector changed from `.sample:hover` to `.sample:hover .value` to match the border-bottom moving onto `.value`. Visually identical for non-indicated probes — same pulse, same color, same trigger; just targets the element that now owns the border-bottom.

src/web/www/style/projectors/proj-table-probe.css
```diff
@@ -0,0 +1,216 @@
...
+  margin-top: 5px;
+}
+
+/* When the rich-probe modal is open, pull the active sample menu into
+   normal flow so the modal's `top: 100%` anchor includes the menu's
+   height — modal stacks below the menu instead of overlapping it. */
+.projector.probe:has(.modal-backdrop)
+  .live-offside
+  .sample-context-menu.dropdown-active {
+  position: relative;
```
7h3kk1d, line 94: This rule and the `align-items: flex-start` one below are both gated by `.projector.probe:has(.modal-backdrop)` — a non-rich probe (or a rich probe with the modal closed) has no `.modal-backdrop` in its subtree and matches neither rule, keeping the original absolute/baseline/center layout. Flipping the menu to `position: relative` while the modal is open lets `.live-offside`'s height grow to include the menu, so the modal's `top: 100%` anchor stacks below it instead of overlapping.

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -495,7 +626,6 @@ let value_view =
 
   div(
```
7h3kk1d, line 516: `value_view` used to wrap this `.value` div in an outer flex container and prepend zero-or-more renderer badges as left-aligned siblings (the old badge click opened the rich-probe modal). The badges moved into the sample-context-menu as "View as …" action-items, so the wrapper is gone and `value_view` returns the `.value` div directly. Net effect for every probe: one fewer wrapping div; no class/attribute change on `.value` itself.

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -902,8 +1073,11 @@ let sample_view =
       sample: Sample.t,
     ) => {
   let hide_env = hide_env(ctx.statics);
+  let has_rich =
+    ctx.core_settings.evaluation.rich_probes
+    && List.exists(r => r.can_handle(ctx.sort, sample.value), renderers);
```
7h3kk1d: `has_dropdown` gains a `has_rich` term so the dropdown trigger attaches for samples whose value has a compatible rich-probe renderer even when there's no env/ap_id/call_stack to otherwise justify opening the menu. For a probe with no compatible renderer (or rich-probes disabled), `has_rich` is `false` and this gate is unchanged from `dev`.

7h3kk1d (commented):
@disconcision I want to make sure I don't break any of your probe stuff outside of the rich probes so if you have time I'd like you to at least play with it and feel free to claude-review if you want.

Inline notes on the DOM and style changes that touch every probe sample (not just rich-probe ones), since the restructure has wider surface area than the PR title suggests. The short version: every probe paints visual fills on `.value` now instead of `.sample`, which keeps the sample's background/outline from wrapping the sample-context-menu when it flows in-line (for modal stacking). `.value` was already the only visible child of `.sample`, so pixels match `dev` for non-rich samples.

src/web/www/style/cell.css
```diff
@@ -363,4 +363,5 @@
 .point-num-input {
   width: 50px;
   font-size: 12px;
-}
\ No newline at end of file
+}
```
7h3kk1d: Revert this

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -644,25 +663,52 @@ let step_into_action = (ctx: probe_ctx, sample: Sample.t, ap_id: Id.t) =>
...
+      ),
+    ],
+    [r.badge, text("View as " ++ r.id)],
+  );
+
+let rich_probe_items =
+    (ctx: probe_ctx, local, sample: Sample.t): list(Node.t) =>
+  renderers
+  |> List.filter(r => r.can_handle(ctx.sort, sample.value))
+  |> List.map(rich_probe_action(ctx, local, sample));
```
7h3kk1d: filter_map

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -1368,50 +1450,156 @@ type a = action;
...
+                    Attr.on_click(_ => Effect.Stop_propagation),
+                  ],
+                  [
+                    div(
+                      ~attrs=[
+                        Attr.classes(["modal-close-btn"]),
+                        Attr.title("Close"),
+                        Attr.on_click(_ => local(ToggleModal(None))),
+                      ],
+                      [text({js|×|js})],
```
7h3kk1d: Why is this `js`

src/haz3lcore/projectors/ProjectorPerform.re
```diff
@@ -23,20 +23,60 @@ open OptUtil.Syntax;
...
+    switch (MakeTerm.for_projection(seg)) {
+    | None => None
+    | Some(any) => ProjectorInit.init(kind, orig_piece, any)
+    };
+
+  switch (raw_result) {
+  | Some(_) => raw_result
+  | None =>
+    /* For elaborate_syntax projectors, validate against elaborated form */
+    if (P.elaborate_syntax) {
```
7h3kk1d: Can wee flatten these switches better.

src/web/app/common/ProjectorView.re
```diff
@@ -442,23 +498,37 @@ let split_views =
...
+      font_metrics,
+      ~core_settings,
+      projector_data,
+      projector_list,
+    );
+  /* status comes from statics (sort/error/warning/indication); view_error
+   * is a projector's per-render decision (e.g. TableProj can't render the
+   * current elaborated shape). Keep them separate: status is what the
+   * ViewCache keys on once the Probes III cache lands upstream, so we
+   * don't want to mutate it here. */
```
7h3kk1d: What is this comment

src/web/www/style/projectors/proj-probe.css
```diff
@@ -386,33 +390,37 @@
...
 } */
 
-/* TOKEN BKG */
+/* TOKEN BKG. The .sample element is the positioning container for the
+ * context-menu dropdown and hosts the CSS variables / layout transforms
+ * (top/translate/scale/height) that move the sample as a whole. All
+ * visual fills (background, outline, border-radius, box-shadow) live on
+ * the .value element instead, so they don't extend to cover the
+ * sample-context-menu when it flows in-line (e.g. when the rich-probe
+ * modal is open). */
```
7h3kk1d: This comment seems verbose and confusing

7h3kk1d:
Table context menu should keep column button highlighted. The add column icon is highlighted when leaving the button.

View table button should have different text or indication when it's already viewed as table.

Add new numeric types to transform operations.

Add filter with other option

7h3kk1d:
Add markdown files about rich probes so future 🤖s know how to create them

7h3kk1d:
Value display on by default

src/haz3lcore/CachedSyntax.re
```diff
@@ -74,10 +87,61 @@ let mark_old: t => t =
     old: true,
   };
 
-let calculate = (z: Zipper.t, info_map, dyn_map, old: t) =>
-  old.old
-    ? mk(z, ~info_map, ~dyn_map)
-    : {
+/* Recompute only the statics-derived fields (shape_map, projector_errors,
```
cyrus-, line 90: @disconcision @Negabinary fyi this is worth being aware of / taking a look at

src/util/WebUtil.re
```diff
@@ -147,4 +147,28 @@ module TextArea = {
   };
 };
 
+let empty_hole_svg = (~attrs=[], ()) =>
```
cyrus-, line 225: seems likely to be a duplicate of logic elsewhere

src/web/app/common/RefractorView.re
```diff
@@ -41,10 +41,32 @@ let mk_data =
   let {measured, term_data, selection_ids, _}: CachedSyntax.t = syntax;
   List.filter_map(
     ((id, entry)) => {
-      /* Construct full Base.projector on demand for rendering */
-      let p = Refractors.to_projector(id, entry);
+      /* Construct full Base.projector on demand for rendering,
```
cyrus-, line 44: @disconcision maybe take a look at this

test/Test_ProblemCollection.re
```diff
@@ -49,7 +49,8 @@ let has_structural =
     (p: ProblemCollection.problem) =>
       switch (p.source) {
       | Structural(d) => d == desc
-      | FromInfo(_) => false
+      | FromInfo(_)
```
cyrus-, line 52: maybe add a test for projector error problem collection

cyrus- (changes requested):

src/util/ColumnMenuListener.re
```diff
@@ -0,0 +1,190 @@
+open Js_of_ocaml;
```
7h3kk1d: @disconcision heads-up — this file is copy/pasted from src/util/ContextMenuListener.re with minor edits (target class is .column-menu, plus a keydown listener and a scrollIntoView helper). Worth a follow-up to factor out a shared listener utility parameterized on class name + key/scroll behavior, so both menus share one implementation.

src/haz3lcore/projectors/implementations/TableTransforms.re
```diff
@@ -89,29 +89,48 @@ let can_move_column =
   | None => false
   };
 
-let conversion_functions = (cls: Atom.cls) =>
+/* Extract the underlying Atom.cls from a Typ.t, if any. */
```
7h3kk1d: Unnecessary comment

src/haz3lcore/projectors/implementations/TableTransforms.re
```diff
@@ -89,29 +89,48 @@ let can_move_column =
...
+let conversion_builtin = (~from_: Atom.cls, ~to_: Atom.cls): string =>
+  Atom.cls_string_lower(to_) ++ "_of_" ++ Atom.cls_string_lower(from_);
+
+/* All conversion options from a source class, one entry per other class. */
+let conversion_functions = (cls: Atom.cls): list((string, string)) =>
+  Atom.all_of_cls
+  |> List.filter(target => target != cls)
+  |> List.map(target =>
+       (Atom.show_cls(target), conversion_builtin(~from_=cls, ~to_=target))
+     );
```
7h3kk1d: Can we call these functions from the Builtins modules?

src/haz3lcore/projectors/implementations/TableTransforms.re
```diff
@@ -89,29 +89,48 @@ let can_move_column =
...
-      ("Float", "float_of_bool"),
-    ]
-  | _ => []
+  | Int => Some("int_compare")
+  | SInt => Some("sint_compare")
+  | Nat => Some("nat_compare")
+  | Float => Some("float_compare")
+  | String => Some("string_compare")
+  | Bool => None
+  };
```
7h3kk1d: Use the Builtins functions

src/haz3lcore/projectors/ProjectorInfo.re
```diff
@@ -13,6 +13,11 @@ let utility: ProjectorBase.utility = {
         ...ExpToSegment.Settings.of_core(~inline, CoreSettings.off),
         show_unknown_as_hole: false,
         fold_fn_bodies: `NoFold,
+        /* Disable auto-table projection inside projector-rendered segments:
+         * probe sample values render via Code.view with an empty shape_map,
+         * so a Table projector piece would surface as invisible whitespace
+         * (collapsing the sample's value to nothing). */
```
7h3kk1d: Comment too verbose

src/web/app/editors/decoration/EmptyHoleDec.re
```diff
@@ -1,84 +1,14 @@
...
-    scale_x(s_x, scale_y(s_y, tip_l)),
-    [H_({dx: -. s_x})],
-    scale_x(s_x, scale_y(s_y, tip_r)),
-    [Z],
-  ]);
-
-let path_convex = path(right_tip_path_convex, left_tip_path_convex);
-let path_concave = path(right_tip_path_concave, left_tip_path_concave);
+/* Path geometry lives in Util.WebUtil.EmptyHole so non-decoration callers
+ * (e.g. TableRenderer table headers) can render the same hexagon. */
```
7h3kk1d: Useless comment

src/haz3lcore/projectors/implementations/TableRenderer.re
```diff
@@ -96,76 +110,100 @@ let build_column_menu =
   // If we're in a submenu, show that submenu
   switch (menu_path) {
   | ["Filter"] =>
+    /* Comparators offered for numeric columns. Each lifts to the cls-correct
+     * BinOp via TableTransforms.numeric_bin_op. */
```
7h3kk1d: Not a useful comment

src/haz3lcore/projectors/implementations/TableRenderer.re
```diff
@@ -292,19 +330,18 @@ let build_column_menu =
       | None => []
       };
 
+    /* Show the Filter entry whenever something useful is offered inside —
+     * any atom column can use Equals/Not equal/Custom, numeric columns get
+     * comparators, strings get regex matching. */
```
7h3kk1d: Not a useful comment

src/haz3lcore/projectors/implementations/TableRenderer.re
```diff
@@ -358,16 +395,46 @@ let build_column_menu =
...
+let render_menu =
+    (
+      ~selected_idx: int,
+      ~on_hover_action: int => Ui_effect.t(unit),
+      menu_data: menu_data,
+    ) => {
+  let actions = action_items(menu_data);
+  /* selected_idx counts only Action items; render walks the original list
+   * so separators stay in place. */
+  let action_idx = ref(0);
```
7h3kk1d: Can we do this without a ref?

7h3kk1d:
Combine column menu with ContextMenu

cyrus- (approved):

PR #2007, opened 2025-11-07
Vim-like Keybindings

Cryplo:


cyrus-:
closing in favor of a future design based on #2166

PR #2009, opened 2025-11-10
fix(statics): correct self parameter in uexp_to_info_map for Parens and Probe

### LINKED ISSUE #1997: Inconsistent highlighting of parentheses in errors
Low priority bug relating to parentheses around erroneous tuples/labelled tuples.

If a tuple is labelled the parentheses is included in the error, but otherwise it is not. Ideally it should be in both cases (it's weird selecting the tuple via its parentheses stating that 'this term is consistent' when it obviously looks like it isn't). 

Image

In fact, I think it makes sense to highlight parentheses in general if their inner term has an error, even in a situation like `(e) + 3` like below:
Image

### LINKED ISSUE #2011: Error shows up on let pattern when errored expression doesn't get expected type
Image

If there's an error on a expression that still gives a fixed type that's inconsistent with the expected type we drop the inconsistent error and it pops up in the pattern. I think this is related to the fix I made here and we should just use the ana type for errors that aren't inconsistent https://github.com/hazelgrove/hazel/commit/fac57cda9405516cdbf87c4915c9514b36eda915.

Originally found from the parens_self branch with 

let extra_label : (Int, a=String) = (c=1, a="hello") in true

### LINKED ISSUE #2012: Duplicate labels in labeled tuple pattern give duplicate labels in synthesized type
For expressions `(a=1, a=2)` we mark the duplicate labels as an error and give the type `(a=?)`

For patterns `fun (a=a, a=b) -> 1` is giving the type `(a=?, a=?) -> Int`. It should instead give `(a=?) -> Int`

### LINKED ISSUE #2014: InHole(Redundant(impossible_err)) is raised when invalid label is present in a redundant pattern
```hazel
case () | _ => A| [], a=? => ""
```

Crashes the editor with the `InHole(Redundant(impossible_err))` exception. The issue is that a is in invalid label and the code assumes you can't be a redundant pattern and have a labeled tuple error. This doesn't happen if the labeled tuple is parenthesized since the error isn't on the direct pattern. As part of #2009 this happens for parenthesized tuples. This wasn't caught by the pbt since we wrap all generated tuples in parens when converting from the menhir representation.

7h3kk1d:
Parens and probe now mirror the info of their enclosed expression/pattern/type. Cursor inspector messages and marks appear on the parens themselves, and the parens’ self type matches the inner node’s self type (important for the type projector).

Fixes uncovered while enabling this:
-  Redundant patterns can carry tuple errors; these are no longer thrown or hidden by parentheses. #2014 
-  Duplicate pattern labels were synthesizing the wrong type; this is corrected. #2012 
-  Certain type-preserving errors in lets were incorrectly moving diagnostics from the expression to the pattern because the analytical type on the expression wasn’t preserved; this now stays on the expression. #2011 
- Remove `WithArrow` and `IncompleteVariant` errors that are never used

Behavioral impact
- Diagnostics (messages and marks) now show up on parentheses/probes exactly as on their contents.
- Type projection over parentheses reflects the same self type as the enclosed node.

### DIFF
--- a/test/Test_Coverage.re
+++ b/test/Test_Coverage.re
@@ -1074,6 +1074,114 @@ in ?|},
     ],
   );
 
+let labeled_tuple_additional_error =
+  has_errors(
+    "Labeled Tuple Additional Error Test",
+    {|case () | _ => ""| {{{([], a=?)}}} => "" end|},
+    [
+      Info.Pat(
+        Redundant(
+          Some(
+            Common(
+              TupleLabelError({
+                malformed_labels: [],
+                duplicate_labels: ["a"],
+                invalid_labels: [],
+                typ:
+                  Test_Statics_Prelude.FTemp.Typ.(
+                    prod([list(unknown(Internal)), unknown(Internal)])
+                  ),
+              }),
+            ),
+          ),
+        ),
+      ),
+    ],
+  );
+
+let labeled_tuple_additional_error = {
+  test_case("Labeled Tuple Additional Error Test", `Quick, () => {
+    Test_Statics_Prelude.(
+      annotated_tree_test(
+        {|case () | _ => ""| {{{([], a=?)}}} => "" end|},
+        FTemp.Typ.(string()),
+        FIError.(
+          Exp.(
+            match(
+              tuple([]),
+              [
+                (Pat.wild(), string("")),
+                (
+                  Pat.(
+                    tuple(
+                      ~ann=
+                        Some(
+                          Pat(
+                            Redundant(
+                              Some(
+                                Common(
+                                  TupleLabelError({
+                                    malformed_labels: [],
+                                    duplicate_labels: [],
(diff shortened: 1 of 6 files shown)


7h3kk1d:
@cyrus- I'd like your opinion on this. This will copy the cursor inspector message and marks from whatever is inside the parens to the parens. I'm doing it for probe reasons but I also think it's an improvement in general (especially with tuples).

test/Test_Coverage.re
```diff
@@ -1074,6 +1074,89 @@ in ?|},
...
+                ),
+              ],
+            )
+          )
+        ),
+      )
+    )
+  });
+};
+
```
7h3kk1d, line 1159: I don't think there's a good way to do this for indicated since we'd need to indicate the tuple and parens and the menhir parser doesn't have those as distinct.

PR #2015, opened 2025-11-12
cleanup: remove closure environment from FixF terms

7h3kk1d:
- Updated FixF constructor in MakeTerm.re to omit the third argument
- Adjusted pattern matching and pretty printing in ExpToSegment.re
- Simplified fixpoint evaluation logic in Evaluator.re by removing closure handling
- Removed extraneous None arguments in builtin definitions

This change simplifies the FixF representation by eliminating the optional closure environment, reducing complexity in term construction and evaluation.

@Negabinary I removed it and all the tests seem to pass. I haven't reasoned through anything so not sure why it was there.


cyrus-:
@Negabinary and I decided we should keep the closures, for things like future de bruijn indexed shadow references

PR #2026, opened 2025-11-21
Log Replay

### LINKED ISSUE #2023: Undo is not logged

Negabinary:
Adds a log replay feature to Hazel, when the setting is enabled, a log replay panel appears on the right of the screen


cyrus- (changes requested):
- log replay should stop when it runs out of messages (I believe fixed during meeting)
- add export button back to the log panel
- fix indentation of the import & reset button

okay with merging once those are addressed

cyrus- (approved):

PR #2027, opened 2025-11-24
Rename Typ.join to Typ.meet to be more consistent with other papers

### LINKED ISSUE #2018: Clarify naming of meet/join for types
We _should_ commit to an order for our type precision relation. The code and the 2019 hazelnut live paper use join as the ordering of the lattice places more precise types as higher:

Image

POPL2024 does the opposite
Image

There are certain situations where we may want the opposite of what is implemented so we should make sure that we have some consistency in naming otherwise it's going to get very confusing.

7h3kk1d:
type system's 'join' operation was actually computing the meet
(greatest lower bound) since Unknown represents the highest element in the precision ordering.

- Rename Typ.join → Typ.meet
- Rename Typ.join_all → Typ.meet_all
- Rename join_type_provenance → meet_type_provenance
- Update comment to reference 2019 Hazelnut live paper join

Fixes #2018


Negabinary (approved):

test/statics/Test_Statics_Properties.re
```diff
@@ -11,7 +11,7 @@ let qcheck_statics_does_not_crash =
     | exception Stack_overflow => true // TODO https://github.com/hazelgrove/hazel/issues/1622
     | exception (Failure(f) as e) =>
       switch (f) {
-      | "Type join of ap" => true // TODO https://github.com/hazelgrove/hazel/issues/1459
+      | "Type meet of ap" => true // TODO https://github.com/hazelgrove/hazel/issues/1459
```
cyrus-: the linked issue here is closed

cyrus- (dismissed):
minor comment, go ahead and merge once you're ready

test/statics/Test_Statics_Properties.re
7h3kk1d: Resolved in https://github.com/hazelgrove/hazel/pull/2027/commits/eceb0ade8cfa50b97c383a8fb19a17a4d6de6719

Issue #2004, opened 2025-11-06
Exporting/Importing Scratchpads Incompatible


MaxCarroll0:
Scratchpads are exported as S-Expressions. But importing scratchpads seems to expect .json. 
I'm assuming that the code for importing submissions was rewritten to use json, but this code was also used for importing scratchpads, which should be using S-expressions.

I'm not familiar with this part of code but the relevant functions seem to be: `Widgets.file_select_button_named` `ScratchMode.View.import_button` `JsUtil.read_file` and `ScratchMode.Update.update`

PR #2033, opened 2025-11-25
Fix #1974 #2004

### LINKED ISSUE #1947: 'Export Scratchpad' in nut menu available but ineffective for docs
This is a regression. It exports a file but that file doesn't actually contain the editor data. Not sure this command is necessary for docs but we should either make it work or disable it for non-scratchpads

dm0n3y:
Scratch import button used to only take .json files, despite export button producing .txt files. Now scratch import takes .txt files. I considered exporting/importing .sexp but this isn't a standard MIME type and noticed that this broke file preview in my macOS finder and left it as .txt for now.

This PR also fixes a serialization/deserialization bug where scratch export would serialize all scratch slides while scratch import only deserialized a single slide. Now scratch export serializes only the current scratch slide.

Fixes https://github.com/hazelgrove/hazel/issues/1947 and https://github.com/hazelgrove/hazel/issues/2004

### DIFF
--- a/src/util/JsUtil.re
+++ b/src/util/JsUtil.re
@@ -78,6 +78,13 @@ let read_file = (file, k) => {
     });
 };
 
+let reset_file_input = (input_id: string): unit => {
+  switch (get_elem_by_id_opt(input_id)) {
+  | Some(elem) => Js.Unsafe.set(elem, "value", Js.string(""))
+  | None => ()
+  };
+};
+
 let set_localstore = (k: string, v: string): unit => {
   let local_store =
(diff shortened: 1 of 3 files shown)


dm0n3y:
Noticed that Firefox doesn't add the .txt file extension by default when exporting. Not sure how to change that so I've removed the expectation that the imported scratch file has .txt extension.

dm0n3y:
@cyrus- I can't repro the broken import. Could you try again and report back?

cyrus-:
@dm0n3y It's still happening to me, though the first time I loaded the file it did seem to work but not any subsequent times?

dm0n3y:
@cyrus- fixed

PR #2038, opened 2025-11-25
Fixes #2014

### LINKED ISSUE #2014: InHole(Redundant(impossible_err)) is raised when invalid label is present in a redundant pattern

7h3kk1d:
Fixes #2014 

Allow for tuple errors with redundant pattern errors


7h3kk1d:
Duplicate of https://github.com/hazelgrove/hazel/pull/2009

PR #2048, opened 2025-11-29
Project Mode

russell-rozenbaum:
# Project Mode with Coding Agent

## Before Merge
- [ ] Split out project mode code and have a PR that just enables agentic assistant in docs/scratch mode
- [ ] Resolve Andrew's comments (special fold - unit test, deleted backpack code)

## After Merge
- [ ] Tab to view actions/documentation for actions
- [ ] Improve header readability and descriptiveness of tool calls
- [ ] Add ability for agent to write tests for its own code
- [ ] Allow for agent to write probes
- [ ] Add check for concave grout after edit actions
- [ ] General prompt improvement
- [ ] Working project mode
- [ ] Additional UI work on stepping through / exploring chats
- [ ] Make agent available in tutorial and exercises mode


Desktop/Programmierung/ResearchProjects/assistant-actions/.gitignore
disconcision, line 1: what is this folder 'Desktop/Programmierung'? there are a few files from it

src/haz3lcore/tiles/Base.re
```diff
@@ -39,33 +39,37 @@ let rec segment_to_string =
         (
           ~holes=" ",
           ~concave_holes=" ",
+          ~special_folds=false,
```
disconcision: are these for the code maps? can we see if it's possible to add these by putting folds directly on the segment structure before sending it to the printer to avoid a special case at this level? if it's easier to put them on the term instead this should now be possible, as maketerm and exp_to_segment now retain whitespace/comments with appropriate settings. lmk if there's complexity here.

(also nbd but ideally avoid re-ordering existing params like projector_to_segment to simplify merge conflicts)

src/haz3lcore/tiles/Base.re
```diff
@@ -39,33 +39,37 @@ let rec segment_to_string =
...
+
 and piece_to_string =
     (
       ~holes: string,
       ~concave_holes: string,
-      ~refractors: list((Id.t, _)),
-      ~refractor_seg_to_seg,
       ~projector_to_segment,
+      ~special_folds,
+      ~refractors: list((Id.t, _))=[],
```
disconcision, line 68: can you say why you added a default here? piece_to_string is an internal function that we should be cautious about calling independently; probably better to be explicit about its arguments. ideally we avoid having to call it at all but sometimes special cases must be made. open to discussion here.

src/haz3lcore/tiles/Base.re
```diff
@@ -74,31 +78,51 @@ and piece_to_string =
...
   | Projector(p) =>
-    segment_to_string(
-      ~holes,
-      ~concave_holes,
-      ~refractors,
-      ~refractor_seg_to_seg,
-      ~projector_to_segment,
-      projector_to_segment(p),
-    )
+    if (special_folds) {
```
disconcision: ah okay i get this now... nevermind what i said earlier about putting folds on through seg/term. but instead i think what you're doing here can be done by wrapping projector_to_segment and special-casing what happens with folds. then just call the regular printer with that special projector_to_segment; i don't think any modification to the printer here is necessary. the special projector_to_segment should basically replace projectors which single convex tiles whose label is ["⋱"]; lmk if you don't get what i mean

src/haz3lcore/zipper/action/Action.re
```diff
@@ -82,6 +82,35 @@ type paste =
...
+  | UpdatePattern(string, string) // Path, Code
+  | UpdateBindingClause(string, string) // Path, Code
+  | DeleteBindingClause(string) // Path
+  | DeleteBody(string) // Path
+  | InsertAfter(string, string) // Path, Code
+  | InsertBefore(string, string); // Path, Code
+
+// AddToolLabel_1.0: Make the action types (above) and add their cases to the funs (below)
+[@deriving (show({with_path: false}), sexp, yojson, eq)]
+type agent_editor_action =
```
disconcision, line 111: we can probably call these something like structural_actions instead as in principle they are not agent specific

src/haz3lcore/zipper/action/Action.re
```diff
@@ -82,6 +82,35 @@ type paste =
...
+  // Lists all the use sites of the indicated variable
+  | ShowUseSites(string)
+  // Displays the typing context/scope at the current let expression in the AST
+  // | ShowContext //todo: technically this is accomplished via showing sibs/parent
+  // Displays the entire definition of the current node, with no child/sub definitions abstracted away
+  | ShowReferences(string);
+
+// --- Edit Actions ---
+
+[@deriving (show({with_path: false}), sexp, yojson, eq)]
```
disconcision, line 97: these could ideally use a little more commentary explaining what the various actions do. also these types are a bit suspect, ideally a path is something more structured than a string, but we can discuss that.

src/haz3lcore/zipper/action/Action.re
```diff
@@ -82,6 +82,35 @@ type paste =
   | String(string)
   | Segment(Segment.t);
 
+[@deriving (show({with_path: false}), sexp, yojson, eq)]
+type read_action =
```
disconcision, line 87: not totally obvious to me that these should be `Actions` here as they do not change the editor state. we definitely want a type for them but editor actions specifically have so far been for things that produce changes not what i think is basically a side effect? but may change my mind as i read on.

src/haz3lcore/zipper/action/Action.re
```diff
@@ -121,7 +151,9 @@ module Failure = {
     | CantAccept
     | Cant_undo
     | Cant_redo
-    | CantIntroduce;
+    | CantIntroduce
+    | Composition_action_failure(string)
+    | Cant_derive_local_AST_information;
```
disconcision, line 157: lets me more specific about what the latter failure means, or at least talk about it

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -6,6 +6,15 @@ type state = {
   col_target: option(int),
 };
 
+let mk_statics = (z: Zipper.t): Language.StaticsBase.Map.t =>
```
disconcision, line 9: why does this helper need to be here instead of being part of CompositionGo? if for dep cycle reasons lets say so, otherwise probably should move. let's also be specific about what this is going to get called on. if we want parent editor statics, we should use the cached statics. if not, we want to be deliberate about what the initial ctx is, so the builtin ctx may not be appropriate. not saying this helper is wrong, but just seeing it like this raises questions

src/haz3lcore/zipper/Printer.re
```diff
@@ -74,6 +74,7 @@ let of_segment =
     (
       ~holes=" ",
       ~concave_holes=" ",
+      ~special_folds=false,
```
disconcision: should be able to eliminate this parameter, see prev

src/language/proof/ProofCtx.re
disconcision, line 1: curious what causes the permissions change here; should probably reset these for cleanliness unless there's something wrong with existing perms?

src/llm-tests/tasks/booking-base.yaml
disconcision, line 1: not sure about what this folder should be called or where it should be. maybe in 'tests' directory. not sure we want to commit results to dev. maybe fine but discuss with @cyrus-

src/util/API.re
```diff
@@ -76,6 +76,64 @@ module Json = {
     let* pairs = get_kvs(json);
     List.assoc_opt(key, pairs);
   };
+
```
disconcision, line 79: could maybe use a comment. what is this?

src/util/ListUtil.re
```diff
@@ -305,6 +305,15 @@ let rec zip_defaults =
     ]
   };
 
+let rec opt_zip = (xs: list('x), ys: list('y)): option(list(('x, 'y))) =>
```
disconcision, line 308: add at least a unit test or two for any listutil fns added

src/util/ListUtil.re
```diff
@@ -502,6 +511,12 @@ let rec fold_left_opt =
   };
 };
 
+let intersection = (xs, ys) =>
```
disconcision, line 514: might want to note that this is O(mn) (hash version would be O(m+n)). this is fine but worth noting explicitly in case it gets on a hot path.

src/util/StringUtil.re
```diff
@@ -138,6 +138,85 @@ let prefixes = (s: string): list(string) => {
   };
 };
 
+/* Compute edit distance between two lists of strings using the Levenshtein algorithm */
```
disconcision, line 141: cool! what are we using this for?

src/web/app/sidebar/Sidebar.re
```diff
@@ -189,12 +189,11 @@ let view =
...
-      ~assistant_inject,
-      ~signal,
       ~explainThisModel: ExplainThisModel.t,
-      ~assistantModel: AssistantModel.t,
-      ~editor,
-      info: option(Language.Info.t),
+      ~editors_inject,
+      ~editors: Editors.Model.t,
+      ~editor: CodeWithStatics.Model.t,
+      ~signal,
```
disconcision, line 210: sending in both editors and editor is confusing, as the former contains the latter. why do we need editors holistically?

src/web/app/Page.re
```diff
@@ -32,12 +31,10 @@ module Store = {
         ~instructor_mode=globals.settings.instructor_mode,
       );
     let explain_this = ExplainThisModel.Store.load();
-    let assistant = AssistantModel.Store.load();
```
disconcision, line 49: how is state persistance for the assistant model being handled now? (i may answer this for myself later, just noting in case i dont)

src/web/app/Page.re
```diff
@@ -115,7 +98,10 @@ module Update = {
       |> Updated.return_quiet(~scroll_active=true)
     | Set(action) =>
       let* settings =
-        Settings.Update.update(~action, ~settings=model.globals.settings);
+        Settings.Update.update(
```
disconcision: why is settings being given a callback to call itself?

src/web/app/Page.re
```diff
@@ -525,7 +492,46 @@ module View = {
           if (is_input_field(elId)) {
             ();
           } else {
-            copy(cursor);
+            // Check if the target is within a system-message or agent-message
```
disconcision, line 529: let's break this out to a helper and also explain what's going on a bit better (i don't exactly get what this is doing, and the depth special case it a bit scary)

src/web/init/Init.re
```diff
@@ -35,6 +35,8 @@ let startup: PersistentData.t = {
   ),
 };
 
+let startup_project = Maps.StringMap.empty;
```
disconcision, line 38: if this is meant to live here it could probably use a comment; very unclear what this is

src/web/view/ProjectModeCore/AgentCore/Agent.re
disconcision, line 1: why is AgentCore in ProjectModeCore? what is the conceptual relationship between ProjectMode and Agent?

src/web/view/ProjectModeView/OLD_ASSISTANT/AssistantModel.re
disconcision, line 1: do we want to retain these commented-out files in OLD_ASSISTANT?

src/web/view/NewChat_openrouter_1.76523455245e+12.json
disconcision, line 1: what is this suspiciously named file?

src/web/www/style/nut-menu.css
```diff
@@ -286,7 +286,7 @@
   justify-content: center;
 }
 
-#top-bar #editor-mode .icon:hover {
```
disconcision, line 289: why was this removed? not necessarily bad but I wonder if it might have adverse consequences, ie this now adds a hover effect to icons not within the editor-mode element, which is not obviously desirable

src/web/Main.re
```diff
@@ -153,11 +153,6 @@ let start = {
       >= 0;
     NinjaKeys.initialize(Shortcut.options(schedule_action));
     JsUtil.focus_clipboard_shim();
-    /* Setup scroll listener for floating elements (backpack) */
-    FloatingElement.setup_scroll_listener();
```
disconcision, line 160: i don't think this should be removed; this will break the backpack z-index i think

src/web/Settings.re
```diff
@@ -332,6 +298,17 @@ module Update = {
           ...settings, //TODO[Matt]: Make sure instructor mode actually makes prelude read-only
           instructor_mode: !settings.instructor_mode,
         }
+      | AgentGlobals(agent_globals_action) =>
+        // AgentGlobals updates are handled at Page level with proper async scheduling
+        // This case should not be reached, but we include it for completeness
```
disconcision: given that this mandates a significant chance to the settings API (including the scheduling callback), we probably shouldn't have this if it isn't reached?

test/Test_Info.re
disconcision, line 1: more mode changes

disconcision (changes requested):
Here's a basic review of the integration layer. I think we should make some changes but probably nothing too substantial i dont think. Short notice for this afternoon's meeting; if you have a chance to glance over we can discuss, otherwise can discuss later in the week or next if you are available. some comments are just questions i'm not totally sure about it, mostly about functionality that has moved. 

A few things to break out:
 - I think the internal printer changes you made are probably both unnecessary and may have broken the probe tests; or at least something did. lmk if you don't know what my in-file comments about this mean.
  - I didn't look at the agent-specific stuff in depth but Agent.re is 1723 lines, effectively a god module containing Model + Update + Store for the agent, chat system, and messages all in one file. Should be split soon if not now
  - OLD_ASSISTANT/ directory — 2048 lines of fully commented-out dead code
  - 5 accidentally committed JSON chat logs (NewChat_openrouter_*.json) scattered in source directories — debug artifacts, not test fixtures (maybe add to .gitignore if these are getting generated automatically?)
  - HighLevelNodeMap is 786 lines with complex diff logic — could use documentation (not necessary immediately tho)
  - CSS files are large (~1272 lines for chat messages) — likely fine for now but suggest component-scoped styles later
  - empty files committed (Test_Info.re, Test_MatchExp.re, Test_Stepper.re, DebugConsole.re), just mode changes, no content

dm0n3y:
I can't seem to switch to project mode (testing this on hazel.org/build/project-mode). Keep getting an unhandled exception with the message "No current file" in console. This persists after clearing local storage. Has anyone else encountered this?

src/haz3lcore/tiles/Base.re
russell-rozenbaum: > by putting folds directly on the segment structure

I'm not sure what this means. Maybe we can talk about it in person/call.

src/haz3lcore/tiles/Base.re
russell-rozenbaum: sounds good, still might be helpful to talk over a call, this isn't trivial to me. id still be happy to try to understand more about it

russell-rozenbaum:
> I can't seem to switch to project mode (testing this on hazel.org/build/project-mode). Keep getting an unhandled exception with the message "No current file" in console. This persists after clearing local storage. Has anyone else encountered this?

this comes from Page.Update.get_editor. dev seems to use this function, but our branch never had this. When resolving the merge conflicts, I attempted to remove it, but the Probes tool in the Sidebar seems to deeply depend on it. @disconcision do you know if we need get_editor in Page.Update? I believe Probes is the only thing that depends on it now.

package.json
disconcision: are these load bearing or a merge artifact?

disconcision:
@russell-rozenbaum: Made the folds change I described, and also re-added a dropped line from dev which was causing some probes tests to fail. see: https://github.com/hazelgrove/hazel/pull/2048/changes/386090f378eca65b9206aed4518b7bef2c6b7e7f

cyrus-:
closing in favor of #2131 -- can open a separate PR for project mode things

disconcision:
@russell-rozenbaum @dm0n3y  re: get_editor/"no current file" yeah that's load-bearing to the extent that part of the sidebar functionality is to show a list of the probes in the current editor. is it the case that for project mode that sometimes there is no current editor? if that's the case, we can change this function signature to an option type, and update the sidebar to not show the probes list if its None.

PR #2053, opened 2025-12-17
Gradescope update

thomasporter522:
Updated gradescope related logic in dev


thomasporter522:
The logic in the gradescope directory are LLM translations of the prior perl scripts

src/gradescope/autograders/README.md
```diff
@@ -0,0 +1,110 @@
+# Python Autograder System
+
+This directory contains the Python equivalent of the Perl-based Gradescope autograder system.
```
cyrus-: there is no Perl-based system in the repo anymore

INSTRUCTORS.md
```diff
@@ -37,12 +37,8 @@ between the two without clearing your local storage (in browser dev tools).
...
 
-1. Open the exercise in instructor mode and export a grading version (button in top bar) which generates an OCaml file.
+1. Update the `src/haz3lschool/Specs.re` module with `<module_name>.exercise`.
 
-2. Move the file to `src/haz3lschool/specs`.
-
-3. Update the `src/haz3lschool/Specs.re` module with `<module_name>.exercise`.
-
-4. Run `dune exec ./src/haz3lschool/gradescope.exe <path_to_student_json>` under project root to print the grade report.
+2. Run `make gradescope ZIP=<path to submissions zipfile> COURSE=<course_id> ASSIGNMENT=<assignment_id>` under project root to generate grade reports and upload them to Gradescope.
```
cyrus-: add instructions or pointer to the README for how to print the autograder report to the terminal rather than immediately sending it out to Gradescope

src/gradescope/autograders/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Advantages of Python Version
+
+1. **Simpler Dependencies**: No need for complex module embedding
+2. **Better Error Handling**: Python's exception system
+3. **More Readable**: Python syntax is generally clearer
+4. **Rich Ecosystem**: Easy access to scientific/data processing libraries
+5. **JSON Native**: Built-in JSON handling
+
+## Migration Notes
```
cyrus-: not necessary

src/gradescope/autograders/README.md
```diff
@@ -0,0 +1,110 @@
...
+      "max_score": 10,
+      "output": "Test details..."
+    }
+  ],
+  "stdout_visibility": "visible",
+  "output": "General feedback"
+}
+```
+
+## Advantages of Python Version
```
cyrus-: not necessary

src/gradescope/single_upload.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
```
cyrus-: move to a gradescope subdirectory of a base grading directory

src/gradescope/autograders/AG.zip
cyrus-, line 1: this needs to be generated with a make target, rather than included as a binary blob

cyrus- (changes requested):

thomasporter522:
It currently can grade individual submissions. I think we should merge it so the grading logic remains in sync with dev. The missing features are batch grading, export to gradescope, and autograder generation, and these are hard to fix without access to submissions that are in-sync with current dev (last semesters I think are using different exercise formats), so I propose that we just wait until we have such submissions and then port over these capabilities from the previous commit.

cyrus- (approved):

PR #2055, opened 2026-01-12
Don't serialize worker messages

disconcision:
Changes:
- WorkerClient.re: Changed worker type to Worker.worker(Request.t, Response.t), removed serialize/deserialize calls
- WorkerServer.re: Removed sexp serialization, receives/sends OCaml values directly
- Environment.re: Changed cached_search_tree from Core.Map.t to Maps.StringMap.t (Core.Map embeds a comparator function that cannot be structured-cloned)

### DIFF
--- a/src/language/term/Environment.re
+++ b/src/language/term/Environment.re
@@ -9,8 +9,7 @@ type t('a) =
       id: Id.t,
       binding: binding('a),
       prev_env: t('a),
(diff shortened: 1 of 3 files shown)


Negabinary (approved):
This is huge

disconcision:
Merged as part of #1879.

PR #2056, opened 2026-01-12
Hazel CLI LSP

disconcision:
Features:
- [x] Add autoprobe to command palette
- [x] Multiprobe empty hole behavior supoptimal for writing exercises
- [x] delimiter prefixes shouldn't get probes
- [x] AutoProbe toggle moved to sidebar
- [x] Multi-probe: Make manual auto probes definition/test specific, rename to multi probe
- [x] Toggle probe from selection (round up if not term)
- [x] StepInto: Set manual probe, and make sure pin is preserved, especially in autodef mode
- [x] StepInto: Don't show for built-in functions
- [x] Make undo/redo pass through probe focus
- [x] Pin enclosing call: Add 'Pin enclosing call' sample dropdown. Similar to 'Pin call' but works on anything.
- [x] Printarium: word-wrap printouts
- [x] Printarium: better toggle
- [x] Add DynBar: Better dynamic cursor UI
- [x] DynBar: Don't filter out aps in builtins.
- [x] DynBar: Truncate long call stacks with middle ellipses
- [x] DynBar: tooltip should indicate when syntax jump will round up
- [x] DynBar: Allow unpin via dynamic cursor pin icon
- [x] DynBar (or elsewhere): a remove-all affordance for probes
- [x] Dyncursor: Align to new ephemerals as you type: In the context of EmojiPaint-extend writing task: consider setting the dyncursor when new autos are added or something like that. when start writing a new case, there are no live values if not already aligned.
- [x] DynCursor should reset when there are no probes, or maybe something more specific
- [x] DynCursor: Adding a probe should set dyncursor to that probe's initial sample. In particular, align dyncursor to it if no samples would be otherwise shown
- [x] Sample Dropdown: Show function argument values for aps. When an argument is just a variable reference, append the name as a label and remove it from the environment display
- [x] AutoMode: Top-level definition level autoprobe that follows caret, automatically auto-probing the top level definition that the caret is in. Removes option from context menu
- [x] AutoMode: In automode Pin should add manual probe
- [x] AutoMode: should always show dyncursor bar in autodef mode; otherwise jumps on and off as  you arrow between functions. also try to somewhat integrate auto toggle into position of the bar
- [x] Autoprobe: For TupLabel on line, autoprobe probes TupLabel instead of just the value (nbd but suboptimal)
- [x] Feature: Probe all mode (refactor probe sample gathering in evaluator)

Bugs:
- [x] Bug: ExpToSeg tests indicate regression around spacing of tuplabels after the =
- [x] Bug: StepRange collection is busted for recursive functions
- [x] Bug: memory leak in auto mode (MAYBE fixed by limiting undo stack to 50)
- [x] Bug: In tutorial slide 3, some of the top level data tuples show multiple samples....
- [x] Bug: When probing a function literal, body with let always shows no samples
- [x] Bug: Failure nth on tut slide 3 when clicking back on autoprobe samples after clicking out of the def (when they go blue); stale state i think they should dissapear?
- [x] Bug: Abbreviate considers emoji to be of width 1 
- [x] Bug: builtins showing up as <function> in env
- [x] Bug: clicking on sampels triggers selection sometimes
- [x] Bug: Auto mode dyncursor is borked (try on-screen arrows, or just clicking sample)
- [x] Bug: Add statics to fn literal and look at ctx menu options
- [x] Bug: Autoprobe on go in ADTs reveals incorrect no samples for case (see screenshot on desktop)
- [x] Bug: 'Remove autoprobe' -> multi
- [x] Bug: incomplete `then` delimiter state parsing as a tight multi is suboptimal from feedback perspective
- [x] Bug: In clamp task, new probes aren't getting focussed (or something seemingly inconsistent is happening at least)


disconcision:
Closed in favor of #2154 and #2155.

PR #2061, opened 2026-01-14
Probes III

disconcision:
Now #2056.


cyrus-:
- show sample selector on hover not requiring a click on a sample

disconcision:
Closed in favor of #2056.

PR #2071, opened 2026-01-19
Fix BigInt serialization through web worker postMessage

### LINKED ISSUE #2066: Exception when entering large ints
Image

disconcision:
Closes #2066

BigInteger.js objects lose their prototype methods (like .lt()) when passed through postMessage's structured clone algorithm. This caused crashes when displaying large integers in evaluation results. Starting in v0.17.0, `zarith_stubs_js` switched to using **native JavaScript `BigInt`** instead of BigInteger.js. Native `BigInt` is a primitive type that fully survives structured clone. However, we cannot simply upgrade to zarith_stubs_js v0.17.0 due to versioning issues described in `bigint-webworker-fix.md`, so instead for now we'll vendor the `runtime.js` from zarith_stubs_js v0.17.0 and copy it into the opam switch at build time.


cyrus- (changes requested):
- licensing of the vendored library

cyrus-:
update the PR description

disconcision:
@cyrus- (MIT) license added, PR description changed

cyrus- (approved):

PR #2075, opened 2026-01-21
Sort-dependent expansion for Modules

disconcision:
This implemented sort-dependent expansion. Multi-delimiter forms now only expand when inserted in an appropriate sort context. This is a slightly subtle notion; it currently means that the nearest leftward non-secondary piece has a nib of the appropriate sort. This has a complex but well-tested special case around case rules.

The new functionality is demonstrated below by creating four new forms of pattern sort. They are patterns because I didn't want to bother creating a new sort, but are intended to be representative of what we'd need to implement module syntax. The `let` and `type` forms below reuse `Exp` keywords, but now in a pattern context they expand different. `;` is an infix operator below, not a trailing delimiter. AFAICT this suffices to implement module syntax; see appendix below.

Screenshot 2026-01-21 at 12 47 14 AM

Appendix:

Example of module syntax which the above enables:
```
	let M : {
	  type T = Int;
	  let a: T
	} = {
	  type T = Int;
	  let a = 4;
	  let b = a
	}
```

Required for modules: 2 new sorts: `Mod` and `Sig` and 8 new forms:
```
Typ:= ...
  | "{" Sig "}"
Exp ::= ... 
  | "{" Mod "}"
Mod ::=
  | Mod ";" Mod
  | "type" TPat "=" Typ
  | "let" Pat "=" Exp
Sig ::=
  | Sig ";" Sig
  | "type" TPat "=" Typ
  | "let" Pat
```

Remaining issues:
- If you copy an expanded form to a new sort context, or deconstruct its parent sort context, it will not remold or disintegrate; i.e. to turn a module `let` into a regular `let` or vice versa you'll need to delete all delimiters. Getting this to work is a larger change; I'm not sure on a good approach. Copy-paste could be hacked to work somewhat straightforwardly though (just skip the clipboard cache for segments containing forms with multiple expansions) 
- Probably want an error for reserved words (expanding delimiters used out-of-sort-context)
- Much more functional testing needed; implementing modules should handle this though.


disconcision:
Subsumed by #2075.

disconcision:
Incorporated into https://github.com/hazelgrove/hazel/pull/2123.

PR #2079, opened 2026-01-25
Canonical completion

disconcision:
This assigns a canonical completion of every incomplete segment, laying the groundwork for round-tripping incomplete segments to terms and back (after adding an annotation describing which shards are down; forthcoming).

Unlike the previous dump logic, canonical completions do not depend on caret position. However, ergonomic completions cannot depend simply on term structure. When inserting a new form, e.g. a let expression above existing code, the term structure alone is ambiguous as to whether the new code is wrapping the term below or is purely above it. Previously the caret was used to discriminate here, but the caret-based heuristic is (a) only useful for incomplete forms in the caret segment, and (b) necessitates either re-completing the code on caret movement (expensive and potentially confusing) or having a hidden history dependence (retain the completion of the last edit; current dev behavior; potentially confusing).

The new approach involves addressing two problems of potential history-sensitivity by smooshing them together. The other issue is indentation around incomplete code, which has the same issue of uncertainty around whether or not we're wrapping the thing below when inserting something above. On dev, this currently leads to repeated indentation thrashing, where typing on one line causes lines below to indent and unindent repeatedly. The (partial) dev solution involves resetting indentation level after a blank line (ie two consecutive linebreaks), which is okay-ish for top level definitions but still creates jank when say inserting a new let inside a function literal above existing code.

My candidate solution is this:

1. Remove the existing row-indentation system in favor of 'traditional' indentation, i.e. indentation is now simply regular whitespace.
2. Like most conventional editors, we now don't re-indent everything on every edit action. Instead, we only re-indent in two cases: (a) when pressing enter, insert indentation after the new linebreak only, and (b) when pressing cmd/ctrl-S, reindent everything i.e. autoformat.
3. The enter behavior uses the same indentation logic as before, so you get the indentation you'd expect during left-to-right entry. Nothing below the caret re-indents unless you autoformat before completing the syntax. If you do do this, then you can autoformat after completing the syntax to get correct indentation. (We could also add logic to re-indent ranges selectively on certain actions, e.g. re-indenting segments when a delimiter bounding them is completed, but from experimentation, the new system already feels fine without this)
4. The presence of this history-dependent indentation now allows us a new form of intent-based discrimination: When calculating the canonical completion of an incomplete form, only suck in following lines if they are indented relative to the line the incomplete form is on.
5. This allows the user to indent/dedent lines to get the behavior that they want, similar to the old blank line heuristic, but the beauty of this in conjunction with the enter-based line-specific indentation insertion is that they don't have to; the system inserted indentation seems to suffice for the completion logic to work as expected!
6. The upshot here is that so far the system seems to just work; indentation and completion seem to do what i expect them to do, in the left-to-right entry case, whether writing new code, or code above other code. And if you do pass through a weird incomplete case, when you're back to complete you can just `cmd+s` to get correct indentation again. Could optionally make this automatically happen when the editor is complete, or more granularly.
7. The general principle here to me seems to be: Move all history sensitivity into secondary, and all system history dependence auto-formatting behavior. The user has full control over secondary so none of this is (fundamentally) hidden or out of the control of the user. When the code is complete, autoformat erases any trace of historical dependence; it exists purely during incomplete states.

Still todo:
- [ ] Canonical completion of missing leading delimiters
- [ ] Canonical completion of missing middle delimiters
- [ ] Preserve/restore incompleteness of terms as annotations during maketerm/exptoseg


cyrus-:
the `in` after the `fun` is indented too far:

```
let x =
  let map : ([Int], Int -> Int) -> [Int] =
    fun   -> 
      in 
in 
```

disconcision:
problematic scenarios for round-tripping:

Minor:
- Float normalization
- Sum type syntactic normalization (whether or not it begins with a prefix)

Major:
- Grout UUIDs

disconcision:
Note to self: indentation issue from cyrus above is due to convex holes not being considered terms for indentation

disconcision:
subsumed by #2374

PR #2093, opened 2026-01-29
Fix issue when pasting code with linebreaks in windows

### LINKED ISSUE #2090: Pasting multiline code has weird behaviour
paste any multiline expression in to the editor: e.g. 

```
let x = 5 in
 5
```

and you get these weird red things

Image

Image

disconcision:
Normalize \r\n (Windows) and \r (old Mac) line breaks to \n in the trim_leading function to prevent parsing errors that appear as red error markers when pasting multiline code.

Closes #2090


Negabinary (approved):
Works for me

PR #2106, opened 2026-02-05
Select by range not id in stepper

Negabinary:
Small internal stepper change


Negabinary:
This is part of the next pr actually

PR #2111, opened 2026-02-05
Write out steps

Negabinary:
Adds a setting that forces the user to write out steps in the prover


cyrus-:
reflexivity isn't working in the inductive case

```
let id = fun x -> x in 
let map = 
fun (xs, f) -> 
case xs
| [] => []
| hd::tl => f(hd)::map(tl, f)
end
in
theorem test_thm = 
forall xs: [  ] -> 
map(xs, id) == xs
in map```

cyrus-:
also the proof doesn't seem to work with built in map

cyrus- (changes requested):
- the up triangle in the pop up UI has a different height causing jitter in the UI when clicked on
- proofs about built in map (and presumably other built ins) not going through
- when check is valid, insert it immediately (also for algebrite)
- run checks on every edit rather than when clicked

cyrus-:
@Negabinary the write out steps option seems to disable stepping in the normal stepper entirely, since the pop-up doesn't happen. either have this option only work in the prover, or add a pop up for stepping manually in the normal stepper (which could be useful in lecture sometimes, so preference for this if it isn't a ton of work)

Negabinary:
Stepper: should be just evaluation judgement, so no stepping inside functions, induction etc. Write out steps when setting is on.

cyrus- (changes requested):
Can't undo a step bug: Image

- let fib = <anon fun> should be let fib = <fib> 
- Need to be able to use function names in steps

cyrus-:
image

word gets cut off at bottom

cyrus- (changes requested):
1. Would like to be able to write an explicit fixpoint without type annotations (works with Int -> Int annotation on the fixpoint):

Image

2. Backpack is showing up inside the editor so it is cut off:

Image

3. Make issues related to stepper + filters crashes

4. Make issue about allowing the next step to be checked for equality modulo fixpoint unrolling (how many? where?)

5. Make the "From" editor show folds as in a normal editor

6. Get rid of angle brackets on folded functions

7. Some sort of style distinction rather than postfix `+` for folded fixpoints

PR #2115, opened 2026-02-09
Hazel HTML

disconcision:
Elm-style web apps as Hazel programs. Write a 4-tuple `(init, update, view, subs)` anywhere in a program, project it, and it runs:

```
type Action = Int in
let update(model: Int, action: Action): (Int, Cmd) = (model + action, CmdNone) in
let view(model: Int): HTML =
  Div([], [Button([OnClick(1)], [Text("+")]), Int(model)]) in
(0, update, view, fun _ -> SubNone)
```

Editing `update` while it runs rebinds the live app and keeps its state, as long as the new `view` still accepts the old model.

`HTML`, `Attr`, `Cmd`, `Sub` and `KeyEvent`/`MouseEvent` are builtin types; `HazelDOM` renders values of them to vdom. App state lives in `AppStore`, an id-keyed store in page state rather than in the syntax — models hold closures, and `SetModel` is an edit action, so putting it in the tree would mean whole-document statics per message. Dispatch is `is_edit=false` and non-historic, so clicking a button doesn't touch statics or undo.

The model comes first in `update`, matching `fold_left`'s `(acc, elem)` and Hazel's subject-first convention, so `update` folds directly over a list of actions. `update` always returns `(Model, Cmd)`; a program that issues no commands can return a bare `Model` and lift it with a two-line `noCmd` at the tuple, which keeps the 4-tuple's shape uniform. (The runtime tolerates a bare model via a warning fallback, but relying on it would put two `update` types through one slot.)

There is one HTML projector, not two. It picks its commit target per render from the projected expression's live value: a 4-tuple gets **state** commit, where the model lives in `AppStore`; bare HTML gets **syntax** commit, where the expression itself is the model and an action is an `Html -> Html` transform spliced back in. So "self-modifying HTML" is not a separate feature, just what the projector does when handed HTML — and it only works on closed HTML, since evaluation erases variables before the splice.

To facilitate displaying HTML apps alongside code, all projectors gain a generic placement toggle (`Inline | Sidebar`): Alt+S on any projector docks it to a new Projectors panel, leaving a chip at the code site.

Ten example apps in `hazel-programs/mvu/`, each also shipping as a documentation slide under **MVU / …** with its app already running and docked. `regen-slides.sh` re-encodes the slides from the `.hz` sources; nothing detects a stale encoding, so that script is the only correct way to edit them. The generated slides in `src/mvu/` are most of this PR's diff — ~54k lines, in line with the existing `src/b2t2` (5.8 MB) and `src/web/init/docs` (1.3 MB) slide libraries.

Also here because the ~50-variant recursive builtin types made it necessary: `ConstructorMap.venn_regions` was O(n^2), builtin constructor annotations now stay in compact `Var` form instead of expanding, and physical-equality fast paths were added to `Typ.meet`/`normalize`/`Equality`.

`docs/mvu.md` is the feature doc: program shape, the type and event reference, the design decisions and the known limitations.

Known caveats: `init` accepts any `Var`/`Ap` (the value can only be checked after evaluation), so the HTML context-menu entry offers itself broadly; `AppStore.gc` is implemented but unwired; the Projectors panel shows only the reported editor's projectors; subscriptions are torn down and recreated on every model change, so timers restart their countdown each tick. Opening a slide shows the code plus a chip, and the app appears once the Projectors panel is revealed — the panel choice then persists, so that is one click per browser profile rather than per slide.

Not caused by this branch but worth flagging, since interactive apps make it visible: an MVU dispatch is dominated by `IncrEval.remove_pat_bindings` (~50%) and UUID minting during evaluation (~30%). Measured, with a writeup, but untouched here.

---
### Fix brought over from the aria-demo branch (2026-09-09)
- f21ac5f8dd `let j : JSON = Assoc(..)` stuck as a cast since the compact builtin types: the constructor's own type is an alias of a recursive builtin sum and the ascription rule only unrolled it structurally. Weak-head-normalize it first. New test in Test_Evaluator_Builtins fails on this branch without the fix; dev is unaffected.


disconcision:
Screenshot 2026-02-10 at 9 21 46 PM
Screenshot 2026-02-10 at 9 21 30 PM
Screenshot 2026-02-10 at 9 21 12 PM
Screenshot 2026-02-10 at 9 20 52 PM
Screenshot 2026-02-10 at 9 20 40 PM
Screenshot 2026-02-10 at 9 20 30 PM
Screenshot 2026-02-10 at 9 20 17 PM
Screenshot 2026-02-10 at 9 20 09 PM
Screenshot 2026-02-10 at 9 20 01 PM
Screenshot 2026-02-10 at 9 19 50 PM
Screenshot 2026-02-10 at 9 19 28 PM
Screenshot 2026-02-10 at 9 19 14 PM

src/language/dynamics/transition/Ascriptions.re
```diff
@@ -160,8 +186,7 @@ let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
         ),
         Sum(m) as sumt',
       )
-        when
-          Typ.is_consistent(Ctx.empty, Typ.unroll(sumt), sumt' |> Typ.temp) =>
+        when Typ.is_consistent(ctx, Typ.unroll(sumt), sumt' |> Typ.temp) =>
```
7h3kk1d: I realize this is still in draft. I just didn't want you to run across it again.
🤖 *Generated with Claude Code.*

`sumt` is the **constructor's own** result type, and it isn't normalized here. `compact_builtin_recs` leaves builtin aliases compact (`Var("HTML")`), and `Typ.unroll` only unrolls a `Rec` — it doesn't resolve a `Var`. So this compares `Var("HTML")` against the already-unrolled `Sum(...)` on the right, the guard fails, and we fall through to no transition. The ascription never pushes into the constructor, and `case` ends up sitting on an ascribed value it can't see through.

Net effect: annotating a parameter with a builtin **recursive** alias (`HTML`, `JSON`, `Cmd`, `Sub`) makes pattern matching on that argument silently stop reducing. User-defined recursive types and non-recursive builtins (`Option`) are unaffected.

```
let f : JSON -> JSON = fun j -> case j | Null => Null | _ => Null end in
f(Null)
```

**Expected** `Null`. **Actual** on this branch:

```
case Null:JSON
| Null => Null:JSON
| _ => Null:JSON
end
```

This is the invariant the file header sets out — "All structural pattern matching on types must go through weak_head_normalize first" — and line 54 already obeys it for the *ascription* type. This guard and the nullary-constructor one at line 204 are the two sites that missed it.

Resolving at the comparison keeps the compaction and its perf win:

```suggestion
        when
          Typ.is_consistent(
            ctx,
            Typ.unroll(Typ.weak_head_normalize(ctx, sumt)),
            sumt' |> Typ.temp,
          ) =>
```

Introduced in `bce50ac5fc` ("Fix post-eval statics performance: compact builtin types + meet fast path") — it adds the compaction, adds the header invariant, and applies it to the ascription type but not the constructor type, so the invariant and its violation land together. Bisected against its parent `5b1bf9e2cb`, where the same program reduces. Only `HTML` was affected then; `JSON` was pulled in later by `b18b27d8bd` (a dev merge with conflicts in `BuiltinsADT.re`), since the compaction keys off `BuiltinsADT.type_aliases` — which is why this now regresses pre-existing functionality.

Both cases reduce correctly with these two changes, and the evaluator / statics / elaboration suites pass. I have it on a branch with regression tests in `Test_Evaluator_Sum_Types`, next to the existing `Br annotation is Var("HTML") after eval` test that pins the compaction itself — happy to open a PR against `hazel-html`.

**How I hit it:** building a rich-probe renderer for `HTML` values on top of this branch. A handler committing an `Html -> Html` transform wrote the stuck `case` back into the user's program via `SetSyntax` — so anything that evaluates a transform and splices the result should check that what it's about to commit is still html, independently of this fix.

src/language/dynamics/transition/Ascriptions.re
```diff
@@ -176,7 +201,7 @@ let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
       | None => None
       };
     | (Constructor(_, Some(Some(t))), t')
-        when Typ.is_consistent(Ctx.empty, Typ.unroll(t), t' |> Typ.temp) =>
+        when Typ.is_consistent(ctx, Typ.unroll(t), t' |> Typ.temp) =>
```
7h3kk1d: 🤖 *Generated with Claude Code.*

Same as line 189: `t` is the constructor's own type and needs resolving before the comparison, or a nullary builtin constructor (`Null`, `Br`) never matches its own sum. This is the guard the `JSON` repro above trips.

```suggestion
        when
          Typ.is_consistent(
            ctx,
            Typ.unroll(Typ.weak_head_normalize(ctx, t)),
            t' |> Typ.temp,
          ) =>
```

PR #2118, opened 2026-02-11
Rujula's tests: photoalbum and flowershop

RujulaAdwant:


RujulaAdwant:
Additional files beyond my yaml test files have somehow been edited, and I want to find a way to "uncommit" these. Is there a way for the PR to be approved, but only accepting the test files? For example, package-lock.json is resulting in a merge conflict.

cyrus- (changes requested):
- in the future make a branch rather than a fork (should now have an invite to the org)
- check the syntax of your code in Hazel before pasting it into the yaml files
-

cyrus-:
note: move to the new llm-tests repo

PR #2104, opened 2026-02-05
Baby Modules

disconcision:
Potemkin modules. Cargo modules. I can believe theyre not modules. Subsumes #2075.

PR #2123, opened 2026-02-13
Modules I

disconcision:
Replaces #2104.

This adds a basic version of modules, suitable for namespacing and code organization, which is really just a syntactic gloss over labelled tuples. This is intended to be transitional progress toward a full modules system including proper statics and dynamics. However this does now include module keyword syntax, capitalized module variables, and type member access; see docs slide and docs.

See [docs here](https://github.com/hazelgrove/hazel/blob/1739b3f4d46aae5fac72a159879ff7f49c2a22d2/docs/modules.md) for a comprehensive description


cyrus- (changes requested):
- auto probe on a module binding shouldn't probe the signature
- auto probe on a module binding shouldn't probe the body
- make referencing a bare module name an error

cyrus- (changes requested):
additional:
- hole in module bodies says "Whitespace or Comment" in the cursor inspector
-

cyrus-:
additionally either in this PR or make an issue for removing the type alias shadowing restriction:

```
module M = {
type T = Int;
let x : T = 3;
let y : T = 4;
module Q = {
type T = String;
let x:T = "abc" # should not have an error #
}
} in 
```

disconcision:
@cyrus- all fixes made and tested except for "make referencing a bare module name an error" which makes sense to wait until either modules aren't tuples or a more precise relationship between them is established

cyrus-:
@disconcision the failed test (statics does not crash) looks like its pointing to a potentially real problem here

disconcision:
@cyrus- not evident to me... last run gives `^^fold(fun (_) -> (()))` i've seen similar before on dev

disconcision:
@7h3kk1d is there a quick path to deriving a failing unit test from a qcheck failure like the above? i pasted the above into the editor and it doesn't cause a statics crash there so i'm assuming it's down to something in the Term structure created that isn't reflected in the printed representation?

7h3kk1d:
> @7h3kk1d is there a quick path to deriving a failing unit test from a qcheck failure like the above? i pasted the above into the editor and it doesn't cause a statics crash there so i'm assuming it's down to something in the Term structure created that isn't reflected in the printed representation?

The seed is printed and you can rerun the seed to get the same failure. There's no quick way to export a unit test afaik. Originally I was printing the full expressions but it was unreadable and now prints. Not sure if that's the issue or not.

cyrus- (changes requested):
- unused variable warnings in signatures
- unused variable warnings should account for uses of module names in types

disconcision:
@cyrus- changes made

cyrus-:
@disconcision ReparseBackuptext for Modules seems to be failing

disconcision:
@cyrus- it was that empty hole at the bottom... hopefully fixed now

cyrus- (approved):

PR #2131, opened 2026-02-20
Coding Agent

russell-rozenbaum:


disconcision:
FYI @russell-rozenbaum  I did some cleanup. In particular, in Action.re, the agent actions, I consolidated to the Structural (action) submodule. I made some changes to the API and added some comments about what the different actions do. I also factored out the read actions, now called LanguageServerActions, and they're in CompositionAction. I also factored out Initialize because it wasn't actually implemented in perform. So I've just left it in composition action. Insofar as the editor is concerned, it's just select-alling and deleting. Let me know if you have any other questions about the changes I've made or any concerns.

disconcision (approved):
code lgtm from a hazel systems integration perspective. we should keep tabs on how big the assistant model is getting; as it stands it looks like it can only grow; could become a memory issue. but won't come up unless the assistant is being used so we can defer this for now

dm0n3y (commented):
tested the agent on a few tasks:
- implement a stutter function that takes a list of numbers and returns a new list with elements of the original duplicated
- implement an arithmetic calculator that takes an arithmetic AST and returns the evaluated output
- implement matrix multiplication
- implement a sudoku solver (barely tested this one bc matrix multiplication was challenging enough depending on model)

mostly tested using gemini flash 2.5, tried out sonnet 4.6 at the end

a lot of persistent syntax issues at first, esp writing `case ... of`, not knowing how to write sum types or tests, sometimes still writing `let rec`. i updated the syntax notes to give examples of new type definitions, sum types, tests, and to be more specific about `case` and that seems to have helped. gemini flash 2.5 still will mess up `case ... of` and `let rec` sometimes tho.

there are also recurring path errors. earlier in my testing it seemed to happen after any bindings were removed (eg when i asked it to refactor some plain function invocations bound to variables into standalone tests). even when i wasn't removing bindings later tho, i saw these occurring later too, often with the empty path.

these issues generally wouldn't be blockers, agent would keep trying and figure it out, but they'd fill up the context and local storage such that gemini flash 2.5 would consistently run out of local storage memory before it could finish the matrix multiplication task. sonnet 4.6 on the other hand killed it, no failed steps afaict, so it completed matrix multiplication no problem. 

btw trying to use any anthropic models would consistently 400 so i had claude diagnose and fix our openrouter api calls

disconcision:
Functional review (task: implement Fibonacci). 
Went pretty well. One issue:

Here, the details for the 'update body' tool doesn't actually show the body in the before/after:

Screenshot 2026-03-19 at 4 37 26 PM

disconcision:
General comments on the UX of using the workbench:
- When you press enter to send a message:
- (a) your message disappears, and 
- (b) there is often considerable time before any other change happens in the UI
- I feel like optimally (a) would not happen so abruptly; instead it would move into a bubble in the log or something
- Some kind of thinking entry should probably immediately appear in the UI to show something is in progress.
- we should indicate when something is in-progress, and what is in progress. In general, probably the current step could use some kind of simple animation.

disconcision:
Experience: Fixing buggy 'Crop Plotter' program from probes user study. (gemini 3 flash)

The assistant succeeded in finding and fixing the bug. It also put probes on, which is cool... not clear if it used them meaningfully though.

Bug: leading whitespace not being trimmed for code insertion:
Screenshot 2026-03-19 at 4 47 29 PM
(This seems to be a persistent issue so might be worth trying to unit test)

Looking at the log, the first thing that struck me is that the use of probes doesn't seem to be reported as a tool use? How does probe use/placement work exactly?

Screenshot 2026-03-19 at 4 53 37 PM

The next thing that strikes me is that although the first of three tasks is just called 'investigate', both the investigation and the fix are done here. Not sure how much of this is just down to how the model chose to label things versus the harness? 

Also I'm curious what if anything the subsequent two tasks actually did. For example did the verify tests task actually verify anything?

Expanding the actions:

Screenshot 2026-03-19 at 4 51 08 PM

Minor notes: When this action suceeds i'm not sure how useful the 'result' entry is. not sure we need to show 'no diff to display' either.

More substantively: Reporting the paths here i great, but I feel a bit of uncertainty given that I can't see the code map. I'd like to know what the model is seeing when it makes the decision to request those paths, as well as what it sees after it does.

Screenshot 2026-03-19 at 4 51 17 PM

Again in the success case here I'm not sure we need to show the result. also, in the success case the 'code' field is redundant to the 'after' display. I think we want something more like:

- Path
- Before
- After


All in all though this felt pretty good!

PR #2139, opened 2026-02-23
Fix unwanted selection when clicking probe samples

disconcision:
- Clicking a probe sample to focus it would sometimes create a text selection extending from the caret (at the probe term) toward the sample position
- Root cause: the projector wrapper stops `pointerdown` propagation so `MouseState.pointerdown()` is never called, but `mousemove` events still bubble to CodeEditable's `drag_select` handler, which only checked the physical button state (`pointer.buttons`) without verifying `MouseState.is_button_down()`
- Fix: add `MouseState.is_button_down()` guard to `drag_select` so it only creates selections when the editor actually recorded the initial pointerdown


disconcision:
@7h3kk1d you probably want this on your study branch

disconcision:
Merged via #2154

PR #2144, opened 2026-02-24
Fix duplicate probe samples from Dot projection and Asc distribution

### LINKED ISSUE #2142: Repeated evaluation of top level tuple/list elements
```
let f: ? -> (?, ?) = fun t -> t in
f(?, ^^probe(1))
```

```
let f: ? -> (? = ?, ? = ?) = fun t -> t in
f(? = ?, ? = ^^probe(1))
```

```
let f: ? -> [?] = fun t -> t in
f([?, ^^probe(1)])
```

In all above programs the probe is evaluated twice.

disconcision:
Closes #2142.

## Summary

- **Dot projection** on Tuple/TupLabel incorrectly set `is_value: false`, causing re-evaluation of already-final projected values. Fixed by setting `is_value: true`, consistent with other value-extracting transitions (Cons, ListConcat, BinOp, UnOp, TupleExtension).
- **Ascription distribution** through typed functions re-evaluated inner compound elements at a deeper call_stack, creating duplicate samples. Fixed by deduplicating in `add_sample`: if a top-level (empty call_stack) sample already exists for a syntax_id, skip recording deeper samples from the same definition site.
- Added 7 regression tests covering both mechanisms.

**Repro**: `let x = (a = ^^probe(1), b = ^^probe(2)) in let f: (a=Int,b=Int) -> (a=Int,b=Int) = fun t -> t in f(x)` — showed 2 samples per field instead of 1.

## Test plan

- [x] All 328 Evaluator tests pass
- [x] New `Evaluator.Probes.DuplicatePrevention` group: Dot projection, typed identity (labeled/unlabeled tuples, lists), chained functions, multi-call legitimacy

### DIFF
--- a/src/language/dynamics/state/EvaluatorState.re
+++ b/src/language/dynamics/state/EvaluatorState.re
@@ -60,8 +60,32 @@ let add_test = (state: t, instance_report: TestMap.instance_report) => {
     ),
 };
 let add_sample = (state: t, sample: Sample.t) => {
-  ...state,
-  probes: Sample.Map.extend(sample.syntax_id, sample, state.probes),
+  /* Deduplicate: skip recording if a sample at the top level (empty
+   * call_stack) already exists for this syntax_id. This prevents
+   * duplicate samples caused by ascription distribution: when a value
+   * with a probe target ID flows through a typed function, the Asc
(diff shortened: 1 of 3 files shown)


disconcision:
see #2145

PR #2145, opened 2026-02-24
Fix duplicate probe samples from Dot projection and Asc distribution

disconcision:
This fixes 3 evaluator issues resulting in probes collecting duplicate samples during evaluation.

- **Dot projection** on Tuple/TupLabel incorrectly set `is_value: false`, causing re-evaluation of already-final projected values. Fixed by setting `is_value: true`, consistent with other value-extracting transitions (Cons, ListConcat, BinOp, UnOp, TupleExtension).
- **Ascription distribution** through typed functions re-evaluated inner compound elements at a deeper call_stack, creating duplicate samples. Fixed by deduplicating in `add_sample`: if a top-level (empty call_stack) sample already exists for a syntax_id, skip recording deeper samples from the same definition site.
- **Ascription ID preservation**: `Ascriptions.transition` gave fresh IDs to Let, Seq, Parens, and Projector when pushing ascriptions inward, violating the stated ID preservation principle. This caused probes on compound expressions (e.g. `^^probe(let _ = 1 in true)`) inside typed function bodies to produce zero samples, since the Projector (probe wrapper) lost its target ID. Fixed by adding `IdTagged.fast_copy`, consistent with how Fun, Tuple, If, Match, ListLit already handle it.
- Added 9 regression tests covering all three mechanisms.


disconcision:
Merged via #2154

PR #2152, opened 2026-02-26
Fix #2124

### LINKED ISSUE #2124: Active stepper causes large main editor slowdown in ADT doc slide
there's something very slow about the stepper, at least for the ADT doc slide. if I go a couple steps into the last test on that slide, and then try to move the caret in the main editor, versus not having the stepper open, on my desktop machine:

arrow key movement, stepper closed: 10ms
mouse click movement, stepper closed: 15ms
arrow key movement, stepper open: 250ms
mouse click movement, stepper open: 500ms

and trying to step through it in full triggered 'QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of 'SAVE_DOCUMENTATION' exceeded the quota'.

Negabinary:


Negabinary:
I think we should probably archive this one for now - it's not adding too much value.

Issue #2058, opened 2026-01-13
Probes Enhancements


disconcision:
This is a meta-issue for probes features and cleanup that aren't going to make it into #1879 and may not make it into #2061. Will separate these out as they become pertinent.

Stretch goals for study:
- [ ] Multi-line sample value view, either offside or in sample dropdown
- [ ] Try to see if we can replace step-into by jump to definition, but if used on a probed ap, or the fn of a probed ap i guess, it accurately sets the dyncursor
- [ ] Consider doing actual space calculations to permit many mode samples to take up more space if there is space instead of having a hard limit

Bugs:
- [ ] DynCursorBar width abbreviation can lag behind

Features:
- [ ] Support multiple probes on a line, offsetting the views vertically, with arrows routing from the probes to their samples
- [ ] Make step into work for function literals in function position
- [ ] DynBar: Dropdown like sample dropdown for actions (pin) and env/args
- [ ] Sidebar: Legend: Add indet?

Cleanup:
- [ ] Streamline refractor map to store just ids
- [ ] Garbage collect refractors map when persisting to localstore or exporting
- [ ] Decide if refractors/probes should still exist in projector panel
- [ ] Code cleanup: Clarify what does in Refractors vs DynCursor vs DynCursorPerform vs ...

Speculative tweaks:
- [ ] consider setting or toggle to send sample dropdown hover to sidebar below quick reference instead of being a hover
- [ ] Cyrus: If selection is non-empty and has no probes, probe/autoprobe keybinding should add autoprobe or bulk probe to using same logic as auto probing
- [ ] Consider that instead of Single/Many mode applying everywhere, only ever show multiple samples on current top-level definition
- [ ] Consider reinstating hover effect for small samples in many-mode
- [ ] Autoprobe vs. Manual Probe Interaction: Prevent adding manual probe inside autoprobe, or un-autoprobe on manual add?

Speculative features:
- [ ] When samples are focussed, grey out lines of code not associated with cursor-unrelated probes
- [ ] Debugging breadcrumbs: Add option to mark a sample as 'wrong', perhaps auto-populated if probe is on a test
- [ ] Subtractive pins

PR #2154, opened 2026-02-28
Probes III

### LINKED ISSUE #2062: horizontal scroll bar always appears

### LINKED ISSUE #2122: Livelits are missing
Livelits are missing

Image

something in DOM says no livelit found

### LINKED ISSUE #2129: copy-paste into external applications does not preserve indentation

### LINKED ISSUE #2142: Repeated evaluation of top level tuple/list elements

### LINKED ISSUE #2151: Indentation Memory Leak
The indentation cache appears to be un unbounded hash table which can grow pretty fast.

### LINKED ISSUE #2158: bad delimiter matching when entering let before let on same line
given `|let x = 1 in x`
entering `let y =`
results in nonsense

### LINKED ISSUE #2160: Code decoration gets more out of sync the longer the buffer
For a really long file the lines start getting out of sync and the arms stop matching the lines. Tested file was about 700 lines

Image

### LINKED ISSUE #2174: Broke Type parse issue
https://github.com/user-attachments/assets/47baf42a-c9ad-4d5d-b289-bbf5611a5dc3

Starting with ("1" : Int)

if you delete the string and left paren then type (1 it gets a broken parse.

disconcision:
Note: This incorporates fixes #2133, #2139, and #2145 which may or may not get merged first.

Closes #2122
Closes #2129
Closes #2142
Closes #2151
Closes #2158
Closes #2160
Closes #2062
Closes #2174

---

# Probe System Overhaul

This continues work on the probes system (#1420 and #1879), incorporating richer call stack navigation, smarter automatic probe placement, a redesigned value abbreviation engine, and numerous correctness/performance fixes throughout the evaluation and editing pipelines. See #2058 for features and feature concepts deferred to future work.

In addition, this PR expands the Hazel CLI with improved static error reporting and the ability to get probe results inline in text form. The main goal is to permit agentic Hazel program development via external harnesses.

---

## Nomenclature Changes

- "Auto probe" → "Multi probe" for the per-line probe data structure
- "Auto probe mode" is the new cursor-following feature (the toggle that automatically places multi probes)
- "Dynamic cursor" → "Sample focus" internally, "Probe focus" in user-facing UI

---

### Unified Probe Action (Cmd+E)

Replaces the old two-shortcut system (manual probe vs auto probe) with a single **context-sensitive Cmd+E**:

- On **definition forms** (let bindings, tests): adds a **multi probe** that expands to per-line probes across the definition body
- On **other expressions**: adds a **manual probe** on exactly that expression
- On a **selection**: probes the enclosing term of the selection (rounds up if the selection is not a complete term)

The context menu adapts its label accordingly. The old Cmd+Shift+E shortcut for auto probe is removed.

### Auto Probe Mode (Cmd+P)

A new toggle mode that **automatically places a multi probe on whichever top-level definition the cursor is inside**, following the cursor as it moves between definitions. Walks the ancestor chain to find the enclosing let/test body, and reconstitutes the probe only when the target changes.

### Sample Focus Bar (Breadcrumb Navigation)

Replaces the old "Dynamic Cursor" sidebar panel with a horizontal **breadcrumb bar** showing the call stack as navigable function names:

- **Windowed display**: Slides a visible window around the focused entry for long call stacks, with ellipsis markers
- **Function name resolution**: Shows resolved names from evaluation, lambda symbol for anonymous functions
- **Navigation**: Arrow keys move call depth, Enter/click jumps to call sites, body icon jumps to definition
- **Pin indicators**: Visual pins on pinned entries with click-to-unpin
- **Clear all**: Button to remove all probes at once

### Enriched Call Stack Frames (Supporting the Sample Focus Bar)

Call stack frames now carry function names and definition-site IDs (not just application IDs). This enables the breadcrumb display, jump-to-definition, and named call displays throughout the UI.

### Argument Value Display

Probes on function applications now show **argument values** inline: `f(x = 5)` or multi-line for tuple arguments. The environment dropdown filters out variables already shown in the call display.

### Pinning Improvements

- **Pin Call** (meta-click on application probes): Pins a specific call sample and creates a manual probe (if not already present) so pins persist when the cursor moves in auto-probe mode.
- **Pin Enclosing Call** (meta-click on non-application probes with call stacks): A new action, not function-application-specific, which pins the enclosing call of any sample from any expression.
- Visual distinction: The pinned call sample gets a red pin icon; a smaller red dot marks samples within the pinned call, making pinning visible even when the pinned application is off-screen.

### Sample Color Schemes

Three selectable schemes color samples relative to the current focused sample:

- **Calls**: Colors based on relative call stack position (previous default).
- **Steps**: Colors based on relative step range — before, after, contains, or inside.
- **Hybrid**: Mostly Steps, except samples in the same closure as the focused one are colored green (as in Calls) for direct association.

A **segmented control** (Calls / Hybrid / Steps) in the sidebar switches between schemes, with tooltips.

### Redesigned Probe Sidebar

- **Legend**: Tooltips explaining each color, adapts to active scheme
- **Toggle controls**: Auto Probe on/off, Samples Shown (One/Many), keyboard shortcut badges
- **Quick reference panel**: Keyboard shortcuts and icon meanings
- **Print panel**: Toggle at the top to activate an alternate sidebar mode for print statements. Structured entries with sequence numbers, source lines, word wrap; Manual/Auto eval modes.

### Improved Keyboard Controls

- **Cmd+Enter** and end-of-line bounce to switch focus from editor to probes
- **Up/Down arrow** navigation between probes when focused
- All probe functions including showing environment now keyboard-controllable

### Improved Manual/Multi Probe Interactions

When a manual probe exists on a term, the multi-probe system no longer creates a duplicate ephemeral probe on the same term. Ephemeral probes are suppressed when they would overlap with manual ones. Additionally, ephemeral probes and can be toggled on/off the same way manual probes are with the lifetime of the containing multi-probe.

### Sample Focus Fixes

Turns out it is hard to maintain sample focus alignment across all reasonable interactions. This is now well-tested.

- **Recursive probe sample navigation fix** (`59100ce019`): Fix for effective-first suffix scan in recursive probe navigation.
- **Probe alignment lag fix** (`2ef5bd1d7a`): Detects stale focus and forces a second pass to fix alignment lag.
- **Intent preservation fixes** (`212b959bad`, `1f5bfebc79`): Fixes for sample focus intent preservation at 3+ nesting levels.

---

## Abbreviation System Rewrite

Complete overhaul of the value abbreviation engine.

### Hard-Cap Budget

To enable predictable resizing, rendered output **never exceeds the budget** (for budget >= 1). Enforced via safety-net retry: if a form overshoots, retry at budget-1 (preserves monotonicity).

### Improved Sequence Abbreviation

- **Pre-computed item count**: Chooses largest count where items + separators + annotation fit
- **Count annotations**: Truncated sequences show `...+3` instead of bare `...`
- **Even distribution for records**: Budget split evenly across labeled tuple fields so all field names appear simultaneously
- **Two-tier fallback**: Items can degrade to bare ellipses before full collapse

### Additional Improvements

- Proper Unicode width-aware accounting (display columns, not bytes/graphemes)
- `Invalid` form classification for truncated labels (avoids backtick quoting overhead)
- Coverage expanded to Module, Match, FixF, Theorem, ProofObject, Use, TyAlias, Closure, TypAp
- Unit-cost atoms (EmptyHole, Deferral) no longer abbreviated to ellipsis
- Font-metrics-based drag resizing instead of hardcoded pixel values

---

## Hazel CLI

Command-line interface (`src/CLI/Cli.re`, `src/CLI/Run.re`):

### `hazel test <file>`

Runs all `test...end` blocks, reports PASS/FAIL/INDET with line numbers and hints. Non-zero exit on failure. `--verbose` flag for passing tests.

### `hazel analyze <file>`

Static analysis with Rust-style error formatting: file path, line/column, source line with `^` caret underlines.

### `hazel probe <file>`

Probe output with Unicode brackets `⟦⟧`. `--auto` flag auto-probes all expressions (one per line) without manual annotations.

### `hazel bench-parse <file>`

Benchmarks parsing performance: unsegmented vs segmented parsing, slow vs fast paste. Reports per-file timings and speedup ratios.

---

## Multi-Probe Placement Algorithm

Rewrite of the probe placement logic (`MultiProbe.re`) that selects which expression to probe on each line of a definition:

- Priority-based pipeline: rightmost-ending, largest term at that position
- Predicate filtering chain: term sort, delimiter prefix rejection, module declaration skip, hole avoidance, function-type avoidance, container redundancy, let-hole skip, avoid targeting sub-structural forms like case rule tiles
- Special adjustments for multi-line if (promote else branch) and incomplete binding forms (promote trailing sibling)
- Extensively documented with examples

---

### Nested Probe Start Stack

`pending_probe_starts` changed from single int to stack per probe_id, correctly handling nested recursive calls.

### Transient State Cleanup

`clear_transient` removes `app_args` (can be 100MB+), `pending_probe_starts`, and `targets` before worker serialization.

### Fix Probe Sample Doubling from Nested Return Type Annotations

When a function with a return type annotation calls an inner function that also has a return type annotation, the outer Asc distribution would re-evaluate inner results, recording probe samples at a different (shorter) call stack. Fixed by using `Ascriptions.transition_multiple` with `is_value: true` to fully resolve all Asc layers in one step.

---

### Parser Optimization: Segmented Parsing

Batch parser operations now use segmented parsing for significantly faster performance on large files. Also included: CLI benchmark (`bench-parse`)

### TyDi Improvements

- Minimum 2-character prefix before suggestions appear
- Type-aware ordering: keywords first when expected type is unknown, variables first when type is known
- Exact match suppression: all suggestions hidden when any candidate exactly matches the token

### InfixDelimiterPrefix Precedence

Lowered from `Precedence.max` to `Precedence.concave_grout` — fixes multi-probe incorrectly selecting partial keyword tokens.

### Editing Robustness

- **Shard theft guard**: Prevents shard theft when typing before complete multi-shard tiles
- **Shard-index monotonicity**: Enforced in rescan to prevent out-of-order reassembly crash
- **Inner-caret remolding**: Fix parent tile not remolded when caret inside bidelimited form
- **Space suppression through grout lifecycle**: Tracks suppressed spaces correctly

### Performance: General

- **Undo stack capped** at 250 entries (prevents unbounded memory growth)
- **Dirty-tracking autosave**: Only re-persists slides marked dirty, eliminating expensive segment comparisons on every autosave cycle
- **Original doc segment caching**: Avoids re-parsing original slides on each autosave

### Performance: Throttle Statics During Typing

Statics are now throttled across all editor modes during typing, reducing latency and CPU usage on fast keystroke sequences.

### Performance: Per-Projector View Cache

Projector and refractor views are cached per-projector, keyed on map identity + status + model + settings_version. Avoids redundant re-renders when only one projector's state changes. `ChangeLength` and `ToggleShowEnv` bump settings_version to invalidate correctly.

### Performance: Remove Ineffective Caches

- Removed `Core.Memo.general` memoization caches that leaked memory without providing measurable speedup
- Removed a hashtable-based indent cache that wasn't providing benefit

### Copy/Paste

- **Indentation preservation**: Editor copy now includes indentation so pasted text preserves structure
- **Fast paste path**: Segment splice for faster paste operations. The existing segment paste cache is retained, but compeletely internalized to avoid poluting the log with the verbose Paste(Segment(...)) action.

### Select Term Improvement

Cmd/Ctrl+D now selects the next containing term up when a term is already selected.

---

## UI/CSS

- Redesigned sidebar header and layout
- Sample focus bar overlaid on editor
- Keyboard shortcut badges in sidebar
- Box-sizing normalization (`border-box`). Should help with mystery scroll bar issues
- Phantom horizontal scrollbar fix

---

## Test Coverage

| Feature                        | Tests                              | Scope                        |
| ------------------------------ | ---------------------------------- | ---------------------------- |
| Multi-probe placement          | `Test_MultiProbe.re`               | 13 categories, ~40 tests     |
| Sample selection (unit)        | `Test_SampleSelection.re`          | 7 categories, ~25 tests      |
| Sample selection (integration) | `Test_Evaluator_ProbeSelection.re` | 5 categories, ~12 tests      |
| Abbreviation hard cap          | `Test_Abbreviate.re`               | 22 expression types verified |
| Duplicate sample prevention    | `Test_Evaluator_Probes.re`         | 9 dedup tests                |
| Ascription ID preservation     | `Test_Evaluator_Probes.re`         | constructor, list, let/seq   |


cyrus- (changes requested):
- [x] when you resize the window horizontally to the point where the code buffer needs a horizontal scroll, you get a tan background in the background of the window to the right of the scroll boundary
- [x] make the top bar only show when the dynamic focus is non-empty
- [x] mutual recursion doesn't provide you the pop up behavior I would expect from function calls (andrew: can't reproduce)
- [x] in right click menu, Select Term is always shown as selected (andrew: this was actually happening because of keyboard selection in the menu, the highlight is load-bearing due to this. however i made i so hovering over the menu pre-empts this) 
- [x] increase default undo history size to 100 and make it configurable

cyrus- (approved):

PR #2155, opened 2026-02-28
Probes study

disconcision:
BUGS:
- [ ] In case rule body, when typing a let above existing code, the dynamic focus can become misalisgned with the let swallows the rest of the case rules after putting down `=` but before putting down `in`. better dump logic might fix.

TODO before STUDY:
- [ ] Disable unused warnings (if not fixed)
- [ ] Turn probe sidebar on by default
- [ ] Set probe focus bar to always on
- [ ] Remove unused slides, rename modes
- [x] All debugging tasks should use model/action nominative types
- [x] Add quickdocs to sidebar

TUTORIAL:
- [x] Populate tutorial folder with tutorial
- [ ] Finish pinning slide
- [ ] Finish step into slide
- [ ] Figure out what to do with print slide
- [ ] Use Tutorial Mode?
- [ ] Image to show focus bar, sidebar, etc?

TASKS:
- [x] Re-theme/cleanup tasks
- [x] Finalize Debug: Crop Plotter Debug
- [ ] Finalize Debug: Harvest Streak
- [ ] Pick two more debug tasks (one easy one hard) and rm the rest
- [x] Finalize Writing: Crop Plotter Extend
- [ ] Finalize Writing: Garden Path
- [ ] Pick two more writing tasks and rm the rest

REVIEW:
- [ ] Revisit tasks in terms of RQs
- [ ] Revisit tutorial in terms of tasks


disconcision:
rebuilt as #2364

Issue #2157, opened 2026-03-03
Unused variables are used


7h3kk1d:
<img width="1058" height="1521" alt="Image" src="https://github.com/user-attachments/assets/eb703264-ac6e-454f-a00b-ca51b09eee43" />

In the tables documentation slide `average` and `e` are marked as unused even though they're used.

cyrus-:
Closed by #2159

PR #2159, opened 2026-03-05
fix: propagate co-contexts through TupleExtension (#2157)

dm0n3y:
The tuple spread operator (...) was discarding co-contexts from both operands, causing variables used inside spread expressions to be incorrectly flagged as unused.

### DIFF
--- a/test/Test_UnusedWarnings.re
+++ b/test/Test_UnusedWarnings.re
@@ -203,5 +203,25 @@ let tests = (
       "let m : { let x : Int } = { let x = 1 } in 5",
       ["m"],
     ),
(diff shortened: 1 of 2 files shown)


test/Test_UnusedWarnings.re
```diff
@@ -203,5 +203,25 @@ let tests = (
...
+    check_unused(
+      "tuple extension: variable used in extension of spread",
+      "let x = 1 in (a=1, b=2) ... (a=x)",
+      [],
+    ),
+    check_unused(
+      "tuple extension: fun param used in spread",
+      "fun e -> e ... (a=1)",
+      [],
+    ),
```
7h3kk1d, line 225: Can we have a test with an unused argument?

7h3kk1d (approved):

PR #2161, opened 2026-03-05
Performance benchmarking

dm0n3y:
Introduces a GitHub action + comment-based trigger to run performance benchmarks comparing base and current branches. Replaces some instances of structural memoization with pointer-based memoization to test benchmarking.


src/haz3lcore/derived/Measured.re
```diff
@@ -396,10 +396,22 @@ let of_segment_inner =
...
 
-/* Memoized for perf. We use an inner function with positional args
-   because Core.Memo.general doesn't preserve labeled argument types.
-   The wrapper provides the nice labeled argument interface. */
-let of_segment_memo = Core.Memo.general(of_segment_inner);
+/* Memoized on the segment pointer via WeakMap for O(1) cache lookup.
+ * This is a major perf win: Core.Memo.general's structural comparison
+ * on Measured.t took ~25ms even on cache hit; WeakMap reduces this to ~0.1ms.
+ * We store and verify all arguments for correctness. */
+let of_segment_cache:
```
disconcision, line 403: curious if you compared this versus just removing the memoization

7h3kk1d:
Noticed while auditing CI for an unrelated cleanup PR stack. Recording it here so
it's on record as known — mainly so future cleanup passes, agent-assisted or
otherwise, don't re-report it.

Both workflows from this branch are on `dev` without the things they call:

- `perf.yml:120,132` run `bash bench/build-and-run.sh`, but `dev`'s `bench/` has
  only `compare.js` — that script exists only on `perf-bench`.
- `bench-compare.yml` `core.setFailed`s at `:65`/`:96` looking for `bench-${sha}`
  artifacts, but no `upload-artifact` step exists on `dev` — it's in `bench.yml`,
  also only on `perf-bench`.

So `/perf` and `/bench-compare` fail for anyone who runs them on `dev`. Invisible
in normal CI, since both are `on: issue_comment`. No action needed if this is
already known.

PR #2165, opened 2026-03-08
Virtual Grout

disconcision:
This is an experimental branch which removes grout from the edit state (no more Grout in Piece). There is no longer a concept of regrouting; consumers of segments (that is, MakeTerm, Code.view, and Printer) simply detect shape conflicts between adjacent tiles and act accordingly. The logic to place hole decorations depends on the contents of the run of secondary (possibly empty) at the conflict boundary. It is fully functional, but with some bugs and questionable behaviors for grout drawing, caret decoration, and cursor inspector indication.

Benefits:
1. No more grout fidgeting
2. No more reparse failures due to grout/whitespace positioning
3. Allows removing over a dozen error-prone special cases in editing involving avoiding jank or molding errors due to grout insertion /removal

Drawbacks:
1. Need to decide where to draw grout (in such a way that the user is happy)
2. Need to handle the zero-width grout case
4. Grout not being material complicates certain aspects of the view layer including indication / selection

I think the current version handles (2) okay and (1) is getting there, but I'm not entirely happy with it. There's a long tail of bugs related to (3), and decisions still to be made about e.g. whether grout gets a selection decoration.

Known Bugs:
- [ ] Tab to next/prev hole works, but is overly granular, going through every secondary position near a hole
- [ ] Caret shape is wrong in some positions near holes
- [ ] Cursor inspector is wrong when on a delimiter after a hole
- [ ] Hole in case scrutinee after a linebreak overlaps with rule bar

Questionable Things:
- [ ] How are holes placed within runs of spaces? Right now, I put them in the middle, which causes them to be non-grid-aligned for even numbers of spaces. My only hard and fast policy so far is that the decoration should not overlap text, which brings me to
- [ ] How should holes with no adjacent whitespace be decorated? I came up with a concept here which I think more or less works.
- [ ] How are holes placed when there are linebreaks in the run? So far I put them on the next line if there is a linebreak right after the leftwards conflicting tile, pursuant to certain edge cases about what's on the next line. I never put them further down, regardless of the number of linebreaks.
- [ ] Should holes effect caret shape? Currently they do not, a change from before (I think probably they should?)
- [ ] Should holes appear selected? Currently they do not, a change from before (I think probably they should?)


7h3kk1d:
This I like:
Screenshot 2026-03-19 at 11 23 35 AM

I don't know why but this makes me deeply uncomfortable
Screenshot 2026-03-19 at 11 23 42 AM

disconcision:
agreed on the hole bisection. unfortunately i think i need to do something clever here but i will endeavour to rise to the occasion

disconcision:
Superseded by #2406 (artifact-grout): same goal — holes visible without occupying the buffer — but the edit state is grout-free and holes are derived display material with deterministic ids (statics/probes/inspector anchor), rather than per-consumer zero-width decorations. Leaving this open for comparison until it's confirmed obsolete.

disconcision:
Close in favor of #2406

PR #2175, opened 2026-03-14
Fix crash when Parens wraps inconsistent branches

### LINKED ISSUE #2163: Parenthesized expression cls has no elements
Typing `let x = (if true then 1 else "")` breaks the editor

disconcision:
Closes #2163


disconcision:
Merged via #2154

PR #2176, opened 2026-03-14
Add problem sidebar panel

7h3kk1d:
## Summary

- Adds a new **Problems** tab to the sidebar that lists all problems (errors, warnings, holes) in the active editor
- Problems are categorized into collapsible **Syntax Errors**, **Holes**, **Static Errors**, and **Warnings** sections
- Supports a **Grouped/Flat** view toggle to switch between categorized and document-order views
- Problem rows are expandable (click chevron) and show line numbers, with the active cursor's problem highlighted
- Color-coded legend and per-category styling matching editor decoration colors
- Warnings respect the `display_warnings` setting
- ~~**F8 / Shift+F8** keyboard navigation cycles through all problems (errors, warnings, holes) with wrap-around~~
  **Tab / Shift+Tab** keyboard navigation cycles through all problems (errors, warnings, holes) with wrap-around, replacing the previous hole-cycling behavior
- Sidebar nav buttons (↑/↓) for previous/next problem
- Tab badge shows total problem count matching the panel contents
- Sidebar tab icon doubles as a status indicator (error/warning/ok color)
- Reclassifies `IsMulti` (juxtaposed terms without an operator) from `NotInHole` to `InHole(NoType(MultiHole))`, surfacing broken expressions as syntax errors
- Reuses `CursorInspector` views for error display in the sidebar, with colon prefix suppressed via a `~show_type_colon` parameter. This is a temporary solution — longer term, error message rendering should be factored out of `CursorInspector` into a shared module that both the inspector and sidebar can consume cleanly.

## Architecture

- **`ProblemCollection.re`** (in `haz3lcore`) — data layer: collects and categorizes problems from statics, grout, and incomplete tiles using lazy `Seq`. Takes `~display_warnings`, `~statics`, and `~syntax` directly to avoid web-layer dependencies.
- **`ProblemSidebar.re`** — view layer: renders problem rows, sections, legend, navigation buttons
- **`SidebarModel.re`** — settings/state: category types (re-exported from `ProblemCollection`), collapsed/expanded state, actions
- **`CursorInspector.re`** — adds `~show_type_colon` parameter so sidebar can reuse status views without the leading ":"

## Testing

- `Test_ProblemCollection.re` covers: clean programs, juxtaposed literals, type mismatches, incomplete tiles, MultiHole detection

## Concerns
- Performance: the sidebar rebuilds problem rows on every render by iterating over `error_ids`, `warning_ids`, grout, and incomplete tiles, then sorting by document position. This could be expensive in large files with many problems.

## TODO
- [x] Line numbers are incorrect when projectors are present (projectors affect visual layout but `Measured.find_by_id` returns pre-projection row positions)
- [x] I have seen some errors without a line number — these occur when a problem is inside a folded projector, so `Measured.find_by_id` can't locate it in the visible layout. These now display "L?" with reduced opacity instead of hiding the line number.


7h3kk1d:
@disconcision I removed the status-indicator and put one in the sidebar button. I would like your thoughts on that or if you think there's another place to put all this.

7h3kk1d:
This also does a "fix" adding multihole errors for unknown tokens at the bottom of a program.

7h3kk1d:
Click button selecting text.

Formatting of messages in sidebar spaces around variables. remove colons before types.

src/haz3lcore/ProblemCollection.re
cyrus-, line 1: move to core

cyrus- (approved):

PR #2126, opened 2026-02-16
Pretty printer + formatted value display in probe dropdown

disconcision:
- Adds a Wadler/Lindig-style pretty printer (`PrettySegment.prettify`) that inserts linebreaks into segments to fit a target width
- Adds a formatted 2D value display section in the probe sample dropdown, shown when the value is too large for inline display

PR #2180, opened 2026-03-18
feat: Wadler/Lindig-style pretty printer with Cmd+S and result formatting

7h3kk1d:
## Summary
- Adds a Wadler/Lindig-style segment formatter (`PrettySegment.re`) for better code layout
- Integrates Cmd+S / Ctrl+S shortcut for reparse/reformat in the editor
- **Applies pretty printing to evaluation results**, so multi-line output (e.g., large records, nested expressions) is formatted with proper line breaks instead of rendering as a single long line
- Adds comprehensive test suite (`Test_PrettyPrint.re` — 880+ lines)
- Fixes reverse application (`|>`) parenthesization precedence
- Removes space before `=` in labeled tuple formatting (`a= 1` not `a = 1`)

The main motivation here is having **multi-line result output** — the specific formatting rules are secondary and can be refined over time. This copies the pretty printer from #2126 but without the probe changes so we can merge it sooner.

cc @disconcision

## Test plan
- [x] `make dev` compiles successfully
- [x] All tests pass (`make test`)
- [x] Verify Cmd+S reformats code correctly in the browser
- [x] Verify evaluation results display with proper line breaks for large expressions


disconcision:
seems to lose some module round trip tests? maybe that came from my PR but was accidental if so i think

cyrus- (changes requested):
quibbles:

1. should only put expressions on same line if they are simple forms, not lets and other "block-like" forms

Image

2. trailing holes should be on their own line at the end of "block-like" forms

Image

cyrus- (approved):

PR #2181, opened 2026-03-19
Store slides as .hz plain text, generate .ml at build time

7h3kk1d:
## Summary

- Replaces ~115,000 lines of hand-maintained sexp-serialized `.ml` slide files with human-readable `.hz` plain text sources
- Adds a build-time codegen tool (`src/hazel_codegen/`) that parses `.hz` files through the Hazel parser and emits `.ml` files with `PersistentSegment.t` records
- Dune rules in `src/content/dune` and `src/b2t2/dune` invoke the codegen tool, so `.ml` files are regenerated automatically when `.hz` sources change
- Extracts doc slides into a new `content` library (`src/content/`) shared by `web` and `CLI`
- Adds CLI `export`, `import`, `check-slides`, and `parse-check` commands for working with `.hz` files

## Motivation

The old `.ml` slide files were thousands of lines of opaque S-expression strings containing UUIDs, tile structures, and mold data. They were:
- Impossible to review in PRs (noisy diffs of serialized data)
- Fragile to edit by hand
- A source of merge conflicts

Now, editing a slide means editing readable Hazel code in a `.hz` file. The build system handles serialization.

## How it works

```
.hz source file (human-readable Hazel code)
       │
       ▼  dune rule
node hazel_codegen.bc.js "Slide Name" input.hz output.ml
       │
       ▼
.ml file with PersistentSegment.t (generated in _build/, not in source tree)
```

The codegen tool depends only on `haz3lcore` (no web/UI deps) and runs under Node.js via js_of_ocaml. Generated `.ml` files live entirely in `_build/` — nothing is generated into the source tree.

## TODO / Open questions

- [ ] **Documentation for adding new slides**: The process for adding a new slide has changed. Previously you'd serialize through the editor UI; now you write a `.hz` file and add a dune rule. We need to document this workflow — likely in a README in `src/content/` or `src/b2t2/`.
- [ ] **Reduce dune rule boilerplate**: Currently there are 49 explicit dune rules (one per slide). We could instead infer the module name and slide title from the `.hz` filename via a convention (e.g., `Basic_Reference.hz` → module `BasicReference`, title `"Basic Reference"`), and have the codegen tool handle the mapping. This would let us use a single rule or a directory-scanning approach instead of maintaining one rule per slide.
- [ ] **B2T2 Datasheet `.hz`**: The Datasheet slide was exported as `B2T2__Datasheet.hz` but isn't wired into a dune rule (it uses `ppx_blob` for its Markdown content separately). Should it be migrated too?
- [ ] **Non-deterministic IDs**: Each build generates fresh UUIDs via `Id.mk()`. This is fine since `backup_text` is the canonical source, but worth noting.

## Feedback welcome

This is a **draft PR** — I'm soliciting feedback on the overall approach before finalizing. The core question is whether this build-time codegen pattern is the right direction for managing slide content, or if there are concerns about:
- Adding Node.js as a build dependency for the codegen step
- The architecture of the codegen tool and dune integration
- Any other considerations

This doesn't have to be the end state — it's a starting point to get human-readable slide sources into version control and eliminate the opaque sexp blobs. Happy to iterate based on feedback.

## Test plan

- [x] `make dev` builds successfully (all 49 slides generated from `.hz`)
- [x] `make test` passes (1715 tests, 0 failures)
- [ ] Manual: open documentation slides in browser, verify they render correctly
- [ ] Verify `hazel export` → `hazel import` round-trip produces stable output


7h3kk1d:
Confirm that files don't get reparsed every compile. Also time the parsing. Try to go to a directory based approach.

7h3kk1d:
To consider

compile/parsing time and make sure its cached
If we could just have a directory structure that would be great
expect-style tests for grout/static errors.

PR #2185, opened 2026-03-22
Move agent data from localStorage to IndexedDB

### LINKED ISSUE #2188: QuotaExceededError thrown as soon as any large slide persists
Due to empty agent chats (sysprompt etc) blowing up localstore

disconcision:
Closes #2188

This is a triage change for the QuotaExceededError issue; it introduces a new db for simplicity rather than using the one used for logging. I'll follow it up with a broader migration away from localstorage which will use a more consolidated/structured storage approach.

- Fixes the `QuotaExceededError` caused by agent data (system prompts + tool definitions ~60KB per scratchpad) filling localStorage when serialized across 52 documentation slides
- Agent chat data now stored in IndexedDB (`hazel_agent_db` database), localStorage gets a minimal empty stub (~few hundred bytes)

### DIFF
--- a/src/web/StoreIDB.re
+++ b/src/web/StoreIDB.re
@@ -1,0 +1,80 @@
+/* IndexedDB-backed key-value store.
+   Async counterpart to Store.F (which uses localStorage).
+   Each store instance gets its own IndexedDB database to avoid
+   version-coordination issues between independent stores. */
+
+module F =
+       (
+         STORE_KIND: {
+           [@deriving (show({with_path: false}), sexp, yojson)]
+           type t;
+           let default: unit => t;
+           let db_name: string;
+         },
+       ) => {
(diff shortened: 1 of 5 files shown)


disconcision:
@7h3kk1d I've verified that basic agent stuff still works; could you give this a smell test?

disconcision:
@russell-rozenbaum: The coding-agent merge is blowing up localstorage;  even the empty state was taking up about half of the maximum size, which is only 10MB. This PR both reduces the storage footprint and moves it to indexedDB which has less limited storage. The agent still seems to work, but can you verify that it seems to retain data as you'd expect between reloads?

disconcision:
@7h3kk1d, @russell-rozenbaum nm, closed in favor of #2186, please take a look there instead.

PR #2189, opened 2026-03-23
Selection wrapping, quote unwrapping, and comment toggle

disconcision:
- **Selection wrapping**: Typing an opening delimiter (`(`, `[`, `{`, `"`, `` ` ``, `#`) with a selection wraps the selected text in the corresponding pair. Balanced delimiter wrappings can be done on any selection. Quote wrappings can be done on any selection that can be quoted by that quote delimiter; this basically means any selection that (a) does not contain any instance of the quote delimiter and (b) does not contain a linebreak.
- **Quote unwrapping**: Backspace/delete on a string, backtick label, or comment delimiter unwraps its content instead of deleting everything. The effect is the same as if pasting the formerly quoted contents as text.
- **Comment toggle** (Cmd+/ / Ctrl+/): Toggle line comments on single or multi-line selections, with correct sort remolding on uncomment. Keeps lines selected after multi-line comment toggle for easy repeat toggling


7h3kk1d:
Super nitpicky: since we do delimited comments can the behavior be to comment selection and only do line if there's no selection. edit: maybe since # does this it's fine.

Also I'm getting stack overflows from something when adding non-balanced parens.


https://github.com/user-attachments/assets/4761f5d1-66ee-4767-a7f0-404201f84179

7h3kk1d:
Also for certain cases " is adding quotes and others it's replacing with empty string. If we can't do this one in general I'd skip it.

https://github.com/user-attachments/assets/9030d232-7807-4169-ae34-dd9443cdd84d

disconcision:
@7h3kk1d 

re: commenting selection / delimited comments: the tradeoff here is the ability to quickly comment out a range of lines without needing to be super specific about your selection... that was the motivation for making it as is for me. I'm open to changing it but I wanted to be clear that's why the design is the way it is (ala vscode). i might feel different if our comments could contain linebreaks (this is still a bit of work unfortunately).

re: string quote sometimes resulting in empty string: the precise criteria for this is documented in the PR description... does knowing the logic change your opinion one way or the other? we can relax this in the future by allowing escaping / line breaks, but both are larger changes than I can do atm. I guess an alternative to defaulting to the insertion behavior for situations where we can't wrap would be to no-op? In any case, these considerations apply to all wrapping forms so backing off on this would be to remove wrapping and comment toggling entirely

stack overflow fixed

disconcision:
Merged via #2154

PR #2191, opened 2026-03-23
Variable highlighting cleanup

disconcision:
- Pure refactor. Avoids allocation.


src/language/statics/CoCtx.re
```diff
@@ -52,9 +52,21 @@ let mk = (ctx_before: Ctx.t, ctx_after, co_ctx: t): t => {
   );
 };
 
-/* Note: this currently shadows in the case of duplicates */
+/* Merge co-contexts, combining entry lists for the same variable name. */
 let union: list(t) => t =
-  List.fold_left((co_ctx1, co_ctx2) => co_ctx1 @ co_ctx2, []);
+  co_ctxs => {
+    let all = List.concat(co_ctxs);
```
7h3kk1d: Extremely nitpicky but this is where using streams would not allocate

src/web/app/editors/code/CodeWithStatics.re
```diff
@@ -101,7 +101,7 @@ module Update = {
         ~dynamics: Language.Dynamics.Map.t,
         ~is_dynamic_term,
         ~ana=?,
-        {editor, statics, context_menu, dynamics: _}: Model.t,
+        {editor, statics, context_menu, dynamics: _, _}: Model.t,
```
7h3kk1d: Should either make dynamics part of the _ or name the other one

src/language/statics/Ctx.re
```diff
@@ -159,23 +159,32 @@ let add_ctrs = (ctx: t, name: string, id: Id.t, ctrs: TermBase.Typ.sum_map): t =
...
-              },
-            custom_statics: None,
-          }),
-        )
+      | ConstructorMap.Variant(ctr, ann, typ) => {
+          let ctr_id =
+            switch (ann.ids) {
+            | [id, ..._] => id
+            | [] => id
+            };
```
7h3kk1d: I don't have a real understanding about what's going on here. Did we have the wrong id before and because we're using type-directed inference it should be the id of the analyzed type?

7h3kk1d (commented):
I finished a functional review and it seems great. Haven't finished looking at the code. But feel free to merge if you don't care about that.

src/language/statics/CoCtx.re
disconcision: are we using these anywhere atm?

7h3kk1d: Whoops I meant Seq and not Stream. There's a few calls to List.of_seq in ListUtil and StringUtil

cyrus- (approved):

PR #2192, opened 2026-03-23
Inward indication bias + Cmd+D for selecting enclosing term

disconcision:
- Change indication from left-biased to inward-biased for more intuitive cursor behavior.
- Exception: All leftwards-concave forms (including infix operators) have their left side as indicated position
- Convert Indicated.piece to record type for clearer field access
- Rename Indicated `piece'` and `piece''` member functions for clarity
- Add Cmd+D (select enclosing term): when a term is already selected, selects the next containing term up


7h3kk1d:
Undoubtedly do to type-directed autolabeling this isn't working for the table buffer.

https://github.com/user-attachments/assets/a665ac07-bca5-4e4c-9399-63e4cf81d2e1

I think I like the behavior overall though.

disconcision:
Merged via #2154

PR #2198, opened 2026-03-30
Keyboard architecture refactor: Key.handler, ContextualAction, focus decentralization

Negabinary:
## Summary

 1. Change the way keyboard handling works. (Previously the root element was always focussed and it passed key events down, now key events happen locally at the editor)
 2. Make NinjaKeys use contexual actions - so now the NinjaKeys actions can change based on what's selected.

### What changed

**Key.handler / Key.listener** — Reusable keyboard event wiring in Key.re. `handler` adds tabindex(0) for focusable components; `listener` is for containers that catch bubbled events (like #page).

**ContextualAction.t** — Unified type for actions that serve both keyboard shortcuts and command palette entries. Uses Effect.t(unit) so it's decoupled from Page.Update.t. Shortcut.re keeps its Page.Update.t type internally with a to_contextual_action converter. NinjaKeys.re consumes ContextualAction.t directly.

**Editors handle their own keyboard events** — Each CodeEditable editor now has its own Key.handler. Events are handled locally (context menu -> projector handoff -> Keyboard.handle_key_event) with Stop_propagation. Unhandled events bubble to Page.re for page-level shortcuts (undo/redo/F7/meta).

**Boundary escape** — Arrow keys at the true editor boundary (caret Outer, no neighbor, no ancestors) call `~escape(direction)` so parents can navigate between cells/projectors. Direction means "side to escape TO" not "key pressed".

**Projector escape focus** — TextAreaProj now focuses the parent .code-editor element after blur during escape, so the editor regains keyboard focus.

**cursor.contextual_actions** — The cursor type has a new contextual_actions field and with_actions helper, enabling future per-component dynamic actions.

**EditMode.re** — New type bundling inject, escape, take_focus, and focus for editor views.

### Non-obvious gotchas (from splices trial-and-error)

- **Arrow keys trigger scroll with tabindex(0)** — must Prevent_default on handled arrow events
- **Boundary escape must check ancestors == []** — without this, cursor gets stuck inside nested expressions (let/in, parens) because caret==Outer with no neighbor is also true at edges of nested groups
- **Projector shortcuts must be re-registered** — when moving keyboard handling into component-level handlers, projector shortcuts (Alt+F etc.) don't come along for free
- **Escape direction = side to escape TO** — not which key was pressed. Getting this backwards breaks projector navigation
- **Projector blur loses focus** — after TextAreaProj blurs itself, focus goes to body. Must explicitly focus parent .code-editor
- **First click needs tabindex always** — editor elements need tabindex(0) even when not selected, otherwise first click gives DOM focus to the element but Key.handler only appears after MakeActive re-render, requiring a second click


disconcision:
patchwork detection??? claude may be lost in the multiverse again

Negabinary:
> patchwork detection??? claude may be lost in the multiverse again

lol I was hoping you'd know what it meant by that

Negabinary:
## Known issues (in progress)

- **Keys don't work after first click, only after second** — First click dispatches MakeActive (re-renders with selected=true and Key.handler), but DOM focus isn't on the editor element until a second click
- **Selection highlighting shows black boxes** — tabindex(0) on the editor div is causing browser-native focus/selection outlines
- **Handoff out of projectors doesn't work** — Arrow key escape from projectors not wired up yet

Working on fixes.

Negabinary:
## Latest: Distribute contextual actions, wire EditMode, remove dead code

This commit cleans up the dead code from the initial refactor and moves the architecture closer to the `projector-unsegment` pattern.

### What changed

**Contextual actions distributed to their layers** (replaces `Shortcut.re`):
- **Page level** — undo/redo, benchmark, settings toggles, navigation, selection, projection, TyDi, reparse, introduce, export-for-init
- **ScratchMode** — export, encode, buffer management (add/rename/delete slide)
- **ExercisesMode** — export submission + instructor exports
- Each layer builds its own `ContextualAction.t` list using `ContextualAction.mk`, added to the cursor via `Cursor.with_actions`
- `NinjaKeys.initialize` now called from `Page.View.view` using `cursor.contextual_actions` (dynamic per-render, not static at startup)

**EditMode wired into CodeEditable**:
- `CodeEditable.View.view` takes `~edit_mode: EditMode.t(Update.t, unit)` instead of separate `~inject`, `~selected`, `~escape` params
- All 7 callers updated (CellEditor, InductionStep, InductionCase, MissingStep, CodeSelectable, StepperEditor, EvalResult)
- `EditMode.ReadOnly` used by CellEditor when locked

**Dead code removed** (net -11 lines):
- `EditorView.re` — unused scaffolding (56 lines, zero callers)
- `Shortcut.re` — fully replaced by per-layer actions (311 lines)
- `Selection` modules from `History.re` and `Logged.re` (no external callers)
- `Page.Selection.handle_key_event` (editors handle their own keys now)

### Architecture direction

This follows the `projector-unsegment` pattern where each layer owns its contextual actions. The `~inject` parameter is threaded through `get_cursor_info` at each level (Editors → ScratchMode/ExercisesMode/TutorialsMode) so each layer can build effects for its own actions. The cursor accumulates actions as it flows up through the layers.

Negabinary:
## Follow-up: Remove dead handle_key_event chain

Removed 305 lines of dead `handle_key_event` routing across 23 files. Every removed function was pure delegation (call next level's handler + wrap result in local update type) — no logic lost.

**Removed from:**
- **Module type signatures:** `StepInterface.STEP`, `StepInterface.STEPPER`
- **Stepper dispatch:** `StepperBase.StepKind`, `StepperBase.Stepper` (recursive dispatcher)
- **Stepper focus layer:** `StepperView.Focus`, `Theorems.Focus`
- **Step implementations:** SingleStep, InductionStep, InductionCase, MissingStep, ForallStep, AxiomStep, AlgebriteStep, AxiomsBox
- **Editor routing:** Editors, ScratchMode, TutorialsMode, TutorialMode, ExercisesMode, ExerciseMode, TheoremExerciseMode, CellEditor, EvalResult, CodeSelectable, StepperEditor

**What's still live:**
- `CodeEditable.Selection.handle_key_event` — called from `key_handler_attr` (the new per-editor handler)
- `Keyboard.handle_key_event` — maps `Key.t` → `Action.t`
- `Page.View.handle_key_event` — page-level keys (undo/redo/benchmark)

disconcision:
@Negabinary is this a pure refactor or does it add functionality? the boundary stuff sort of implies you can use arrows to move between cells now?

Negabinary:
@disconcision pure refactor, you can't use keyboard to go between cells - though I guess that would be easier to implement now

Negabinary:
TODO:
- [ ] Make it clear when focus is lost

cyrus-:
@Negabinary when I open up the command palette and do a Cmd+A to select all, it seems to do the select all in the last editor rather than in the command palette. this doesn't seem to happen for other text boxes, e.g. in the Hazel Assistant sidebar, but maybe points to something that isn't working right.

cyrus-:
Escape key not working + conflicts

PR #2199, opened 2026-03-30
Character-level selections

disconcision:
TODOs:

- [x] Modifier key not working to do token-based selection
- [x] Make default that char ranges only inside tokens and modifier is required 
- [x] Setting to invert that
- [x] crash on do_towards

---------------

- **Char-level selection precision**: Shift+Arrow now grows/shrinks selections one character at a time instead of one token at a time (Alt+Shift+Arrow for token-level). Adds anchor_caret field to Selection.t tracking the anchor end position within a boundary token.
- **Selection operations**: Destruct, insert, copy, cut, and paste all work correctly with partial-token selections. Highlight rendering clips to sub-token boundaries.
- **Cross-boundary paste fix**: When cutting a selection that spans a delimiter (e.g. = in let...=...in) and pasting back, the re-inserted delimiter now reconnects with its ancestor tile via rescan_parent_shards. This is a pre-existing bug fix also ported to deep-reassociate-clean and probes-III.

| File | Changes |
|------|---------|
| Selection.re | anchor_caret field, push/pop/map/toggle plumbing |
| Select.re | local_by_char grow/shrink, vertical/point-to-point char-level |
| Zipper.re | rescan_parent_shards, trim_selected_text, caret/anchor point fixes, no-progress guard |
| Insert.re | Inner-caret handling for insert-over-selection |
| Highlight.re | Sub-token clipping for selection highlight rendering |
| Keyboard.re | Shift+Arrow to ByChar, Alt+Shift+Arrow to ByToken |
| Test_Editing.re | ~700 lines of new tests

## Todo

- [x] Verify char-level selection with projectors at boundaries
- [ ] Test on PC keyboard layout (Ctrl+Shift for ByToken)


cyrus-:
cutting and pasting the following selection does not round trip 
image

cyrus-:
modifier key not working to do token-based selection

cyrus-:
- [x] if I start a selection inside a variable and go in one direction I can't cross back over to the other direction
- [x] I can't empty out a selection, buggy
- [x]  shift-select with mouse doesn't seem to be working yet
- [x]  perhaps relatedly, having shift down when I start a selection makes it so I can't select at all

(I can show you all of these in our meeting, just noting it down.)

cyrus- (changes requested):
see PR comment

disconcision:
@cyrus- All issues addressed, except for the partition_at round trip issue, which is an existing dev issue (it is similarly borked when the selection includes the whole `tl` token) #2308

cyrus- (changes requested):
- Can't hold shift-option and go back up once further down. This was the code I was using:
```
let fib(n : Int) = 
  if n < 2 then 1 else fib(n - 1) + fib(n - 2) in 
module M1 = {
  let fib = (fun (n : Int) ->
    if n < 2 then 1 else fib(n - 1) + fib(n - 2));
  let x = 1201; 
  let r = x + x + x + x
} in 
module M2 = {
  let r = M1.x + M1.x + M1.x + M1.x;
  let z = fib(21) 
} in M2
```
- also some issue with shift-option dragging on first character of token and going to beginning of line and back right

cyrus-:
- default for mouse should be chunky, option to be character-level (current behavior)
- default for keyboard should be character-level, option to be chunky
- setting changes default for mouse to character-level ("Character-level mouse")

cyrus- (changes requested):
see comment

cyrus- (approved):

PR #2206, opened 2026-04-03
Coding agent UI updates v2

russell-rozenbaum:
- System prompt and developer notes use full-screen-doc (left-aligned, readable)
- Hide edits summary when there are no edit tools in a turn
- Wire Agent Tools to tools-view-* CSS (flex row, hover, toggle states)

- Scroll chat to bottom on updates only when already pinned to bottom (still buggy)


russell-rozenbaum:
Closing in favor of #2210

PR #2209, opened 2026-04-07
Tydi tuple scaffolding

disconcision:
When you're either inside the argument parens in a function application, or a parens in a context which is expecting a tuple, this scaffolds the tuple by previewing appropriate commas/holes. Unlike previous TyDi completions, this scaffolding is reflected in statics, i.e. statics is re-run after scaffolding is inserted, so that you won't necessarily have static errors while entering a tuple as you would have had previously.


disconcision:
image (2)

also alo thing with two holes next to each other

disconcision:
Closed in favor of #2398

PR #2210, opened 2026-04-09
Coding Agent Expansion (Added Tools, New Slash Commands, Rewritten Prompts, Improved Chat UX, and more)

russell-rozenbaum:
## Summary

Major coding-agent expansion. Scope: agent surface (`src/web/view/agent{Core,View}/`, agent CSS/tests, `OpenRouter.re`, `HighLevelNodeMap.re`, prompt factory).

### What's new

- **Structure-editor tools** — path-addressed syntax-projector tools via `HighLevelNodeMap`, statics tools, generalized `insert_before`/`insert_after`, per-tool + per-category toggles, `ProjectorCatalog`.
- **Chat loop & control** — send queue, stop, multi-tool abort, auto + manual `/compact`, context meter, sticky scroll, centralized empty/API retry path.
- **Session modes & workbench** — Converse / Edit / Plan top-bar toggle; tasks + subtasks with ordering and status tracking.
- **Slash commands** — typed payloads + `SlashCommandOutput` view; `/help`, `/session-usage`, `/key`, `/key-usage`, `/account-usage`, `/show-thinking`, `/compact`.
- **Model & API** — thinking view, reasoning-effort dropup, model picker (FP Lab recs, fuzzy search, only-free), OpenRouter account/key endpoints, safer error paths.
- **Display** — plain-text tool-call rows + signifiers + jump-to-node, `ToolCallSummary`, agent-message Markdown via Omd → vdom, chat export/copy.
- **Prompts & docs** — structure-editor-first identity, composition/compaction overhaul, `^^ projector` syntax, `AGENTS.md` + handoff doc.
- **Refactors** — `AgentCore`/`AgentView` → `agentCore`/`agentView`; top-bar above textbox; dropped empty-reply workbench nudge (plan-mode loop fix).
- **Tests** — new `Test_AgentControlFlow`, `Test_AgentMultiTool`; expanded `Test_AgentUX`, `Test_AgentTools` (incl. `HighLevelNodeMap`).
- **Streaming** — tokens are now streamed in via OpenRouter
- **Prompt caching & cost** — a fixed *floor* breakpoint on the stable prefix (tools + system prompt + dev-notes) caches ~93% of a typical payload at ~0.1× input (Phase 1, verified live ~95% reads). Phase 2 adds an *advancing* breakpoint on the last history message to also cache the growing conversation, plus a per-conversation `session_id` for OpenRouter sticky routing; a diagnostic harness (`OpenRouter.CacheDiag`, off by default) proved the advancing marker is **dropped by OpenRouter's OpenAI-compat path to Anthropic** (it honors `cache_control` only on `system` content), so cumulative history caching is currently a no-op pending native-Anthropic routing — full writeup, evidence, and next steps in `agent-docs/prompt-caching-findings.md`. `cache_control` gated to explicit-breakpoint providers (Anthropic, Gemini, Qwen); auto-caching providers get the field stripped. Per-message token + cache-read/write metadata chip.
- **Chat UX** — collapsible top-bar toggle; lighter slash cards; tool name/def separator.

### Files outside agent scope

11 narrow integration touches (icons, Page route, sidebar hook, JsUtil helpers, prompt-string trim, projector/probe perform plumbing, test-runner reg, `run_tests`/`.gitignore`). Rest of `git diff origin/dev..branch` is artifact from dev moving 421 commits since divergence at `ab5ba95da25` — no branch-side deletions.

### Tour — new bottom-bar controls

- Toggle agent modes: *Edit* / *Plan* / *Converse*
- Select reasoning effort
- Change model inline (current model shown)

Screenshot 2026-04-18 at 3 02 18 AM

Screenshot 2026-04-18 at 3 20 06 AM

Screenshot 2026-04-18 at 3 20 35 AM

 ### New slash commands

Screenshot 2026-04-18 at 3 04 31 AM

### Agent has ability to place probes and understand their intermittent results

Along with this, it can also place projectors. 

Screenshot 2026-04-18 at 3 10 05 AM

### More informative collapsed tool descriptions

Also comes with the ability to Cmd/Ctrl+Click to jump to the definition of the respective definition that was edited

Screenshot 2026-04-18 at 3 06 20 AM

### Thinking blobs now displayed as their own UI chunks

Screenshot 2026-04-18 at 3 07 42 AM

### Refactored model selection UI

We can recommend some models (this requires maintaining code though :/ might need to remove).

Also added the ability to grep for models in the master view.

Stopped displaying free models first, instead added a filter toggle. 

Screenshot 2026-04-18 at 3 11 42 AM

### User can stop chats early (and also queue up messages)

Screenshot 2026-04-18 at 3 14 50 AM

### Batched Tool Calls

Along with error handling—if tool call *n* fails, all tool calls scheduled after  *n* are not executed to avoid cascading failures, rogue bugs, and unintentional logic flow.

Screenshot 2026-04-18 at 3 15 19 AM


disconcision:
The system prompt refers to the old `forall` keyword; this should be replaced by `poly`.

russell-rozenbaum:
Okay, verified that main code this hits external to agent infrastructure are `*Perform.re` files, where it mainly adds to them. 
This was for the projector and probe placement tools we gave the agent.

russell-rozenbaum:
> The system prompt refers to the old `forall` keyword; this should be replaced by `poly`.

resolved

src/haz3lcore/projectors/ProjectorPerform.re
```diff
@@ -283,3 +283,190 @@ let go =
...
+    | _ => None
+    }
+  );
+};
+
+/** Re-validate projectors under a segment after the underlying syntax may have
+    changed. Strips a projector (exposing underlying syntax) when
+    [[MakeTerm.for_projection]] or [[ProjectorInit.init]] fails, migrating
+    refractors to the underlying term id when possible. */
+let sanitize_projectors_in_segment =
```
disconcision, line 473: what does sanitize mean? re-validate? what does it mean for maketerm/init to fail? what does this have to do with refractors, which aren't in syntax the same way projectors are?

I don't really get what this is doing and the call site doesn't clear things up for me either

AGENTS.md
```diff
@@ -0,0 +1,170 @@
+# AGENTS.md - Hazel Development Guide
```
disconcision, line 1: not sure if we should commit this or not; seems fine; discuss with @cyrus-

src/web/app/Page.re
```diff
@@ -662,7 +662,20 @@ module View = {
         switch (target) {
         | Some(el) =>
           let elId = Js.Opt.to_option(Js.Unsafe.coerce(el)##.id);
-          if (is_input_field(elId)) {
```
disconcision: This hopefully won't be an issue when you're up to date with dev; @Negabinary has some keyboard handling changes that should obviate the need for these workarounds. lmk if that's not the case and we'll find a better way of doing this.

src/util/StringUtil.re
```diff
@@ -105,17 +105,18 @@ let sanitize_filename = (s: string): string => {
 
 let trim_trailing_whitespace = (str: string): string => {
   let lines = String.split_on_char('\n', str);
+  let is_trailing_ws = (c: char): bool => c == ' ' || c == '\t' || c == '\r';
   let trim_line = (line: string): string => {
     let chars = String.to_seq(line) |> List.of_seq;
-    let rec drop_leading_spaces = (chars: list(char)): list(char) =>
```
disconcision, line 110: was this just mis-named before?

src/util/JsUtil.re
```diff
@@ -595,3 +632,71 @@ let has_ancestor_class =
...
+  };
+
+/** If the current window selection is non-collapsed and its anchor sits
+    (within [max_depth]) under an element with any of [classes], write the
+    selected text into [evt.clipboardData] as text/plain, preventDefault the
+    event, and return [true]. Otherwise return [false] and touch nothing.
+
+    Used to rescue native copy when a focused hidden element (e.g. the editor's
+    clipboard shim) would otherwise intercept Cmd/Ctrl+C. */
+let try_copy_window_selection_in_classes =
```
disconcision, line 718: hopefully won't be necessary when current with dev... see comment below

src/web/view/agentCore/AgentGlobals.re
```diff
@@ -6,20 +6,56 @@ module Model = {
...
+    | Edit
+    | Plan;
+
   [@deriving (show({with_path: false}), sexp, yojson)]
   type t = {
     active_screen: screen,
     api_key: option(string),
     active_llm: option(OpenRouter.AvailableLLMs.Model.llm_info),
     available_llms: OpenRouter.AvailableLLMs.Model.t,
+    [@yojson.default ""] [@sexp.default ""]
```
disconcision, line 22: why are these fields but not others defaulted? trying to make sure this isn't just papering over something

src/haz3lcore/ProbePerform.re
```diff
@@ -854,6 +854,48 @@ let toggle_statics =
     };
   };
 
+/** Like [[toggle_statics]], but for path-resolved ids (agent tools). */
+let toggle_statics_at = toggle_statics;
```
disconcision, line 869: not sure this alias is necessary?

src/haz3lcore/projectors/ProjectorPerform.re
```diff
@@ -283,3 +283,190 @@ let go =
   | SampleFocus(a) => Ok(SampleFocusPerform.go(z, a))
   };
 };
+
```
disconcision, line 352: I don't really get what this set of functions is doing. they seem to be for generic projector placement, but internally they call migrate_refractor, which only applies to probes/statics, which unlike projectors are not written into the syntax tree.

disconcision (changes requested):
Looks solid in general. I skimmed most of the agent-specific code, focussing on the interface with the rest of hazel; see code comments for detail.

Functional issues:
1. New code insertion still seems to have leading spaces. Probably worth trying to make tests around this as it seems to be a recurring issue.
2. Pressing enter to send a message in a long-ish conversation often lags... probably worth identifying why, but regardless, it would be nice to get some immediate feedback, just to make it obvious something is happening (eg message gets added to log and disappears from entry box), even if actual processing is subject to the lag. otherwise the user is left doubting what's going on.
3. When inserting new let bindings into the body below the last let binding, it doesn't seem to always insert a linebreak or space. so you sometimes get stuff like `inlet empty_playlist : PlayList = ([], NoSongSelected) in`, which is problematic as if it's copy-pasted it'll break.

Aesthetic:
1. Slash cards (.slash-card) are a bit dark... lets use `background: var(--T2)` instead.
2. Tool calls are currently displayed as `tool_name def_name`. despite using different colors these kind of run into each other, let's stick some kind of separator between the two words. Also, it seems like def_name is slightly vertically elevated over tool_name; check the css.

Broader issues:
- Not specific to this PR, but I think I'm running into issues based on how we're only keeping the current code view/map. I had the agent do one of the test tasks; he also wrote some tests. I then pasted in our test suite over the agent's, and told him that I did so and that a test was failing. The agent got confused though as all he saw were 'the tests he added'. I'm not actually totally clear on what was going on here, but I think it was that the agent is only ever seeing the current version, plus agent edits, so the agent gets confused when referring to 'updated'/'new' versions resulting from user edits, as those don't show up in the log. something like that. I copy/pasted the history here (https://gist.github.com/disconcision/cdf2950ee8b49820ce36f8540a169669) but it's not too readable. We should discuss this. I think this showcases the importance of a way to see what the model saw at each stage of a job.

disconcision (approved):
lgtm pending patch PR

PR #2212, opened 2026-04-10
(Just a) Stalaborator

Negabinary:
Zips up statics and elaborator into one pass. Makes our statics file look a lot more like the rules.


cyrus-:
closing in favor of #2213

PR #2213, opened 2026-04-10
(All-out) Elastatics

Negabinary:
Bigger more ambitious version of #2212

- combines statics & elaborator
- replaces self with separated Mark.t, Warning.t, and Message.t
- applies heavy cleaning to all these files

Statics rules will look much more like rules we'd write in a paper, putting types, marks, and the like all in one place.

I encourage you to read Statics.re for yourself and drink it in.

todos
 - [x] tidy up the pattern cases of statics
 - [x] do a pass over custom statics

Left for a future pr
 - There's more work that needs to be done around Message.re: at the moment it's not very well defined.


src/language/statics/Statics.re
```diff
@@ -11,7 +11,7 @@
    (ana:Typ.t): Defines the Mode.t type which is used to represent the
    typing expectations imposed by a term's ancestors.
 
-   SELF.re: Define the Self.t type which is used to represent the
+   SELF.re: Define the Mark.t type which is used to represent the
```
7h3kk1d: Remove reference to Self.re

src/language/statics/Info.re
```diff
@@ -265,15 +31,17 @@ type label_inference('a) =
 
 [@deriving (show({with_path: false}), sexp, yojson)]
 type exp = {
-  term: Exp.t, /* The term under consideration */
+  user_term: Exp.t, /* The term under consideration */
+  elab_term: Exp.t,
```
7h3kk1d, line 37: Is there a reason we have elab_term for exp but not for the others?

src/web/app/sidebar/ProblemSidebar.re
7h3kk1d, line 1: This PR might not be the right place to handle it but introducing the mark abstraction underneath the weak `problem` abstraction might require some more thought. The live typing pr https://github.com/hazelgrove/hazel/pull/1988 is also going to add the _live typing errors_ to the problem sidebar.

It seems reasonable that there are multiple marks for a given form and maybe each of these marks should have a direct correspondent in the problem list. Maybe the cursor inspector itself shows the highest priority problem in the case of multiple.

Alternatively if we want to remove the problem notion entirely maybe everything could live on marks, but then we need some way of re-incorporating the live typing marks post-dynamics.

src/language/statics/Mark.re
```diff
@@ -0,0 +1,102 @@
...
+
+[@deriving (show({with_path: false}), sexp, yojson)]
+type t =
+  /* “Core” marks (meet / label / hole shape); shared by exp and pat statics */
+  | NoMeet(meet_type, list(Typ.source))
+  | DuplicateLabel(LabeledTuple.label, Typ.t)
+  | CompareFun(Typ.t)
+  | DuplicateVar(string, Typ.t)
+  | BadToken(string)
+  | BadLabel(Any.t)
```
7h3kk1d: Is it worth pulling out the marks that are only available on certain forms to separate types just for organization reasons?

src/language/statics/MarkSelection.re
```diff
@@ -0,0 +1,56 @@
+let rank = (m: Mark.t): int =>
```
7h3kk1d: My personal preference would be to use `ppx_deriving.ord` or `ppx_variants_conv` and leave a comment saying that declaration order is load-bearing. This also works but it's been annoying juggling the numbers in Precedence. I also don't know if there's any reason it shouldn't be a total order

src/web/app/sidebar/ProblemSidebar.re
7h3kk1d: In the same sense I think the warnings being conjoined with marks also makes sense.

src/CLI/Run.re
```diff
@@ -1,13 +1,11 @@
...
-let elaborate = (exp: Exp.t): Exp.t =>
-  fst(
-    Elaborator.elaborate(
-      Statics.mk(CoreSettings.on, Builtins.ctx_init(Some(Int)), exp),
-      exp,
-    ),
-  );
+let elaborate = (exp: Exp.t): Exp.t => {
+  let (_, elab) =
+    Statics.mk(CoreSettings.on, Builtins.ctx_init(Some(Int)), exp);
```
7h3kk1d: unrelated to PR but I see below we have Operators.default_mode

src/language/statics/Message.re
7h3kk1d, line 1: I feel somewhat unsure of the Message separation for being exclusively well formed payloads. In some sense these feel more like informational non-error marks as they're not the message as shown to a user but stuff that would be useful in constructing the happy path message (only in the context where there are no marks).

src/language/statics/Info.re
```diff
@@ -265,15 +31,17 @@ type label_inference('a) =
...
   ancestors, /* Ascending list of containing term ids */
   ctx: Ctx.t, /* Typing context for the term */
   ana: Typ.t, /* Parental type expectations  */
-  self: Self.exp, /* Expectation-independent type info */
+  syn_ty: Typ.t, /* Principal synthetic type before hole fixing (statics) */
```
7h3kk1d: I don't know if we want to call this self or syn_ty but. Ithink we should clarify that this is not purely synthetic.

It's computed after analysis-directed rewrites like literal replacement (Int(1) → Nat(1) via ana/use_mode) and automatic label introduction for labeled tuples — so it's not context-free nor is it pure syn.

src/language/statics/Info.re
```diff
@@ -498,638 +270,69 @@ let pat_ty: pat => Typ.t = ({ty, _}) => ty;
...
-      if (Typ.is_syn_plus(ana)) {
-        Unknown(Internal) |> Typ.temp;
-      } else {
-        ana;
+  | InfoExp({marks, _}) =>
+    marks != []
+    && (
+      switch (MarkSelection.highest_ranked_mark(marks)) {
+      | Some(Mark.BadToken(_) | Mark.IsMulti) => true
+      | _ => false
```
7h3kk1d: Long term this might live better with Mark code

test/Test_Info.re
```diff
@@ -1,29 +1,68 @@
...
-let status_typ_tests = (
-  "Info.status_typ",
+let typ_status_testable =
+  testable(
+    Fmt.using(Statics.show_typ_status, Fmt.string),
+    Statics.equal_typ_status,
+  );
+
+let derive_typ_tests = (
+  "Statics.derive_typ_status",
```
7h3kk1d: Can we put this in a different file if it's not testing the info?

src/language/statics/StaticsBase.re
```diff
@@ -5,12 +5,36 @@ module Map = {
...
+  let equal_errors: (errors, errors) => bool =
+    Id.Map.equal((a: list(Mark.t), b: list(Mark.t)) => a == b);
+
+  let show_errors = (m: errors): string =>
+    Id.Map.bindings(m)
+    |> List.sort((a, b) => Id.compare(fst(a), fst(b)))
+    |> List.map(((id, marks)) =>
+         Id.show(id) ++ " => " ++ [%derive.show: list(Mark.t)](marks)
+       )
+    |> String.concat("\n");
```
7h3kk1d: This seems to be unused
```suggestion
```

src/language/statics/StaticsBase.re
```diff
@@ -5,12 +5,36 @@ module Map = {
...
 
   let empty = Id.Map.empty;
   let lookup = Id.Map.find_opt;
   let filter = Id.Map.filter;
+  let add_info = (ids: list(Id.t), info: Info.t, m: t): t =>
+    ids |> List.fold_left((m, id) => Id.Map.add(id, info, m), m);
```
7h3kk1d, line 12: This is the most nitpicky but I don't like the pipelining when we're already passing multiple args.

7h3kk1d: Feel free to ignore me though

src/language/statics/StaticsBase.re
```diff
@@ -33,18 +57,24 @@ module Map = {
...
   let collect_errors = (map: t): errors =>
     Id.Map.filter_map(
-      (_: Uuidm.t, info: Info.t) => {Info.error_of(info)},
+      (_: Uuidm.t, info: Info.t) =>
+        switch (Info.marks_of(info)) {
+        | [] => None
+        | ms => Some(ms)
+        },
       map,
     );
```
7h3kk1d: These are exclusively used in tests now so we should consider moving them or removing them.

src/language/statics/StaticsBase.re
```diff
@@ -60,10 +90,56 @@ module Map = {
...
+      Id.Map.fold(
+        (id, info, acc) =>
+          switch (Info.get_binding_site(info)) {
+          | Some(id') when id' == binding_id => [id, ...acc]
+          | _ => acc
+          },
+        m,
+        [binding_id],
+      )
+      |> List.sort_uniq(Id.compare)
```
7h3kk1d: We should just accumulate to a set rather than allocate a list then sort_uniq. This code is also directly duplicated below.

src/language/statics/StaticsBase.re
```diff
@@ -99,218 +175,319 @@ module Map = {
...
-    switch (lookup(binding_id, _m)) {
-    | Some(InfoPat({term: {term: Var(name), _}, co_ctx, _})) =>
-      switch (Util.VarMap.lookup(co_ctx, name)) {
-      | Some(entries) => List.map((e: CoCtx.entry) => e.id, entries)
-      | None => []
-      }
-    | _ => []
-    };
-  };
+  let let_definition_path = (~statics: t, ~id: Id.t): list(Pat.t) => {
```
7h3kk1d: Unrelated to pr

If you add the Id.Set to Id.re
```
module Set = Set.Make(Uuidm);
```
you can use a set for the accumulator and we should just use a fold rather than full recursion
```reason
let let_definition_path = (~statics: t, ~id: Id.t): list(Pat.t) => {
    let rec contains_id = (target: Id.t, ids: list(Id.t)): bool =>
      switch (ids) {
      | [] => false
      | [head, ...tail] =>
        Id.equal(head, target) || contains_id(target, tail)
      };

    let rec gather =
            (remaining: list(Id.t), seen: list(Id.t), acc: list(Pat.t))
            : list(Pat.t) =>
      switch (remaining) {
      | [] => acc
      | [current_id, ...rest] =>
        let acc' =
          switch (lookup(current_id, statics)) {
          | Some(InfoExp({user_term: {term: Let(pat, def, _), _}, _})) =>
            contains_id(IdTagged.rep_id(def), seen) ? [pat, ...acc] : acc
          | _ => acc
          };
        gather(rest, [current_id, ...seen], acc');
      };

    switch (lookup(id, statics)) {
| Some(info) =>
      let ancestors: list(Id.t) = Info.ancestors_of(info);
      let collected: list(Pat.t) = gather(ancestors, [id], []);
      List.rev(collected);
| _ => []
};
};
```

src/language/statics/StaticsBase.re
```diff
@@ -99,218 +175,319 @@ module Map = {
...
 let map_m = (f, xs, m: Map.t) =>
   List.fold_left(
     ((xs, m), x) => f(x, m) |> (((x, m)) => (xs @ [x], m)),
     ([], m),
     xs,
   );
 
-let add_info = (ids: list(Id.t), info: Info.t, m: Map.t): Map.t =>
-  ids |> List.fold_left((m, id) => Id.Map.add(id, info, m), m);
+let map_m2 = (f, xs, ys, m: Map.t) =>
```
7h3kk1d, line 271: Might be worth moving these to a map helper eventually. I also think leaving a comment on map_m saying it's a traverse over a state monad may make it easier for people to understand what's going on. Or alternatively add the types.

src/language/statics/StaticsBase.re
```diff
@@ -99,218 +175,319 @@ module Map = {
...
+/* Strip TupLabel(ExplicitNonlabel, _) wrappers on expectation. */
+let rec ana_skip_explicit_nonlabel = (ty_ana: Typ.t): Typ.t =>
+  switch (ty_ana.term) {
+  | TupLabel({term: ExplicitNonlabel, _}, ana_inner) =>
+    ana_skip_explicit_nonlabel(ana_inner)
+  | _ => ty_ana
+  };
+
+let should_emit_nomeet_mark = (ctx: Ctx.t, ana: Typ.t, syn_ty: Typ.t): bool =>
+  switch (Typ.meet(ctx, ana_skip_explicit_nonlabel(ana), syn_ty)) {
```
7h3kk1d: I don't remember how this used to work but it's worrisome to remember to call ana_skip_explicit_nonlabel everywhere

src/haz3lcore/derived/CachedStatics.re
```diff
@@ -92,19 +92,15 @@ let init_from_term =
       ~default=Builtins.ctx_init(is_dynamic_term ? None : Some(Int)),
       ctx,
     );
-  let info_map = Statics.mk(~ana?, settings, ctx_init, term);
+  let (info_map, stalaborated) = Statics.mk(~ana?, settings, ctx_init, term);
```
7h3kk1d: I vote we can continue to call the term elaborated

src/language/statics/CustomStatics.re
7h3kk1d, line 1: I can't tell where it's happening yet but the label argument like:
```
omit_labels((a=1,b=2), `a`)
```

if you have the cursor on the `a` it's showing a synthesized type of hole in the cursor inspector. This is a regression.

7h3kk1d (commented):
🎉‼️😍
The PR is pretty huge so I haven't had the chance to review it in entirety, but I wanted to leave intermediate comments (about 1/4 done).

Overall architecturally this seems like a massive improvement. The removal of the intermediate layers of Self/Info make this much easier to make changes and to read the current code. In addition the interleaving of elaboration/statics makes it much easier to keep the two in sync and should drastically improve type/static-directed elaboration without having to sidechannel the information.

I have a few comments below that range from small code-style stuff to some architectural questions especially centered around error/problem/message types.

src/web/app/sidebar/ProblemSidebar.re
Negabinary: Hm yeah I don't think I have enough background on the concept of a "problem" to make a call here - I'm tempted to just leave marks as they are and leave it to future work to define this abstraction.

src/language/statics/Message.re
Negabinary: I agree; Message is a little ill-defined in this PR. I believe we didn't used to have information displayed to the user outside of errors. I think that's a pretty good next step but I'd like to leave it beyond the scope of this PR I think.

src/language/statics/Mark.re
Negabinary: I don't think we can do this and do the priority ordering at the same time. I think I like priority ordering more.

src/language/statics/StaticsBase.re
Negabinary: hm yeah I think I don't mind it because it's a fold, and I feel like the list is the obvious thing to put into a pipeling

src/language/statics/StaticsBase.re
Negabinary: This function already existed but I deliberately renamed it to something more worrisome so it's clear how worrisome the contents are. It appears to be having the desired effect, I think I want to leave it as a wontfix though.

src/haz3lcore/derived/CachedStatics.re
Negabinary: Agreed; good catch - I didn't notice this codexism

src/language/statics/Message.re
7h3kk1d: >  I believe we didn't used to have information displayed to the user outside of errors.

I don't understand what you mean by this since we have the cursor inspector outside of errors. Unless you mean for the agent.

src/language/statics/Statics.re
Negabinary: Updated comment.

src/language/statics/Info.re
Negabinary: only exp and pats have elaborations; just added it to pat.

src/language/statics/Info.re
Negabinary: we discussed this; calling it elab_syn_ty

src/language/statics/Info.re
Negabinary: Moved.

src/language/statics/StaticsBase.re
Negabinary: Moved.

src/language/statics/CustomStatics.re
Negabinary: hm looks like I forgot patch_elab_syn_ty_exp

src/language/statics/Info.re
7h3kk1d: @disconcision you may want to be looped in here. Matt and I talked and it seems like the self type could be the elaborated syn type if we make sure during elaboration we add all the implicit type-directed stuff to the term: empty list's type, variant constructors type, etc.

src/haz3lcore/projectors/implementations/TypeProj.re
```diff
@@ -13,8 +13,8 @@ let expected_ty = (info: option(Info.t)): option(Typ.t) =>
 
 let self_ty = (info: option(Info.t)): option(Typ.t) =>
```
7h3kk1d, line 14: We should probably rename this and change the hover text if there is any.

PR #2222, opened 2026-04-22
Incremental evaluation

Negabinary:
Working on a formalism at https://typst.app/project/r1zZKHltWXXBRyJcVkTc5k.
Everything up to "The Tuple Bottleneck" is proven and implemented in Hazel. 

image

During evaluation, this pr "freezes" any parts of the previous evaluation that haven't changed, and re-uses the values.

You can turn on some visual "ice" to show which sub-expressions are frozen by toggling "Incremental Reuse" in the nut menu.

The evaluator is passed a map of the previous evaluation at each top-level id. It is able to re-use an expression iff:

1. **The call stack is empty** (This is our notion of top-level)
2. **The elaboration has not changed** (Uses deep equality; every id in the program is checked at most once, so this check is quadratic in the length of the program)
3. **The co-context does not contain any dirty variables** (Variables are dirtied if the thing they are bound to has changed)
4. **The ids associated with the co-context has not changed since last time**
5. **The probe targets in the sub-expression have not changed**


cyrus-:
This one isn't freezing cross-module: 
```let fib = fun n -> 
  if n < 2 then 1 else fib(n - 1) + fib(n - 2) in 
  let a = {
    let a = 13;
    let xy = fib(a);
    let yy = fib(a);
    let zy = xy + yy
  } in
  let c = {
    let b = 12;
    let x = fib(b);
    let y = fib(b);
    let z = x + y
  } in a, c```

also doesn't seem to be working correctly for desugared functions:

```
let fib(n) = ...
```

**Comment by Negabinary** [comment]
Note to self: things left to double-check

1. This relies heavily on the relationship between pre-and-post-elaboration ids, we should make sure that all pre-elaboration ids survive into the elaboration (we could probably make a pbt for this - I think other things rely on it); this is probably what's causing the module issues.

2. I'm not convinced by the targets_stable logic. I think I'd rather generate a list of targets in the statics map or something.

3. Check ice logic

File: src/language/dynamics/IncrEval.re [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+   * None when the run was made with `probe_all` on: every InfoExp is then
+   * a probe, so any change to the probe set coincides with a change to
+   * the elaboration tree — `Exp.fast_equal(prev_elab, info.elab_term)`
+   * already catches it, and a separate witness would be redundant. */
+  prev_probe_targets: option(SubexpProbeTargets.t),
+  value: DHExp.t,
+  state: StateSlice.t,
+};
+
+module EvalInfoMap = {
```
cyrus-: move into its own file

cyrus- (changes requested):
- bug(s) related to shadowing:
```
let fib(n : Int) = 
  if n < 2 then 1 else fib(n - 1) + fib(n - 2) in 
let x = fib(21) in 
#let x = 5 in#
x + x + x + x + x```
- also happening for modules with members that share a global name:
```
```
let fib = (fun (n : Int) ->
  if n < 2 then 1 else fib(n - 1) + fib(n - 2)) in 
module M1 = {
  let fib = (fun (n : Int) ->
    if n < 2 then 1 else fib(n - 1) + fib(n - 2));
  let x = 11111; 
  let r = x + x + x + x
} in 
module M2 = {
  let r = M1.x + M1.x + M1.x + M1.x;
  let z = fib(21) 
} in M2
```

cyrus- (approved):

cyrus- (approved):

MarisaKirisame:
Hi, is there any benchmark numbers of this vs hazel (without this)?

cyrus-:
@MarisaKirisame no, we haven't run systematic evaluations here, though that would be useful.

Negabinary:
@MarisaKirisame yeah we should run the numbers to check that I haven't made it too much, but I'm hoping it should just be:

- no improvement if it's changed
- subexpressions that haven't changed at all should be instant
- a small overhead, O(size of program * avg number of free variables), scanning all the top-level stuff to see what needs to be re-calculated.
- a possibly annoying overhead in saving intermediate results, especially if they're big, because they have to be part-serialized to cross the web worker boundary

I suspect in order to see an improvement in a realistic scenario we'd need a reasonably big program with lots of parts that don't interact with each other

PR #2225, opened 2026-04-23
Refactor Problems sidebar; fix jumps in Exercise mode and folds

### LINKED ISSUE #2223: problem sidebar does not allow jumping to problems in other editors
in the following you can only click to jump to problems in the same buffer you are currently in

Image

7h3kk1d:
Issue
-----
In ExerciseMode the Problems sidebar was broken in several ways:
  1. Clicking a problem row didn't jump to its tile.
  2. Clicking a hole never jumped, even after (1) was addressed.
  3. After a successful jump the target cell's caret was invisible.
  4. Problems from non-`user_impl` cells (notably Test Validation / your tests) never appeared in the sidebar at all.
  5. Read-only cells (e.g. Prelude in student mode) could have problems listed that were non-jumpable.
  6. Errors inside a fold projector had no jump target — the underlying ids aren't in `measured` because the projector replaces them with the folded view.
  7. None of the above worked for derivation / theorem slides (Exercise, Documentation, or Scratch). Tree judgement errors weren't surfaced and `Drv`-sort error messages rendered as empty rows.

Fixes
-----
1. Sidebar jump now uses the mode-aware dispatcher

   `ProblemSidebar.jump_to` used `ActiveEditor(Move(Goal(TileId id)))`, which only targets whichever cell currently has the cursor. In ExerciseMode the problem's tile almost always lives in a different cell than the focused one, so the action dropped silently.

   Switched to `Globals.Action.JumpToTile(id)`, which Page routes through `Editors.Selection.jump_to_tile` -> each mode's `Selection.jump_to_tile`. CodeExerciseMode's implementation scans `CodeExercise.positioned_editors` and emits both the correct editor update and the correct selection, so jumps land in the right cell. Scratch/Documentation/Tutorial modes already had their own `Selection.jump_to_tile`, so the change is a no-op there.

2. Editor-membership predicate now matches holes as well as tiles

   The `jump_to_tile` predicates in `CodeExerciseMode`, `CodeEditable`, `TutorialMode`, `DerivationExerciseMode`, and `TheoremExerciseMode` used `TermData.root_tile` to decide whether an id lived in a given editor. `root_tile` returns `None` for grouts, so hole ids (the `Hole`/`Syntax` categories in the sidebar) never matched any editor and the jump failed.

   Added `TermData.root_piece: Id.t -> t -> option(Piece.t)` returning the root piece for any variant (tiles, grouts, secondaries, projectors), and switched the predicates to use it. `root_tile` is refactored to delegate, so existing callers are unaffected.

3. Sidebar jump transfers DOM focus to the target cell

   ~~Caret visibility driven by model selection, not DOM focus. The caret CSS rule gated visibility on `:focus`, but DOM focus stays on the sidebar row that was clicked, so after a jump the caret stayed hidden. Switched the rule to `.selected` (model selection); DOM focus is not forcibly transferred.~~

   Reverted: gating the caret on `.selected` left it blinking on a cell that had lost DOM focus (e.g. after clicking elsewhere in the sidebar) — it looked active but ignored keystrokes. The caret rule is back to `:focus`. Instead, a jump now moves DOM focus to the target cell: the active `.code-editor` carries a stable id (`JsUtil.active_cell_id`) and the `JumpToTile` handler schedules `ProbePerform.FocusEffect.schedule_cell()`, executed in `after_display` (focus with `preventScroll` so it doesn't fight scroll-into-view). The editor now receives keystrokes after a jump and the caret correctly tracks focus.

4. Problems sidebar aggregates across all jumpable cells in every mode

   Sidebar.view built one `problem_context` from a single editor (`user_impl` in Exercise mode via `Update.get_editor`), so problems in `your_tests`, `prelude`, etc. never showed up. Drv slides surfaced only `setup`, missing `prelude` and every tree judgement.

   Replaced with a per-editor data structure owned by `haz3lcore` (see "Architecture" below). `Page.Update.get_problem_editors` now returns labeled editors. A cell's problems are listed iff the cell is *rendered* (regardless of read-only), so the lists can't drift from what's on screen. Coverage:
   - **Exercise / Code**: every cell that is rendered for the current user — `CodeExercise.shown_in(~instructor_mode)` — which now includes the read-only Prelude. ~~`CodeExercise.visible_in(~instructor_mode)`~~ (`visible_in` gated *editability*, not rendering, so it dropped the shown-read-only Prelude; it's been renamed `is_editable` and the sidebar/jump now use the new `shown_in`).
   - **Exercise / Derivation**, **Documentation / Drv**, **Scratch / Drv**: `Prelude` (exercise mode only — it isn't rendered on scratch/documentation slides), `Setup`, and every rendered tree judgement editor — enumerated by `DerivationExerciseMode.Model.get_problem_editors`.
   - **Exercise / Theorem**: `Prelude`, `Lemmas`, `Theorem` (all rendered, all jumpable).

   Read-only cells are still navigable (they accept `Move` via `CodeSelectable`), so listing and jumping to them works. The sidebar renders one labeled, collapsible section per editor when there are multiple groups.

5. Fold projectors now have a jump target

   Errors located inside a folded expression had no jump target before — the inner ids aren't present in `measured`, and the predicate (`root_tile` / `Measured.find_by_id`) returned `None`, so the jump silently dropped.

   `problem_group` now exposes a `pos` function that walks ancestors to the nearest measured id when the original id has no measured entry. Since a projector itself *is* in `measured` (occupying the folded line), the walk lands on the projector — meaning jumps inside a fold now scroll/select the folded expression rather than failing. Sidebar `L#` labels and sort order use the same resolution, so problems inside a fold cluster on the projector's line instead of falling back to `(0, 0)`.

6. Drv error rows render their messages

   `ProblemSidebar.problem_status_view` previously rendered `InfoDrv` (and `InfoMod` / `InfoSig` / `InfoMPat`) as an empty `<div>`, so Drv-sort errors like "Unexpected term for sort DrvProp" surfaced as blank rows. Wired `InfoDrv` to `DrvCursorInspector.drv_view` (the same renderer the cursor inspector uses) and added a class-name fallback for the others.

7. Derivation tree judgements collapse into one labeled section

   Every tree judgement editor is enumerated under the shared label `"Derivation"`. The sidebar clusters consecutive `problem_group`s with the same label into a single section: shared header, summed count, per-category subsections aggregated across all constituent groups in Grouped mode. Per-row line numbers are suppressed in merged sections (L# would otherwise refer to a different editor's geometry). Tree walk is postorder so within-tree order matches the visual top-to-bottom layout.

8. Switching/toggling the sidebar no longer steals editor focus

   Clicking a sidebar tab (switch section) or the collapse/expand toggle moved DOM focus off the editor, hiding the caret (gated on `:focus`). `switch_to` and `switch_assistant` now add `Effect.Prevent_default` to the mousedown so focus stays on the editor. Selecting text or clicking inputs in a panel still works and still moves focus (the panel's `tabindex` is retained).

Architecture
------------
`ProblemCollection` (in `haz3lcore`) now owns:

- `editor_input` — labeled `CachedStatics + CachedSyntax`.
- `problem_group` — problems attributed to one editor, pre-sorted per category, with counts and the `measured`/`row_to_line` needed to render `L#` labels and resolve jump positions.
- `problem_collection` — list of groups plus aggregated counts.
- `ProblemCollection.make` — builds the collection, de-duplicating shared `(id, category)` pairs in caller-provided order (first-wins).

`Sidebar.re` is now thin plumbing: build `editor_input`s and call `make`. `ProblemSidebar.view` renders zero/one/many groups appropriately, merging consecutive same-label groups into one section.

This replaces the old hand-rolled dedup (excluding `test_validation`) with a principled first-wins ordering: `test_validation` is now listed *before* `user_tests` so shared hole/syntax ids land in the "Test Validation" group. To keep jump-to-tile consistent, `CodeExercise.editor_positions` was reordered so `YourTestsValidation` precedes `YourTestsTesting`. `idx_of_pos` / `pos_of_idx` are unchanged, so persistence is unaffected.

Static-error scoping: `make_problem_context` now filters `error_ids`/`warning_ids` so each editor surfaces only problems for ids in its own segment. Previously `user_tests`'s static errors leaked into "Your Tests" for ids whose tokens lived in `your_impl`.

Per-editor collapse state: collapse keys are now `(editor_label, category)` for category-collapse and `editor_label` for editor-group collapse. Existing persisted `collapsed` settings will fall back to default (everything expanded).

Visibility is a single source of truth: `CodeExercise.shown_in(pos, ~instructor_mode)` decides which code-exercise cells are rendered, and the view, `get_problem_editors`, and `jump_to_tile` all consult it. The view's `vis_marked` (`Always`/`InstructorOnly`) markers — which duplicated this fact — were removed; each editor cell is gated directly on `shown_in` (a thunk keeps instructor-only cells from being built for students). The misnamed `visible_in` (it gates *editing*) was renamed `is_editable` in `CodeExercise` and `Tutorial`.

Deferred / not fixed
--------------------
  - `NextProblem` navigation buttons still dispatch via `ActiveEditor(Move(Goal(NextProblem dir)))`, so they only walk the currently-active cell's problems, not the full aggregated sidebar list. After a sidebar jump the cursor lands in the target cell and NextProblem walks that cell coherently. A fully cross-cell walk would need a new action variant or per-cell cycling in the sidebar itself.
  - ~~DOM focus is not transferred on sidebar jumps (see (3)).~~ Fixed in (3).
  - Some folded expressions surface duplicate rows in the problem sidebar — when an error id resolves to a projector that already had its own error, both versions appear.
  - ~~`DerivationExerciseMode` / `TheoremExerciseMode` `jump_to_tile` still use `root_tile`; same hole-click bug applies.~~ Fixed in (2).
  - Derivation slides can have **duplicate ids across tree judgement editors** (created by `push_premise` inlining an abbreviation without regenerating ids). Jumps to a duplicated id always resolve to the first matching editor, which is often wrong. Tracked as #2277 — a real bug to be fixed in a follow-up PR, not this one.

Notes
-----
Rebased onto dev after the Derivation Trees merge, which split `view/ExerciseMode.re` into per-kind files and reorganized scratchpads into a `kind` sum (`Code | Drv`). `Page.Update.get_problem_editors` dispatches over `Scratchpad.kind`, mirroring the existing `get_editor` helper.

Closes #2223


src/web/app/editors/mode/ExercisesMode.re
```diff
@@ -221,6 +213,25 @@ module Model = {
...
     };
   };
+
+  /* Editors whose problems should appear in the Problems sidebar, each
+     paired with a display label shown as a section header when multiple
+     groups are present. For Code we surface the visible-in-mode cells
+     (Prelude, Test Validation, etc.). For Derivation we surface Prelude,
+     Setup, and every reachable tree judgement node (each labeled
+     "Derivation N.path"). For Theorem we surface Prelude, Lemmas, and
+     Theorem. */
```
7h3kk1d: The derivation part of this comment is stale. We shouldn't be overprescriptive in the comment anyways just keep the part talking about what the function is for.

src/web/app/sidebar/ProblemSidebar.re
```diff
@@ -62,21 +62,31 @@ let problem_status_view = (~globals, ci: Language.Info.t): Node.t =>
     CursorInspector.typ_view(~globals, cls, ~marks, ~message)
   | InfoTPat({cls, marks, message, _}) =>
     CursorInspector.tpat_view(~globals, cls, ~marks, ~message)
-  | Secondary(_)
-  | InfoMod(_)
-  | InfoSig(_)
-  | InfoMPat(_)
-  | InfoDrv(_) => div([])
+  | InfoDrv(ci) => DrvCursorInspector.drv_view(~globals, ci)
```
7h3kk1d, line 68: This is special because we don't do marks on InfoDrv and errors are handled differently.

src/web/app/sidebar/ProblemSidebar.re
```diff
@@ -316,55 +327,157 @@ let view =
...
+          };
+        },
+        categories,
+      );
+    };
+  };
+  let non_empty_groups = List.filter(group_has_problems, collection.groups);
+  /* Cluster consecutive groups with the same label into one section. The
+     Derivation flow uses this to flatten every tree judgement editor into
+     a single "Derivation" section without per-row line numbers. */
```
7h3kk1d: Think about whether this is the right way to do it.

src/web/app/Page.re
```diff
@@ -85,6 +85,34 @@ module Update = {
...
+     Labels are only shown when there are multiple groups, so single-editor
+     modes pass an empty label. Drv scratchpads / documentation slides
+     reuse DerivationExerciseMode.Model.get_problem_editors so Prelude /
+     Setup / each tree judgement node all contribute their problems. */
+  let get_problem_editors =
+      (model: Model.t): list((string, CodeEditable.Model.t)) => {
+    let scratchpad_editors = (m: ScratchMode.Model.t) => {
+      let sp = List.nth(m.scratchpads, m.current);
+      switch (sp.kind) {
+      | Code({editor, _}) => [("", editor.editor)]
```
7h3kk1d: Should we use an option type instead of empty string?

src/haz3lcore/ProblemCollection.re
```diff
@@ -63,20 +76,35 @@ let make_problem_context =
...
       }
     };
+  /* An editor's statics may be computed over a stitched term that spans
+     multiple editors (e.g. Exercise mode's `user_tests` statics covers
+     `user_impl_term ⊕ your_tests.tests`). Filter error/warning ids to
+     those whose piece actually lives in *this* editor's segment; otherwise
+     a type error in `your_impl` would leak into the "Your Tests" group.
+     Using TermData (not Measured) so ids hidden under a projector — which
+     are absent from `measured` but still part of the term — are correctly
+     recognized as belonging to this editor. */
```
7h3kk1d: Comment is too verbose

src/haz3lcore/ProblemCollection.re
```diff
@@ -200,17 +230,172 @@ let collect_category =
...
+  syntax: CachedSyntax.t,
+};
+
+/* Input for one section of the Problems sidebar. `label` is the display
+   name shown as a section header when there is more than one group;
+   `None` for single-editor modes that don't need a label. `sources` is a
+   non-empty list of editors that all contribute to this section. When
+   multiple sources are provided (e.g. every Derivation tree judgement
+   editor under a single "Derivation" label), L# labels are suppressed
+   since "L1" would mean different things in each source. */
```
7h3kk1d: Comment too verbose

src/haz3lcore/ProblemCollection.re
```diff
@@ -200,17 +230,172 @@ let collect_category =
...
+        ~statics=source.statics,
+        ~syntax=source.syntax,
+      );
+    let problems_by_category =
+      List.map(
+        cat => {
+          let deduped =
+            collect_category(ctx, cat)
+            |> List.of_seq
+            |> List.filter((p: problem) => {
```
7h3kk1d: Seq.filter?

src/haz3lcore/ProblemCollection.re
```diff
@@ -200,17 +230,172 @@ let collect_category =
...
+            sort_by_pos(ctx, deduped)
+            |> List.map(p =>
+                 {
+                   problem: p,
+                   measured: ctx.measured,
+                   row_to_line: ctx.row_to_line,
+                   nearest_measured_id: ctx.nearest_measured_id,
+                   pos: ctx.pos(p.id),
+                 }
+               );
```
7h3kk1d: Would it be better to do this map and then sort by `pos` since it's already calculated?

src/haz3lcore/ProblemCollection.re
```diff
@@ -200,17 +230,172 @@ let collect_category =
...
+    problems_by_category;
+  };
+  let groups =
+    List.map(
+      (input: editor_group_input) => {
+        let per_source = List.map(collect_source, input.sources);
+        /* For each category, concat the sources' problems in order.
+           Within each source, problems are already pos-sorted; across
+           sources we keep input order so "Tree 1" appears before "Tree 2"
+           in a merged Derivation section. */
```
7h3kk1d: Comment too verbose

src/web/app/sidebar/ProblemSidebar.re
```diff
@@ -316,55 +327,107 @@ let view =
...
+    | [g] =>
+      /* Single group: no header, just render its body directly. */
+      render_group(g)
+    | gs =>
+      /* Multiple groups: each is a collapsible labelled section with a
+         count. */
+      List.map(
+        (g: problem_group) => {
+          let label = Option.value(g.label, ~default="");
+          let total = List.fold_left((n, (_, c)) => n + c, 0, g.counts);
```
7h3kk1d: can we do a seq map and sum

src/web/app/Page.re
```diff
@@ -85,6 +85,36 @@ module Update = {
     };
   };
 
+  /* Editors feeding the Problems sidebar, paired with display labels.
+     `None` for single-editor modes that don't need a section header. Drv
+     scratchpads / documentation slides reuse
+     DerivationExerciseMode.Model.get_problem_editors so Prelude / Setup /
+     each tree judgement node all contribute their problems. */
```
7h3kk1d: Too verbose

src/web/view/CodeExerciseMode.re
```diff
@@ -86,6 +86,61 @@ module Model = {
...
+     Ordering matters: `ProblemCollection.make` de-duplicates by
+     `(id, category)` with first-wins. `test_validation` and `user_tests`
+     share the same underlying zipper (`your_tests.tests`), so we list
+     `test_validation` first. This makes shared hole/syntax ids land in
+     the "Test Validation" group, matching where jump-to-tile routes
+     (CodeExercise.editor_positions lists YourTestsValidation before
+     YourTestsTesting for the same reason). `user_tests`'s unique static
+     errors (when the tests are evaluated against the student's
+     implementation) still surface in the "Implementation Validation"
+     group. */
```
7h3kk1d: Too verbose

src/web/view/DerivationExerciseMode.re
```diff
@@ -156,6 +156,36 @@ module Model = {
...
+     paired with a display label. Every tree judgement editor shares the
+     label "Derivation"; the sidebar merges runs of same-label groups
+     into one section without per-row line numbers. Trees are walked in
+     postorder so within-tree order matches the visual top-to-bottom
+     layout (premises above the conclusion); trees themselves are in
+     display order. */
+  let get_problem_editors =
+      (model: t): list((option(string), list(CodeEditable.Model.t))) => {
+    let rec postorder = (Tree.Node(v, c)) =>
+      (c |> List.map(postorder) |> List.concat) @ [v];
```
7h3kk1d: concat_map

7h3kk1d:
Read only editors still showing up in the problem sidebar for exercise mode.

cyrus- (changes requested):
- sidebars steal focus but the caret does not disappear (ideally, sidebars shouldn't steal focus at all)
- clicking on problems no longer works

src/haz3lcore/ProbePerform.re
```diff
@@ -9,6 +9,7 @@ module FocusEffect = {
    * and we can't dispatch actions from after_display without causing loops. */
   type target =
     | Editor
+    | Cell
```
7h3kk1d, line 12: Does this belong in ProbePerform.re? It doesn't seem to have to deal with probes.

src/web/exercises/CodeExercise.re
```diff
@@ -406,7 +406,7 @@ let eds_of_spec =
   };
 };
 
-let visible_in = (pos, ~instructor_mode) => {
+let is_editable = (pos, ~instructor_mode) => {
```
7h3kk1d, line 409: This seemed to be a misnomer before and was used for read-only not for visibility

src/web/exercises/CodeExercise.re
```diff
@@ -418,6 +418,17 @@ let visible_in = (pos, ~instructor_mode) => {
   };
 };
 
+/* Whether a cell is *rendered* for the current user. Differs from
+   `is_editable` (which gates *editing*) only for the Prelude, which is
+   shown read-only to students. Drives which cells feed the Problems
+   sidebar and can be jumped to — read-only cells are still navigable
+   (CodeSelectable permits Move), so jumping into them works. */
```
7h3kk1d: Comment to verbose

src/web/exercises/CodeExercise.re
```diff
@@ -418,6 +418,17 @@ let visible_in = (pos, ~instructor_mode) => {
...
+let shown_in = (pos, ~instructor_mode) =>
```
7h3kk1d, line 423: We could call this visible_in but I'm a bit afraid of merge issues

7h3kk1d: A lot of this also feels like it should be across modes. @Negabinary or @disconcision  I'm not sure if you've thought about that.

src/web/view/CodeExerciseMode.re
```diff
@@ -645,7 +633,7 @@ module Selection = {
     CodeExercise.positioned_editors(model.editors)
     |> List.find_opt(((p, e: Editor.t)) =>
          TermData.root_piece(id, e.syntax.term_data) != None
-         && CodeExercise.visible_in(
+         && CodeExercise.shown_in(
```
7h3kk1d, line 636: Allows jumping into read only editors

src/web/view/CodeExerciseMode.re
```diff
@@ -955,117 +914,105 @@ module View = {
...
-        @ [
-          mutation_testing_view,
-          your_impl_view,
-          syntax_grading_view,
-          impl_validation_view,
-          hidden_tests_view,
-          impl_grading_view,
-        ],
-      );
+    @ shown(Prelude, prelude_view)
```
7h3kk1d: prelude_view already contains the reference to the Prelude pos so this is duplicated and can be wrong.

src/web/view/CodeExerciseMode.re
```diff
@@ -955,117 +914,105 @@ module View = {
...
+    @ shown(CorrectImpl, correct_impl_view)
+    @ [correct_impl_ctx_view]
+    @ shown(YourTestsValidation, your_tests_view)
+    @ shown(HiddenBugs(0), wrong_impl_views)
```
7h3kk1d: We shouldn't need to use the token HiddenBugs here

src/haz3lcore/ProbePerform.re
7h3kk1d: I only added the `Cell` case for sidebar-jump focus. @disconcision, you created this module — any objection to pulling `FocusEffect` out of `ProbePerform.re` into its own module since it's a general after-render focus mechanism?

src/web/view/CodeExerciseMode.re
```diff
@@ -694,41 +694,55 @@ module View = {
     let score_view =
       CodeGrading.GradingReport.view_overall_score(grading_report);
 
+    /* Render a cell, but only if it is shown for the current user.
+       `shown_in` is the single source of truth — shared with the sidebar
+       and jump-to-tile — and `this_pos` is the cell's only pos reference.
+       Returns a 0-or-1 element list; the `? :` defers building the cell
+       view, so instructor-only cells aren't built for students.
+       `result_kind` is a thunk so a hidden cell's (possibly expensive)
+       result view isn't built either. */
```
7h3kk1d: Comment is too long

cyrus- (approved):

PR #2231, opened 2026-04-27
Preserve caret in zipper transformations

disconcision:
best effort caret preserving zipper transformations. if id isn't found try ancestors


disconcision:
Superseded: this branch's entire delta is CaretPreserving.re with no call sites, and the byte-identical file already lives on completion-provenance (#2374) wired into Perform.re — CaretPreserving.transform backs Format(Spacing), Format(Pretty), and the indent-adjust transform there (commit 8063edf319). So this lands on dev via #2374; closing to avoid a duplicate unwired copy. (The module originated on patchwork via SyncReplace.re, which keeps its own wired copy.)

PR #2233, opened 2026-04-27
Add pre-commit hook for autoformatting

cyrus-:
## Summary
- Adds a shared `pre-commit` hook (under `scripts/git-hooks/`) that runs `dune fmt --auto-promote` and re-stages any *originally-staged* files the formatter touched, so commits never contain unformatted code.
- Files that weren't staged are deliberately left alone — unrelated unstaged WIP won't be swept into the commit.
- `make install-hooks` sets `core.hooksPath` to `scripts/git-hooks`. It's idempotent and is now a dependency of `deps`, `dev`, and `dev-student`, so anyone running setup or a normal build picks up the hook automatically (no manual opt-in needed).
- `git commit --no-verify` still bypasses the hook for cases where it's necessary.

## Test plan
- [x] `make install-hooks` (first run installs, second run is silent — verified)
- [x] Commit with no formatting changes — hook runs, commit goes through
- [x] Stage a misformatted file with no other unstaged changes — formatted version lands in commit
- [x] Stage a misformatted file *while* having unrelated unstaged edits in another file — only the staged file is in the commit (formatted); unrelated WIP stays unstaged
- [x] `git commit --no-verify` — bypasses the hook
- [ ] Reviewer: try `make deps` once on an existing clone to confirm hooks auto-install

### DIFF
--- a/scripts/git-hooks/pre-commit
+++ b/scripts/git-hooks/pre-commit
@@ -13,6 +13,17 @@ set -u
 staged=$(git diff --cached --name-only --diff-filter=ACMR)
 [ -z "$staged" ] && exit 0
 
+# GUI commits (e.g. the VS Code Source Control panel) often run git without a
+# login shell, so the opam switch environment may not be on PATH. The opam
+# binary itself usually is, so bootstrap the environment from it if needed.
+if ! command -v dune >/dev/null 2>&1 && command -v opam >/dev/null 2>&1; then
+    eval "$(opam env 2>/dev/null)" >/dev/null 2>&1 || true
+fi
+if ! command -v dune >/dev/null 2>&1; then
+    echo "[pre-commit] dune not found on PATH; skipping autoformat" >&2
+    exit 0
+fi
+
 # Files with unstaged changes; a staged file also in this list is partially staged.
 unstaged=$(git diff --name-only)
 
--- a/scripts/git-hooks/pre-commit
+++ b/scripts/git-hooks/pre-commit
@@ -1,16 +1,56 @@
 #!/usr/bin/env bash
 set -u
 
-# Capture which files are staged for this commit. After running the formatter,
-# we'll re-add these (and only these) so the commit picks up any formatting
-# fixes — without sweeping in unrelated unstaged changes the developer may have.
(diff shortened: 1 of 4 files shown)


scripts/git-hooks/pre-commit
```diff
@@ -0,0 +1,16 @@
...
+# we'll re-add these (and only these) so the commit picks up any formatting
+# fixes — without sweeping in unrelated unstaged changes the developer may have.
+staged=$(git diff --cached --name-only --diff-filter=ACMR)
+
+echo "[pre-commit] dune fmt --auto-promote"
+dune fmt --auto-promote || true
+
+if [ -n "$staged" ]; then
+    while IFS= read -r f; do
+        [ -e "$f" ] && git add -- "$f"
```
7h3kk1d: This is going to behave interestingly in the case where someone partially stages a file. Resulting in the entire file being committed.

7h3kk1d:
@Negabinary can you try this with vs code + windows. Not sure how it'll work there.

scripts/git-hooks/pre-commit
7h3kk1d: Having thought about this more my personal preference is probably to just abort the commit if there's reformatting. So I would probably stash the unstaged components run formatting (abort if there's changes) then commit then unstash.

7h3kk1d: Alternatively we can have github reformat on the server

cyrus-:
@7h3kk1d  I went with a middle ground between the current behavior and aborting on any reformatting: the hook now hashes each staged file's worktree content before running dune fmt, so it knows precisely which files the formatter changed. Fully staged files are still autoformatted and re-staged. But if the formatter touches a partially staged file, the commit aborts with a message.

Feels good to me, but if you object I'm also fine with just aborting on autoformat.

cyrus-:
> @Negabinary can you try this with vs code + windows. Not sure how it'll work there.

I had claude take a look at this. Should be fine with WSL which is what we recommend. If you use VS Code Git, added an extra check to make sure `dune fmt` is available since it might not run in a login shell -- if it isn't, it tries to do an `eval $(opam env)` to make it available.

PR #2248, opened 2026-04-30
Pin-gated probe navigation, pin icon restyle, TyDi backpack completion

disconcision:
## Bugs

- [ ] 1 char trailing delim completions not working (they do after backspacing delim but not on forward typing)
- [ ] => delim completion doesn't work, bc of pat sort maybe
- [ ] in fact example in probes slide, pinning (say) the `6` sample of `fact(x-1)` doesn't align other samples

## Summary
- Pin-gated strict arrow navigation for probes; preserves sightline when capture perspective extends up an ap chain
- Pin icon restyle: white circle with red 1.5px stroke, shared rule in `proj-probe.css`
- Multi-pin bug fix: pin icon was appearing at multiple breadcrumbs in recursive calls (via `is_suffix_of` in `SampleFocusBar.re`)
- TyDi: allow 1-char completion for backpack trailing delimiters

## Notable commits
- Preserve sightline when capture perspective-extends up an ap chain
- Pin-gated strict arrow navigation for probes
- Extract `is_reachable_pinned` helper + unit tests
- Unify pin icon style: white circle with red stroke, shared rule
- Fix pin icon appearing at multiple breadcrumbs in recursive calls
- TyDi: allow 1-char completion for backpack trailing delimiters

## Known issue
- TyDi backpack 1-char completion: tests pass, but the live editor does not work for the `case` arrow. Tracked locally.

## Test plan
- [ ] `Test_Evaluator_ProbeNav` (new, 217 lines)
- [ ] `Test_TyDi` additions (~64 lines)
- [ ] Visually verify pin icon in recursive call breadcrumbs
- [ ] Manually exercise pin-gated strict arrow nav


disconcision:
Closing: everything substantive here is on dev via Probes IV (#2363) — sightline preservation up an ap chain, pin-gated strict arrow navigation, `is_reachable_pinned` + `Test_Evaluator_ProbeNav`, and the multi-breadcrumb pin fix (dev `fefcd22cd4`, `b4bf47a39f`, `69e9d10545`, `d69bcfeba9`, `4fbcb0463b`). The fact-example pin misalignment listed above is fixed on dev by span-reference pins. Cmd/Ctrl+P landed in #2269.

What remained on the branch: the pin-icon restyle (not wanted) and the TyDi one-character backpack completion, which is superseded by completion-provenance (#2374): its TyDi bypasses the two-character gate for any delimiter the syntax expects, so `if true t` completes to `then` there too, and a typed `=` goes through the pending-delimiter path.

Last tip of the branch: a5c8ffbcf1.

PR #2254, opened 2026-05-02
Parameterized types and multi-binder polymorphism

### LINKED ISSUE #802: Parameterized types

cyrus-:
This PR introduces first-order parameterized type constructors, a tuple-arrow kind system, type-level type functions, and a clean separation between value-level and type-level type abstraction. It also threads multi-binder polymorphism through the existing `poly`/value-level type-application machinery.

### Parameterized types
Hazel now supports parameterized type aliases:

```hazel
type Option(a) = + None + Some(a) in
type Either(a, b) = + L(a) + R(b) in
type List(a) = + Nil + Cons(a, List(a)) in
let xs : List(Int) = Cons(1, Nil) in xs
```

- A declaration with `n` parameters introduces a type constructor of *tuple-arrow* kind `(Type, …, Type) -> Type`. Applying a constructor to its `n` type arguments at once produces an ordinary type.
- Tuple-arrow kinds are atomic: `Either(Int, Bool)` is well-kinded, but `Either(Int)` is rejected as `expects 2 arguments, got 1`. Extra parens disambiguate single-tuple-arg vs multi-arg: `Either((Int, Bool))` is also a 1-of-2 arity error, while `List((Int, Bool))` is a list of pairs.
- Recursive aliases stay implicit, including non-uniform recursion: `type List(a) = + Nil + Cons(a, List((Int, a)))` works and runs to completion.
- Constructors are introduced with polymorphic schemas: `Some : poly a -> a -> Option(a)`, `L : poly a, b -> a -> Either(a, b)`, etc. Implicit instantiation works in analytic positions (`Some(3) : Option(Int)`), and the same constructor can be used at multiple specializations in one program.
- Modules can export parameterized aliases: `module M = { type L(a) = + Nil + Cons(a, L(a)) }` and `M.L(Int)` resolves correctly through the module's exports.

### Type-level type functions: `typfun`

Aliases can be written with the prefix-binder syntax:

```hazel
type Option = typfun a -> + None + Some(a) in
type Either = typfun a, b -> + L(a) + R(b) in
```

This is the same alias as `type Option(a) = …` / `type Either(a, b) = …` (same kind, same constructor schemas).

A *curried* form is also supported and stays curried:

```hazel
type PResult = typfun err -> typfun ok -> + Error(err) + Ok(ok) in
type StringResult = PResult(String) in     // kind Type -> Type
type R = PResult(String)(Bool) in          // fully applied; kind Type
```

`PResult` has kind `Type -> Type -> Type` and accepts curried application. Partial applications are well-kinded and may themselves be used as alias bodies.

### Value-level type abstraction: `abs` (renamed from `typfun`)

To free the `typfun` keyword for the type-level form above, the value-level type abstraction (System F's big-Lambda) is now spelled `abs`:

```hazel
let id : poly a -> a -> a = abs a -> fun x : a -> x in
let pair : poly a, b -> a -> b -> (a, b) =
  abs a, b -> fun x : a -> fun y : b -> (x, y) in
pair@<Int, Bool>(3)(true)
```

Multi-binder `poly`, `abs`, and the value-level type application `e@<T1, T2>` are now first-class: a single `Poly(TPat.Tuple([a, b]), …)` binds both at once, and `e@<T1, T2>` reduces in one step. Curried explicit nesting (`poly a -> poly b -> …` / `abs a -> abs b -> …`) keeps its curried structure.

A new `Mark.TypAbsApplyArityMismatch` is emitted at the `@<…>` node when the type-arg count doesn't match the function's `Poly` arity (e.g. `e@<Int>` for a 2-binder `Poly`); the result type is `Unknown` so downstream value-arg type-checking doesn't see a body with leaked free type variables.

### Prelude updates

- `Option(a)` and `Either(a, b)` are now parameterized aliases (kinds `(Type) -> Type` and `(Type, Type) -> Type`).
- `Either`'s constructors are renamed `Left`/`Right` → `L`/`R`.
- The `_opt`-family list functions and `partition_map` use `Option(?)` and `Either(?, ?)` in their signatures so the cursor inspector shows the parameterized form rather than the expanded `+ None + Some(?)` sum.
- `string_compare` returns `Ord` (the named alias) instead of the expanded sum.

### Cursor inspector & error messages

- Type aliases now display as declarations: `type Either(a, b) = + L(a) + R(b)` rather than `Either :: typfun a, b -> + L(a) + R(b)`.
- A type-pattern's class label distinguishes "Type alias" from "Type parameter" based on its statics classification.
- New focused marks: `TypKindMismatch`, `TypParamApplyNonArrowKind`, `TypParamApplyArityMismatch` (callee + arity), `TypAbsApplyArityMismatch`, `TPatParamNotAtAliasHead` (e.g. `poly T(a) -> …` is rejected with a clear message).
- Kind errors stay localized to the offending node (no longer spread to surrounding `+` / `,` tokens).
- Free type variables in `TypParamAp` callees produce a single mark (the free-variable mark on the var), not a redundant "cannot apply" mark too.
- Constructor highlighting / jump-to-definition is type-directed: when two sum types in scope each declare `B`, `B(true) : OneOfThree(_, _, _)` resolves to `OneOfThree`'s `B`, not the innermost shadowing one. Constructors declared inside a module also highlight correctly.
- Hovering on `M.PList` lights up `PList`'s declaration site inside `M`.

### Identifier syntax
Capitalized identifiers (constructors, module names) accept trailing primes — `Result'`, `M''`, etc.

### ExplainThis
- Dedicated forms for parameterized types, multi-binder `poly` / `typfun` / `abs`, multi-arg type application, and module / signature decls, with arity-dispatched explanations for arity-1/2/3 forms.
- Per-form highlighting for module/signature item decls.
- `Module Definitions` items get correct dropdowns and highlighting.

### Other surface changes
- Sum types render as `+Lt + Eq + Gt` (space after the leading `+`, spaces on both sides of interior `+`s) instead of the previous `+ Lt + Eq+ Gt`.
- Constructor arguments in sum types render without doubled parens: `Cons(a, List(a))` instead of `Cons((a, List(a)))`.

### AST additions

`src/language/term/Grammar.re`:

- `Typ.TypFun(tpat, typ)` — type-level type function.
- `Typ.TypParamAp(typ, typ)` — type parameter application.
- `Typ.TypTuple(list(typ))` — multi-arg bundle (only ever the second arg of a `TypParamAp`; no kind on its own).
- `TPat.Param(tpat, list(tpat))` — alias-head form `T(a, b, …)`.
- `TPat.Tuple(list(tpat))` — multi-binder list, used as the binder of `Poly` / `TypAbs` / `TypFun` / `Rec` for comma-separated forms.
- `TPat.Parens(tpat)` — transparent parens wrapper so user parens get their own tile id.
- `Exp.TypAbs` — renamed from `Exp.TypFun` (the value-level type abstraction).

### New module: `Language.TypKind`

```ocaml
type t =
  | Unknown
  | Type
  | Arrow(list(t), t);
```

`apply_all` consumes a tuple-arrow kind's argument list in one step (atomic; no partial application of multi-binder kinds), but a curried `Arrow([Type], Arrow([Type], Type))` accepts arguments one per `TypParamAp`. `Unknown` is consistent with everything and absorbs application. `to_string` renders single-arg arrows as `k -> r` and multi-arg as `(k1, …, kN) -> r`.

### Single source of truth for kinds

`Statics.kind_of_typ` is now a top-level helper. `TyAlias` calls it once on the alias body and threads the result through `Ctx.extend_alias(~typ_kind=…)` (so downstream `Var(name)` references see the right kind via `lookup_tvar_typ_kind`) and `utpat_to_info_map(~alias_kind=…)` (so the cursor-inspector message at the alias name agrees).

### `TypExpectation.AnyKindExpected`

A new variant on `TypExpectation.t` distinguishes alias-body positions (any kind is fine — the alias inherits the body's kind) from value-annotation positions (must be kind `Type`). This lets the strict kind-`Type` check live where it belongs without firing on legitimate higher-kinded constructs (a curried alias body, a partial parameterized application used as an alias body, a `TypFun` alias body, …).

### Higher-kinded recursive types

`type List(a) = + Nil + Cons(a, List(a))` is stored as `Rec(List, TypFun(a, Sum[Nil, Cons(a, TypParamAp(Var("List"), a))]))` — a fixed point at kind `* → *`. `weak_head_normalize` intentionally treats `TypParamAp(Rec(_, TypFun(_, _)), _)` as a normal form so recursive types don't infinitely expand. `Typ.unfold_one` performs one step of μ-unrolling for callers (constructor matching, sum extraction, structural meet across `Sum`/`Rec` shapes) that need to peer inside. `Typ.apply_args` peels one `TypFun` per `TypTuple` element for multi-arg applications.

### Multi-binder substitution and reduction

- `Typ.subst_many` and `DHExp.ty_subst_many` substitute a list of types for a corresponding list of binders in one step.
- `Transition.TypAp` zips a `TypTuple` argument against a `TPat.Tuple` binder element-wise. Single-binder forms substitute as before.
- Polymorphic constructor schemas are a single `Poly(TPat.Tuple([a, b, …]), ty)`, never a curried chain.
- `Equality` and `Typ.meet` flatten multi-binder `Poly`s element-wise via `TPat.binders_of`. A binder list of all holes is treated as a wildcard so the `Poly(EmptyHole, _)` placeholder used as `TypAp`'s callee analysis target matches any `Poly` regardless of binder shape.

### Module exports

`ExpandModule.collect_type_exports` now handles `Param` tpats: it builds a single `TypFun(TPat.Tuple([a, b, …]), body)` over the params (wrapped in `Rec` for self-referential definitions) and `Ctx.extend_alias`-es it with the matching `(Type, …) -> Type` kind. The exported field's tile is annotated with the original alias declaration's id so var-highlight on `M.T` recovers the binding site.

### Parsing

- The new keywords (`abs`, `typfun`, `poly`, `typlam`, `forall`, `rec`) are listed in `Token.keywords` so `should_add_space` keeps the space between keyword and a leading `(`.
- `MakeTerm.binder_of_tpat` lifts a comma-separated tpat `MultiHole` into a `TPat.Tuple` (preserving an outer `Parens` wrapper if present).
- `MakeTerm.apply_typ_param_args` normalizes `T(a, b, …)` into `TypParamAp(T, TypTuple([a, b, …]))` while leaving `T((a, b))` as `TypParamAp(T, Prod([a, b]))`. The sum-variant extractor reverses this locally so `Cons(a, b)` in a sum-type position remains a constructor with a `Prod` payload.
- The Menhir parser, AST, and conversion layer all gain matching nodes (`TypParamAp`, `TypTuple`, `TypFunType`, `ParamTPat`, `TupleTPat`).
- `Precedence.dot = 10` (was 22) so `M.T(Int)` parses as `(M.T)(Int)`. Slide files were migrated via `scripts/migrate_dot_precedence.sh`.

### Substitution bug fix
`Substitution.in_exp` previously substituted into the RHS of `Let` / `Theorem` using the *extended* env, which would alpha-rename free occurrences of the bound name in the RHS itself. The non-recursive RHS now uses the outer env. A regression test `Test_Stepper` covers the case (`{ let y = 1; let y = y }` and similar).

### Property-based testing

The QCheck random term generator now exercises modules, parameterized type aliases (`type T(a, …) = body` and the typfun-bodied form), `TypFun`, `TypParamAp`, multi-binder `poly` / `abs`, `e@<T1, T2>`, and `T.x` projections — round-trip and statics-preservation properties cover all the new forms.

### Test coverage

- `test/statics/Test_Statics_ParameterizedTypes.re` (new) — kind errors, arity mismatches, multi-arg application, constructor elaboration, curried `typfun`, partial application as alias body, etc.
- `test/statics/Test_Statics_Modules.re` — parameterized type aliases inside modules.
- `test/evaluator/Test_Evaluator_TypAp.re` — runtime reduction, including non-uniform recursion and multi-binder TypAp.
- `test/Test_Typ.re` — β-normalization, recursive family lookup.
- `test/Test_VarHighlight.re` — type-directed constructor lookup, module-internal constructors.
- `test/Test_Menhir.re` and `test/Test_MakeTerm.re` — both syntaxes for parameterized types, the renamed `abs` keyword, `T(a, b)` argument normalization, multi-binder forms.

### Documentation

- `docs/parameterized-types.md` — the parameterized-types story (this doc deliberately doesn't cover `abs` / value-level type application; that's described in this PR).
- `docs/modules.md` — interaction with parameterized types.

## Test plan
- [ ] `./run_tests test '' -q` passes (currently 2704 tests, all green)
- [ ] Manual smoke: `type Either(a, b)`, `type Option`, `Some(3) : Option(Int)`, `L("a") : Either(String, Int)`
- [ ] Manual smoke: `abs a, b -> fun x:a -> fun y:b -> (x, y)` and `f@<Int, Bool>(…)`
- [ ] Manual smoke: `type T = typfun a -> typfun b -> body`, `T(String)(Bool)`, `T(String)` as alias body
- [ ] Manual smoke: `module M = { type L(a) = + Nil + Cons(a, L(a)) }`, `M.L(Int)`, jump-to-definition on `M.L`
- [ ] Cursor inspector: kinds shown correctly on alias names, parameter names, `T(args)` callees

Made with [Cursor](https://cursor.com)


7h3kk1d:
Awesome! Stoked to stop doing OptInt and OptString in slides. I haven't reviewed the code yet, just been trying it out on the slides.

  The one thing I'm having to fix in the B2T2 slides is constructors in synthetic position.

  Example from one of the B2T2 slides:
synthetic


  Need to add an explicit Some@<T>(...) TypAp or an : Option(T) ascription.

  Should we open a GitHub issue to track implicit application in synthetic position after this is merged?

src/language/term/Grammar.re
```diff
@@ -103,6 +103,19 @@ and typ_term('a) =
...
+     applications like `Either(a, b)`. It only appears as the second
+     argument of `TypParamAp` and never as a stand-alone type — it has
+     no kind by
+     itself; its arguments must match the tuple-arrow kind of the
+     callee. Single-argument applications and applications whose
+     argument is itself a parenthesized tuple use plain `TypParamAp(T, t)`
+     without `TypTuple` (so `Either((Int, Bool))` is a partial
+     application of `Either` to a `Prod`, distinct from `Either(Int,
+     Bool)` which becomes `TypParamAp(Either, TypTuple([Int, Bool]))`). */
+  | TypTuple(list(typ_t('a)))
```
Negabinary, line 118: Does this need a separate entry in the AST or could TypParamAp just take a list in the AST?

src/language/term/Grammar.re
```diff
@@ -122,6 +135,22 @@ and tpat_term('a) =
...
+  /* Tuple is a multi-binder type pattern, used as the binder of `Poly`,
+     `TypAbs`, `TypFun`, and `Rec` when the user wrote a comma-separated
+     list of binders, e.g. `poly a, b -> t` parses as
+     `Poly(TPat.Tuple([a, b]), t)` and `typfun a, b -> e` parses as
+     `TypAbs(TPat.Tuple([a, b]), e, _)`. The corresponding type-level
+     application supplies a `TypTuple` argument of matching arity, and
+     substitution / reduction zip the pair element-by-element. Tuple
+     never appears as a stand-alone tpat outside a binder position. */
+  | Tuple(list(tpat_t('a)))
+  /* Parens is a transparent wrapper for displaying the user's
```
Negabinary, line 148: Superfluous comment

src/language/statics/Statics.re
```diff
@@ -11,6 +11,120 @@ include StaticsBase;
 let add_info = Map.add_info;
 let add_missing_info = Map.add_missing_info;
 
+/* Compute a type's kind without descending into descendants. The
```
Negabinary, line 14: I'd prefer if this was part of the main traversal instead of tacked on like this - you can add a TypeKind field to the type info record, and then access it directly from info, the same way elab-syn-tys are accessed in expressions.

(Also Statics helpers like this should probably go in a separate file, possibly in Ctx with the other Ctx helpers)

src/language/statics/Statics.re
```diff
@@ -2193,70 +2426,346 @@ and uexp_to_info_map =
...
+         (params extension, polymorphic constructor schemas,
+         tuple-arrow kind `(Type, …) -> Type`).
+
+         A *curried* form `type T = typfun a -> typfun b -> body`
+         is left intact: each unary `TypFun` stays as its own
+         binder so the alias has the curried kind
+         `Type -> Type -> kind(body)` and accepts curried
+         applications `T(a)(b)`. `peel_typlams` refuses to peel a
+         TypFun whose body is itself a TypFun. */
+      let rec strip_parens = (t: Typ.t): Typ.t =>
```
Negabinary, line 2446: Shouldn't be in statics - risks later duplication; I'm surprised this helper doesn't exist already

src/language/statics/Statics.re
```diff
@@ -2193,70 +2426,346 @@ and uexp_to_info_map =
...
+        | Sum(_)
+        | Poly(_)
+        | ProofOf(_)
+        | Sig(_)
+        | TypTuple(_)
+        | TypFun(_)
+        | TypParamAp(_)
+        | Rec(_) => t
+        };
+      let peel_typlams = (typ: Typ.t): (list(TPat.t), Typ.t) => {
```
Negabinary, line 2471: Again, ideally this helper function lives outside statics

docs/parameterized-types.md
Negabinary, line 1: I've seen some chat about currying here - 1) do we want to be able to curry these given that we don't automatically curry normal functions? 2) would we be able to simplify the statics logic if we didn't have currying?

src/language/term/Equality.re
```diff
@@ -764,10 +778,39 @@ let equality =
   }
   and tpat = (tp1: TPat.t, tp2: TPat.t): option(Alphas.t) => {
     switch (tp1 |> Annotated.term_of, tp2 |> Annotated.term_of) {
+    /* Parens is transparent — a parenthesized tpat is equal to the
```
Negabinary, line 781: We have a setting for ignoring parens

src/language/term/Typ.re
Negabinary, line 1008: This meet looks like it does some sort of implicit auto abstraction now - it might be nice if that behaviour was documented at the top of the function now.

src/language/dynamics/Substitution.re
```diff
@@ -47,8 +47,17 @@ let rec in_exp = (env: Environment.t(Exp.t), exp: Exp.t) =>
 
         // Cases with binders: remove binder from env
         | Let(p, e1, e2) =>
+          /* `let p = e1 in e2` is non-recursive: `p` is bound only in
```
Negabinary, line 50: Comment largely references previous version

Negabinary (commented):
Did a quick code style pass during takeoff, haven't checked the functionality.

src/language/term/Grammar.re
7h3kk1d: I talked to my claude about this and it was partially about the need to have somewhere to hang the `Id.t` on. I personally find that really frustrating as an answer though.

Negabinary: does it need a separate id in statics? or MakeTerm can we just add all those ids to the same TypeParamAp? idk if the commas need to have a different cursor inspector message to the application itself.

7h3kk1d: This may no longer be true but I tried that with the label/projection at one point with labeled tuples and there was some sort of uniqueness issue. Possibly related to ancestors in the statics map.

docs/parameterized-types.md
7h3kk1d: I don't know if I'm understanding it correctly but I don't think this does automatic currying. It just allows for manual currying the same as normal functions.

Negabinary: oh you're right, mb

7h3kk1d:
typfun explain this doesn't work

MaxCarroll0:
From a user perspective I think the distinction between `typfun`, which introduces a _higher kinded_ type vs. `poly`, which is a (possibly higher rank) _polymorphic_ type, is a bit difficult to notice.

Both of these do different things, but look/sound annoyingly similar. Distinction:
- Higher kinded types can be applied at the type level, and must be fully applied.
- A polymorphic type is a singleton kind `Type`, cannot be applied at the type level, but its values can be partially applied at the value level (`_@<Typ>`)

Yet writing `type Id = typfun A -> (A -> A)` vs `type Id = poly A -> (A -> A)` look quite similar. I feel it makes sense to entirely disallow writing `typfun` explicitly to make this a little more obvious. Especially given the `explainThis` doesn't say anything about how `typfun` construct works.

Anyway, here's a little example we could use as a base for some docs for demonstrating the difference:
image

(ok technically these are order 1 kinded types, not higher kinds, but you get what I mean)

MaxCarroll0:
Also, allowing writing `typfun` allows some very weird annotations which should definitely be erroneous, but aren't caught.

Despite higher kinds not being allowed in annotations, I can literally annotate a value with a higher kind directly (`typfun`). Both uses below should error:
image

I think it makes sense to entirely disallow writing `typfun` in the surface syntax. Is there any genuine use for it? Since it doesn't make sense to write type aliases containing higher kinded types not in prenex normal form. 
(Technically our current function def syntax doesn't allow things like `type Either(A)(B) = ...` for curried types, or `let f(a)(b) = ...` for curried functions, so you would probably need to add this synrax.)

MaxCarroll0:
Here is the relevant scratch file (.txt so github can upload)
[poly_vs_typfun.txt](https://github.com/user-attachments/files/30467066/poly_vs_typfun.txt)

MaxCarroll0:
Also, reminder to add an ExplainThis entry for `typfun`

MaxCarroll0:
There is also an issue with constructor resolution for curried typfuns: 
image
(D is not found)

PR #2267, opened 2026-05-05
Fix probes on custom-static tuple ops (#2264)

### LINKED ISSUE #2264: Probes not working on custom static tuple operations
Image

recent regression

7h3kk1d:
## Summary

- Fixes #2264. Probes on applications of builtin tuple ops with `custom_statics` (`to_lvs`, `project_labels`, `select_labels`, `omit_labels`, `omit_all_labels`, `group_by_label`) recorded zero samples.
- Root cause: `mk_builtin_ap_elab` built the elab `Ap` with `Exp.fresh`, minting a new id. Probe targets are keyed by `user_term` ids in `info_map`; the evaluator looks targets up by elab term id; the two drifted, so the lookup missed.
- ~~Fix: thread the source `Ap`'s ids from the call site in `Statics.re` through `custom_statics_ap` and its helpers into `mk_builtin_ap_elab`, which now stamps the elab `Ap` with those ids via `IdTagged.IdTag.mk_internal`.~~ Fix: thread the source `Ap`'s full annotation from the call site in `Statics.re` through `custom_statics_ap` and its helpers into `mk_builtin_ap_elab`, which now constructs the elab `Ap` record directly with that annotation (preserving ids and secondary runs). `custom_statics_deferred_ap` was already fine — its elab term flows through `rewrap` from the call site.

## Tests

- Added 7 probe sample-count tests in `test/evaluator/Test_Evaluator_Probes.re` (one per custom-statics builtin plus a multi-call shape) under `Evaluator.Probes.CustomStaticsTupleOps`. Validated by reintroducing the bug locally: all 7 failed with `expected 1 sample, received 0`; reverting restored 2697 passing tests.

~~- Added a statics invariant `id_alignment_violations` and two QCheck properties in `test/statics/Test_Statics_Properties.re`:~~
~~- generic over `arb_exp`-generated programs;~~
~~- targeted by wrapping a generated body in `Ap(Var(<custom-statics-builtin>), _)` at the test boundary, with no shrinker.~~

The PBTs were dropped before merge: at higher iterations the broad invariant surfaced multiple pre-existing id-drift sites unrelated to #2264 (`TupLabel` rebuilt with parent `rewrap`, `TyAlias` erasure, `Constructor` / `List literal` reconstruction in case branches), and the narrower custom-statics PBT was redundant with the manual probe tests above.

## Test plan

- [x] `make dev` clean
- [x] `make test-quick` passes
- [x] Full `dune runtest test --profile dev` passes (2697 tests)
- [x] Manual UI: place a probe on `to_lvs((a=1, b=2))`, `project_labels(t, a)`, `omit_all_labels(t)`, `group_by_label(rows, k)` — sample appears in each
- [x] Manual UI: probes on partially applied versions still work (regression check on `custom_statics_deferred_ap` path)


src/language/statics/CustomStatics.re
```diff
@@ -10,8 +10,19 @@ type tuple_type = list(tuple_entry);
...
+   elab `Ap` had a fresh id (as it would from `Exp.fresh`), the lookup would
+   miss and probes on builtin calls would never fire. */
+let mk_builtin_ap_elab =
+    (~ids: list(Id.t), fn_info: Info.exp, arg_elab: Exp.t): Exp.t => {
+  let ap = Ap(Forward, fn_info.elab_term, arg_elab) |> Exp.fresh;
+  {
+    ...ap,
+    annotation: IdTagged.IdTag.mk_internal(ids),
+  };
+};
```
7h3kk1d: Don't call fresh here if we're providing the id

7h3kk1d: alternatively we can pass in the full annotation

PR #2268, opened 2026-05-06
fix: Shift+Backspace and Shift+Delete behave like plain Backspace/Delete (#2253)

### LINKED ISSUE #2253: Shift-Backspace is a no op
I'm expecting it to just do normal backspace.

7h3kk1d:
## Summary
- Backspace/Delete handlers in `src/web/Keyboard.re` previously required `shift=Up`, so holding shift made them no-ops.
- Match either shift state since users expect shift to be ignored for destructive key actions.

Closes #2253.

## Test plan
- [x] In an editor, place caret in code and press Shift+Backspace — should delete left like plain Backspace.
- [x] Press Shift+Delete — should delete right like plain Delete.
- [x] Plain Backspace and Delete still work.


Negabinary (approved):
idk why we want this but it works

cyrus-:
> idk why we want this but it works

Text editing elsewhere behaves like this and I have run into having shift held down incidentally and backspace isn't working so we might as well make our behavior consistent.

7h3kk1d:
non-serious take: I like the idea of shift backspace deleting forward and shift delete deleting backwards. Or shift backspace replacing the previous character with the capitalized version.

PR #2281, opened 2026-05-08
Fix all_ctrs_of_typ infinite recursion on vacuous Rec bindings

### LINKED ISSUE #1624: Infinite recursion in all_ctrs_of_typ
`fun ((()): ((rec x -> (rec y -> x)))) -> [] ` breaks the coverage checker in statics with non-termianation in all_ctrs_of_typ

### LINKED ISSUE #2235: Recursion limit exceeded in all_ctrs_of_typ error
`fun ((()):((rec x -> (rec ? -> x)))) -> 132032.832758` crashes the statics/elaborator with
`Failure("Recursion limit exceeded in all_ctrs_of_typ")`

Found from the elaborator doesn't crash pbt

7h3kk1d:
## Summary
- `all_ctrs_of_typ` crashed with `Failure("Recursion limit exceeded in all_ctrs_of_typ")` on types like `rec x -> (rec ? -> x)`. Root cause: `Typ.unroll` is non-productive on Recs whose binder is a hole (or otherwise unused), so unrolling cycles between two equivalent representations.
- Fix: at a `Rec` head in `all_ctrs_of_typ`, strip vacuous Recs from the body's head before deciding what to do. This makes `rec x -> (rec ? -> x)` collapse to `rec x -> x`, which the existing self-loop rule maps to `Unknown`.
- Same fix also resolves the long-standing `rec x -> (rec y -> x)` case (#1624), which had been parked as `skip_known_bug`.

Closes #2235
Closes #1624

## Test plan
- [x] `make dev` builds clean
- [x] `make test` passes (2690 tests, including qcheck property tests)
- [x] New regression tests added for both #2235 and #1624 (the latter promoted from `skip_known_bug` to a passing test)
- [x] Manual verification in editor

### DIFF
--- a/src/language/statics/Coverage.re
+++ b/src/language/statics/Coverage.re
@@ -94,21 +94,6 @@ module Ctr = {
       }
     };
 
-  /* Strip Recs whose binder isn't free in the body — they're equivalent
-     to the body. Without this, `Typ.unroll` cycles on types like
-     `rec x -> (rec ? -> x)`. See hazelgrove/hazel#2235, #1624. */
-  let rec strip_vacuous_rec = (ty: Typ.t): Typ.t =>
-    switch (ty.term) {
-    | Rec(tp, body) =>
-      let vacuous =
-        switch (TPat.tyvar_of_utpat(tp)) {
-        | None => true
-        | Some(name) => !List.mem(name, Typ.free_vars(body))
-        };
-      vacuous ? strip_vacuous_rec(body) : ty;
-    | _ => ty
-    };
-
   let rec all_ctrs_of_typ = (~rec_count=0, ty: Typ.t): all_ctrs => {
     if (rec_count > 1000) {
       failwith("Recursion limit exceeded in all_ctrs_of_typ");
@@ -128,15 +113,11 @@ module Ctr = {
            )
         |> Map.of_list,
       )
-    | Rec(tp, body) =>
-      let body = strip_vacuous_rec(body);
-      switch (TPat.tyvar_of_utpat(tp), body.term) {
-      | (Some(w), Var(v)) when v == w => Unknown
-      | (None, _) => all_ctrs_of_typ(body)
-      | (Some(name), _) when !List.mem(name, Typ.free_vars(body)) =>
-        all_ctrs_of_typ(body)
-      | _ => all_ctrs_of_typ(Typ.unroll(ty))
-      };
+    | Rec(_) =>
+      switch (Typ.unroll_to_non_rec(ty)) {
+      | None => Unknown
+      | Some(ty') => all_ctrs_of_typ(ty')
+      }
     | Prod(elts) =>
       Finite(Map.singleton(tuple_ctr(List.length(elts)), elts))
     | ProofOf(_) => Infinite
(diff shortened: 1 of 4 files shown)


cyrus-:
This issue may already be fixed / manifest differently on #2254, maybe check that

7h3kk1d:
> This issue may already be fixed / manifest differently on #2254, maybe check that

I confirmed it manifests the same. I'll look and see if it changes related code

7h3kk1d:
Checked — it manifests the same on `parameterized-types-4`. `Coverage.all_ctrs_of_typ` still falls through `Rec(_) => all_ctrs_of_typ(Typ.unroll(ty))`, and `Typ.unroll` is unchanged. The new `TypParamAp(Rec(_, TypFun(_,_)))` arm with `unfold_one` is for higher-kinded recursive families and doesn't catch `rec x -> (rec ? -> x)`.

The fix should port over cleanly. PT4 adds multi-binder `TPat.Tuple` for `poly`/`typfun`/`abs`, but `rec` is single-binder — only the `Rec(Var, _)` arm in statics binds anything. So `tyvar_of_utpat` only needs to handle `Var` as a binder. The one thing I worried about, is`Parens` wrappers around the binder, but PT4 already handles them in `tyvar_of_utpat` (recurses through `Parens(inner)`).

src/language/statics/Coverage.re
```diff
@@ -94,6 +94,21 @@ module Ctr = {
       }
     };
 
+  /* Strip Recs whose binder isn't free in the body — they're equivalent
+     to the body. Without this, `Typ.unroll` cycles on types like
+     `rec x -> (rec ? -> x)`. See hazelgrove/hazel#2235, #1624. */
+  let rec strip_vacuous_rec = (ty: Typ.t): Typ.t =>
```
7h3kk1d: Can this be put in type normalization

7h3kk1d: Check to see if this bug happens in Typ.normalize

src/language/statics/Coverage.re
```diff
@@ -113,8 +128,15 @@ module Ctr = {
            )
         |> Map.of_list,
       )
-    | Rec({term: Var(w), _}, {term: Var(v), _}) when v == w => Unknown
-    | Rec(_) => all_ctrs_of_typ(Typ.unroll(ty))
+    | Rec(tp, body) =>
+      let body = strip_vacuous_rec(body);
```
7h3kk1d: strip_unbound_recs

7h3kk1d: unroll_to_non_rec

test/statics/Test_Statics_Types.re
```diff
@@ -473,10 +473,14 @@ let tests = (
       "Typ.weak_head_normalize infinite recursion", // https://github.com/hazelgrove/hazel/issues/1621
       "type y = y in type ? = y in ?",
     ),
-    skip_known_bug(
-      "Coverage.all_ctrs_of_typ infinite recursion", // https://github.com/hazelgrove/hazel/issues/1624
+    statics_does_not_crash(
```
cyrus-: put the actual errors here, remove statics_does_not_crash helper

cyrus- (changes requested):

cyrus- (approved):

Issue #1487, opened 2025-01-29
Build step for building Init.ml editors


disconcision:
Currently we have a somewhat brittle, laborious, and diff-intensive process of populating the initial editors by exporting the whole app state and inserting zipper serializations into the code. What I want instead is something like a folder in the codebase with a plain-text file for each doc slide, which, at compilation time, is turned into the Init.ml serialization. Or alternatively if the parser is fast enough we could just do this at run time. In any case whatever we build could reasonably be part of the git ignore so we no longer get those chonky commit diffs.

Complexities:
- Doc slides with additional text at the top (not sure how this is implemented)
- Projectors (would need to add textual syntax for them)

I think I have a simple-ish story for the latter at least.

7h3kk1d:
Just curious for your thoughts. If we don't want to textify the projector state can we do a sexp-serialization and just have a better interface for downloading it from the app? I can't tell if you want the text version for updating the files in vs code or just for the git diffing.

disconcision:
Should be more doable when if #1808 is merged.

disconcision:
@7h3kk1d my primary motivation for this is not to have to regenerate each slide when there is a change to the zipper data type or any contained data types, which includes e.g. precedence changes, which has been slowing down making basic fixes. secondary benefit is cleaner diffs

7h3kk1d:
Just talked with @cyrus- about a proposed plan for this:

- When copying hazel code out of the editor we can manifest convex grout as an explicit character such as `¿` like I did in https://github.com/hazelgrove/hazel/tree/whitespace_preserving_reparse. 
- When pasting code back into hazel Interpret every `¿` as an explicit hole and track the Id of all the implicit holes
- After the parse replace every tacked Id in the segment structure with convex grout to recover the original positions
- If pasting into a text editor the `¿` represent the convex grout

We can then store the documentation slides as text files exported from copied from the editor and import the text the same way paste works and validate that the serialization of the resulting text matches the original file.

@disconcision do you have any concerns with this approach or a simpler one that retains whitespace and allows for explicit ?s in syntax?

disconcision:
I understand what you're saying as a plan to preserve relative whitespace/grout positioning in the text representation.

1. This seems like a relatively minor aspect of the overall ask here. When does this import happen? I.e. Is there a compile-time step?
[...]

PR #2282, opened 2026-05-09
Add slide-* CLI subcommands for inspecting and rewriting Hazel slides

7h3kk1d:
## Summary

Adds three small `hazel` CLI subcommands for working with the documentation slides linked into the binary (the `let out : ... PersistentSegment.t` modules under `src/web/init/docs/` and `src/b2t2/slides/`), plus a new core module — `Haz3lcore.TextRoundtrip` — that makes the slide round-trip preserve **implicit holes (Grout)**, plus supporting tweaks to `format` and `analyze` and a small CLI refactor.

The design is deliberately minimal: slides are addressed by **name**, not by `.ml` file path, and the only slide-specific operations exposed are `decode` (slide → plaintext) and `encode` (plaintext → slide). Every other transformation (prettify, suppress unused-var warnings, static analysis) is done generically on the plaintext via the existing `format` and `analyze` commands.

> [!NOTE]
> This is intended as a stopgap. Once #1487 ("Build step for building Init.ml editors") lands — replacing the serialized `let out : ... PersistentSegment.t` modules with plain-text slide files compiled into `Init.ml` at build time — these commands will likely be obviated or need to be reworked. At that point `slide-decode` collapses to `cat`, `slide-encode` collapses to writing the plain-text file, and only the trigger-syntax round-tripping (which would presumably move into whatever textual syntax #1487 adopts for projectors/refractors) carries over.

### Preserving implicit holes through the text round-trip

**The problem (raised in review).** Slides contain Grout pieces — implicit empty holes the editor inserts to maintain shape invariants. A naive `to_text` rendered Grout as `" "` (whitespace), which the parser couldn't recover on the way back: re-parsing would either drop the Grout entirely or place it differently, breaking `decode → encode` for any slide that contained one.

**The fix.** A new core module `Haz3lcore.TextRoundtrip` (`src/haz3lcore/zipper/TextRoundtrip.re`) renders Grout as a distinguishable marker token, then on `of_text` walks the resulting zipper and `Destruct.go`s each marker — `remold_regrout` re-inserts Grout in the canonical position. The marker character is settable via a new `--implicit-hole CHAR` flag on `slide-decode`, `slide-encode`, and `format` so it survives a full `decode | format | encode` pipe.

**The marker is `¿`** (U+00BF). It has to be:

1. A single self-contained token (so `slide-encode` can find one Tile, destruct it, and let `remold_regrout` re-place Grout).
2. Not an identifier character (otherwise it would glue with adjacent keywords — an identifier-shaped marker next to `in` would tokenize as one Var, swallowing the keyword).
3. Not an operator character (otherwise `[1, ¿, 3]` would merge `¿,` into a single operator token).
4. Distinct from `?` (so user-typed empty-hole tiles round-trip distinct from Grout).

`¿` is wired in `Token.re` (excluded from the operator-class regex; recognized as a standalone atomic token) and `Form.re` (new `ImplicitHoleMarker` atomic form, same Convex molds as `ExplicitHole`).

**Why this matters beyond slides.** `TextRoundtrip` is independent of slides — any caller that wants a parser-canonical text representation of a `PersistentSegment` can use it. The slide CLI commands are now thin shims over it.

### New slide commands

- **`hazel slide-list`** — print the names of every slide statically linked into the binary.
- **`hazel slide-decode <NAME>`** — look up a slide by name and print its program as plaintext. Manual refractors are rendered with `^^probe(...)` / `^^statics(...)` trigger syntax; projectors as `^^fold(...)` (etc.); implicit holes (Grout) as `--implicit-hole CHAR` (default `¿`). Output is reparseable.
- **`hazel slide-encode --title T <TEXT-or-->`** — build a slide `.ml` from a title and plaintext program, emitting the `let out : ... PersistentSegment.t` module form. Use `-o` to write to a file (otherwise stdout). Refractors are rebuilt by the parser's `Triggers` module on insertion; `¿` markers are stripped via `Destruct` and Grout is re-inserted by remold/regrout.

The slides are looked up out of `Web.Init.documentation_slides : list((string, PersistentSegment.t))`, a new top-level binding extracted from the existing inline list in `Init.re`. No on-disk `.ml` parsing is needed.

### Supporting changes

- **`hazel format` now preserves refractors and the implicit-hole marker.** It parses to a zipper instead of just a segment and threads `z.refractors.manuals` through `Printer.of_segment`, and accepts `--implicit-hole CHAR` so the marker survives a `decode | format | encode` pipe.
- **`hazel analyze` gains `-W` / `--warnings`.** When set, warnings (currently just unused variables) are reported alongside errors using the same Rust-style source context.
- **Rust-style diagnostic rendering extracted to `src/CLI/Diagnostic.re`.** Shared by `analyze`'s error and warning paths; replaces ~140 lines of duplicated formatting in `Cli.re` with one helper plus two ~15-line wrappers.
- **`slide-decode` uses `print_string` instead of `print_endline`** so the CLI round-trip doesn't accumulate a trailing newline per pass.
- `src/CLI/README.md` updated for the new commands, the `--warnings` flag, the `¿` marker (with a "Why `¿`?" subsection laying out the four constraints), the corrected `format` / `analyze` behavior, and a note recommending `NODE_OPTIONS="--max-old-space-size=4096"` for `analyze` on larger inputs.

### Tests

- **`test/Test_TextRoundtrip.re`** (new):
  - *DocSlides* (`Slow`) — text fixed-point on every documentation slide, including projector slides (Tuples, Livelits). **50/50 pass.**
  - *TextReproducers* (`Quick`) — small inputs covering edge cases: holes in pattern / type positions, holes inside list literals and tuples, refractors wrapping a hole, nested refractors, backtick identifiers, multiple adjacent holes, comments adjacent to holes, string literals, projector-wrapped contents, `?((x))`-style shapes, etc. Runs on `-q`.
  - *Property* (`Slow`) — QCheck generates arbitrary `Exp.t`, renders to text, parses via the round-trip to establish a parser-canonical baseline, and checks the text fixed-point.
- **`test/Test_ReparseDocSlides.re`** — gets a cross-reference comment noting that it covers the *structural* segment equality that text fixed-point can't (text is a projection; multiple segment shapes can print the same).

### How they compose

A slide-level pretty-print is still the same three-command pipe — now with the implicit-hole marker carried through:

```sh
./hazel slide-decode "Probes" \
  | ./hazel format -w 60 - \
  | ./hazel slide-encode --title "Probes" - -o src/web/init/docs/Probes.ml
```

The same shape (`decode → transform → encode`) covers every workflow the earlier draft hard-coded into bespoke `slide-prettify` / `slide-update` / `slide-suppress-warnings` subcommands.

## Test plan

- [x] `make dev` builds cleanly
- [x] `./hazel slide-list` prints all expected names
- [x] `./hazel slide-decode "Probes"` emits plaintext with `^^probe(...)` markers intact and Grout rendered as `¿`
- [x] `./hazel slide-decode "Probes" | ./hazel format -w 60 -` preserves refractor count and `¿` markers
- [x] Full `decode | encode` round-trip on Basic Reference is byte-identical text and structurally identical `.ml` (Tile=585, Secondary=802, Grout=0 on both sides)
- [x] All 50 documentation slides text-fixed-point under `TextRoundtrip` (Test_TextRoundtrip.DocSlides)
- [x] 19 quick reproducer cases pass on `-q`
- [x] QCheck property holds across 50 generated Exps
- [x] `./hazel analyze -W` on a program with an unused let-binding prints a `warning: unused variable: <name>` diagnostic with location
- [x] Full `-q` test suite (2761 tests) passes

### DIFF
--- a/src/CLI/README.md
+++ b/src/CLI/README.md
@@ -5,7 +5,7 @@ The Hazel CLI is a command-line interface for working with Hazel programs. It pr
 ## Features
 
 - **Run Hazel Programs**: Execute Hazel programs and print their evaluated results.
-- **Format Hazel Code**: Reconstruct Hazel code from its abstract syntax tree (AST), using tylr to ensure syntactic correctness while removing original whitespace and comments.
+- **Format Hazel Code**: Reconstruct Hazel code from its abstract syntax tree (AST), using tylr to ensure syntactic correctness. Comments and refractor trigger syntax are preserved; original whitespace is replaced with structured formatting.
 - **Static Analysis**: Perform static analysis on Hazel code to identify and report errors (with Rust-style source locations). Optionally also report warnings such as unused variables.
 - **Grade Submissions**: Grade an exported submission JSON and emit either the raw grading data as JSON (`grade-json`) or a human-readable text report (`grade-report`).
 - **Slide tooling**: List, decode, and encode the documentation slides linked into the binary, so they can be edited as plaintext and re-emitted as `.ml` modules.
@@ -45,14 +45,14 @@ To format a Hazel program, use the `format` command followed by the path to the 
 ```sh
 $ ./hazel format path/to/hazel_file.hz
 ```
-This command will reformat the Hazel code, inserting line breaks to fit within a target width (`-w`/`--width`, default 60). Comments are preserved, but original whitespace is replaced with structured formatting. Manual refractors written with `^^probe(...)` / `^^statics(...)` trigger syntax round-trip through the formatter. The formatter performs explicit hole insertion using `?` instead of grout.
+This command will reformat the Hazel code, inserting line breaks to fit within a target width (`-w`/`--width`, default 60). Comments are preserved, but original whitespace is replaced with structured formatting. Manual refractors written with `^^probe(...)` / `^^statics(...)` trigger syntax round-trip through the formatter. Implicit holes (Grout) are rendered using the marker character set by `--implicit-hole` (default `¿`); see [Slides — round-trip](#round-tripping-a-slide-through-the-formatter) for why a marker is needed.
 
 You can also use `-` instead of a file path to read from standard input. For example:
 
 ```sh
 $ echo "let  = 5 in  + 3" | ./hazel format -
-let ? = 5 in                            
-? + 3
+let ¿ = 5 in
+¿ + 3
 ```
 ### Static Analysis
 To perform static analysis on a Hazel program, use the `analyze` command followed by the path to the Hazel file. For example:
@@ -91,16 +91,18 @@ $ ./hazel slide-list
 Basic Reference
 Projectors
 ADTs
+Tuples
 ...
 Probes
 Livelits
-B2T2: example tables
+B2T2 / Datasheet
+B2T2 / Example Tables
 ...
 ```
 
 #### Decoding a slide to plaintext
 
-`slide-decode` prints a slide's program as plaintext. Manual refractors are rendered with `^^probe(...)` / `^^statics(...)` trigger syntax so the output is reparseable:
+`slide-decode` prints a slide's program as plaintext. Manual refractors are rendered with `^^probe(...)` / `^^statics(...)` trigger syntax, projectors as `^^fold(...)` (etc.), and implicit holes (Grout) as a marker token — `¿` by default, overridable with `--implicit-hole`. The marker character is what lets `slide-encode` recover Grout positions on the way back; see [round-trip](#round-tripping-a-slide-through-the-formatter) below for the reasoning. The output is reparseable:
 
 ```sh
 $ ./hazel slide-decode "Probes" | head -3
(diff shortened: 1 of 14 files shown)


cyrus- (changes requested):
- empty holes aren't round tripped. proposed solution from our meeting: use a separate character for implicit holes (settable via a parameter). add a pass to delete the implicit holes after parsing back. leave a note saying this might cause issues with a few forms like `(<hole>)` and `[<hole>]`

src/haz3lcore/zipper/TextRoundtrip.re
7h3kk1d, line 1: @disconcision virtual_grout probably obviates this but this is aiming to do the round-tripability assuming parsing is consistent

src/CLI/README.md
```diff
@@ -5,13 +5,24 @@ The Hazel CLI is a command-line interface for working with Hazel programs. It pr
...
+- **Static Analysis**: Perform static analysis on Hazel code to identify and report errors (with Rust-style source locations). Optionally also report warnings such as unused variables.
 - **Grade Submissions**: Grade an exported submission JSON and emit either the raw grading data as JSON (`grade-json`) or a human-readable text report (`grade-report`).
+- **Slide tooling**: List, decode, and encode the documentation slides linked into the binary, so they can be edited as plaintext and re-emitted as `.ml` modules.
 
 ## Usage
 The Hazel CLI can be invoked from the command line using the `hazel` script located in the root of the repository. The script accepts various commands and options to perform different tasks.
 
+### A note on memory
+
+The CLI is compiled with `js_of_ocaml` and runs under Node, which defaults to a fairly small old-space heap. On bigger inputs (e.g. the larger documentation slides) commands like `analyze` can blow the default heap with no helpful error. If you see a sudden silent failure, bump the heap:
```
cyrus-, line 18: change this in the script

cyrus- (changes requested):
slide-decode on basic reference is still producing `?` for the empty hole at the top, would expect this issue to be caught by the property-based test.

7h3kk1d:
@cyrus- was an issue with the slide itself. Should be fixed.

cyrus- (approved):
Looks good

PR #2291, opened 2026-05-12
Add Void built-in type (#2289)

### LINKED ISSUE #2289: better support for void
Internally, I think we have everything we need for a void type (nullary sum), but we need syntactic support for it. I would suggest just defining a built in type called `Void` where all patterns are redundant by definition (no cases).

7h3kk1d:
## Summary
- Closes #2289 by adding `Void` as a built-in type keyword that parses to `Sum([])` (the nullary sum).
- The internal void infrastructure already existed (`Typ.is_void`, the exhaustiveness/redundancy check at `Coverage.check_matrix`); the missing pieces were the surface keyword, the `MakeTerm` mapping, and unblocking the pretty-printer which previously `failwith`'d on empty sums.
- Adds a dedicated Explain This entry for `Void`, with an example of an absurd eliminator (`fun (v : Void) -> case v end`).

## Changes
- `Token.re` / `Ctx.re`: register `Void` as a base type keyword (kept the two in-sync lists in sync).
- `MakeTerm.re`: parse `Void` as `Sum([])`.
- `ExpToSegment.re`: pretty-print `Sum([])` as the text `Void` (replaces the `failwith("Empty Sums are not allowed")`).
- `ExplainThis.re`: dispatch empty `Sum` to a Void-specific entry ahead of the general sum arm.
- `TerminalTyp.re` + `ExplainThisForm.re`: new `VoidTyp` form/group and `VoidAbsurd` example.
- `Test_ExpToSegment.re`: two new tests — pretty-print round-trip and `MakeTerm` round-trip for `Void`.

## Test plan
- [x] `make dev` clean.
- [x] `make test-quick` — all 2625 tests pass, including the two new Void tests.
- [x] Manual: `Void` keyword recognized in type position; `let f : Void -> Int = fun (v : Void) -> case v end in ...` typechecks and renders without crashing the pretty-printer.
- [x] Manual: Explain This on a `Void` keyword shows the Void-specific text and the absurd-eliminator example; non-empty sums still show the generic sum doc.

### DIFF
--- a/src/language/statics/Coverage.re
+++ b/src/language/statics/Coverage.re
@@ -1082,14 +1082,16 @@ module CheckMatrix: CheckMatrix = {
     switch (col_tys) {
     | [] => failwith("Empty column types.")
     | [first_col_ty, ...rem_col_tys] =>
-      if (Typ.is_void(first_col_ty)) {
+      switch (Ctr.all_ctrs_of_typ(first_col_ty)) {
+      | Finite(map) when Ctr.Map.is_empty(map) =>
+        /* No constructors (e.g. Void): no value of this type can exist, so
+           every row is unreachable and the match is vacuously exhaustive. */
         {
           is_exhaustive: true,
           unseen_pattern: UnseenPatternList.empty,
           redundant_rows: List.init(List.length(m), i => i),
-        };
-      } else {
-        let all_ctrs = Ctr.all_ctrs_of_typ(first_col_ty);
+        }
+      | all_ctrs =>
         let Submatrices.{
           ctrs,
           first_col_exhaustive,
--- a/src/web/app/explainthis/ExplainThis.re
+++ b/src/web/app/explainthis/ExplainThis.re
@@ -2705,7 +2705,9 @@ let get_doc =
       default
     }
   | Some(InfoTyp({user_term: term, _} as typ_info)) =>
-    switch (bypass_parens_typ(term).term) {
+    let typ = bypass_parens_typ(term);
+    switch (typ.term) {
+    | _ when Typ.is_void(typ) => get_message(TerminalTyp.void)
     | Unknown(SynSwitch)
     | Unknown(Internal)
     | Unknown(Hole(EmptyHole)) => get_message(HoleTyp.empty_hole)
@@ -2967,7 +2969,7 @@ let get_doc =
       simple(
         "`DrvTPat` is the type of ALFA type patterns, used in binding positions within ALFA type abstractions.",
       )
-    }
+    };
   | Some(InfoTPat(info)) =>
     switch (info.user_term.term) {
     | Invalid(_) => simple("Type names must begin with a capital letter")
--- a/src/web/init/docs/ADTs.ml
+++ b/src/web/init/docs/ADTs.ml
@@ -720,7 +720,7 @@ let out : string * Haz3lcore.PersistentSegment.t =
          45))(sort Exp))))))(shards(0 1 2))(children(((Secondary((id \
          54e4c70a-cfe2-45d1-8424-857bcfa34d74)(content(Whitespace\" \
          \"))))(Tile((id \
-         3c4d6a4a-549f-44a1-8176-c86a257e11bc)(label(_absurd))(mold((out \
+         3c4d6a4a-549f-44a1-8176-c86a257e11bc)(label(absurd))(mold((out \
          Pat)(in_())(nibs(((shape Convex)(sort Pat))((shape Convex)(sort \
          Pat))))))(shards(0))(children())))(Tile((id \
          d08db6e9-e6df-4b6e-a85b-8abe69809698)(label(:))(mold((out \
@@ -792,7 +792,7 @@ let out : string * Haz3lcore.PersistentSegment.t =
          45))(sort Exp))))))(shards(0 1 2))(children(((Secondary((id \
          3c621334-8497-4fb7-93b1-64cb990cc3eb)(content(Whitespace\" \
          \"))))(Tile((id \
-         73c70482-be2b-474d-9b7f-6f248bbfc931)(label(_diverge))(mold((out \
+         73c70482-be2b-474d-9b7f-6f248bbfc931)(label(diverge))(mold((out \
          Pat)(in_())(nibs(((shape Convex)(sort Pat))((shape Convex)(sort \
          Pat))))))(shards(0))(children())))(Tile((id \
          08f0b112-d2ad-41f2-b226-d3ad6c2fc599)(label(:))(mold((out \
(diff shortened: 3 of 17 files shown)


src/web/app/explainthis/ExplainThis.re
```diff
@@ -2931,6 +2931,8 @@ let get_doc =
           ),
         TerminalTyp.var(v),
       )
+    | Sum(sm) when ConstructorMap.is_empty(sm) =>
+      get_message(TerminalTyp.void)
```
7h3kk1d: Could this instead use `Typ.is_void`?

7h3kk1d:
Failing tests were from unrelated failures resolved in https://github.com/hazelgrove/hazel/pull/2281

cyrus- (changes requested):
1. Needs some more testing -- I'd suggest at least that the following has type `Int`:

```
let diverge : () -> Void = 
  fun () -> diverge()
in
let absurd : Void -> Int = 
  fun v -> case v end
in 
in absurd(diverge())
```

2. Mention Void in the language reference and ADTs slides briefly.

7h3kk1d:
> 1. Needs some more testing -- I'd suggest at least that the following has type `Int`:
> 
> 
> ```
> let diverge : () -> Void = 
>   fun () -> diverge()
> in
> let absurd : Void -> Int = 
>   fun v -> case v end
> in 
> in absurd(diverge())
> ```

Adding the unit test now but this does seem to have type `Int`. Did you notice a bug just commenting for a test?

The `case v end` has `elab_syn_ty` `?`. We could also make this `Int` but then we probably want to add the type a case statement is for onto it. @Negabinary do you have any thoughts on this?

src/web/init/docs/ADTs.ml
```diff
@@ -3,1098 +3,1985 @@ let out : string * Haz3lcore.PersistentSegment.t =
...
+        \      + Ok(Exp)\n\
+         in\n\n\
+         # A sum type with no constructors is the built-in Void type. #\n\
+         # Since Void has no inhabitants, an eliminator on a Void value can #\n\
+         # return any type -- here, any Result -- without ever producing one. #\n\
+         let _absurd: Void -> Result =\n\
+        \      fun v -> case v end\n\
+         in\n\n\
+         # Since Void has no inhabitants, any complete expression of type Void #\n\
+         # must fail to terminate -- there is no Void value to produce. #\n\
```
7h3kk1d, line 1960: I don't know how we feel about saying it like this. In Hazel it still does terminate but it can be indet. I can say fails to produce a value instead? @cyrus- let me know if you have thoughts or want other prose/examples.

Negabinary:
> > 1. Needs some more testing -- I'd suggest at least that the following has type `Int`:
> > 
> > ```
> > let diverge : () -> Void = 
> >   fun () -> diverge()
> > in
> > let absurd : Void -> Int = 
> >   fun v -> case v end
> > in 
> > in absurd(diverge())
> > ```
> 
> Adding the unit test now but this does seem to have type `Int`. Did you notice a bug just commenting for a test?
> 
> The `case v end` has `elab_syn_ty` `?`. We could also make this `Int` but then we probably want to add the type a case statement is for onto it. @Negabinary do you have any thoughts on this?

This feels equivalent to the empty list situation - options are either add a type to the case statement, or wrap it with an ascription in elaboration I guess?

I'm not too against adding the type here

src/web/init/docs/ADTs.ml
cyrus-: If its a complete expression it can't be indet so the comment is accurate.

cyrus-:
> > > 1. Needs some more testing -- I'd suggest at least that the following has type `Int`:
> > > 
> > > ```
> > > let diverge : () -> Void = 
> > >   fun () -> diverge()
> > > in
> > > let absurd : Void -> Int = 
> > >   fun v -> case v end
> > > in 
> > > in absurd(diverge())
> > > ```
> > 
> > 
> > Adding the unit test now but this does seem to have type `Int`. Did you notice a bug just commenting for a test?
> > The `case v end` has `elab_syn_ty` `?`. We could also make this `Int` but then we probably want to add the type a case statement is for onto it. @Negabinary do you have any thoughts on this?
> 
> This feels equivalent to the empty list situation - options are either add a type to the case statement, or wrap it with an ascription in elaboration I guess?
> 
> I'm not too against adding the type here

There should be a property to guide us here. I'm not sure what the intended meaning of elab_syn_ty is -- is it the type synthesized by the elaboration, inclusive of ascriptions added? How does this work with empty lists, for example?

src/web/init/docs/ADTs.ml
7h3kk1d: I'm interpreting that as expressions with marks are "incomplete". So `1 : Void` would not be complete.

cyrus-: Right, "incomplete" in the sense defined in the POPL 2017 paper so both empty and non-empty holes (i.e. marks) make something incomplete.

7h3kk1d:
> There should be a property to guide us here. I'm not sure what the intended meaning of elab_syn_ty is -- is it the type synthesized by the elaboration, inclusive of ascriptions added? How does this work with empty lists, for example?

I think elab_syn_ty is always [?] for the empty list and we aren't storing the Typ.t on the ListLit. That matches what the CursorInspector shows for the empty list in [] : [Int] where the first type shownhere is the elab_syn_ty:

Screenshot 2026-05-14 at 11 13 28 AM

elab_syn_ty is a per-node concept — roughly node.elab_syn_ty = Typ.syn(node.elab_term) locally. Any ascription inserted during elaboration lives on the parent's elab_term, not on the child's info entry, so a child node's elab_syn_ty never reflects an ascription its parent inserted around it. So the elab_syn_ty of a node does not include the ascription, because ascriptions are only added in places where the TDOS ascription transitions would not automatically propagate ascriptions — for example in:
```hazel
# We use the type meet for if expressions so we need to add an ascription to the right branch even though there is no literal one in the program #
if true then 1 else ?
```
or in
```hazel
# We give this list a type of `[Int]` but we want the elab_syn_ty of the last element to be `?` #
[1, 2, "" : ?]
```

To me that suggests empty list and empty case should probably behave differently from the general ascription path, they're more like complete expressions that are polymorphic in what they can produce, rather than treating them like they contain holes. One option is to add a Typ.t to their constructors so the synthesized type is recorded on the form itself. I don't think we want the CursorInspector to show holes for an empty list of a known type either so that not supports the idea.



edit: I didn't say it but the current behavior is consistent with how we handle the empty list but I'm saying that probably isn't ideal.

7h3kk1d:
add holes to the end of the slides and make the variables not underscored

src/language/statics/Coverage.re
```diff
@@ -1082,14 +1082,16 @@ module CheckMatrix: CheckMatrix = {
...
+           every row is unreachable and the match is vacuously exhaustive. */
         {
           is_exhaustive: true,
           unseen_pattern: UnseenPatternList.empty,
           redundant_rows: List.init(List.length(m), i => i),
-        };
-      } else {
-        let all_ctrs = Ctr.all_ctrs_of_typ(first_col_ty);
+        }
+      | all_ctrs =>
```
7h3kk1d, line 1094: @aasmart I switched the implementation of this so that it's using the constructors as opposed to `Typ.is_void` so we don't have to walk the type to see if it's a `rec a -> Void` or something like that. Let me know if that seems fine with you.

cyrus- (approved):

Issue #1593, opened 2025-04-07
Can't select text normally on any page elements


disconcision:
As revealed when debugging an issue from #1464. 

Regardless of user-select and pointer-event properties, DOM text cannot be drag-selected. Oddly, if these properties are appropriately set, you can select single words, for example by two-finger-clicking on touch, but you can't drag-select ranges.

This seems to have been an issue for a while, we just haven't noticed it. It looks to stem from the introduction of the clipboard shim. Disabling the clipboard shim focus action (the Attr.on_focus handler in Page.re) eliminates the issue, but also prevents copying code from code editors.

I haven't been able to find a working solution. Probably because (a) I don't get how exactly the focus event is screwing up dragging and (b) I don't fully grasp what's load-bearing with the clipboard-shim in the first place. I tried moving the on_focus handler more local to code editors (adding it to CodeEditable instead), but the focus event doesn't seem to fire if I put it here.

disconcision:
@cyrus- any thoughts?

cyrus-:
Hmm, not sure why the clipboard shim on focus handler would be preventing dragging....

disconcision:
@cyrus- as best as i can tell, the shim gaining focus interrupts the drag

PR #2306, opened 2026-05-14
fix: Cmd+C/V/X in Firefox after editor focus

### LINKED ISSUE #2314: Copying code causes editor to lose focus
Pressing Ctrl+C causes focus loss (caret disappears, insert actions don't work). Same for Ctrl+X. Paste still seems to work when in that state though (despite the focus being apparently gone)

7h3kk1d:
## Summary

Firefox copy/paste shortcuts (Cmd+C, Cmd+V, Cmd+X) silently failed inside the editor after clicking into a cell.

Root cause: PR #2198 (keyboard architecture refactor) decentralized focus into per-editor `tabindex(0)` divs. Firefox only dispatches native `copy`/`paste`/`cut` events to **editable** elements (input/textarea/contenteditable). Once focus landed on a `.code-editor` div, the page-level `Attr.on_copy`/`on_paste`/`on_cut` handlers in `Page.re` never fired, so the clipboard was never written. Chrome is more permissive and fires the events on any focused tabindex element, which is why we didn't notice sooner.

## Changes

- **`src/util/JsUtil.re`** — add `write_clipboard` / `read_clipboard` using the async Clipboard API (`navigator.clipboard.writeText` / `readText`). Feature-detected with a fallback to the existing `execCommand`-based `copy` for browsers without the API.
- **`src/web/app/editors/code/CodeEditable.re`** — handle Cmd/Ctrl + C/X/V directly in the active editor's `Key.handler`. The handler serializes the current zipper selection with `Printer.of_segment` (refractor-aware, same as the old `Page.View.copy`), updates the segment cache when there are no refractors, and writes to the clipboard directly. Cmd+X additionally dispatches `Destruct(Right)`. Cmd+V calls `navigator.clipboard.readText()` and schedules `Perform(Paste(...))` on resolve via `Bonsai.Effect.Expert.handle`.
- **`src/web/app/Page.re`** — ~~narrow the page-level `Attr.on_copy` to only hijack copy when the event target is inside a `.code-editor`...~~ Per review (4eb3b559ea): page-level `on_copy`/`on_cut`/`on_paste` removed entirely, along with the `is_input_field` / `selection_has_refractors` / `copy` helpers that only supported them. Editor-level handlers in `CodeEditable` are now the single clipboard path; sidebar / inputs / agent messages fall through to the browser's default DOM-selection copy.
- **`src/util/JsUtil.re`** — drop now-unused `has_ancestor_class`.

~~The page-level `on_copy`/`on_cut`/`on_paste` handlers remain as a fallback for cases where focus is on `#page` rather than an editor (e.g., before any editor click).~~

## Related

- PR #2198 — the keyboard refactor that triggered the regression.
- Issue #1593 — drag-select interference from the clipboard shim's `on_focus`. Not fixed here, but decoupling editor clipboard from native events opens the door to softening that focus call in a follow-up.
- Fixes #2314.

## Test plan

Manual (no e2e harness).

In **Firefox**:
- [x] Load page, click into an editor cell, keyboard-select a term, Cmd+C, paste into another app → selection text on clipboard.
- [x] Cmd+X → text removed from editor and on clipboard.
- [x] Copy text from another app, click into editor, Cmd+V → text inserted.
- [x] Repeat after blur/focus cycle (click outside, back in) — keeps working.
- [x] Select text in the sidebar, Cmd+C → sidebar text on clipboard (not editor content).
- [x] Cmd+C with refractors in selection → expanded `^^refractor_name(...)` form lands on clipboard.
- [x] Focus an input field (title/module/prompt/hint inputs, API key), Cmd+C/X/V → normal text input behavior, no editor action.
- [x] Fresh load, Cmd+C before clicking any editor → no crash; browser default copy.

In **Chrome**:
- [x] Same scenarios above — no regression.

cc @Negabinary

### DIFF
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -731,20 +731,18 @@ module View = {
         Effect.Ignore;
       }),
       Attr.on_copy(evt => {
+        /* Only hijack copy when the target lives inside a code editor —
+           editor-level Cmd+C now handles its own clipboard write, but the
+           native event still bubbles in cases where focus is on #page or
+           the editor's tabindex div without our key handler having run.
+           For sidebar / system-message / agent-message / inputs, fall
+           through to the browser's default DOM-selection copy. */
         let target = Js.Opt.to_option(evt##.target);
         switch (target) {
         | Some(el) =>
-          let elId = Js.Opt.to_option(Js.Unsafe.coerce(el)##.id);
-          if (is_input_field(elId)) {
-            ();
-          } else {
-            let el = Js.Unsafe.coerce(el);
-            if (JsUtil.has_ancestor_class(el, "system-message")
-                || JsUtil.has_ancestor_class(el, "agent-message")) {
-              ();
-            } else {
-              copy(cursor);
-            };
+          let el = Js.Unsafe.coerce(el);
+          if (JsUtil.has_ancestor_class(el, "code-editor")) {
+            copy(cursor);
           };
         | None => ()
         };
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -569,61 +569,7 @@ module Selection = {
 };
 
 module View = {
-  let is_input_field = (elId: option(string)) => {
-    switch (elId) {
-    | Some("title-input-box")
-    | Some("module-name-input")
-    | Some("prompt-input-box")
-    | Some("test-required-input")
-    | Some("point-max-input")
-    | Some("agent-api-key-input") => true
-    | Some(id) when String.starts_with(~prefix="hint-input", id) => true
-    | Some(id) when String.starts_with(~prefix="syntax-hint-input", id) =>
-      true
-    | Some(id) when String.starts_with(~prefix="impl-hint-input", id) => true
-    | _ => false
-    };
-  };
-
-  let selection_has_refractors =
-      (
-        refractors: Haz3lcore.Zipper.Refractor.t,
-        selection: Haz3lcore.Segment.t,
-      )
-      : bool =>
-    if (List.is_empty(refractors.manuals)) {
-      false;
-    } else {
-      let ids = Haz3lcore.Segment.ids(selection);
-      List.exists(
-        id =>
-          List.exists(((id2, _)) => Id.equal(id, id2), refractors.manuals),
-        ids,
-      );
-    };
-
-  let copy = (cursor: Cursor.cursor(Editors.Update.t)): unit => {
-    let str = (cursor.selected_text |> Option.value(~default=() => ""))();
-    let should_set =
-      switch (cursor.editor, cursor.selection) {
-      | (Some(editor), Some(selection)) =>
-        /* If the selection contains refractors, we forgo the segment cache
-         * for the sake of preserving refractors in the copy via expanding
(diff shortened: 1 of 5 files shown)


Negabinary (approved):
lgtm

cyrus-:
@7h3kk1d this seems to work and fix #2314, anything remaining to do here?

7h3kk1d:
@cyrus- just that I'm not positive how it works

src/web/app/editors/code/CodeEditable.re
```diff
@@ -776,6 +776,44 @@ module View = {
         );
       } else {
         let z = model.editor.state.zipper;
+        /* Editor-level clipboard helpers. Bypass the page-level
+           on_copy/on_paste path because Firefox refuses to dispatch
```
disconcision, line 786: I think we simply shouldn't have page level on_copy / on_paste. after @Negabinary's changes? tagging to confirm. I think this PR might be bypassing a system that it should be replacing instead.

src/web/app/editors/code/CodeEditable.re
```diff
@@ -807,8 +845,75 @@ module View = {
                 && z.relatives.ancestors == []
                 && snd(Siblings.neighbors(z.relatives.siblings)) == None =>
             Effect.Many([Effect.Prevent_default, escape(Right)])
+          /* 2. Cmd/Ctrl + C/X/V handled here rather than via the page
```
disconcision, line 854: presumably these exist elsewhere like on Page. we should probably get rid of those, or if they're load-bearing, have a polyfill or whatever at a single location. i don't think we need this functionality to exist at two different DOM locations.

src/web/app/Page.re
```diff
@@ -731,20 +731,18 @@ module View = {
         Effect.Ignore;
       }),
       Attr.on_copy(evt => {
```
disconcision: can this just be removed now?

disconcision (changes requested):
"The page-level on_copy/on_cut/on_paste handlers remain as a fallback for cases where focus is on #page rather than an editor (e.g., before any editor click)."

Do we actually need these page level handlers now? @Negabinary? If not let's get rid of them

src/web/app/editors/code/CodeEditable.re
7h3kk1d: Done in 4eb3b559ea — page-level handlers and their helpers removed; editor-level handlers are now the single path.

src/web/app/editors/code/CodeEditable.re
7h3kk1d: Yes — `Page.re` was the only other location for actual clipboard handling, removed in 4eb3b559ea. The remaining `on_copy`/`on_cut`/`on_paste` references in the tree (`TextAreaProj`, AgentView) are just `Stop_propagation` shields against the old page handler; happy to strip those too in a follow-up if you want.

src/web/app/Page.re
7h3kk1d: Yes — removed in 4eb3b559ea.

cyrus- (approved):

disconcision (approved):
lgtm

PR #2307, opened 2026-05-18
Reparenthesization in Stepper

nskh:
Working towards hiding left-associativity when evaluating arithemetic and other math in the stepper, introduces a "reparenthesize" step that wraps sections of the AST in parens so it can be evaluated. Exposed as either "Step here" or "Parenthesize" depending on what the user has highlighted in the Stepper.

Builds on #1890 and ideally should be merged into `dev` after that PR lands.

Screenshot 2026-06-01 at 5 03 28 PM
Screenshot 2026-06-01 at 5 04 18 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,4 +70,3 @@ plans/
 
 # Cursor IDE (local)
 .cursor/
-AGENTS.md
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,4 @@ plans/
 
 # Cursor IDE (local)
 .cursor/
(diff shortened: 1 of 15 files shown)


nskh:
Failing tests now pass.

Negabinary (commented):
If I put 

(fun x -> x)(1) * 5

in the stepper and select it, it crashes

Also I wonder if it might be helpful to be able to unparenthesize as well as parenthesize?

Implementation looks reasonable

PR #2309, opened 2026-05-19
Add `explore` keyword to invoke an exploratory stepper

nskh:
Adds an `explore ... in` keyword that invokes a new stepper, like `theorem`, but for non-booleans. The explore body itself doesn't evaluate to anything. However, unknown variables are assumed to be "free" and of their assumed type in the stepper, suppressing warnings and supporting algebraic variables. 

Screenshot 2026-05-19 at 1 21 28 AM

Screenshot 2026-06-01 at 6 17 46 PM


Negabinary:
Looks good to me - only thing I thought worth flagging was that it looks like all the explores always come after all the theorems. Moving all the theorems to the bottom is kind of a "temporary fix" (for a long range definition of temporary) anyway so maybe it's fine.

nskh:
Fixed bug where a `theorem` prover removed an `explore` stepper.

src/haz3lcore/derived/CachedStatics.re
```diff
@@ -147,11 +147,27 @@ let init =
...
+  let statics =
+    init_from_term(
+      ~settings,
+      ~ctx?,
+      ~is_dynamic_term,
+      ~ana?,
+      ~probe_ids,
+      term,
+    );
+  is_complete
```
cyrus-: what's happening here?

src/web/app/explainthis/ExplainThis.re
```diff
@@ -1987,6 +1987,7 @@ let get_doc =
             ),
           TheoremExp.tests,
         );
+      | Explore(_) => default
```
cyrus-: add some docs

cyrus- (changes requested):

src/haz3lcore/derived/CachedStatics.re
nskh: removed

src/web/app/explainthis/ExplainThis.re
nskh: added in latest update.

nskh:
Updated to address comments, base on `dev`, and clean up code.

PR #2315, opened 2026-05-20
Add Hazel build info to top bar of editor

CyrusD123:
Displays the git information for the current Hazel build as an unobtrusive, always-visible indicator in the top-right of the top bar, plus in the page title, so it's clear which build you're looking at when juggling multiple branches/tabs. Originally written by @7h3kk1d and developed by @CyrusD123; design revised per review discussion (the earlier git-icon slide-out panel was heavier than this information warrants).

## What it shows

`branch@short-sha[-ahead][-dirty] · #PR`, in a color only slightly darker than the top-bar background — legible if you look for it, invisible if you don't.

- **branch** links to the branch on GitHub; **short-sha** links to the commit (links always present; an unpushed commit's link starts working after push).
- **`-ahead`** marks commits not yet pushed to the upstream branch; **`-dirty`** marks uncommitted local changes (addressing the request to distinguish these).
- **`· #NNNN`** appears when the branch has an open PR, linked to it. Detected at build time via the `gh` CLI (`scripts/build-info.sh`) — an *optional* dependency: if `gh` is missing, unauthenticated, offline, or there's no open PR, the segment is simply omitted. Since the `BuildInfo.ml` rule re-runs on every rebuild and the `gh` call costs ~0.4s, the lookup is cached per branch in `.git/hazel-pr-cache` (steady-state rebuilds pay ~0.06s, same as the git-only rule); a `pre-push` hook deletes the cache since the answer most often changes around pushes. The hook activates once the `core.hooksPath` installation from #2233 lands; until then (or without `gh`), staleness is bounded by branch switches, and deleting the cache file forces a re-query.
- The **page title** becomes `hazel · branch@sha[-suffix]`, except on a clean, pushed `dev` build where it stays plain `hazel`.

## Selectability

The label is drag-selectable for copying to a collaborator — you can start the selection anywhere in the top bar's empty space and sweep across. This took some spelunking: mousedown over non-focusable page regions moves focus to `#page`, whose `on_blur` handler re-focuses the clipboard shim mid-gesture, destroying the nascent selection (the sidebar avoids this only because it has a `tabindex`). Fixes: `tabindex="-1"` on the indicator so it swallows focus like the sidebar; exemptions in the `#page` copy/cut handlers so Cmd+C copies the selected label rather than the editor selection (and Cmd+X doesn't delete editor content); sidebar-style `user-select` descendant rules; non-draggable anchors; and the label kept as a single inline span (selection across flex items is unreliable). The same tabindex + exemption treatment could make other UI text selectable in future PRs.

## Removed

The bottom-bar git-icon slide-out panel, its `build_info_expanded` setting, the git icon asset, and the associated CSS.

## Test plan
- [x] Indicator renders with branch/sha links; `-ahead`/`-dirty` suffixes appear and disappear as commits are made/pushed (observed live during development)
- [x] Open-PR detection: `BuildInfo.ml` generated with `Some 2315` + URL on this branch; `None` fallback verified with `gh` off PATH
- [x] PR cache lifecycle: first build writes `.git/hazel-pr-cache`; cache hit serves the PR with no network (verified with `gh` off PATH); `pre-push` hook invalidates; post-invalidation build without `gh` degrades to `None`
- [x] Drag-selection from empty bar space, glyphs (both directions), padding, and bar edge — verified via automated CDP tests in headless Chrome
- [x] Copy event over a build-info selection passes through natively (clipboard shim untouched); editor copy/cut behavior unchanged elsewhere
- [x] Page title: `hazel · branch@sha-suffix` on this branch; plain `hazel` on clean dev
- [ ] Reviewer: sanity-check on a machine without `gh` installed

https://claude.ai/code/session_01RXiW9jhHvYfwpwBT15MCwW


7h3kk1d:
My ideas: It would be nice if the branch was linkable separate to the commit so you can get to the branch url in github or the specific commit.


@disconcision's original notes to me: 

> i feel like it maybe doesn't need to show all the time? like maybe it's normally just a git icon, and hovering over it slides out to the above? or if you want it for screenshots maybe make it click icon to toggle to full. i don't think that long term we want to burn an entire corner on just this, so icon plus fold out feels like a good stub to me for suggesting a future menu or whatever. also radius is off, use the one from the gamma icon. width equal to the sidebar menu width. or just have the icon w/o outline, either works

I think that on `dev` we probably don't want this visible all the time so making the toggle change andrew suggested sounds good. On other branches/commits I would probably make it default show so you know somewhere other than the url what version you're on.

src/web/app/inspector/CursorInspector.re
```diff
@@ -1075,18 +1075,45 @@ let inspector_view = (~globals: Globals.t, ci): Node.t =>
...
 
+let build_info_view = () => {
+  let label =
+    BuildInfo.branch
+    ++ "@"
+    ++ BuildInfo.commit_short
+    ++ (BuildInfo.dirty ? "-dirty" : "");
+  let inner = text(label);
+  let body =
+    BuildInfo.dirty || BuildInfo.ahead
```
7h3kk1d: I think we should consider doing another class/symbol to represent that this is ahead rather than just using `dirty` since it's helpful to have a link to an existing commit on github even if there's local changes.

I'm also somewhat ambivalent of removing the link if the commit isn't on github since it only happens in dev workflows and the link starts working after you push. I could see it either way though.

CyrusD123:
The build info now slides out from a git icon on the bottom right, animated and styled to be similar to the cursor inspector:

Screenshot 2026-05-24 at 11 09 11 PM

Screenshot 2026-05-24 at 11 08 51 PM

The branch and commit are also linked separately now. In the case where a commit is local but not on GitHub yet, I'm still using the "dirty" CSS styling (though the commit hash doesn't have the "-dirty" suffix). If you prefer that we just not handle that case I'm pretty ambivalent as well. Or maybe we could link to the most recent commit on GitHub as well as show the current commit hash?

7h3kk1d:
I like the look of the bar now. It's a little confusing that if you click on it it doesn't go away because you're still hovering it. I wonder if we can make it go away and just reappear when you hover again. @disconcision might have more styling opinions though.

Sidenote: this also links to the local branch name on github which could not be the upstream branch name. I don't think many people are doing this so I'm not really concerned but it could be something we try to handle.

cyrus-:
I think the UI is a bit heavyweight for a relatively unimportant piece of information most of the time. How about just putting it on the top right of the top bar, always visible but in a color that is only slightly darker than the background?

cyrus-:
make it show up on `dev` in the same color

PR #2325, opened 2026-05-28
Force statics refresh on probe-id change

### LINKED ISSUE #2276: Auto-probe doesn't work in exercise mode

### LINKED ISSUE #2323: Probe values don't appear until follow-up edits
Neither (a) placing a manual probe nor (b) moving to a definition with autoprobe on shows sample values. Instead, the empty-set icon is shown until a substansive edit is made. Afterwards, the probe values update on edits as expected. This is true in all modes.

disconcision:
## Summary

Closes #2323.
(manual probes (and autoprobe placements) showed the empty-set indicator until the next edit instead of showing samples immediately.)

## Root cause

`StaticsDebounce` defers statics for 225ms on any `is_edited` action. `CachedStatics.with_targets` refreshes `statics.targets` from the new refractors each cycle but leaves the per-id `probe_targets` in `info_map` stale. The eval_info_map sent to the worker therefore has empty `probe_targets` at the top-level expression, so `IncrEval.reuse_check` sees the cached entry's `prev_probe_targets` match the current (stale) `info.probe_targets` and reuses the old state slice, which does not contain a sample for the newly placed probe.

The probe only updates on the next edit because the debounce timer eventually fires `RefreshStatics` (or another edit forces statics), recomputing `info_map` correctly.

## Fix

`CodeWithStatics.Update.calculate` now compares the zipper's current probe ids to the cached `statics.targets` keys. If they differ, it forces a full `CachedStatics.init` regardless of debounce state. Normal keystrokes (no probe change) still debounce as before.

Closes #2276

### DIFF
--- a/src/web/app/editors/Editors.re
+++ b/src/web/app/editors/Editors.re
@@ -323,6 +323,7 @@ module Update = {
         TutorialsMode.Update.calculate(
           ~schedule_action=a => schedule_action(Tutorial(a)),
           ~settings,
+          ~autoprobe_mode,
           ~is_edited,
           m,
         ),
@@ -332,6 +333,7 @@ module Update = {
         ExercisesMode.Update.calculate(
           ~schedule_action=a => schedule_action(Exercises(a)),
           ~settings,
+          ~autoprobe_mode,
           ~is_edited,
           m,
         ),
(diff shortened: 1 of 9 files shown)


disconcision:
Now part of #2363

PR #2336, opened 2026-06-04
Revert "fix: bump conf-pkg-config lock pin to 5 to unbreak CI"

7h3kk1d:
Reverts hazelgrove/hazel#2333

They revert we revert


disconcision (approved):
once more unto the breach

PR #2339, opened 2026-06-10
Evaluator streaming

Negabinary:
This pr adds streaming to the evaluator. This means results come through and are displayed to the user as the web worker is running. 

https://github.com/user-attachments/assets/8e5871d2-6e4a-4904-9b96-7ea778171992

I've also done a lot of work cleaning up the evaluator - this may create some conflicts, but I'm hoping it'll be a lot easier to work with. 

This pr commits to making evaluator state write-only (module steps), I think that's fine because we're a pure language.

We also switch the web worker to being always alive, instead of the current regime where we spin up a new web worker for each evaluation.


Negabinary:
**Guide for reviewing this pr**

**Helpful Background for Reviewing**

A key change we've made here is to the protocol for discussing back and forth between the webworker and the main thread. You can read these protocol specifications in the web worker server file.

The new protocol is as follows:

1. Main thread sends over new expression + previous evaluation cache
2. Worker performs a "reuse pass" over the whole expression to work out what files can be re-used
3. Worker sends an acknowledgement to the main thread, with a list of ids the worker plans to re-evaluate
4. Main thread draws these ids as shaded so the user can see what's evaluating
5. every 5000 steps, the worker pauses and sends back any top-level ids that have completed, and the current partially complete step
6. when the main thread receives updates, it adds it to its own previous evaluation cache.
7. repeat upon new edits

WorkerServer.re is mostly GPT-generated but should give you an overview of how this works.

Because this changes the way the webworker works, I've tried my hardest but wouldn't be surprised if there are weird browser states you can get into here. A functional review where you try ot break it might be good.

In addition I've done some focussed human cleanup on EvaluatorState.re and Evaluator.re, these should be significantly more readable now, and I would encourage you to have a read through and see if they match your understanding of how the Evaluator should work.

I tried to keep my commits clean so reading through them should also give you a sense for what I've added.

PR #2341, opened 2026-06-12
Add Cut/Copy/Paste to the editor context menu

7h3kk1d:
## Summary

Adds a clipboard section (Cut / Copy / Paste) to the editor's right-click context menu, with platform-appropriate shortcut chips (⌘X/⌘C/⌘V on Mac, Ctrl+X/C/V elsewhere).

- **Cut / Copy** are grayed out when the selection is empty. They write the system clipboard using the same helpers as the existing Cmd/Ctrl+C/X keyboard path (including the segment-cache fast-paste optimization), which were hoisted out of the key handler so both paths share them.
- **Paste** is always available. ~~Menu items carry an `Action.t` payload, so the menu's Paste uses a placeholder `Paste("")`; the new `perform_from_menu` inject intercepts it, reads the system clipboard asynchronously, and dispatches the real `Paste(text)` when it arrives, closing the menu immediately.~~
  **Revised** ([review](https://github.com/hazelgrove/hazel/pull/2341#discussion_r3476994230)): rows carry `ContextMenu.command = Perform(Action.t) | PasteFromClipboard` instead of a bare `Action.t`, so no placeholder action exists at all. `action_item` wraps its argument in `Perform` — every row builder is unchanged — and the Paste row carries `PasteFromClipboard`, which `perform_from_menu` turns into the real `Paste(text)` once the async clipboard read lands, closing the menu immediately.
  Paste is never grayed out, because the web can't report whether the clipboard is empty short of reading it, and that read is permission-gated and async (Chrome prompts once per origin, Firefox and Safari on every read) — testing it on menu-open would prompt on every right-click. Google Docs and VS Code for the Web leave Paste enabled for the same reason and find out at click time.
- **Right-click now preserves selections.** Previously right-click always dispatched `Move(Point(...))`, collapsing the selection before the menu opened — which would have made Copy useless. A new `click_in_selection` hit test keeps the selection when the click lands inside its measured extent; clicks outside still move the caret as before.
- Added `.disabled` styling for context-menu rows (dimmed, no hover highlight) and bumped the menu height estimate used for open-direction flipping, since the menu gained three permanent rows.

Known approximation: the hit test uses the selection's first/last piece measurements, so for char-level selections that split a token the "inside" region can be slightly wider than the highlighted text.

## Testing

- `make dev` builds clean.
- Manual: select code → right-click inside the selection → Cut/Copy act on it; right-click elsewhere → caret moves, Cut/Copy disabled, Paste inserts clipboard contents. Keyboard menu navigation (arrows/Enter) drives the same actions.
- The `command`-type revision above still wants a manual re-check — same flows, different dispatch path.

### DIFF
--- a/src/web/app/editors/code/CodeEditable.re
+++ b/src/web/app/editors/code/CodeEditable.re
@@ -576,34 +576,40 @@ module View = {
       /* Cache for paste reuse only when nothing was trimmed: a trimmed
          sub-token string must re-parse on paste, not round-trip to the
          full segment. */
-      if (str == full && !selection_has_refractors(z.refractors, segment)) {
-        Haz3lcore.Parser.set_segment_cache(Some(segment), str);
-      };
-      JsUtil.write_clipboard(str);
+      let cache_for_paste =
+        str == full && !selection_has_refractors(z.refractors, segment)
+          ? Effect.of_sync_fun(
+              () => Haz3lcore.Parser.set_segment_cache(Some(segment), str),
+              (),
+            )
+          : Effect.Ignore;
+      Effect.Many([cache_for_paste, JsUtil.write_clipboard(str)]);
     };
     let paste_from_clipboard = () =>
-      JsUtil.read_clipboard(text => {
-        let action =
-          Haz3lcore.Action.Paste(Util.StringUtil.trim_leading(text));
-        Bonsai.Effect.Expert.handle(inject(Perform(action)));
-      });
+      Effect.bind(JsUtil.read_clipboard(), ~f=text =>
+        inject(
+          Perform(
+            Haz3lcore.Action.Paste(Util.StringUtil.trim_leading(text)),
+          ),
+        )
+      );
     /* Inject for context-menu rows. Clipboard rows need view-layer side
        effects the core can't perform: Copy/Cut write the system clipboard
        before dispatch, and PasteFromClipboard starts an async read whose
        result is dispatched as the real Paste, closing the menu
-       immediately. Runs at event-firing time (see Menu.pointerdown_attr),
-       like the keyboard clipboard handlers below. */
+       immediately. Both are Effects, so the clipboard is touched when the
+       row fires rather than when its Effect is built. */
     let perform_from_menu = (c: ContextMenu.command): Ui_effect.t(unit) =>
       switch (c) {
       | Perform(Copy) =>
-        copy_selection();
-        inject(Perform(Copy));
+        Effect.Many([copy_selection(), inject(Perform(Copy))])
       | Perform(Cut) =>
-        copy_selection();
-        inject(Perform(Cut));
+        Effect.Many([copy_selection(), inject(Perform(Cut))])
       | PasteFromClipboard =>
-        paste_from_clipboard();
-        inject(ContextMenu(ContextMenu.Model.Close));
+        Effect.Many([
+          paste_from_clipboard(),
+          inject(ContextMenu(ContextMenu.Model.Close)),
+        ])
       | Perform(a) => inject(Perform(a))
       };
     /* Sync document-level listeners (click-outside + keyboard) for the
@@ -995,8 +1001,11 @@ module View = {
               alt: Up,
               _,
             } =>
-            copy_selection();
-            Effect.Many([Effect.Prevent_default, Effect.Stop_propagation]);
+            Effect.Many([
+              copy_selection(),
+              Effect.Prevent_default,
+              Effect.Stop_propagation,
+            ])
           | {
               key: D("x" | "X"),
               sys: Mac,
@@ -1015,12 +1024,12 @@ module View = {
               alt: Up,
               _,
             } =>
-            copy_selection();
             Effect.Many([
+              copy_selection(),
               Effect.Prevent_default,
               Effect.Stop_propagation,
               inject(Perform(Destruct(Right))),
-            ]);
+            ])
           | {
               key: D("v" | "V"),
               sys: Mac,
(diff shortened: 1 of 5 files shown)


7h3kk1d:
Not sure if we want to wait for #2199 or not

src/web/app/editors/code/ContextMenu.re
```diff
@@ -192,6 +195,29 @@ let introduce_data =
...
+      "Cut",
+      Action.Cut,
+    ),
+    action_item(
+      ~shortcut=Shortcuts.copy(),
+      ~enabled=has_selection,
+      "Copy",
+      Action.Copy,
+    ),
+    action_item(~shortcut=Shortcuts.paste(), "Paste", Action.Paste("")),
```
7h3kk1d, line 217: Why does paste have an empty string here? We should disable if clipboard is empty

7h3kk1d:
Stop context menu for spanning lines

src/web/app/editors/code/ContextMenu.re
7h3kk1d: Both halves addressed in 0f0e93398a.

**The empty string is gone.** Menu rows no longer carry a bare `Action.t`:

```reason
type command =
  | Perform(Action.t)
  | PasteFromClipboard;
```

`action_item` wraps its argument in `Perform`, so no row builder changed, and the Paste row carries `PasteFromClipboard`. `CodeEditable.perform_from_menu` is what turns that into a real `Paste(text)` once the async read lands — the placeholder payload existed only because the row's type demanded an `Action.t` before the text was knowable.

**On disabling it:** I left Paste enabled. The web can't report whether the clipboard is empty without actually reading it, and that read is permission-gated and async — Chrome prompts once per origin, Firefox and Safari on every read. Checking on menu-open would mean a clipboard-permission prompt on every right-click. (`navigator.permissions.query({name: 'clipboard-read'})` only reports permission state, not contents.)

That's also what rich web editors settle on: Google Docs and VS Code for the Web keep Paste enabled and find out at click time. Docs falls back to a "press ⌘V" dialog when the read is refused; we have no equivalent to offer, since our ⌘V path goes through the same API. Desktop apps (Word, Finder) do gray it out — native clipboard access is synchronous and ungated, which has no web equivalent.

Cut/Copy still gray out on an empty selection, which *is* synchronously knowable. The reasoning is recorded in the `clipboard_data` comment so it doesn't get re-litigated later.

src/web/app/editors/code/CodeEditable.re
```diff
@@ -543,6 +543,69 @@ module View = {
...
+      if (str == full && !selection_has_refractors(z.refractors, segment)) {
+        Haz3lcore.Parser.set_segment_cache(Some(segment), str);
+      };
+      JsUtil.write_clipboard(str);
+    };
+    let paste_from_clipboard = () =>
+      JsUtil.read_clipboard(text => {
+        let action =
+          Haz3lcore.Action.Paste(Util.StringUtil.trim_leading(text));
+        Bonsai.Effect.Expert.handle(inject(Perform(action)));
```
Negabinary: Scares me - Bonsai.Effect.Expert.handle shouldn't be this deep in the codebase, but I think read_clipboard requires it because of a pre-existing bad pattern so happy to leave it.

Negabinary (commented):
Generally happy, but would want @disconcision to sign off on the "Right-click now preserves selections" part before merging because he understands the implications of that decision better than I do

src/web/app/editors/code/CodeEditable.re
7h3kk1d: Yeah, this was just the moved code. But it would be nice to do another way.

disconcision: what is this exactly? previously i recall such things being handle via schedule_action?

7h3kk1d: In this pr it just got moved from https://github.com/hazelgrove/hazel/pull/2341/changes/BASE..0f0e93398a493f08750bf7c566b58fc75f74ffc5#diff-0f2adc2969e77b473081774030968b78f37c8a13b70f2c4e111463b137a1460dL890 . Let me see if I'm the one that added the original

7h3kk1d: Happened when I fixed the firefox paste https://github.com/hazelgrove/hazel/pull/2306

Looking to see how it worked before.

disconcision:
re: right-click preserving selections: this is fine.

from functional testing, it might be nice to have a case where, in the case where the focus is on a side of the selection which is not visible in the viewport, but the other side is visible in the viewport, to swap focus before putting up the context menu, to prevent an unnecessary scroll. i was encountering this when selecting a range of lines, where the last line finishes under the sidebar. this may be an epicycle though with the real issue being my decision to situate the context menu at the caret position... i sort of like that conceptually though and i don't think it's entirely sunk yet.

7h3kk1d:
> re: right-click preserving selections: this is fine.
> 
> from functional testing, it might be nice to have a case where, in the case where the focus is on a side of the selection which is not visible in the viewport, but the other side is visible in the viewport, to swap focus before putting up the context menu, to prevent an unnecessary scroll. i was encountering this when selecting a range of lines, where the last line finishes under the sidebar. this may be an epicycle though with the real issue being my decision to situate the context menu at the caret position... i sort of like that conceptually though and i don't think it's entirely sunk yet.

Is there any reason it is at the caret position? if you have many lines and are in the middle of the scroll it's going to scroll no matter what if we keep it adjacent to the caret. So I'm not sure what the best answer is.

disconcision:
@7h3kk1d not a strong reason. i like conceptually the caret being at a kind of canonical position for the action being performed, but yeah, it has the issue you suggest. i'm happy just leaving it as is for now, or seeing what it feels like with positional restriction relaxed entirely.

7h3kk1d:
@disconcision I'm sure claude can do the flipping of the caret but I'm more nervous about escaping the context menu then having my selection flipped so I'm probably going to leave it alone. If you feel strongly though I can add it in

src/web/app/editors/code/CodeEditable.re
7h3kk1d: Defined a new effect in https://github.com/hazelgrove/hazel/pull/2464

PR #2346, opened 2026-06-15
Automerge read/write projectors

dm0n3y:
Introduces a pair of Automerge projectors that can read from and write to Automerge documents.

The read projector pulls in an Automerge document specified by URL and converts the raw JSON into a Hazel JSON value that can then be manipulated programmatically within Hazel. It includes buttons for hot-reloading and, if not hot-reloading, manual refreshing. These are instantiated via `^^Automerge(Null)` (where `Null` is the initial JSON value getting projected over and will be replaced by whatever Automerge JSON doc is loaded in).
image

The write projector takes an Automerge url and pushes whatever JSON value it is projecting over to that url. These are instantiated via `^^AutomergeWriteBack(<json>)` where `<json>` is whatever Hazel JSON value you want to write back - in the screenshots below, this is just the empty `Null` value.
image
image

Validating behavior here requires creating Automerge documents through tiny.patchwork.inkandswitch.com (or some other Automerge doc interface), copying and pasting Automerge URLs into the read/write projectors, and assessing the JSON values read into Hazel and written back to the Automerge doc.

### DIFF
--- a/.changetours/2346-automerge-read-write-projectors.changetour.md
+++ b/.changetours/2346-automerge-read-write-projectors.changetour.md
@@ -1,0 +1,2248 @@
+---
+schemaVersion: 1
+prNumber: 2346
+prOwner: hazelgrove
+prRepo: hazel
+baseSha: 63261b75c3d5ea05b55c78f7e8c1823cf14dc09e
+headSha: d30895c9f65258697f913904e1e9d32796e81783
+---
+# Automerge read/write projectors
+
+This PR introduces a pair of Automerge projectors that can read from and write to remotely stored Automerge documents.
+
+Validating behavior here requires creating Automerge documents through tiny.patchwork.inkandswitch.com (or some other Automerge doc interface), copying and pasting Automerge URLs into the read/write projectors, and assessing the JSON values read into Hazel and written back to the Automerge doc.
+
+The read projector loads in an Automerge document specified by URL and converts the raw JSON into a Hazel JSON value that can then be manipulated programmatically within Hazel. It includes buttons for hot-reloading and, if not hot-reloading, manual refreshing. These are instantiated by typing ^^Automerge(Null) (where Null is the initial JSON value getting projected over and will be replaced by whatever Automerge JSON doc is loaded in).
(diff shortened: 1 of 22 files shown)


dm0n3y:
This PR introduces a pair of Automerge projectors that can read from and write to remotely stored Automerge documents. 

The read projector (`AutomergeProj.re`) loads in an Automerge document specified by URL and converts the raw JSON into a Hazel JSON value that can then be manipulated programmatically within Hazel. It includes buttons for hot-reloading and, if not hot-reloading, manual refreshing. These are instantiated by typing `^^Automerge(Null)` (where `Null` is the initial JSON value getting projected over and will be replaced by whatever Automerge JSON doc is loaded in).
image

The write projector (`AutomergeWriteBackProj.re`) takes an Automerge URL and pushes whatever JSON value it is projecting over to that url. These are instantiated by typing `^^AutomergeWriteBack(<json>)` where `<json>` is whatever Hazel JSON value you want to write back - in the screenshots below, this is just the empty `Null` value.
image
image

Validating behavior here requires creating Automerge documents through tiny.patchwork.inkandswitch.com (or some other Automerge doc interface), copying and pasting Automerge URLs into the read/write projectors, and assessing the JSON values read into Hazel and written back to the Automerge doc.

The standard approach to incorporating projector values into the Hazel runtime is to expand projector values into the underlying editor representation (ie `Segment.t`) and then let the regular parsing/typechecking/evaluating pipeline do its thing. This approach is inefficient in that typically the projector values start in AST form (ie `Term.t`) and get serialized to `Segment` only to be parsed back into `Term`. This is fine for small values but there's noticeable latency when parsing edit states encoding real Automerge docs. Hence, this PR introduces a bypass system by which the parser can skip parsing these large values. Projectors may store their expanded values (in already parsed `Term` form) in a table separate from the edit state, while leaving behind holes in their place. Later, the parser encounters these holes and puts the stored projector values in their place when constructing the overall term. See `ProjectorCore.re` and search for uses of `ProjectorCore.get_bypass` and `ProjectorCore.set_bypass` in `MakeTerm.re` and the projector implementations.

`Automerge.re` provides helpers for connecting to remote Automerge repos, reading and writing Automerge docs by URL, and translating between raw and Hazel JSON values.

PR #2321, opened 2026-05-26
Multiline probe drawer display and dockable sample context menu

disconcision:
### Probe drawer mode

New per-probe display mode: pretty-printed multi-line samples shown below the code line, full editor-pane width, in addition to the existing single-line inline display.

- **Toggle** via `Cmd/Ctrl+↓` (open) / `Cmd/Ctrl+↑` (close), or the chevron in the row-level nav-bar control
- **Per-probe state**: different probes can be in different modes simultaneously; persisted across reloads via the projector model
- **Per-sample wrap width**: `Shift+drag` on a sample resizes its pretty-print wrap target; falls back to `settings.drawer.width = 80` for unset samples
- **Layout**: uses projector framework's `Tab(n)` placeholder to reserve rows below the line. `DrawerHeight.compute` measures the pretty-printed rows via `Measured.of_segment`. Capped at 15 rows; taller samples scroll vertically inside a wrapper.
- **Nav-bar control**: consolidated single SVG control pad with three regions (left arrow / right arrow / down chevron drawer-toggle). Lives at end-of-line in both inline and drawer modes (controls stay put when toggling).
- **Two-stage Esc** in drawer mode: first Esc collapses the drawer (stays focused), second Esc blurs.

### Misc fixes / polish

- `PrettySegment.decompose_first_element` for comma-separated segments — fixes layout for tuples in pretty mode
- Reactive caret-anchor scroll compensation (replaces the older predicted-delta approach)
- Sample-focus anchor scroll compensation on Left/Right navigation
- Refractor `Tab(n)` no longer consumed by internal linebreaks
- Refractor model-change detection via cache reference equality (avoids spurious recomputes)
- Lowered test-checkmark z-index to code level (was overlapping the probe sample dropdown)
- Bumped `ProjectorView.offside_offset` from 4 → 5 → 4 (settled at 4 once the consolidated nav-bar reduced horizontal footprint)

## Test plan

- [ ] Drawer mode: focus a probe, press `Cmd+↓` to open / `Cmd+↑` to close. Confirm focus persists across toggles for keyboard shortcuts.
- [ ] Drawer mode: `Shift+drag` a sample horizontally to resize wrap width.
- [ ] Drawer mode with a very large value (e.g. a deeply nested record): confirm capping at 15 rows + vertical scroll inside the drawer.
- [ ] Nav-bar control: hover each of the three regions, confirm individual red hover; click each to confirm correct action.
[...]

PR #2363, opened 2026-06-29
Probes IV

### LINKED ISSUE #2067: Linebreaks before Tab projectors on the same line break text rendering
Image

### LINKED ISSUE #2217: Reinstate viewport culling for projectors
Reinstate viewport culling for projectors. This is implemented, just commented out.

- Major issue: Projectors are universally culled in tutorial/exercise model.
- Minor issue: When probes show up there appears to be some sort of animation/shimmer maybe involving the depth indicators.

### LINKED ISSUE #2276: Auto-probe doesn't work in exercise mode

### LINKED ISSUE #2323: Probe values don't appear until follow-up edits

disconcision:
Probes features and fixes from the study branch. Subsumes #2321 and #2325 (now closed).

Closes #2067
Closes #2323
Closes #2276
Closes #2217

## Features

- **Drawer mode.** Expand a probe's sample into a multi-line, pretty-printed block below the code line (`Cmd/Ctrl+↓` open / `Cmd/Ctrl+↑` close, or the row's chevron; Esc exits). Per-probe state, capped at 15 rows (scrolls beyond). Built on a new `below` projector view slot reserving rows via `Tab(n)`.
- **Enhanced Auto-probe** Auto-probe now probes top-level expressions as well ad defiitions. Auto-probe now has an `All` mode. Modes are `Off | Caret | All` (was on/off). _Caret_ probes the top-level definition under the cursor; _All_ probes the whole program, one probe per source row (a single root multiprobe). `Cmd/Ctrl+P` cycles Off↔All; the sidebar's segmented control also reaches Caret.
- **"Simple" color scheme (new default).** Two colors: focal-call green vs. off-focus washed out green. replacing the old Hybrid default. Use the toggle in the sidebar to access the old default.

## Fixes (pre-existing behavior)

- **Probe samples no longer stale until the next edit:** Placing a probe (or autoprobe) showed the ∅ indicator until the next edit, because the statics debounce left `probe_targets` stale and `IncrEval.reuse_check` reused a sampleless slice. Now a probe-id change forces a full statics refresh. (Also makes auto-probe work in exercise mode.)
- **Step-into for indirect calls.** Extends the existing step-into to higher-order calls, partial applications, and `let f x = …` sugar, by carrying a resolved call frame (`fn_def_id`) through evaluator state into each sample.
- **More stable Sample-ids to prevent collisions on deep recursion** — fixes double-indication, twin resize handles, and broken scroll anchors via collision-resistant ids.
- **Accurate Tab projector/refractor decoration offset** — Code/decoration row desync from out-of-order deferred-linebreak counting is fixed.
- **Projector action index crash** — a stale render-time index now drops the action instead of raising.
- Smaller: duplicate manual-refractor entry (assoc-list vs map disagreement), stuck pending probe cursor, a sibling-cleanup toggle that re-added forever, and pointer/drag guards.

## Modifications / polish

- **Sample-count indicator moved** from an overlay circle on the projector to an inline badge at the start of the samples row (hidden for a single sample).
- **Sample dropdown now uses right-/alt-click.** The sample dropdown (which shows environment variables and pin/step options) opens on right-click or alt-click (was hover-only), with click-outside / scroll dismissal. Hover activation proved to hyperactive.
- **Function-sugar probing.** Putting a multi-probe on `let f(args) = body` now probes the arguments on the header line, not just the body.
- **Sample-focus navigation now scrolls horizontally:** When using ←/→ move the focused sample it should scroll-follow it into view; this previously only worked vertically.
- Indeterminant samples e.g. holes no longer colored differently than value samples; this proved confusing to multiple people.
- Removed the dashed connector line between probes and samples; added visual noise doesn't pay for itself.
- Sidebar quick-reference updated for new interactions (alt-click for env, right-click for pin/step, an expand-probe row).
- Focus bar: "Clear all" also turns auto-probe off; the editor caret stays visible after a focus-bar / sidebar jump.
- Scroll stability for drawer refractors: drawer-height changes above the caret are prevented from shifting it via scroll trigger.
- Tuple pretty-printing breaks symmetrically
- The test pass/fail badge no longer floats above probe sample menu.

## Performance

- Viewport culling now re-enabled, but only for auto-probe All mode (infra already existed on dev but was inert): off-screen probe views are culled in the single-editor modes (Scratch / Documentation / Tutorial); Exercises stays unculled.
- Cull-before-build for probe views, memoized multi-probe expansion + placeholder calcs, and phys-equality gating on rebuilds.

## Late additions: cards + rich-view UX (2026-08-22)

- Cherry-picks Alexander's card RichProbe renderer (authorship preserved) (from `card-rich-probe`), forward-ported to this branch's registry API (`empty`, `drawer_rows`), with card/hand shape-matching intact.
- Rich views render **inside** the sample chip (green backing, inert content) rather than replacing the row — the chip keeps the dropdown (Hide), sample capture, dbl-click, and resize. The anchored modal is retired: small rich views (≤4 rows) embed in the row; taller ones auto-open the drawer on activation. Many mode shows one card/fan per sample.
- **Rich Views** toggle in the probes sidebar (default on): probes auto-render the first applicable renderer when it fits; double-click flips an individual probe back to text. Tables remain opt-in.
- Empty hands render a card-footprint silhouette (dotted outline, ∅) instead of nothing.
- Offside **stagger**: an after-display pass floors each display at the end of every code line it spans and first-fits it among displays from rows above, so multi-row content (card fans) overlaps neither program text nor other probes.


disconcision:
@Negabinary I broke out dynamics / incremental eval related fixes / changes below (claudeslop but edited for correctness), lmk if anything smells to you.

## Incremental-eval staleness fix (the ∅-until-next-edit bug)
- **`src/web/app/editors/code/CodeWithStatics.re`** — placing a probe (or autoprobe) showed the empty-set indicator until the next edit: the statics debounce left per-id `probe_targets` stale, so `IncrEval.reuse_check` matched the cached entry and reused a state slice with no sample for the new probe. Fix: force a full statics re-init when the zipper's probe ids differ from the cached targets (bypassing the debounce). Closes #2323/#2276.

## Eval state slice now carries call frames
- **`src/language/dynamics/state/StateSlice.re`** — the incremental-eval state slice's `app_data` is now `Id.Map.t(list((call_stack, elided_value, stack_frame)))` (was `app_args`, a 2-tuple with no frame); `diff_app_args` → `diff_app_data` retyped. Changes what's diffed / carried across incremental eval. This data was necessary for step-into to work properly for indirect function calls (eg function passed as param).
- **`src/language/dynamics/state/EvaluatorState.re`** — `app_args`→`app_data` rename + frame storage; `lookup_app_arg`→`lookup_app` merged (returns `(elided_value, stack_frame)`). `clear_transient` still wipes it before postMessage.

## Transition / Evaluator (step-into for indirect calls)
- **`src/language/dynamics/transition/Transition.re`** — step-into through a function **cast** now builds the inner `Ap` with `rewrap` (reuse the redex id) instead of `fresh`, so the call keeps identity across the cast; partial-application (`DeferredAp`) steps now emit `RecordStackFrame`; `get_fn_name`/`get_fn_def_id` made recursive through `Closure`/`DeferredAp`.
- **`src/language/dynamics/Evaluator.re`** — the probe-on-`Ap` path looks up `(args, frame)` via `lookup_app` and carries `~frame` into `Sample.mk`.

## Sample identity / ordering
- **`src/language/dynamics/Sample.re`** — sample-id collision hardening: `Hashtbl.hash_param(64, 256, (List.length(stack), syntax_id, stack))` (cheap discriminators first) replaces `Hashtbl.hash`, fixing collisions on deep recursive stacks sharing a prefix.

Negabinary:
> Incremental-eval staleness fix (the ∅-until-next-edit bug)

This is the right fix - I think I saw this on a separate pr 

> Eval state slice now carries call frames

Concerns me greatly and doesn't feel like it should be a thing, though it will be made redundant by #2339 anyway so I guess we can keep it to fix whatever Claude thought it was fixing with that

> step-into through a function cast now builds the inner Ap with rewrap

uh probably harmless but I wanna know why - and whether this suggests we need to document a new invariant in transition, notably it goes against the comment already in Transition.re

```
  /* Note[Matt]: For IDs, I'm currently using a fresh id
     if anything about the current node changes, if only its
     children change, we use rewrap */
```

Asking Claude this question reassured me a little that this might actually be just for one very specific circumstance.

> partial-application (DeferredAp) steps now emit RecordStackFrame

seems reasonable

disconcision:
@Negabinary: 

re frame in state slice: yeah it's a hack but needed to get the closure's fn_def_id, which i can only get when the fn steps, which isn't when i get the probe sample... added a comment saying it'll be deprecated by #2339

re cast rewrap: "if only its children change, we use rewrap" isn't that this case? the step is cast distribution. freshening the id of the ap here seems incorrect both for probes and for incremental eval (might falsely invalidate?)

.gitignore
```diff
@@ -18,6 +18,10 @@ Version_autogenerated.re
 
 # ocamlbuild working directory
 _build/
+# local dev-server staging dir (built bundle + copied statics).
+# Never commit: a broad `git add` swept 84MB bundles into history
```
7h3kk1d, line 22: Are these removed from the history at least?

7h3kk1d:
Rich probes/tables rendering is broken if you use the inline view. I'd also be okay to switch to something closer to the version I have in https://github.com/hazelgrove/hazel/tree/resizable-probe-modal which replaces the sample with the rich version if that's easier. Also the vertical scrolling for tall inline samples doesn't have any visual indicator that it's scrollable.


https://github.com/user-attachments/assets/f4f51691-58c4-43c2-b7ff-a6382a81ec15

.gitignore
disconcision: i want to run the repo surgery command by you first

disconcision:
@7h3kk1d both are fixed. in drawer mode, just the table display is used

src/web/app/editors/decoration/ProbeStagger.re
```diff
@@ -0,0 +1,120 @@
...
+  row: int,
+  origin_col: int,
+  el: Js.t(Dom_html.element),
+  w_px: float,
+  h_px: float,
+};
+
+let update = (~measured: Measured.t, ~font_metrics: FontMetrics.t): unit => {
+  let nodes =
+    Dom_html.document##querySelectorAll(Js.string(".offside-wrapper"));
```
disconcision: Correctness: this pass selects offside wrappers from the entire document, while Main passes measured from only Page.Update.get_editor. Scratch derivation slides, Tutorial, and Exercises can mount multiple CodeEditable cells, and row numbers are local to each cell. Wrappers in other cells therefore use the wrong line ends and collide through one global occupied set, so probes can be pushed unnecessarily or fail to clear their own code. Please run the pass per code container with that editor measurement, or attach an editor-scoped floor and identity to each wrapper.

src/web/RefractorShift.re
```diff
@@ -0,0 +1,80 @@
+/* Cause-driven compensation for refractor (probe-drawer) height changes above
+ * the caret: when a drawer above the locus grows/shrinks, scroll #main by the
+ * exact row delta so the caret stays put. Gated on refractor_rows ref
+ * identity (zero work on unrelated frames). Replaces the old symptom-driven
+ * CaretAnchor, which conflated drawer shifts with unrelated caret motion. */
+
+open Js_of_ocaml;
+open Haz3lcore;
+
+let prev_rows: ref(option(Id.Map.t(int))) = ref(None);
```
disconcision: Correctness: prev_rows is process-global, but update receives the one editor returned by Page.Update.get_editor. When the current scratch slide, tutorial or exercise, or mode changes, the new editor map is compared with the previous editor map; drawers already open above the new caret are treated as newly opened and #main is scrolled even though no height changed. In multi-cell derivation views this also never compensates the actually focused non-primary cell. Key this cache by editor or cell identity and reset the baseline when identity changes, or run it per editor.

src/web/app/editors/Editors.re
```diff
@@ -22,6 +22,14 @@ module Model = {
     | Tutorial(_) => "Tutorial"
     | Exercises(_) => "Exercises";
 
+  /* true only for single-active-editor modes; Exercises (multi-editor) is unculled */
+  let supports_viewport_culling: t => bool =
+    fun
+    | Scratch(_)
```
disconcision: Correctness: Scratch and Documentation are not always single-editor: a Drv scratchpad renders many CellEditor views, and Tutorial exercises also render multiple cells. visible_rows is one global range derived from document.querySelector(.code-container), which selects the first cell, and CodeEditable then applies that range to every cell. Visible probes in later cells can therefore be culled using the first cell local row range. Until bounds are editor-scoped, this predicate should return false for multi-cell content, at minimum Drv scratchpads and Tutorial.

disconcision (commented):
I reviewed head b90c310 against dev, including the adjacent probe and editor-layout code. The current fast checks, full tests, preview, and Codecov checks are green. I found three browser-side correctness issues that I would address before merge; the inline comments include the failure mode and a suggested direction.

Maintainability and structure:

- ProbeProj.re is now 2,544 lines and owns persisted projector state, mutable global display settings, sample caches, navigation/focus, menus and listeners, rich-renderer selection, and drawer rendering. Those are useful seams for ProbeSettings, ProbeNavigation, ProbeSampleMenu, ProbeRichView, and ProbeDrawer modules, leaving the projector implementation mostly as orchestration.
- ProbePerform.re is 1,424 lines and similarly combines target discovery, automatic-probe reconciliation, manual/multi-probe lifecycle, collision cleanup, focus, and step-into behavior. ProbeTargets, AutoProbePerform, and ProbeFocus would make the invariants and tests much easier to isolate.
- CardTypes/CardSyntax/CardView and MenuListener are good extractions in this branch; I would continue that pattern. In particular, ProbeProj.vacuous_value currently knows about CardSyntax to decide whether automatic rich rendering has evidence. That policy belongs on the RichProbe renderer interface as an auto-applicability/evidence predicate, so future list-shaped renderers do not add more domain checks to the generic probe module.
- ProbeStagger, RefractorShift, ScrollWidth, SampleAnchor, and FocusEffect are individually small, but together form a global mutable after-display subsystem. An editor-scoped layout-effects coordinator would give these caches an explicit owner and directly prevent the first two findings below.

Naming and comments:

- Settings.s and Settings.go hide whether code is reading state or dispatching an update; current and update/dispatch would be clearer, ideally behind a non-global interface.
- The strongest comments explain invariants, such as raw versus finalized sample maps, observation-span delegation, and agreement between reserved and measured drawer rows. Keep those close to the code.
- The long design-history comments are less durable. ProbeStagger links an implementation plan, AutoProbe describes Caret only as a top-level definition even though bare expressions are supported, and the RichProbe module comment ends mid-thought. I would trim these to current behavioral contracts and move the exegesis/history to maintained design docs.
- git diff --check reports trailing whitespace in hazel-programs/docs/reference/cards.hz on lines 8 and 38.

Overall, the card/view and listener extractions are heading in a good direction, but the new layout code needs editor identity to be a first-class input rather than relying on document-global selectors and refs.

PR #2370, opened 2026-07-01
Fix eval worker postMessage stack overflow on deep results (#2368)

### LINKED ISSUE #2368: 'DedicatedWorkerGlobalScope': Maximum call stack size exceeded
[long.txt](https://github.com/user-attachments/files/29554309/long.txt) is a hazel program that has a long list (~100 rows of labeled tuples) bound to a variable and just computes 3 + 1. On chrome this is failing due to a stack issue with the worker:

```
Uncaught RangeError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': Maximum call stack size exceeded
    at post_message (worker.ml:76:3)
    at caml_call1 (blackbox.ml:1:1)
    at on_request (WorkerServer.re:65:3)
    at caml_call1 (blackbox.ml:1:1)
    at js_handler (worker.ml:70:65)
    at caml_call_gen (+stdlib.js:27:23)
    at +jslib.js:390:15
+sys.js:96 Uncaught RangeError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': Maximum call stack size exceeded
    at post_message (worker.ml:76:3)
    at caml_call1 (blackbox.ml:1:1)
    at on_request (WorkerServer.re:65:3)
    at caml_call1 (blackbox.ml:1:1)
    at js_handler
```

7h3kk1d:
## Problem

Evaluating certain programs threw in the eval worker on **Chrome only**:

```
Uncaught RangeError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': Maximum call stack size exceeded
    at post_message (worker.ml)
    at on_request (WorkerServer.re)
```

Repro: a large list bound to a variable (e.g. ~100 rows of labeled tuples) even when the computed result is trivial, like `length(x)` or `3 + 1`. Does not reproduce in Firefox.

## Root cause

`WorkerServer.on_request` / `WorkerClient.request` posted the raw OCaml value graph across the worker boundary. `postMessage` hands it to the browser's structured-clone serializer, which in Chrome is **recursive** and overflows its (relatively small) stack on a deep — but finite, acyclic — payload. Firefox's limit is higher, so it tolerated the same payload.

## Fix

Encode the request/response with jsoo's **`Marshal`** (iterative → a flat string that structured clone copies without recursing), so a deep-but-finite payload crosses the boundary without hitting the clone serializer's recursion limit. This is the active encoding — `WorkerServer.Active`, a one-line alias `module Active = MarshalEncoding`.

The boundary stays typed `Worker.worker(request, response)` via an abstract `WorkerServer.ENCODING` module signature: only `encode`/`decode` can cross it and the two directions can't be swapped.

## Worker Messaging debug panel

To choose an encoding on real workloads, the diff keeps an **`encoding` variant** (`Direct | Marshal | Sexp`, with display strings and `all_of_encoding` ppx-derived — no hand-maintained name lists) behind the abstract `ENCODING` signature, plus a **"Worker Messaging" section in the Debug sidebar** — how the main thread talks to the eval Web Worker — that benchmarks each enabled encoding per request (encode / structuredClone / decode timing + size), correlated by request id.

- Metrics are typed: durations are `Core.Time_ns.Span`, sizes `Core.Byte_units` (each encoding reports its own size via the `ENCODING` signature). A stage that throws renders as `—` with the error on a tooltip, not a misleading `0`.
- **Defaults:** the section starts **collapsed** (so no benchmarking runs) and only **`marshal`** is enabled; the enabled set (`worker_encodings`) and per-encoding toggles persist in settings, so a slow encoding like `sexp` stays off.
- Benchmarking is gated on the panel being visible and synced once per cycle via `WorkerMetrics.sync` (the settings that gate it aren't reachable at the `WorkerClient.request` call site).
- The section lives in its own `WorkerMessagingSection.re` behind a small `DebugSection.S` interface, rendered by a generic driver in `DebugSidebar`. Swapping the active encoding is the one-line `module Active = …` alias.

## What we explored, then removed

Two additional encodings were benchmarked during the analysis and have since been **removed from the code** (preserved in git history on this branch). Documenting them here for the record:

- ~~**`columnar`** — was briefly the active protocol.~~ Packs the jsoo value graph into a handful of flat typed arrays (`flatten`/`unflatten`, custom JS via `Js.Unsafe`): a fixed set of `Float64Array`s regardless of depth, so structured clone never recurses *and* no intermediate string is built. Blocks are numbered by identity, so sharing/DAG structure is preserved. Benchmarks put it at **~2× faster than Marshal and fully depth-proof**. **Dropped** because it leaned on `Js.Unsafe`/`Obj.magic` and jsoo's block-representation internals for a payload where Marshal is already fast enough — Marshal's simplicity and safety won once the metrics settled the question. (added in `4e47e876fe`)
- **`array`** — converts only the top-level request/response list spine to an OCaml array. Everything *inside* each entry still crosses as its natural nested graph, so a payload that is deep inside an entry still overflows structured clone — i.e. it does **not** fix #2368. Kept only as a benchmark contrast to the columnar packing; removed.

## Testing

- `Test_WorkerServer` (4 tests): **Marshal is depth-proof** — a 20k-deep payload round-trips through `structuredClone`; and **every encoding is isomorphic** — one round-trip-equality test each for `direct` / `marshal` / `sexp`. (Deep failure of direct/sexp isn't asserted: raw `structuredClone` on a deep graph segfaults node uncatchably, whereas in-browser it throws a catchable `RangeError`.)
- `make dev` clean, no compiler warnings.
- **Chrome (manual):** the original `Marshal`-based fix (`a5c49701eb`) evaluated the repro correctly. Re-verification of the current active path + the Worker Messaging panel (durations/sizes render via `Span`/`Byte_units`; a failed `direct` shows `—`) is pending against the latest rebuild.

Closes #2368

### DIFF
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -406,10 +406,19 @@ module Update = {
 
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
-    /* Gate worker wire benchmarking on the debug panel (the flag isn't
-       reachable at the WorkerClient.request call sites, which see only
-       settings.core). This runs before the modes below queue any request. */
-    WireMetrics.enabled := model.globals.settings.show_debug_panel;
+    /* Gate worker wire benchmarking on the Wire Metrics panel being visible:
+       the debug panel is on AND its section is expanded. The flag isn't
+       reachable at the WorkerClient.request call sites (which see only
+       settings.core), so we sync it here, before the modes below queue any
+       request. Skipping it while collapsed avoids the (potentially large)
+       per-request measurement cost when nothing is watching. */
+    WireMetrics.enabled :=
+      model.globals.settings.show_debug_panel
+      && !
+           SidebarModel.Settings.is_debug_collapsed(
+             DebugSidebar.wire_metrics_title,
+             model.globals.settings.sidebar,
+           );
     let editors =
       Editors.Update.calculate(
         ~settings=
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -412,14 +412,15 @@ module Update = {
        settings.core), so we sync it here, before the modes below queue any
        request. Skipping it while collapsed avoids the (potentially large)
        per-request measurement cost when nothing is watching. */
-    WireMetrics.enabled :=
+    WorkerMetrics.enabled :=
       model.globals.settings.show_debug_panel
       && !
            SidebarModel.Settings.is_debug_collapsed(
              DebugSidebar.worker_messaging_title,
              model.globals.settings.sidebar,
            );
-    WireMetrics.disabled_wires := model.globals.settings.sidebar.wire_disabled;
+    WorkerMetrics.enabled_encodings :=
+      model.globals.settings.sidebar.worker_encodings;
     let editors =
       Editors.Update.calculate(
         ~settings=
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -406,9 +406,9 @@ module Update = {
 
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
-    /* Gate worker wire benchmarking on the Wire Metrics panel being visible:
-       the debug panel is on AND its section is expanded. The flag isn't
-       reachable at the WorkerClient.request call sites (which see only
+    /* Gate worker-messaging benchmarking on the Worker Messaging panel being
+       visible: the debug panel is on AND its section is expanded. The flag
+       isn't reachable at the WorkerClient.request call sites (which see only
        settings.core), so we sync it here, before the modes below queue any
        request. Skipping it while collapsed avoids the (potentially large)
        per-request measurement cost when nothing is watching. */
@@ -416,7 +416,7 @@ module Update = {
       model.globals.settings.show_debug_panel
       && !
            SidebarModel.Settings.is_debug_collapsed(
-             DebugSidebar.wire_metrics_title,
+             DebugSidebar.worker_messaging_title,
              model.globals.settings.sidebar,
            );
     WireMetrics.disabled_wires := model.globals.settings.sidebar.wire_disabled;
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -406,6 +406,10 @@ module Update = {
 
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
+    /* Gate worker wire benchmarking on the debug panel (the flag isn't
+       reachable at the WorkerClient.request call sites, which see only
+       settings.core). This runs before the modes below queue any request. */
+    WireMetrics.enabled := model.globals.settings.show_debug_panel;
     let editors =
       Editors.Update.calculate(
         ~settings=
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -406,12 +406,8 @@ module Update = {
 
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
-    /* Gate worker-messaging benchmarking on the Worker Messaging panel being
-       visible: the debug panel is on AND its section is expanded. The flag
-       isn't reachable at the WorkerClient.request call sites (which see only
-       settings.core), so we sync it here, before the modes below queue any
-       request. Skipping it while collapsed avoids the (potentially large)
-       per-request measurement cost when nothing is watching. */
+    /* Sync worker-messaging benchmark gating here (settings aren't reachable at
+       the WorkerClient.request call sites); only run when the panel is open. */
     WorkerMetrics.sync(
       ~enabled=
         model.globals.settings.show_debug_panel
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -419,6 +419,7 @@ module Update = {
              DebugSidebar.wire_metrics_title,
              model.globals.settings.sidebar,
            );
+    WireMetrics.disabled_wires := model.globals.settings.sidebar.wire_disabled;
     let editors =
       Editors.Update.calculate(
         ~settings=
--- a/src/web/app/Page.re
+++ b/src/web/app/Page.re
@@ -412,15 +412,16 @@ module Update = {
        settings.core), so we sync it here, before the modes below queue any
        request. Skipping it while collapsed avoids the (potentially large)
        per-request measurement cost when nothing is watching. */
-    WorkerMetrics.enabled :=
-      model.globals.settings.show_debug_panel
-      && !
-           SidebarModel.Settings.is_debug_collapsed(
-             DebugSidebar.worker_messaging_title,
-             model.globals.settings.sidebar,
-           );
-    WorkerMetrics.enabled_encodings :=
-      model.globals.settings.sidebar.worker_encodings;
+    WorkerMetrics.sync(
+      ~enabled=
+        model.globals.settings.show_debug_panel
+        && !
+             SidebarModel.Settings.is_debug_collapsed(
+               WorkerMessagingSection.title,
+               model.globals.settings.sidebar,
+             ),
+      ~encodings=model.globals.settings.sidebar.worker_encodings,
+    );
     let editors =
       Editors.Update.calculate(
         ~settings=
--- a/src/web/app/sidebar/DebugSidebar.re
+++ b/src/web/app/sidebar/DebugSidebar.re
@@ -718,8 +718,8 @@ let toggle_bar = (~globals: Globals.t): Node.t => {
 };
 
 /* ---- Worker Messaging: how the main thread talks to the eval Web Worker ----
-   Per-request benchmarks of the candidate wire protocols (WireMetrics) that
-   encode payloads crossing the main-thread <-> Web Worker boundary, populated
+   Per-request benchmarks of the candidate encodings (WorkerMetrics) that
+   pack payloads crossing the main-thread <-> Web Worker boundary, populated
    only while the debug panel is open. All requests share one table so columns
    line up across them; each request is separated by a bold `#N` group row
    (request) and a lighter `response` sub-row, a row per variant under each.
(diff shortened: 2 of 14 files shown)


disconcision:
am i reading right that this stringifies at the boundary? this is presumably quite slow. instead maybe we could just not use a linked list for sample etc, instead using something flat

7h3kk1d:
> am i reading right that this stringifies at the boundary? this is presumably quite slow. instead maybe we could just not use a linked list for sample etc, instead using something flat

That's correct. I'll performance test it and see if we can use something else.

disconcision:
i think we should try hard to avoid stringification, i did benchmark for probe stuff and it's considerably slower for large payloads. i think it's worth trying some kind of array or immutable vector for both this and more broadly. or collabing with matt's streaming stuff to pull things into the main thread on-demand as opposed to in bulk

7h3kk1d:
> i think we should try hard to avoid stringification, i did benchmark for probe stuff and it's considerably slower for large payloads. i think it's worth trying some kind of array or immutable vector for both this and more broadly. or collabing with matt's streaming stuff to pull things into the main thread on-demand as opposed to in bulk

Did you serialize using Marshal or with the sexp/show functions? I think no matter what is going to the web worker it's going to have to go through the structured clone which is what's hitting the stack overflow right now because the jsoo lists are hiearchically linked. 

I have a version that coerces everything in plain js to a columnar layout and it's not much faster because of the conversion. Let me see if I just store things as arrays.

disconcision:
the current version i think doesn't explicitly serialized? other than the structured clone itself. previously it was strings, which degrades badly with lots of samples. i was thinking we'd just have the results as an array or whatever; generate and keep them that way for consumption, rather than doing the conversion. but i might not be thinking that through enough?

7h3kk1d:
> the current version i think doesn't explicitly serialized? other than the structured clone itself. previously it was strings, which degrades badly with lots of samples. i was thinking we'd just have the results as an array or whatever; generate and keep them that way for consumption, rather than doing the conversion. but i might not be thinking that through enough?

Using arrays fixes the issue with the list being long popping the stack but if we have large js objects in side of the expressions for deep programs the structured cloning will still blow up as far as I can tell. If we want to avoid that I think we need a limited depth representation which involves sequentializing the objects which is what Marshal is trying to do it's just doing it as a bytestring. My understanding of what's currently happening is the expression is being sent to the web worker in my example program is a long list which is an ocaml linked list so when cloning the Exp.t it blows up.

update: I just switched to an array as opposed to a list and as expected the long program still blows up because the array is only size 1 and it's a deeply nested expression. Not sure if you had another representation you intended

7h3kk1d:
🤖 Benchmarked the serialization options to follow up on the stringification concern.

**Marshal vs sexp on realistic programs.** Parsed the `Exp.t` of all 50 documentation slides and round-tripped each 100× under node (same V8/serializer family as Chrome). Sexp here is the plain derived path (`Exp.sexp_of_t`/`t_of_sexp` + `Sexp.to_string`/`of_string`):

| | Marshal | sexp | sexp/Marshal |
|---|---|---|---|
| corpus round-trip time | 42.8 ms | 214.3 ms | **5.0×** |
| corpus serialized size | 1.14 MB | 3.46 MB | **3.0×** |

Per-slide ratios sit consistently in the 3.4–5.5× band. Sexp's cost is ~70% decode-side (`Sexp.of_string` re-parsing plus `t_of_sexp`). Separately from speed, the ppx sexp converters recurse once per AST level and stack-overflow somewhere between 1k and 10k of expression nesting (list spines are fine — Sexplib converts those tail-recursively), while jsoo's Marshal is iterative and depth-proof.

**At the worker boundary, we can skip strings entirely.** I prototyped a `Wire` layer that flattens the jsoo value graph into a few flat typed arrays (block lengths / packed fields / ref positions), so `postMessage`'s structured clone copies it as raw bytes with no recursion and no string is ever built. Same harness, medians of 5:

| payload | Marshal enc/clone/dec (total) | flatten enc/clone/dec (total) |
|---|---|---|
| 100k-deep nested exp | 92.5 / 0.8 / 53.7 (**147 ms**) | 39.6 / 4.7 / 12.6 (**57 ms**) |
| 50k-element list | 60.6 / 0.5 / 23.8 (**85 ms**) | 34.2 / 3.7 / 3.9 (**42 ms**) |

Posting the raw value graph directly overflows the clone serializer at these depths (RangeError in Chrome, hard-crashes node in our harness) — that's #2368.

**On "use an array instead of a list":** measured that too. Direct clone of an `array(Exp.t)` payload (50k elements) is 33 ms — statistically tied with flattening the same payload (34 ms), because structured clone's cost is per-object, not per-depth: each element is still ~6 nested blocks of wrappers. Arrays fix the *spine* depth (and compose nicely with flattening — fewer blocks, ~20% faster encode), but deeply nested ASTs still crash the direct path, so a boundary transform is needed regardless. The genuinely zero-copy win would be narrower: numeric sample/probe data in a `Bigarray`/typed array is a transferable `ArrayBuffer`.

So the ranking on these payloads: **flatten (~2× faster than Marshal, no strings) > Marshal (~5× faster than sexp) > sexp**, with sexp additionally depth-unsafe.

7h3kk1d:
wiremetrics
screenshot of the new sidebar panel to compare wire protocols

src/web/app/sidebar/DebugSidebar.re
```diff
@@ -717,6 +717,181 @@ let toggle_bar = (~globals: Globals.t): Node.t => {
...
+   line up across them; each request is separated by a bold `#N` group row
+   (request) and a lighter `response` sub-row, a row per variant under each.
+   Correlated by request id. */
+
+/* Adaptive precision so a slow variant's wide milliseconds (e.g. sexp at
+   ~17000ms) don't blow out the column: drop decimals as the value grows. */
+let ms_str = (f: float): string =>
+  f >= 100.
+    ? Printf.sprintf("%.0f", f)
+    : f >= 10. ? Printf.sprintf("%.1f", f) : Printf.sprintf("%.2f", f);
```
7h3kk1d: We should have a duration type or use something like https://ocaml.janestreet.com/ocaml-core/v0.12/doc/core/Core__Core_time_ns/Span/index.html

src/web/app/sidebar/DebugSidebar.re
```diff
@@ -717,6 +717,181 @@ let toggle_bar = (~globals: Globals.t): Node.t => {
   );
 };
 
+/* ---- Worker Messaging: how the main thread talks to the eval Web Worker ----
```
7h3kk1d: We should use this as an opportunity to have a rough interface for what constitutes a debug sidebar section with a module signature and put this one behind it and store it in another file.

src/web/app/sidebar/DebugSidebar.re
```diff
@@ -717,6 +717,181 @@ let toggle_bar = (~globals: Globals.t): Node.t => {
...
+  f >= 100.
+    ? Printf.sprintf("%.0f", f)
+    : f >= 10. ? Printf.sprintf("%.1f", f) : Printf.sprintf("%.2f", f);
+
+let size_str = (bytes: int): string =>
+  bytes >= 1024 * 1024
+    ? Printf.sprintf("%.1fM", float_of_int(bytes) /. 1024. /. 1024.)
+    : bytes >= 1024
+        ? Printf.sprintf("%.1fK", float_of_int(bytes) /. 1024.)
+        : string_of_int(bytes) ++ "B";
```
7h3kk1d: This should also be using another type or in a util file somewhere like https://ocaml.janestreet.com/ocaml-core/v0.12/doc/core_kernel/Core_kernel__/Byte_units/index.html and have a reusable format string

src/web/app/Page.re
```diff
@@ -406,6 +406,20 @@ module Update = {
...
+       settings.core), so we sync it here, before the modes below queue any
+       request. Skipping it while collapsed avoids the (potentially large)
+       per-request measurement cost when nothing is watching. */
+    WireMetrics.enabled :=
+      model.globals.settings.show_debug_panel
+      && !
+           SidebarModel.Settings.is_debug_collapsed(
+             DebugSidebar.worker_messaging_title,
+             model.globals.settings.sidebar,
+           );
```
7h3kk1d: Is there a better way/place to be doing this?

src/web/util/WorkerMetrics.re
```diff
@@ -0,0 +1,200 @@
...
+  Js.Unsafe.pure_js_expr("(function (x) { return structuredClone(x); })");
+let structured_clone: 'a. 'a => 'a =
+  x => Js.Unsafe.fun_call(clone_fn, [|Js.Unsafe.inject(x)|]);
+
+/* Approximate in-memory footprint of an arbitrary wire form (string, OCaml
+ * value graph, plain object, or typed arrays), by an iterative walk with a
+ * visited set so shared/DAG structure isn't double-counted. Numbers 8B,
+ * strings by length, typed arrays by byteLength, object/array headers ~8B per
+ * slot. Only a relative measure across variants, not an exact byte count. */
+let size_fn =
```
7h3kk1d: We should make this part of the encoding signature and each encoding can report it's own size

src/web/util/WorkerMetrics.re
```diff
@@ -0,0 +1,200 @@
...
+ * throw sets status to Failed and stops (later stages stay 0). */
+let measure:
+  'w 'a.
+  (~name: string, ~encode: unit => 'w, ~decode: 'w => 'a) => dir_metric
+ =
+  (~name, ~encode, ~decode) => {
+    let enc = ref(0.)
+    and cln = ref(0.)
+    and dec = ref(0.)
+    and sz = ref(0);
```
7h3kk1d: Don't report 0 if there's an exception instead return an option

test/Test_WorkerServer.re
7h3kk1d, line 1: These tests are too expensive and the comments a bit too long. We want to confirm marshal is depth proof. And that all the encoders are isomorphic. We don't need to try a bunch of programs for direct/sexp since they're not actually used.

src/web/app/sidebar/WorkerMessagingSection.re
```diff
@@ -0,0 +1,167 @@
...
+  switch (s) {
+  | None => dash
+  | Some(s) => Core.Time_ns.Span.to_string_hum(~decimals=2, s)
+  };
+
+let bytes_str = (b: option(Core.Byte_units.t)): string =>
+  switch (b) {
+  | None => dash
+  | Some(b) => Core.Byte_units.to_string_hum(b)
+  };
```
7h3kk1d: We can have a helper that takes an option(t) and a t -> string and does this once.

src/web/app/Page.re
```diff
@@ -406,6 +406,22 @@ module Update = {
 
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
+    /* Gate worker-messaging benchmarking on the Worker Messaging panel being
+       visible: the debug panel is on AND its section is expanded. The flag
+       isn't reachable at the WorkerClient.request call sites (which see only
+       settings.core), so we sync it here, before the modes below queue any
+       request. Skipping it while collapsed avoids the (potentially large)
+       per-request measurement cost when nothing is watching. */
```
7h3kk1d: Can we have a smaller comment

src/web/util/WorkerClient.re
```diff
@@ -4,10 +4,16 @@ open WorkerServer;
 let name = "worker.js"; // Worker file name
 let timeoutDuration = 20000; // Worker timeout in ms
 
-let initWorker: unit => Js.t(Worker.worker(Request.t, Response.t)) =
+/* Worker exchanges Active-encoding payloads (a flat string produced by jsoo's
+ * iterative Marshal, which structured clone copies without recursing), not live
+ * values, to dodge the structured-clone stack overflow on deep results (#2368;
+ * see WorkerServer.Active). Callers still deal in Request.t/Response.t. When the
+ * debug panel is on, WorkerMetrics benchmarks the other encodings locally on
+ * the side. */
```
7h3kk1d: Comment too long.

src/web/util/WorkerServer.re
7h3kk1d, line 1: Comments too verbose in this file.

disconcision (approved):
lgtm

Issue #2373, opened 2026-07-03
Glomming `(` onto an orphan `)` leaves stale grout; term collapses to a hole and arrow-move crashes


disconcision:
Type `)`, press left-arrow, type `(`. The `(` gloms with the orphan `)` into an atomic `()` token, but the convex grout accompanying the orphan closer is not regrouted away, leaving two adjacent convex pieces. Arrow-moving after the glom then crashes the editor.

Found by the roundtrip fuzzer on the completion-provenance branch, where canonical completion regrouts before parsing and masks the term-collapse symptom but the edit-state invariant break (and the crash) are upstream: the glom/insert path should regrout at the source.

PR #2374, opened 2026-07-03
Canonical completion + Shard/Lexeme Provenance

### LINKED ISSUE #1792: [()] has type [?]
If you type `[()]` it has type `[?]` instead of `[()]`

### LINKED ISSUE #2296: over-parenthesization of unit types in exptosegment
Image

### LINKED ISSUE #2381: Delete and retype the `l` of a non-first `let`: keyword never re-pairs with its orphaned `=`/`in`
In

```
let f = 1 in
let g = 2 in
f + g
```

delete the `l` of the second `let`, then retype it. The text looks right but the structure stays broken: the retyped `let` remains a lone shard-0 remnant and never reassociates with its orphaned `=`/`in` (two incomplete tiles). Doing the same to the first `let` works.

Two composed causes: `Segment.rescan` only matches singleton pieces against an open expectation frame, so the two-shard `=`/`in` remnant is unmatchable; and `Reassociate.flatten_and_repair` starts its ascent at depth 1, so at the top level (no ancestors, limit 0) it bails without ever trying the sibling scope that holds the match.

disconcision:
Closes #2381
Closes #1792
Closes #2296
Subsumes #2079
References #2373 (known-outstanding)

## Goal: round-tripping

This PR makes every editor-reachable state round-trip: the buffer's segment parses to a term, and printing that term (segment-to-segment, via ExpToSegment) reproduces the original (tile ids, labels, shards, secondary) up to a small quotient described below. This is supported by property-based testing. Previously many states didn't survive the trip: incomplete tiles were dumped or dropped, quotes around labels and leading + in sum types disappeared, literal spellings normalized (007 -> 7, 4.0 -> 4.), unknown infix operators created multiholes which lost the operator string, etc.

This enables transforms on terms (eg refactorings, agent edits) while perserving all previously ephemeral editor state. The stacked refactorings PR is the first consumer.

The mechanism has two halves: canonical completion turns every incomplete segment into a well-formed one deterministically, and a small amount of lexical provenance recorded on terms lets the segment printer reconstruct exactly what the user had typed.

## Canonical completion

This assigns a canonical completion to every incomplete segment. Unlike the previous dump logic, canonical completions do not depend on caret position. However, ergonomic completions can't depend simply on
term structure: when inserting a new form, e.g. a let expression above existing code, term structure alone is ambiguous as to whether the new code wraps the code below or sits purely above it. The caret-based heuristic used previously is (a) only useful for incomplete forms in the caret segment, and (b) necessitates either re-completing on caret movement (expensive and potentially confusing) or a hidden history dependence (confusing).

The new heuristic is based on partitioning, meaning which following content an incomplete form absorbs. This is decided by indentation: after a linebreak, content at same-or-lesser indentation than the incomplete tile's line ends the partition. This subsumes the old, somewhat ugly "reset after a blank line" heuristic and is the intent-based discrimination discussed next.

Where the missing delimiters go — placement follows an evidence hierarchy (details in plans/completion-heuristics.md): a leftover prefix token (`i` for a deleted `in`, `-` for `->`) or the deletion's own junction debris pins the delimiter exactly where it was; only evidence-free states fall back to the structural defaults below. The engine completes ONE tile per pass (strongest evidence first, weak ties innermost-first) and recurses on the materialized result, so the suggestion set is a replay script: applying everything shown reproduces the computed result by construction, and jointly-broken forms (delete a case's `end` AND its let's `in`) restore compatibly.
- Missing closers are appended at the end of their partition, stopping before a span-final trailing operator when that strictly reduces synthesized holes (`test 1 == 1 end;` restores with the `end` before the `;`). A single-line form's closer glues back over a trailing linebreak (`let x = 1` + blank line + `2` takes its `in` on the definition line); a multiline form keeps its closer on its own line (a broken multiline case-def restores `end in` exactly where it was).
- Missing openers are placed at the start of their closer's left-operand span in the partition skel — the maximal span the completed form can absorb without crossing enclosing structure — bounded by line-walls: a deleted mid-program `let`/`if`/`case` head stays on its own line instead of absorbing the definitions above. So let a = 1,2] completes to let a = [1,2] in ?, not by hoisting [ above the let. Same-position ties open the later closer outermost: 1) + 2) -> ((1) + 2). Opener spans are also clamped by slot sort: a restored head lands at the nearest span that fits its interior slot (the second `let` of `let a = 1 in let b = 2 in b`, deleted inline, restores at `b` — a complete let can't be a pattern), while Exp-slot wrappers like a deleted `(` keep the maximal wrap.
- Missing middle delimiters (let x in 2, if true else 2; targeted put-down can strand an interior delimiter in the backpack) can't be appended like the others, because reassembly requires shard order. The tile is instead completed in place: each original child stays in the slot opened by its original left shard, and newly created slots get a hole, so let x in 2 completes to let x = ? in 2. The synthesized grout ids derive deterministically from surrounding tile ids, so completion is stable across recomputation.

## Indentation now user-owned

A non-obvious part of this design is that the indentation heuristic requires making indentation user-owned. This isn't
essential for round-tripping itself, but it's what makes canonical completion ergonomic, as well as resolveing a problem of its own.

Indentation around incomplete code has the same wrapping-uncertainty as completion, and on dev it leads to indentation thrashing: typing on one line causes lines below to indent and unindent repeatedly. So the row-indentation system is removed in favor of 'traditional' indentation: indentation is now simply regular whitespace (secondary), owned by the user.

Like conventional editors, we don't re-indent everything on every edit. Pressing enter inserts indentation after the new linebreak only, using the same indentation logic as before, so left-to-right entry gets the indentation you expect and nothing below the caret moves. **Cmd+S formats per a 4-level setting (none / indent / spaces / breaks; default respaces + re-indents); Cmd+Shift+S always pretty-prints.**

There is a nut menu setting, on by default, for limited auto-identation: Completing a delimiter re-indents the lines it affects. Settled (roughly: not containing incomplete tiles) regions get exact re-indentation; unsettled ones a uniform shift.

The presence of user-owned indentation is what allows intent-based discrimination: the completion only absorbs following lines if they are indented relative to the incomplete form's line. The user can indent/dedent to get the behavior they want, but in practice they don't have to, because the enter-inserted indentation suffices for the completion to do what you expect during left-to-right entry, whether writing new code or code above existing code.

The general principle is to move all history sensitivity into secondary, and all system history-dependence into auto-formatting behavior. The user has full control over secondary, so none of this is fundamentally hidden or out of their control. When the code is complete, autoformat erases any trace of historical dependence; it exists purely during incomplete states.

## What round-tripping adds data-structure-wise

Two new fields on the IdTag annotation:
- incomplete: list((Id.t, list(int))): for each canonically completed tile, which shards the user actually typed. The
printer strips synthesized shards on the way back out, so incomplete tiles survive the trip; MakeTerm threads the same masks so statics sees the completed form.
- lexeme: option(string): surface spellings the term normalizes, e.g. 007, 1e3, implicit vs ? vs ?? hole flavor, quoted labels, unknown infix operators (which now elaborate as stuck applications rather than evaluating to their last operand as multi-holes do).

Copy semantics are documented on the type: id-bearing fields drop on copy, value-meaning fields persist.

All other roundtrip gaps filled exploit existing structure rather than new fields. A representative trick: whether a sum type had a leading + is encoded by collected id count, as a leading-plus parse absorbs the separator ids, yielding n ids for n variants, while a bare binary sum carries only the n-1 separators; evaluator-built sums (single id) print bare, which reparses to the same type. Similarly, middle-completion hole ids derive from the tile id instead of being stored.

## Preserved property + fuzzing

The roundtrip property is enforced by a QCheck fuzzer driving Perform.go with random insert/move/destruct sequences from the empty buffer, sampling the actual editor-reachable domain, which parser based generators would undersample (rule tiles only form inside cases; deleted openers, standalone |/=> tokens, some grout cases are edit-reachable only). Also added a corpus and shrunk-counterexample regression suites.

What's compared: canonical completion yields a deeply complete segment whose parse prints back with identical text and piece sequence (tile ids, labels, shards, secondary), modulo a canonical quotient (grout (ephemeral by design), and mold sorts). Spelling, shapes, and precedences compare strictly.

Measured on the fuzzer (5 seeds, 290 states): the grout quotient earns its keep on placement, not ids — grout lands where the parsed term doesn't reproduce it in ~23% of reachable states (dangling boundary grout, stray grout glommed through MakeTerm), while ids agree whenever placement does. (The 4 id-only exceptions traced to the printer minting fresh ids on multihole id shortfalls; padding is now derived, so printing is a pure function of the term.)

The decision to quotient grout ids & sorts is pragmatic; there are currently a number of edge cases, some of which have been addressed, but i think the move it towards making both of these dervied; this is already underway for grout on the virtual-grout branch. If we don't decided to make these derived, we could add more annotations to avoid the need to quotient. In any case you can currently roptionally un the fuzzer with the quotient disabled enumerates to exactly the disagreement states derivation or annotation must handle.

An example of an outstanding sort issue: type : then ]. The editor molds the orphan ] at Typ — the caret sits on the typ side of the :. but canonical completion places the missing [ at the start, reading the state as [? : ?] (an Exp list whose element is the ascription, not ? : [?]), so the reprint derives the ] at Exp. 

Other known generator limitations: it fuzzes single-char insertion, char-wise movement, and backspace (so no selections, backpack pick-up/put-down, or copy/paste) and excludes projectors (follow-up work).

## Later additions (building on the roundtrip)

- **Completion heuristics**: missing delimiters drop into the concave junction their deletion left when shape + sort + uniqueness agree (`let x 1 in 2` -> `let x = 1 in 2`; deleted `else`/`in`/case-bars restore in place); `fun x` over a let-body clips the arrow at the sort frontier instead of absorbing the let into the pattern; a backspaced delimiter's leftover prefix token (`i` for `in`) is completed in place, recorded as a partially-typed shard (`{present, prefixes}` mask) so the roundtrip still closes exactly; leading witnesses restore deleted form heads from their remnants (`le` -> `let`, mid-program); sort-aware symbolic witnesses complete `-` -> `->` and `=` -> `=>`; broken cases repair as one case (wrap/opener arbitration + rule walls), and rules are treated as case-content throughout — mid-entry `case foo |` keeps its `end` after the growing rule.
- **Scoreboard instruments** (Test_CompletionScoreboard, pinned): deletion-inverse restoration over a corpus — full-deletion, last-char-prefix, joint pair-deletion (restoration + delimiter-completeness), and acceptance-stability (applying one suggestion must not move the rest; 96/104 stable, shifts only on known jointly-interacting states). Every remaining miss is a documented accepted-inherent (e.g. `f(3 + f(4)` is byte-identical to a freshly typed wide paren).
- **Quiver chips**: completion suggestions render as interline signposts at their exact placements — a bubble on the line boundary with a ghost-caret pole pinning the insertion column — replacing the old end-of-line boxes. Pins follow the caret through semantically-free whitespace; typed prefixes render cut-out; colorways distinguish tab-territory (pale) from other-segment obligations (gold); at caret coincidence the real caret is authoritative and the chip's pole steps aside.
- **Discharge routes**: `ApplyCompletion(All | One(tile))` materializes the virtual completion into the buffer (caret preserved by id). Alt+M completes everything (replacing the old greedy Dump, now deleted); cmd/ctrl+click on a chip completes exactly that delimiter (pointer cursor while the modifier is held; unmodified clicks pass through to the editor); tab still puts down at the caret. Agent/composition flows materialize through the same engine.
- **TyDi delimiter suffixes**: the ghost suggestion survives on prefix states where completion consumed the token's statics entry (ci-free witness route), 1-char expectation-backed prefixes bypass the length gate (`-`/`=` get arrow completions), and suggestions come from the same witness recognition the engine uses — ghost and quiver agree by construction. Tab at `els` accepts to `else` instead of dropping a fresh delimiter after it.
- **Reassociation**: retyping a deleted keyword char mid-program re-pairs with its orphaned shards (rescan matches singletons only; the repair fallback now explodes remnants and tries the sibling scope); `deep_reassociate` is no longer a setting — reassociation is unconditional, and the harnesses run it (they previously silently didn't, hiding this bug class from every edit-derived test).
- **Formatting**: SpaceNormalize gains an opt-in canonicalize mode (mold-aware within-line respacing: `f (x)` -> `f(x)`, one space around infix ops, comments/linebreaks untouched) backing the new Spaces format level; format actions restore the caret via CaretPreserving (from #2231 — exact piece, else surviving predecessor, else ancestors). Printing is now a pure function of the term (pad_ids derives instead of minting).
- **Auto-reindent triggers**: paste re-indents the pasted lines (settled exact / unsettled uniform shift) and, like TyDi accept, destruct, and comment-toggle, runs the completion-triggered re-indent when a tile completes.
- **Indentation UX** (gated on a core setting, on by default): arrow movement skips leading whitespace (every line keeps one reachable position; clicks/selection stay char-exact); backspace at first-content inverts enter (indent + linebreak, one keystroke, one undo); Cmd+[/] indent/dedent lines from any position; Shift+Backspace dedents at the line-start boundary only.

## Removals

The backpack display is retired (the quiver is the completion display; the put-down mechanism and derived missing-shards machinery remain — internals renamed from "backpack" to missing_shards), along with `Dump`, the `deep_reassociate` setting, and the redundant NutMenu Docs toggle. Persisted settings blobs lose two fields, so browsers may see a one-time settings reset.

## Addendum

- A batch of fidelity fixes found by the fuzzer along the way, including the two closed issues above.
(cross-sort tiles in MakeTerm, crossed synthesized delimiters, adjacent-operator lexemes, typ-side incomplete forms, and more); #2373 remains outstanding.
- docs/segments.md documents the round-trip contract, the provenance fields, and the sort-quotient rationale with the example above.
- Possible future refinements, deferred: sort re-derivation as above; virtual grout (#2165) as the same derive-don't-store program applied to grout.


cyrus-:
- [x] reindent on paste
- [x] think about caret positions in indentation
- [x] Cmd-[ and Cmd-] for dedent/indent

cyrus-:
- [x] indentation in the presence of holes is weird:

```
let f =
    fun x ->
        let x =
            if x < 0 then
                 
                else  
                in  
in
f(3)```

**Comment by cyrus-** [comment]
the decoration in this state seemed strange: 
image

this state seemed to assume I wanted to add a return type annotation rather than annotating the variable:
image

in this state it wants to close out the case analysis and the function definition prematurely which makes variable bindings show up as unbound as I type the rest of the definition:
image

**Comment by cyrus-** [comment]
image

**Comment by disconcision** [comment]
@cyrus- issues addressed

File: src/CLI/Cli.re [hunk 1]
```diff
@@ -50,10 +50,10 @@ let format_hazel = (implicit_hole: string, width, path) => {
       Haz3lcore.Zipper.unselect_and_zip(~erase_buffer=true, zipper);
     let pretty_seg = Haz3lcore.PrettySegment.prettify(~width, segment);
     let output =
+      /* no ~indent: prettify output is self-indented (level_map) */
```
disconcision: unnecessary comment

src/haz3lcore/CompositionCore/CompositionGo.re
```diff
@@ -384,14 +384,63 @@ module Local = {
        re-indents structurally on render), then parse to a segment and
        paste it. Safe: Hazel strings and comments are single-line, so
        no token can span a linebreak. */
+    /* A reserved binder no longer guarantees a parse FAILURE (backup
```
disconcision: rm irrelevant information and tersify

test/CompletionVisualization.re
```diff
@@ -0,0 +1,137 @@
+/* CompletionVisualization: Generate text mockups showing canonical completion.
```
disconcision, line 1: is this file just used for tests? if so maybe should be there

src/haz3lcore/derived/Indentation.re
```diff
@@ -1,3 +1,31 @@
+/* Indentation Calculation
+ * ========================
+ *
+ * This module computes indentation levels for linebreaks in a segment.
+ * The main entry point is `level_map` which returns a map from linebreak
+ * IDs to their indentation level (number of spaces).
+ *
+ * CONTINUATION LINE DESIGN DECISION:
```
disconcision: is this fundamental enough to be header comment content?

src/haz3lcore/derived/Indentation.re
```diff
@@ -7,42 +35,71 @@ let trim_non_content: Segment.t => Segment.t =
     | p => Some(p),
   );
 
-let prev_pieces = (seg: Segment.t): list(option(Piece.t)) => {
-  let rec go =
-          (acc, xs: list(Piece.t), prev: option(Piece.t))
-          : list(option(Piece.t)) =>
+/* Compute context (effective_prev, next, effective_next) for each piece in one pass.
```
disconcision: make sure this info is needed

src/haz3lcore/derived/Indentation.re
```diff
@@ -146,53 +241,246 @@ let is_incrementor = (p: Piece.t): bool =>
...
-          next_pieces(complete_trimmed_seg),
-        ),
-      ),
+      context,
     );
   map;
-}
-and go = (~not_top, base: int, seg: Segment.t) => go'((not_top, base, seg));
+};
+
```
disconcision, line 327: make sure this comment pays for itself, remove irrelevant info that is completely pointless eg ported from blah

src/haz3lcore/derived/CanonicalCompletion.re
```diff
@@ -0,0 +1,2747 @@
+/* CanonicalCompletion: Complete incomplete syntax to enable term creation
```
disconcision: check all comments in this file and make sure they pay for themselves or reduce/eliminate. do make sure though to document the basic process and logic cases of canonical completion

src/haz3lcore/derived/CanonicalCompletion.re
```diff
@@ -0,0 +1,2747 @@
...
+        };
+      add(acc);
+    },
+    [],
+    ins,
+  );
+};
+
+/* Leading missing shard pieces for a tile (openers), natural order */
+let leading_shards = (t: Tile.t): list(Piece.t) =>
```
disconcision, line 206: maybe see if some helpers here and elsewhere should live in the home modules for their types

src/haz3lcore/lang/Form.re
```diff
@@ -562,8 +562,14 @@ let infix_delimiter_ops_prefixes: list(Token.t) =
   |> List.filter_map(((form: compound_form, _)) => {
        let form = get(form);
        switch ((form.mold.nibs |> snd).shape) {
-       /* Could be pickier here, e.g. just trailing delimiters */
-       | _ when List.length(form.label) >= 2 => Some(form.label)
+       /* NON-LEADING tokens only (matching the comment above):
```
disconcision: dont understand this comment pls clarify

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -301,6 +332,56 @@ let is_hole_label = (t: string) =>
...
+
+let all_exp_only = (~host: Sort.t, tm: unsorted): bool =>
+  switch (root_labels(tm)) {
+  | [] => false
+  | labels => List.for_all(is_exp_only_label(~host), labels)
+  };
+
+/* Is this kid a constructor-named type variable? (the head of a
+   potential sum-type constructor application) */
+let is_ctr_headed: Any.t => bool =
```
disconcision, line 388: these helpers might have a better home

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -934,29 +1089,71 @@ and exp_term: unsorted => (Exp.term, list(Id.t)) = {
...
+                {
+                  annotation: IdTagged.IdTag.mk_internal(IdTagged.ids(r)),
+                  term: MultiHole([Exp(r)]),
+                },
+              )
             }
           | (["|>"], []) => Ap(Reverse, r, l)
           | (["@"], []) => ListConcat(l, r)
+          | ([op], []) when op != " " =>
+            /* Unknown infix operator: kids survive as a MultiHole; the
```
disconcision: shorten comment

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -934,29 +1089,71 @@ and exp_term: unsorted => (Exp.term, list(Id.t)) = {
...
   | tm => ret(hole(tm));
 }
 and pat = unsorted => {
   let (term, inner_ids) = pat_term(unsorted);
+  let lexeme = take_lexeme();
   let ids = ids(unsorted) @ inner_ids;
 
   let p =
-    return(p => Pat(p), ids, IdTagged.mk(ids, get_secondary(ids), term));
+    return(
```
disconcision: i feel these plentiful nested return calls with IdTagged.mk are large in this file, esp when formatted vertically. is there a away to streamline?

src/haz3lcore/zipper/action/AutoFormat.re
```diff
@@ -0,0 +1,18 @@
+/* TODO(andrew): once the pretty printer is fixed (it now incorporates
```
disconcision: we should clarify the names etc of this file versus pretty printer

src/web/www/style/editor.css
```diff
@@ -512,44 +512,131 @@ svg:has(.child-line) {
...
+.quiver-chip.chip-live .quiver-chip-body {
+  background: var(--shard-selected);
+  color: var(--token-incomplete);
 }
 
+
+
+/* typed prefix cut-out white on every scheme; remainder takes the
+   scheme's opposite pole */
+.quiver-chip .chip-frac-typed {
```
disconcision: quiver css should maybe be in own file? did we cleanup old backpack css?

disconcision (commented):
Blocking finding:

[P1] Remove the tracked scratch/debug bundle before merge. `probe_tmp/dune` makes `probe2.re` part of the repository-wide `@check` target, but `probe_tmp/probe2.re:191` calls the removed `CanonicalCompletion.chip_at_caret`. The current Fast checks job therefore fails with `Error: Unbound value CanonicalCompletion.chip_at_caret` before quick tests can run. This looks like an accidental artifact bundle rather than product code: the PR also adds two exploratory patch files, probe executables, `CLAUDE.md`, and `plans`, where `plans` is an absolute symlink to `/Users/andrewblinn/Dropbox/projects/hazel/plans` and will be broken for every other checkout. Please remove these local artifacts (or intentionally move/maintain any probe that should survive) so `dune build @check --profile release` passes.

Verification: `dune build @test-quick --profile dev` passes locally with 3,821 tests; the focused Derivation/DerivationCase run also passes (65 tests). GitHub Full tests are still running at review time, while Fast checks fail for the compile error above.

src/haz3lcore/derived/CanonicalCompletion.re
```diff
@@ -0,0 +1,2628 @@
...
+        (Nib.Shape.concave(), Nib.Shape.concave()),
+        reassembled,
+      );
+
+    let insertions =
+      verify_holes(~input=seg, ~completed=completed_seg, insertions);
+    /* materialization can capture still-broken remnants into the new
+       tile's children; recurse until nothing incomplete remains */
+    if (only_tile == None
+        && fuel > 0
```
disconcision: [P1] Remove the fixed ceiling from “complete all”. `complete_segment_deep` defaults to `fuel=24`, and this guard stops recursion after 25 materialization passes even though the comment above says the incomplete count strictly decreases. I parsed two editor-reachable inputs—30 `(` characters followed by `1`, and 30 `let x = ` prefixes followed by `1`. Both had 30 incomplete tiles before canonical completion and still had 5 afterward. That means semantic conversion, `Materialize.all`, and the quiver can all return a partially completed result while claiming to complete everything. Please recurse to a fixpoint with a progress assertion/guard (or derive the bound from the initial incomplete count), and add a regression with more than 25 obligations.

src/haz3lcore/derived/CompletionQuery.re
```diff
@@ -0,0 +1,90 @@
...
+  | [] => None
+  | [d, ..._] =>
+    switch (d.typed_len) {
+    | Some(n) when n < String.length(d.text) =>
+      Some(String.sub(d.text, n, String.length(d.text) - n))
+    | Some(_) => None
+    | None =>
+      let jam_left =
+        switch (z.relatives.siblings |> fst |> List.rev) {
+        | [Tile({label: [tok], _}), ..._] when Token.length(tok) > 0 =>
```
disconcision: [P1] Compute the left spacing boundary from the effective neighboring token, not only single-token tiles. At the caret in `let x = case y | _ => 1 end|`, canonical completion correctly offers the missing `in`, but the neighboring `end` shard belongs to a tile whose full label is `["case", "end"]`, so this pattern does not match. `tab_text` returns `"in "`; applying it produces `let x = case y | _ => 1  endin |` and forces the paste parser fallback. Please use the last visible/effective token of the neighboring piece (or the zipper neighbor-token helper) and add this nested case/let regression.

disconcision (commented):
Second, systematic pass over the updated head found three additional reproducible correctness issues. I consider the first two merge-blocking. The reproductions below were run against this exact head; all current CI checks are green, but the existing tests do not cover these states.

src/haz3lcore/zipper/CaretPreserving.re
```diff
@@ -0,0 +1,308 @@
...
+};
+
+/* Selection anchor for restoration after transform. None for empty or
+   Buffer selections. id_init captures the piece just right of the
+   selection, so transform + reposition normally lands the cursor at
+   the RIGHT edge of the original selection — we track the LEFTMOST
+   selected piece and grow leftward to find it. When the selection
+   abuts buffer end there is no right piece: reposition falls back to
+   the LEFT edge, leftward growth can't reach the anchor, and
+   restore_selection safe-fails to unselect. */
```
disconcision: [P2] Preserve selections that end at the buffer boundary instead of knowingly dropping them. With an identity transform, selecting the final token in `1 + 2` changes `1 + |2§` to `|1 + 2`; a character selection changes `ab|c§` to `ab|c`. `Format(Spacing)`, `Format(Pretty)`, and indentation adjustment all use this transform, so formatting can silently clear the selection and substantially move the caret. Please track/restore both endpoints (including the EOF boundary) and cover whole-token and character-level EOF selections.

disconcision:
@cyrus- changes:

1. optional backpack flag display in settings
2. changed the not-at-caret quiver display to make it slightly less obtrusive
3. quiver now shows leading holes and whitespace padding
4. cursor inspector now show info for leading hole in completion where applicable

PR #2378, opened 2026-07-06
Term-level refactorings

disconcision:
13 context-menu/palette refactorings built on the round-trip (#2374): parse buffer -> transform term -> print -> splice. Ids survive, so probes, statics, and incremental caches do too.

### Gestural - Naming
- **Inline Let**: `let x = 1 + 2 in x + 3` -> `1 + 2 + 3`; offered at the let, its binder, or any occurrence of `x`. Parens only where a reparse oracle proves the bare form changes the program: `let x = 1 + 2 in 3 + x` -> `3 + (1 + 2)`.
- **Add Parameter**: `let f = fun x -> x + 1 in f(2)` -> `let f = fun (x, x1) -> x + 1 in f(2, ?)` — definition and every call site atomically, including recursive calls. Handles `let f(x) = x + 1 in f(2)` -> `let f(x, x1) = x + 1 in f(2, ?)` and annotated bindings (`Int -> Int` becomes `(Int, ?) -> Int`). Gated off when `f` is used as a bare value (`map(f, xs)`).
- **Hoist Let**: `fun n -> let x = 2 in x + n` -> `let x = 2 in fun n -> x + n` — one level per invocation (through chains, out of defs, out of lambdas). The explicit opt-in to evaluation-count changes that Extract avoids.
- **Sink Let**: `let x = 2 in case m | 1 => x | _ => 0 end` -> `case m | 1 => let x = 2 in x | _ => 0 end` — into the single arm/branch/lambda that uses the binding. Both movements are scope-gated and shadow-aware; blocked moves are simply not offered.
- **Extract to Let**: binds at the nearest enclosing line — `let a = 1 in g(f(2))` extracting `f(2)` -> `let a = 1 in let x = f(2) in g(x)`; a def-site hoists above its line (`let a = g(f(2)) in ...` -> `let x = f(2) in let a = g(x) in ...`). Never escapes a lambda or case arm, so evaluation count/conditionality never changes.

### Gestural - Choice
- **Add Missing Case Arm**: `case b | true => 1 end` -> `case b | true => 1 | false => ? end`, one arm per invocation from the Coverage witness — real constructors for finite domains (`Green`), literals wildified for infinite ones (`case n | 1 => 2 end` gains `| _ => ?`). Label names the witness: "Add Arm | false".

### Naming (non-gestural)
- **Rename (bind free occurrences)**: rename a binder by ordinary editing — `let vel = 1 in v + v` (was `v`) marks the old uses free, and the binder token offers "Rename v to vel" -> `let vel = 1 in vel + vel`. Stateless: reconstructed from scope state, no edit tracking. Works at let heads, `f(x)`-sugar names and params, fun params, and case-arm pats; per-occurrence shadow-precise.
- **Remove Unused Let**: `let x = 1 in 2 + 2` -> `2 + 2`; gated on the statics unused-binding warning. Works on annotated heads and unused `f(x)`-sugar fns.

### Equational (syntactic)
- **Convert to Case**: `if a then 1 else 2` -> `case a | true => 1 | false => 2 end`.
- **Convert to If**: the inverse, either arm order: `case a | false => 2 | true => 1 end` -> `if a then 1 else 2`.
- **Negate & Swap Branches**: `if a && b then 1 else 2` -> `if !(a && b) then 2 else 1` — arms swap with their formatting untouched; condition parens by the oracle (`if a ...` gives bare `!a`).
- **Eta-Reduce**: `fun y -> f(y)` -> `f`, gated on `y` not occurring in `f`.

### Equational (semantic)
- **Add Type Annotation**: `let x = 1 + 2 in x` -> `let x : Int = 1 + 2 in x`, type from statics; only offered when the inferred type is hole-free. Oracle-parenthesized: `let p = (1, true) in p` -> `let p : (Int,Bool) = (1, true) in p`.

Formatting is preserved by construction: whitespace runs belong to line slots and transforms move/copy them explicitly, so multiline layouts survive swaps, extractions, and conversions. Menu gating never prints, parses, or runs the oracle (invocation only). ~110 tests in Test_Refactor.re.

Stacked on #2374; retarget to dev when it merges.


7h3kk1d:
Amaze! Amaze! Amaze!

Still looking through. At least one small issue is let-inlining throws away the pattern type annotation

7h3kk1d:
Just getting to testing this and it still looks amazing. It would be great if there was away to add a default param when adding a parameter so it got propagated to all callsites.

For swap params I have a slight preference for move param right/left rather than swap for namings. 

Seeing some issues with inline next use on `subst` in the adt slide. I think next use is recursive and that might not be working.

Some thought should probably go into the behavior of some of these refactors in the presence of holes. eta-reduce can happen when the function is a hole even though that could be filled with a reference. I prefer this but it doesn't match the way we do warnings and such for unused variables.

I think we should offer to add the type params even when there's holes.

Extra asks: 

- convert to case on polymorphic equality with literals should turn it into a case on the actual match.
- a curry using partial application action to turn `let f = fun x -> int_plus(1, x) in  ` into `let f = fun x -> int_plus(1, _)(x) in  ` so you can eta-reduce. Action would be on the ap

7h3kk1d:
Just thought about how the currying one more and I think the right approach is something that pushes function args into a separate application with intermediary partial application.

`f(a, b)` applied to `b` would be `f(a, _)(b)`
`f(a, b)` applied to `a` would be `f(_, b)(a)`
`f(a, _, c)(b)` applied to `a` would be `f(_, _, c)(a, b)`
This coupled with a refactor that can expand partial application into a lambda and a way to fully combine subsequent partial application and applications would be sick for combinator work.

PR #2385, opened 2026-07-20
Fix Skel.push_output crash when reassociating a second in below a nested let

disconcision:
fix: editor will crash if you

write:
```
let a = 
  |
in
```
then:
```
let a = 
  let b = |
in
```
then:
```
let a = 
  let b = 
in
in|
```


cyrus-:
I can't reproduce this on `dev` -- maybe not entering in the correct order, can you post video?

PR #2389, opened 2026-07-22
Make tiles slightly less illegal

disconcision:
Tiles stored their label and mold even though both are determined by form. Each serialized tile carried redundant strings and precedence numbers so operator-precedence changes required migrating every serialized file, and nothing prevented label/mold mismatches.

Now a tile is `{id, form, sort, shards, children}` where `form = Compound(family) | Tok(token) | TokInfix(token)`, with 92 form families (label x shape-role) plus free-text tokens, and label/mold are derived. `TokInfix` is a technical detail that will die with virtual grout.

This isn't quite a pure refactor, but my hope is that there are no observable logic changes.

- Serialized syntax shrinks ~40%, and precedence/mold changes no longer require migrations.
- MakeTerm dispatches on constructors instead of token-string lists. This technically causes some behavior changes in edge cases, but I have never seen the relevant behaviors in the wild.
- Exercise/derivation examples now ship as serialized-zipper text literals rather than OCaml literals, so they survive representation changes; slide files were regenerated id- and refractor-preservingly.
- The stored `sort` is an explicitly isolated local cache (well-sorted <=> `mold_of(form, sort).out == sort`); experimental follow up work will hopefully make sort a derived property, eliminating this field
- The grammar's string layer now lives in the language library (src/language/grammar/), including Token, Label, and FormId (the family enums and the per-family label table), following the existing Sort re-export pattern. This deduplicates the token/string mirrors that src/language previously hand-maintained
- TyDi's suggestion-type tables are now derived from the grammar at startup instead of hand-maintained.


scripts/README_migrate_tile_format.md
```diff
@@ -0,0 +1,49 @@
+# Migrating zipper-embedding exercise modules (tile-datatype flip)
```
disconcision: should be able to get rid of this script now I think since slides are now generated from .hz text

scripts/split_migrate_output.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
disconcision: ibid

src/haz3lcore/derived/Indentation.re
```diff
@@ -115,20 +115,13 @@ let complete_segment = (seg: Segment.t): Segment.t => {
 
 let is_comma = (p: Piece.t): bool =>
   switch (p) {
-  | Tile(t) => t.label == [","]
+  | Tile(t) => Tile.is_comma(t)
   | _ => false
   };
 
 let is_case_rule = (p: Piece.t): bool =>
```
disconcision: this whole thing should be moved to piece maybe

src/haz3lcore/lang/Form.re
```diff
@@ -37,510 +45,318 @@ type sorted_expansions = list((Token.t, Sort.t, Label.t, Direction.t));
...
+       {
+         label,
+         mold,
+         expansion,
+       };
+     });
+};
+
+/* Same-label candidate priority — the only place where row order in
+ * the form table is meaningful. Four labels are spelled by more than
```
disconcision: what is 'spelled' here? make this comment clearer

disconcision: also: Q: are these actually load bearing for tests? examples?

src/haz3lcore/lang/Form.re
```diff
@@ -827,3 +580,202 @@ module Expansion = {
...
+        ? List.map(
+            (m: Mold.t) =>
+              a == InfixDelimiterPrefix ? (TokInfix(t), m) : (Tok(t), m),
+            molds,
+          )
+        : [],
+    atomic_defs,
+  );
+
+/* Memoized per token: accessors derive Tok molds on every read, and
```
disconcision, line 636: does this optimization actually make a difference? same for others below

src/haz3lcore/lang/Form.re
```diff
@@ -827,3 +580,202 @@ module Expansion = {
...
+ * a concrete sort). TokInfix: the InfixDelimiterPrefix bin, uniformly
+ * at any sort. */
+let mold_of = (f: t, sort: Sort.t): Mold.t =>
+  switch (f) {
+  | Compound(fam) =>
+    switch (Hashtbl.find_opt(family_mold_tbl, (fam, sort))) {
+    | Some(mold) => mold
+    /* parens wrap one child at every sort; the family table only
+     * registers rows at the grammar's declared sorts */
+    | None when fam == Parens => Mold.mk_op(sort, [sort])
```
disconcision, line 763: why is this special case necessary?

src/haz3lcore/lang/Form.re
```diff
@@ -827,3 +580,202 @@ module Expansion = {
...
+let classify_label = (sort: Sort.t, label: Label.t): (t, Sort.t) => {
+  let fits = ((_, m): (t, Mold.t)): bool => m.out == sort;
+  switch (List.find_opt(fits, base_candidates(label))) {
+  | Some((id, _)) => (id, sort)
+  | None =>
+    switch (compound_defs(label)) {
+    | [(fam, _), ..._] => (Compound(fam), Sort.Any)
+    | [] =>
+      switch (label) {
+      | [t] => (Tok(t), Sort.Any)
```
disconcision: is this top case redundant to below?

src/haz3lcore/lang/Form.re
```diff
@@ -827,3 +580,202 @@ module Expansion = {
...
+  };
+};
+
+/* The form wrapping a segment in parens at a given sort
+ * (Segment.mk_duo/parenthesize): always the registered Paren form,
+ * never the Ap form classify_label would find first at
+ * Drv(Exp)/Drv(Pat). Sorts with no registered paren row (Rul, Mod,
+ * Sig, MPat, Any, remaining Drv) fall back to classify_label:
+ * (Parens, Any) with the op(Any) fallback mold. */
+let mk_parens_id = (sort: Sort.t): (t, Sort.t) =>
```
disconcision: feels like a bit of a weird function... what's it fore?

src/haz3lcore/zipper/action/Indicated.re
```diff
@@ -19,7 +19,11 @@ type piece = {
    and application forms like Ap parens (Concave, Convex). */
 let has_concave_left_nib = (p: Piece.t): bool =>
   switch (p) {
-  | Tile({mold: {nibs: ({shape: Concave(_), _}, _), _}, _}) => true
+  | Tile(t) =>
+    switch (Tile.mold(t).nibs) {
```
disconcision: mk helper in Tile.re maybe

src/haz3lcore/zipper/action/Triggers.re
```diff
@@ -98,19 +98,18 @@ let apply_refractors = (refractors: list((Id.t, string)), z: t): t =>
...
   switch (z.relatives.siblings |> fst |> List.rev) {
-  | [
-      Tile({label: ["(", ")"], children: [syntax], _}),
-      Tile({label: [name], _}),
-      ...rest,
-    ]
-      when is_refractor_trigger(name) =>
+  | [Tile({children: [syntax], _} as parens), Tile(nt), ...rest]
+      when
+        Tile.is_paren_shaped(parens)
```
disconcision, line 107: same when predicate as this used below i think, maybe pull it out. dunno tho there are other similar ones too so dunno if there's good abstraction

src/haz3lcore/FastParse.re
```diff
@@ -370,14 +367,22 @@ let weave =
             );
           },
           ([], [tok0]),
-          List.tl(t.label),
+          List.tl(label),
           t.children,
         );
+      /* the source spelling must land (float/backtick-equivalence
```
disconcision: vague comment

src/language/grammar/BaseAtom.re
```diff
@@ -0,0 +1,52 @@
+/* The bidirectional association between atomic base-type SPELLINGS and
```
disconcision: improve this comment

src/web/exercises/CodeExercise.re
```diff
@@ -339,6 +339,23 @@ let transition: transitionary_spec => spec =
     };
   };
 
+/* Persistent counterpart of [spec]: each editor is a serialized zipper
```
disconcision: eliminate this comment

src/web/exercises/TheoremExercise.re
```diff
@@ -12,6 +12,68 @@ type spec = {
   theorem: Haz3lcore.Zipper.t,
 };
 
+/* String-based counterpart of [spec], for "transitionary" exercise modules
```
disconcision: Q: why do we have this new spec stuff as part of this PR? what is it necessary for? like here and otherwhere as well

src/web/dune
```diff
@@ -78,6 +79,26 @@
  (preprocess
   (pps js_of_ocaml-ppx)))
 
+; Example-module exporter: rewrites the example exercise modules in the
```
disconcision: why does this exist?

src/web/Migrate_exercises.re
```diff
@@ -0,0 +1,460 @@
+/* DISPOSAL: originally disposable migration tooling for the tile FormId
```
disconcision: ah maybe answering my own question, maybe the spec stuff is for this migration? which maybe we still need as the exercises don't use .hz or .hzt format? maybe we can investigate transitioning them to that though? or a new format ratethre (.hze?)

src/web/app/explainthis/Example.re
```diff
@@ -1,97 +1,120 @@
 open Haz3lcore;
```
disconcision, line 1: Q: should this file be called Example and live here? i've never been quite sure, this might be an opportunity to rethink

src/web/app/explainthis/ExplainThis.re
```diff
@@ -272,8 +272,8 @@ let expander_deco =
 
     let get_clss = segment =>
       switch (List.nth(segment, 0)) {
-      | Base.Tile({mold, _}) => [
-          "ci-header-" ++ Sort.to_string(mold.out) // TODO the brown on brown isn't the greatest... but okay
+      | Base.Tile(t) => [
+          "ci-header-" ++ Sort.to_string(Tile.mold(t).out) // TODO the brown on brown isn't the greatest... but okay
```
disconcision: lets rm this comment

src/menhirParser/AST.re
```diff
@@ -200,6 +200,15 @@ and sig_item =
  *
  * ['A'-'Z'] ['a'-'z' 'A'-'Z' '0'-'9' '_']*
  */
+/* Subset of Language.Token.base_typs the generator avoids; deliberately
```
disconcision: does menhir really not parse the other base types? even with the fast parse enhancements merged to dev? if not we should test and add them i think?

src/haz3lcore/tiles/Base.re
```diff
@@ -8,16 +8,24 @@ and piece =
...
   // - length(shards) == length(children) + 1
   // - sort(shards) == shards
   [@equal (_, _) => true]
   id: Id.t,
-  label: Label.t,
-  mold: Mold.t,
+  form: Form.t,
+  // the local sort guess cached at insertion classification / remold
+  // (the only writers); mold = Form.mold_of(form, sort).
+  // sexp only: sort defaults to Exp and is omitted when Exp;
```
disconcision: some of this comment is strictly redundant; improve please

PR #2390, opened 2026-07-22
Coding agent: review fixes + cleanup pass

disconcision:
Fixes the functional issues from the #2210 review plus problems surfaced by live agent sessions, and does a structural cleanup pass. Example of the headline bug: asking the agent to fill the body hole after `let x = 1 in` produced `let x = 1 inlet e = 2 in x`, token fusion that reparses as the identifier `inlet` when copied. Root cause was in hole-fill path; fixed by padding only where adjacent tokens would lex as one.

Behavioral changes, agent tools:
- Leading whitespace stripped from tool code on all paths. Vertical whitespace normalized after edits: one blank line between top-level definitions, no leading blank, no accumulating trailing blanks.
- Ambiguous paths (shadowing) return an error with `name#k` retry syntax instead of silently editing the earliest match. Function-sugar bindings can now be addressed by bare name (`f`, not `f((x, y))`).
- Pattern rename rejects capture and binder-count changes and re-checks statics. Renaming recursive sugared functions now works (use sites in the definition were previously skipped, so those renames always failed). Note: renaming a recursive `let f = fun ...` (unsugared) now rejects rather than silently leaving the old name dangling in the recursive call.
- Paste failures name reserved keywords (`eval` etc.) instead of a bare `CantPaste`. Node-map builds no longer fail with the cursor inside a sugared binding, and are no longer quadratic.

Behavioral changes, chat UX:
- Enter appends the message and clears the input immediately; the heavy context/payload work runs a tick later. Stop pressed in that gap cancels before anything is sent.
- Streaming replies render as markdown live, with dangling `**`/backticks auto-closed instead of flashing raw markers.
- API failures surface for real: 401s show the provider message instead of empty-reply retry churn, network errors report immediately, a Node-side error path no longer hangs, and per-request debug logging is removed.
- The context meter says "budget used" with a tooltip naming the model window when the 100k compaction cap is what's displayed.

Prompts: function-definition sugar (`let f(x, y) = ...`) documented as preferred; tiered reserved-word warning, empirically verified (e.g. `rec`/`val`/`valid` are not term keywords); audit fixes (nonexistent `string_of_int`, broken doc examples, missing float ops and `|>`, `table` projector kind).

Internals, briefly: `AgentUpdate.re` split 2081 to 385 lines across seven modules; single tool registry replacing four drifting name lists (which exposed four unregistered workbench tools); append-message dedup; dead code and comment trims; all inline comments from the #2210 review addressed or answered on the branch.


disconcision:
@russell-rozenbaum if you're okay with these changes on an architectural level you can merge them into your branch, afterwards i'm fine merging that branch to dev.

PR #2402, opened 2026-07-28
Disallow surface typfun

MaxCarroll0:
Disallowing typfun in the surface syntax as mentioned in #2254. The actual missing errors when using typfun explicitly where proper types are expected is fixed instead in #2401 

Adds curried type function syntax for `type T(A)(B) = L(A) + R(B)` etc. to allow writing curried types


MaxCarroll0:
Decided with @cyrus- this is probably not the way to go from a pedagogical perspective

PR #2404, opened 2026-07-28
Evaluator perf: skip dead reuse-map work, StringMap reuse maps, cheap Id.mk (~1.7x)

disconcision:
Evaluating the same program repeatedly (`./hazel bench-eval -n 20 bench/eval-bench.hz`, new subcommand that parses/elaborates once and times evaluation only, dev profile under node): on current dev (evaluator streaming) plain evaluation goes 811 -> 467 ms/iter and incremental cache-seeding 797 -> 485. Profiling showed two pre-existing evaluator costs dominating every dispatch: `IncrEval.remove_pat_bindings` (~50%) and `Id.mk` (~30%).

Three changes:
- The evaluator skips all reuse-map maintenance (including the streaming-era `ReusePass` whole-program pre-pass) when there is no prior state to reuse and no statics to target (CLI `run`, MVU dispatch): the map is provably never consumed on that path, but its upkeep dominated evaluation.
- `IncrEval.reuse_map` is now a StringMap instead of an assoc list, so `remove_pat_bindings` no longer walks ~180 builtin entries with polymorphic `List.mem` at every binder.
- `Id.mk` mints from a session-tagged monotonic counter instead of drawing a v4 UUID from the OCaml 5 LXM generator (~1.5us/call under jsoo, ~264k calls per bench iteration). Ids stay v4-shaped and unique: counter in the leading bytes (so `Id.str8`/`cls` abbreviations stay distinct), 62 session-random tail bits for cross-runtime uniqueness. Environment ids are load-bearing (closure `fast_equal`, serialization sharing keys) and are still minted.

Pre-streaming this measured 1648 -> 336 / 1539 -> 432 (~4.9x/3.6x); evaluator streaming (#2339) independently captured part of the same win (it gates reuse-map upkeep inside call stacks), which is why the marginal multiplier on top of it is smaller. Streaming also adds ~40% overhead to the fully-optimized plain path (336 -> 467 on the same machine) — a separate future target.

Added 2026-08-10, two upstream-of-evaluator costs found while profiling livelit-heavy programs:
- Statics no longer deep-normalizes types: alias references flow compact and consumers resolve heads on demand and compare names first (`equal_up_to_aliases`, `canonicalize`, `normalize(~expand)` for module exports). Graph-benchmark statics 650 -> 101 ms warm (HTML-heavy synthetic 7.7s -> 155ms); worst in-editor keystroke stall 1.6–2.1s -> ~100ms; plain code unchanged. Design + site table in plans/type-normalization.md.
- Molding's hot token predicates (`is_infix_delimiter_op_prefix`, `is_ambiguous_polymorph`) were `List.mem` scans with polymorphic compare per atomic-form candidate on every molding query — now hash-set membership, ~1.3x on batch parse of a 120-line program (agent insert-text edits, CLI, load).

Full suite green (3297 tests after the dev merge); the pattern-equivalence property's FAIL-then-skip seeds are unchanged vs dev (pre-existing).


disconcision:
@Negabinary this is entirely vibed perf work that improves perf on the hazel-MVU stuff i'm doing... do the first two changes pass the sniff test from an incremental perspective?

Negabinary:
@disconcision 

2/3 pass the sniff test; 

1 feels like it would only be useful for tests / things called from the terminal; in reality in the editor it's always going to be passed an info map and there's almost always going to be a prev; I'd be tempted to undo 1

Negabinary (changes requested):
see above

disconcision:
@Negabinary: r u ok with 1 now? due to CLI path etc. forget where we came down on this

Negabinary (approved):

Issue #2397, opened 2026-07-27
Evaluator.Properties "Pattern equivalence" can OOM and abort the whole test run


disconcision:
from claude:

`QCHECK_SEED=836759744 bash test/run_node.sh test 'Evaluator.Properties'` exhausts node's 4 GB heap and dies with `Abort trap: 6`, taking the whole test run with it — later test groups never execute, so the run looks like a timeout or an unexplained failure rather than a test failure.

It's case 2, "Pattern equivalence" (`test/evaluator/Test_Evaluator_Properties.re:122`). Roughly 1 seed in 6 does it, so `@runtest` aborts intermittently in CI. Reproduces on `dev` (471ca88314); not branch-specific.

Not diagnosed. The test generates 1000 pairs of a size-40 expression and a size-10 type, elaborates and evaluates each twice with step limits of 10,000 and 1,000,000, and `print_endline`s two rendered expressions per case. A million steps of the evaluator retaining intermediates seems a likelier culprit than the printing, but that's a guess.

`QCHECK_SEED=<n>` makes these properties deterministic, which is how to bisect it.

Issue #2372, opened 2026-07-02
Stepper/evaluator confluence PBT flake: duplicate-binder pattern yields alpha-variant results


disconcision:
CI hit a seed-dependent failure of the `Evaluator and stepper are consistent` QCheck property (run 28610859752, on probes-iv — the property and the engines it exercises are identical to dev):

```
Expected: `^^fold(fun (x:: x) -> x)'
Received: `^^fold(fun (x:: x') -> x')'
on `fun (x:: x) -> x` (after 1 shrink steps)
```

One engine freshens the duplicate binder in the cons pattern (`x::x` → `x::x'`), the other doesn't, and `Equality.semantic.exp` doesn't equate the two alpha-variants, so the property fails on whatever run happens to generate a duplicate-binder pattern. The same commit passes the full suite locally under a different seed.

Possible fixes: make `Equality.semantic` alpha-blind for duplicate binders, align the freshening behavior between evaluator and stepper, and/or fix the QCheck seed in CI for determinism.

disconcision:
fable posted this without asking when i asked it why the build was failing on the server, so fyi it does that

disconcision:
Confirming the flake diagnosis: the next probes-iv run (28616005187, same evaluator/stepper code plus only comment/decoration changes) passed the full suite with a fresh seed.

disconcision:
Deterministic repro on dev @a93eaeb0dd: `QCHECK_SEED=262311426` fails `Evaluator.Properties.005` every run, on

`typfun ? -> (let y = ({ let 133032.335046 = B; let y = y }) in ?)`

Same class as the original report: the duplicate `y` binder is freshened to `y'` by one engine but not the other. Found while merge-testing autofocus (#2248) and re-verified in a clean worktree at pristine dev, so it is dev-side rather than branch fallout. Could serve as a pinned regression test once the freshening/equality fix lands.

matthewhammer:
from claude:

Fresh occurrence today, on `fumola-livelit-mvp` at `2e6e2feea8` (a diff that touches no evaluator code), in a local full run — `Evaluator.Properties` case 005:

```
FAIL Small step reduction and big step reduction are equal

   Expected: `let y:: y = () in y'
   Received: `let y:: y' = () in y'''

on `let y:: y = () in y (after 3 shrink steps)`
```

[...]

Issue #2128, opened 2026-02-18
Evaluator and stepper are consistent test failure on CI due to capture avoidance


7h3kk1d:
<img width="1178" height="668" alt="Image" src="https://github.com/user-attachments/assets/0f897182-1470-40f5-8bd4-2049a911927c" />

```
test `Evaluator and stepper are consistent`
raised exception `Alcotest assertion failure
[31mFAIL[0m Small step reduction and big step reduction are equal

   Expected: `[32m^^fold(fun (y::y) -> y)[0m'
   Received: `[31m^^fold(fun (y::y') -> y')[0m'

`
on `^^fold(fun (y::y) -> y) (after 1 shrink steps)`
```

7h3kk1d:
Still live on `dev`, and still intermittent — this property failed for me on a random local seed while I was disabling a *different* flaky property (#2457), then passed on the next run.

Since #2422 made the runner propagate a real exit status, that means an unlucky draw blocks whatever PR it lands on.

So on branch `disable-flaky-eval-pbts` I've **disabled** `qcheck_stepper_confluence` and replaced it with a deterministic pin of the counterexample from this issue and #2372 — `fun (x::x) -> x`, where one engine freshens the duplicate binder and the other doesn't.

The pin asserts the **current, wrong** behaviour, deliberately. That's why it isn't a `skip_known_bug` like the ones in `test/Test_Elaboration.re` and `test/statics/`: a skipped test stays silent forever, whereas this one goes red the day the engines agree and names the property to re-enable:

```
#2128/#2372 appear FIXED: re-enable the disabled `Evaluator and stepper are
consistent` property and delete this expected-failure pin
```

Both engines are now driven through a shared `reduce_both_engines`, called by the disabled property and the pin alike, so re-enabling is a one-line change.

**On #2408:** that PR fixes a *different* confluence cause — uncollapsed casts in the branches of an indeterminate `if` (seed 773972249). It doesn't touch binder freshening, so landing it won't on its own make this property safe to re-enable; the pin here should still be red afterwards. Worth confirming when it lands.

#2372 looks like a duplicate of this issue — same counterexample shape, same diagnosis — so one of them can probably be closed.

---

**Update:** opened as #2482.

[...]

PR #2408, opened 2026-07-28
Fix two seed-dependent QCheck flakes: reserved constructor names, weak_head_normalize allowlist

disconcision:
Two QCheck flakes that reproduce byte-identically on plain dev with `QCHECK_SEED` from CI logs.

**1. `Test_Menhir` properties fail on seeds that generate a constructor named `Nat`** (seed 19038460). The lexer reserves `Nat`, `SInt`, `Void`, `Unknown` and `Internal` alongside the four names `gen_constructor_ident` already avoided; a drawn `Nat` lexes as `NAT_TYPE` and the parse fails. Fix: complete the generator's reserved list in `AST.re`.

**2. `Failure("weak_head_normalize exceeded 1000 recursive calls")` on `type x = y in type y = x in (()) @< x >`** (seed 208099305). The "Elaboration does not crash" allowlist covers the `normalize` wording of the #1627 recursion limit but not the `weak_head_normalize` one. Fix: allowlist the message and pin the counterexample as a `skip_known_bug`.

Previously this PR also changed `Ascriptions.re` to collapse casts pushed into the branches of an indeterminate `if` (stepper/evaluator confluence, seed 773972249). That change is dropped: the semantics were contested in review, and #2482 has since disabled that property on dev. The counterexample is filed as #2514.


MaxCarroll0:
Will this fix #2409? Worth looking at that example too

disconcision:
@7h3kk1d any opinions on this?

src/language/dynamics/transition/Ascriptions.re
```diff
@@ -127,17 +127,24 @@ let rec transition = (~recursive=false, d: DHExp.t): option(DHExp.t) => {
...
+        transition(~recursive=true, d) |> Option.value(~default=d);
       Some(
         IdTagged.fast_copy(
           DHExp.rep_id(e),
           If(
             recur(cond),
-            recur(Asc(e1, t) |> DHExp.fresh),
-            recur(Asc(e2, t) |> DHExp.fresh),
+            collapse(Asc(e1, t) |> DHExp.fresh),
+            collapse(Asc(e2, t) |> DHExp.fresh),
```
7h3kk1d: I don't fully understand this and @Negabinary might have more opinions on how it relates to the stepper. I wouldn't expect to deeply transition into the branches of a stuck if and I would expect those to just be stuck.

If I try pasting the example `[(), (debug () in if ? then () else B)]` into the editor and opening the stepper it crashes

```
Failure("exp_idx: e1 not found in e2 (found 1 matches)")
```

7h3kk1d: Once my local setup is free I'll see if I can figure out what it is the stepper is doing.

disconcision: yeah to be clear i don't really understand this either, claude just brought this up after repeatedly running into flakes. happy to just ditch this if it's slop.

7h3kk1d: I'd feel more comfortable disabling the test with a link to a github issue or waiting for matt to chime in. If I can figure out what it is I'll resolve it. But I'd rather not be doing deep traversals when we don't need to and I don't want to accidentally start being full beta reduction.

7h3kk1d: Or if you'd rather you can change equality for that test to ignore ascriptions and leave the comment on the equality part of it.

disconcision: @Negabinary any thoughts?

7h3kk1d:
Heads-up on a possible collision: I've opened #2482, which **disables** two `Evaluator.Properties` QCheck properties rather than fixing them — including `Evaluator and stepper are consistent`, which this PR touches.

Motivation was narrow: case 7 ("Incremental eval agrees with fresh eval after a literal edit", #2457) reddened three unrelated CI runs in the last two days ([1](https://github.com/hazelgrove/hazel/actions/runs/33513972427), [2](https://github.com/hazelgrove/hazel/actions/runs/33513747011), [3](https://github.com/hazelgrove/hazel/actions/runs/33412342909)), and the confluence property failed on a random local seed while I was doing that. Since #2422 made the runner propagate a real exit status, both now block whatever PR they land on.

Explicitly not trying to pre-empt your fix — just to stop the bleeding until real fixes land. Two things worth knowing:

**1. This PR shouldn't be enough to re-enable the confluence property.** Your fix targets uncollapsed casts in indeterminate `if` branches (seed 773972249). The other known cause is binder freshening — `fun (x::x) -> x` reduces to `^^fold(fun (x::x) -> x)` under one engine and `^^fold(fun (x::x') -> x')` under the other (#2128, dup #2372), and `Equality.semantic.exp` doesn't equate the alpha-variants. #2482 pins that case as a deterministic expected-failure test, so it should still be red after this PR lands. If it isn't, your fix covered more than expected and the property can go straight back on.

**2. Both PRs edit `Test_Evaluator_Properties.re`, so expect a conflict.** #2482 restructures the confluence property fairly heavily: the both-engines run is extracted into a `reduce_both_engines` helper (`option((Exp.t, Exp.t))`, `None` for vacuous cases) that the property and the pin share, and the property body shrinks to a `switch` over it. If your regression test drives the engines directly it'll want to go through the same helper. Happy to defer — if you'd rather land this PR first, I can rebase #2482 on top of it, or drop the confluence half of #2482 entirely and keep only the #2457 disable, which is the one that actually caused the CI failures. Whichever you prefer.

Also flagging #2397 as unaddressed by either PR: "Pattern equivalence" exhausts node's 4 GB heap on ~1 seed in 6 and aborts the whole run, which shows up as a timeout rather than a test failure. Can't be pinned as an expected failure since an OOM kills the process — probably wants a lower `~count`/step limit.

disconcision:
@7h3kk1d i dont have the bandwidth to think about this one so happy to close this in favor of disabling

7h3kk1d:
> @7h3kk1d i dont have the bandwidth to think about this one so happy to close this in favor of disabling

We can just disable them and I can push a change to keep this up to date. Feel free to close if you just want it off your mind though

PR #2414, opened 2026-07-30
ExplainThis PBT

Negabinary:
PBT to check ExplainThis does not crash

### DIFF
--- a/test/Test_ExplainThis.re
+++ b/test/Test_ExplainThis.re
@@ -1,0 +1,46 @@
+open Language;
+
+/* Property-based test ensuring ExplainThis never raises while producing
+   documentation for any sub-term of an expression. The documentation for
+   each form substitutes term ids into its explanation string via a format,
+   so a mismatch between the number of `%s` placeholders and the number of
+   supplied arguments crashes at runtime. This test guards against that. */
+
+let globals = Web.Globals.Model.init(~settings=Web.Settings.Model.init, ());
+let docs = Web.ExplainThisModel.init;
+
+let statics = term =>
+  fst(Statics.mk(CoreSettings.on, Builtins.ctx_init(Some(Operators.default_mode)), term));
+
+let qcheck_explainthis_does_not_crash =
+  QCheck.Test.make(
+    ~name="ExplainThis.get_doc does not crash",
+    ~count=1000,
+    QCheck_Util.arb_exp(~minimal_idents=true, 12),
+    exp => {
+    /* Statics failures are out of scope; we only assert that ExplainThis
+       itself does not raise for any sub-term it is asked to document. */
+    switch (statics(exp)) {
+    | exception _ => true
+    | info_map =>
+      Id.Map.iter(
+        (_id, info: Info.t) => {
+          let _ =
+            Web.ExplainThis.get_doc(
+              ~globals,
+              ~docs,
+              Some(info),
+              Web.ExplainThis.Colorings,
+            );
+          ();
+        },
+        info_map,
+      );
+      true;
+    }
+  });
+
+let tests = (
+  "ExplainThis",
+  [QCheck_alcotest.to_alcotest(qcheck_explainthis_does_not_crash)],
+);
--- a/test/haz3ltest.re
+++ b/test/haz3ltest.re
@@ -67,7 +67,8 @@ let (suite, _) =
     @ [Test_GradingReport.tests]
     @ [Test_Derivation.tests]
     @ Test_DerivationCase.tests
-    @ Test_PromptFactory.tests,
+    @ Test_PromptFactory.tests
+    @ [Test_ExplainThis.tests],
   );
 Junit.to_file(Junit.make([suite]), "junit_tests.xml");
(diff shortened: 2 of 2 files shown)


test/Test_ExplainThis.re
```diff
@@ -0,0 +1,72 @@
...
+   documentation for any sub-term of an expression. The documentation for
+   each form substitutes term ids into its explanation string via a format,
+   so a mismatch between the number of `%s` placeholders and the number of
+   supplied arguments crashes at runtime. This test guards against that. */
+
+let globals = Web.Globals.Model.init();
+let docs = Web.ExplainThisModel.init;
+
+/* Generates core expressions that exercise the derivation/proof ExplainThis
+   forms (Theorem, ProofObject) in addition to ordinary expressions. The base
```
cyrus-: you mentioned this is no longer the case

test/Test_ExplainThis.re
```diff
@@ -0,0 +1,69 @@
...
+open Language;
+open MenhirParser;
+
+/* Property-based test ensuring ExplainThis never raises while producing
+   documentation for any sub-term of an expression. The documentation for
+   each form substitutes term ids into its explanation string via a format,
+   so a mismatch between the number of `%s` placeholders and the number of
+   supplied arguments crashes at runtime. This test guards against that. */
+
+let globals = Web.Globals.Model.init();
```
7h3kk1d: Mostly an aside but the whole HazelDB test configuration that this is "loading" from is sketchy to me. I would kind of rather we just specify `default` here or just Settings.init directly.

test/Test_ExplainThis.re
```diff
@@ -0,0 +1,69 @@
...
+let docs = Web.ExplainThisModel.init;
+
+let arb_drv_exp = (~minimal_idents, size) => {
```
7h3kk1d: Not sure what the drv stands for here. I think this can just use `QCheck_Util.arb_exp` which also has shrinking.

test/Test_ExplainThis.re
```diff
@@ -0,0 +1,69 @@
...
+      |> Printer.of_segment(~holes="?", _)
+    ) {
+    | s => s
+    | exception _ => "<unprintable expression>"
+    };
+  QCheck.make(~print=show, map(to_core, base));
+};
+
+let statics = term =>
+  fst(Statics.mk(CoreSettings.on, Builtins.ctx_init(Some(Int)), term));
```
7h3kk1d: Probably use Operators.default_mode

test/Test_ExplainThis.re
```diff
@@ -0,0 +1,69 @@
...
+
+let qcheck_explainthis_does_not_crash =
+  QCheck.Test.make(
+    ~name="ExplainThis does not crash",
```
7h3kk1d:
```suggestion
    ~name="ExplainThis.get_doc does not crash",
```

7h3kk1d (changes requested):
Left a few comments that should probably be addressed before merge.

On the actual purpose of the PR I'm curious if this is mostly aiming to address the format string issue or if there's a litany of unsafe things we're trying to catch. The generator is not exhaustive and I'm not sure what the chances are of hitting certain edge cases. Also enumerating every subterm for every generated expression is going to take a bit of time. It's a easy test though so it makes sense before we have a better story here.

The two alternative approaches that come to mind are to try to make all of the forms aware of their payload so that it can be written in a more typesafe way by construction or to try to have a way to enumerate through all the forms and test them on examples rather than by random with a PBT (this depends on whether there were some other edge cases you were trying to test).

Negabinary:
> On the actual purpose of the PR I'm curious if this is mostly aiming to address the format string issue or if there's a litany of unsafe things we're trying to catch. 

iirc the main issues I've run into before are to do with segment construction in the explanation and example; format string is also an issue though.

7h3kk1d:
> > On the actual purpose of the PR I'm curious if this is mostly aiming to address the format string issue or if there's a litany of unsafe things we're trying to catch.
> 
> iirc the main issues I've run into before are to do with segment construction in the explanation and example; format string is also an issue though.

Cool, let's just address the surface stuff above then and get this merged in. Cleaning up the ExplainThis code can be a separate project. If we ever make this more correct by construction we should try to remove this test though so we're not paying the cost.

test/Test_ExplainThis.re
Negabinary: not sure I understand this one

Negabinary:
> > > On the actual purpose of the PR I'm curious if this is mostly aiming to address the format string issue or if there's a litany of unsafe things we're trying to catch.
> > 
> > 
> > iirc the main issues I've run into before are to do with segment construction in the explanation and example; format string is also an issue though.
> 
> Cool, let's just address the surface stuff above then and get this merged in. Cleaning up the ExplainThis code can be a separate project. If we ever make this more correct by construction we should try to remove this test though so we're not paying the cost.

cool; sounds like a plan

7h3kk1d (approved):

PR #2415, opened 2026-07-30
Splices

Negabinary:
Makes the tables editable (while introducing splice architecture)

 - Cells in the table are splices
 - Table automatically resizes to fit all cells
 - Columns can be re-named, but only if the names don't come from analytic type
 - Right-click inside a cell to add/remove rows/columns

### DIFF
--- a/src/haz3lcore/CachedSyntax.re
+++ b/src/haz3lcore/CachedSyntax.re
@@ -1,7 +1,14 @@
-type t = {
-  old: bool,
+type splice = {
   segment: Segment.t,
   measured: Measured.t,
+  /* A list of projector IDs in the order they appear in the segment
+   * (allows actions to refer to projectors by index) */
+  projector_list: list(Id.t),
+};
+
+type t = {
+  old: bool,
+  main_splice: splice,
   selection_ids: list(Id.t),
   /* The term-derived data structured below, may differ
    * from the term used for semantics. These terms are identical when
@@ -18,8 +25,6 @@ type t = {
    * certain ids to be present/non-present unexpectedly. */
   term_data: TermData.t,
   terms: TermMap.t,
-  /* A list of projector IDs in the order they appear in the segment
-   * (allows actions to refer to projectors by index) */
   projector_list: list(Id.t),
   /* Since the introduction of shape_map below, caching projectors
    * here is almost vesigial (currently used only for error deco) */
@@ -62,9 +67,12 @@ let mk = (~info_map, ~dyn_map, ~elaborated=None, z): t => {
     Measured.of_segment(segment, projector_shapes, refractor_shape_map);
   {
     old: false,
-    segment,
+    main_splice: {
+      segment,
+      measured,
+      projector_list,
+    },
     term_data,
-    measured,
     selection_ids: Selection.selection_ids(z.selection),
     terms,
     projectors,
@@ -104,12 +112,19 @@ let refresh_shapes =
     );
   let refractor_shape_map = Id.Map.empty;
   let measured =
-    Measured.of_segment(old.segment, shape_map, refractor_shape_map);
+    Measured.of_segment(
+      old.main_splice.segment,
+      shape_map,
+      refractor_shape_map,
+    );
   {
     ...old,
+    main_splice: {
+      ...old.main_splice,
+      measured,
+    },
     shape_map,
     projector_errors,
-    measured,
     shape_info_map: info_map,
     shape_dyn_map: dyn_map,
     shape_elaborated: elaborated,
@@ -145,3 +160,6 @@ let calculate = (z: Zipper.t, info_map, dyn_map, ~elaborated=None, old: t) =>
       selection_ids: Selection.selection_ids(z.selection),
     };
   };
+
+let measured = (syntax: t) => syntax.main_splice.measured;
+let segment = (syntax: t) => syntax.main_splice.segment;
--- a/src/haz3lcore/CachedSyntax.re
+++ b/src/haz3lcore/CachedSyntax.re
@@ -41,6 +41,11 @@ type t = {
   shape_map: ProjectorCore.Shape.Map.t,
   /* Errors reported by projectors (e.g. "can't render as table") */
   projector_errors: Id.Map.t(ProjectorBase.error),
+  /* splice id → (host projector id, row offset of the splice within
+   * the host's placeholder block). Computed alongside shape_map from
+   * each projector's splice_rows; lets main-frame decorations find the
+   * document row a splice's contents are laid out on. */
+  splice_layout: Id.Map.t((Id.t, int)),
   cached_backpack: list(Tile.t),
   /* Inputs last used to compute shape_map/projector_errors/measured.
    * Kept so `calculate` can detect when statics changed and refresh
@@ -84,7 +89,7 @@ let mk = (~info_map, ~dyn_map, ~elaborated=None, z): t => {
   let segment = Zipper.unselect_and_zip(z);
   let MakeTerm.{term: _, terms, projectors, projector_list, term_data} =
     MakeTerm.go(segment);
-  let (projector_shapes, projector_errors) =
+  let (projector_shapes, splice_layout, projector_errors) =
     ProjectorInfo.ShapeMapSemantics.mk(
       projectors,
       z.refractors,
@@ -115,6 +120,7 @@ let mk = (~info_map, ~dyn_map, ~elaborated=None, z): t => {
     projector_list,
     shape_map: projector_shapes,
     projector_errors,
+    splice_layout,
     cached_backpack: Segment.global_missing_shards(segment),
     shape_info_map: info_map,
     shape_dyn_map: dyn_map,
@@ -138,7 +144,7 @@ let mark_old: t => t =
  * elaborated expression (e.g. TableProj placeholder size). */
 let refresh_shapes =
     (z: Zipper.t, info_map, dyn_map, ~elaborated=None, old: t) => {
-  let (shape_map, projector_errors) =
+  let (shape_map, splice_layout, projector_errors) =
     ProjectorInfo.ShapeMapSemantics.mk(
       old.projectors,
       z.refractors,
@@ -162,6 +168,7 @@ let refresh_shapes =
     splices: mk_splice_map(old.main_splice.segment, shape_map),
     shape_map,
     projector_errors,
+    splice_layout,
     shape_info_map: info_map,
     shape_dyn_map: dyn_map,
     shape_elaborated: elaborated,
@@ -203,13 +210,51 @@ let segment = (syntax: t) => syntax.main_splice.segment;
 let splice_opt = (id: Id.t, syntax: t): option(splice) =>
   Id.Map.find_opt(id, syntax.splices);
 
-/* Whether the piece/term [id] lives inside some splice's content.
+/* The splice whose content contains the piece/term [id], if any.
  * Splice interiors are merged into the main measured map in their own
  * (splice-local) coordinate frame, so main-frame decorations must not
- * position themselves from such ids — the owning splice's sub-editor
- * is the frame that can. */
-let id_in_splice = (id: Id.t, syntax: t): bool =>
-  Id.Map.exists(
-    (_, s: splice) => Measured.find_by_id_quiet(id, s.measured) != None,
-    syntax.splices,
+ * position themselves directly from such ids — see doc_row_of_splice
+ * for the vertical remap. */
+let splice_containing_id = (id: Id.t, syntax: t): option((Id.t, splice)) => {
+  let candidates =
+    Id.Map.bindings(syntax.splices)
+    |> List.filter(((_, s: splice)) =>
+         Measured.find_by_id_quiet(id, s.measured) != None
+       );
+  /* Splice contents are measured recursively, so an id inside a nested
+   * splice is found in every enclosing splice's map too; the id's own
+   * frame is the innermost candidate — the one containing no other. */
+  List.find_opt(
+    ((_, s: splice)) =>
+      !
+        List.exists(
+          ((sid, _)) => Measured.has_splice_info(sid, s.measured),
+          candidates,
+        ),
+    candidates,
   );
+};
+
+/* The document row on which the splice's contents begin: the host
+ * projector's origin row, plus the splice's row offset within the
+ * host's block (splice_layout), climbing recursively when the host
+ * itself sits inside another splice. This is a grid-level estimate of
+ * the projector's DOM layout — exact horizontal positions inside the
+ * projector remain unknowable at this level. */
+let rec doc_row_of_splice = (sid: Id.t, syntax: t): option(int) => {
+  open Util.OptUtil.Syntax;
+  let* (host, offset) = Id.Map.find_opt(sid, syntax.splice_layout);
+  let+ host_row =
+    switch (splice_containing_id(host, syntax)) {
+    | Some((outer_sid, outer)) =>
+      let* m = Measured.find_by_id_quiet(host, outer.measured);
+      let+ outer_row = doc_row_of_splice(outer_sid, syntax);
+      outer_row + m.origin.row;
+    | None =>
+      Option.map(
+        (m: Measured.measurement) => m.origin.row,
+        Measured.find_by_id_quiet(host, syntax.main_splice.measured),
+      )
+    };
+  host_row + offset;
+};
--- a/src/haz3lcore/CachedSyntax.re
+++ b/src/haz3lcore/CachedSyntax.re
@@ -202,3 +202,14 @@ let measured = (syntax: t) => syntax.main_splice.measured;
 let segment = (syntax: t) => syntax.main_splice.segment;
 let splice_opt = (id: Id.t, syntax: t): option(splice) =>
   Id.Map.find_opt(id, syntax.splices);
+
+/* Whether the piece/term [id] lives inside some splice's content.
+ * Splice interiors are merged into the main measured map in their own
+ * (splice-local) coordinate frame, so main-frame decorations must not
+ * position themselves from such ids — the owning splice's sub-editor
+ * is the frame that can. */
+let id_in_splice = (id: Id.t, syntax: t): bool =>
+  Id.Map.exists(
+    (_, s: splice) => Measured.find_by_id_quiet(id, s.measured) != None,
+    syntax.splices,
+  );
--- a/src/haz3lcore/CachedSyntax.re
+++ b/src/haz3lcore/CachedSyntax.re
@@ -235,6 +235,18 @@ let splice_containing_id = (id: Id.t, syntax: t): option((Id.t, splice)) => {
   );
 };
 
+/* Whether the frame identified by [frame] (None = the root editor,
+ * Some(sid) = splice sid's sub-editor) owns the on-screen position of
+ * [id]: an id inside a splice is positioned only in that splice's own
+ * frame, so other frames must not draw decorations anchored on it. */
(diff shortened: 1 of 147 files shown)


7h3kk1d:
Just pulled it up for functional testing. Seems great so far.

src/haz3lcore/zipper/action/Move.re
```diff
@@ -2,22 +2,73 @@ open Util;
...
     z |> Caret.set(Outer) |> Option.some
   | (Inner(char), None | Some(_)) =>
     z |> Caret.set(Inner(char - 1)) |> Option.some
   };
 
-let by_char_right = (z: t): option(t) =>
+let by_char_right_with = (mv, z: t): option(t) =>
   switch (z.caret, Caret.nhbr_max_idx(Right, z)) {
-  | (Outer, None) => move(Right, z)
+  | (Outer, None) => mv(Direction.Right, z)
```
disconcision: unnecessary qualification?

src/haz3lcore/zipper/action/Move.re
```diff
@@ -186,11 +246,170 @@ let vertical =
...
+ * publishing splice offsets, this hook will enable direct splice
+ * clicks to translate the goal point into splice-local coordinates. */
+let splice_at_point =
+    (
+      ~splice_offsets: Id.Map.t(Point.t)=Id.Map.empty,
+      ~measured: Measured.t,
+      goal: Point.t,
+    )
+    : option((Id.t, Point.t)) => {
+  ignore(measured);
```
disconcision: ?

disconcision:
@Negabinary is this at a point where code review would be useful? is there an architectural summary somewhere?

7h3kk1d:
Note changing projector from piece to segment breaks persistence so we need to warn people to reset state. I think I'm going to tackle #2454 at some point this year.

Negabinary:
> @Negabinary is this at a point where code review would be useful? is there an architectural summary somewhere?

@disconcision yes, I think this is at a point where code review would be useful. The basic idea is that there is just one zipper which handles all the state; CodeEditables now have multiple CachedSyntaxes for each splice. It also adds a few random extra things like splices being able to add actions to the right-click menu of their editor (for add table row/column)

One decision that may want to be revisited at some point: I think the table projector mostly currently interfaces with segment data structures, I'm pretty sure it is set up so you could have the table projector mostly interfacing with exps, not sure if that's a direction we'll ant to take in the future.

src/haz3lcore/tiles/Base.re
```diff
@@ -106,6 +119,15 @@ and piece_to_string =
...
     )
+  | Splice(s) =>
+    segment_to_string(
+      ~holes,
+      ~concave_holes,
+      ~refractors,
+      ~refractor_seg_to_seg,
+      ~projector_to_segment,
+      s.content,
+    )
```
7h3kk1d, line 130: I need to understand why no splice-oriented structure gets added. Where does the splice id go for example.

src/haz3lcore/projectors/ProjectorBase.re
```diff
@@ -24,6 +26,7 @@ type external_action =
   | Escape(Util.Direction.t) /* Pass focus to parent editor */
   | EscapeToLineEnd(ProjectorCore.Kind.t) /* Pass focus to parent editor, move to end of line */
   | SetSyntax(Base.segment) /* Set underlying syntax */
+  | SetTerm(Any.t, bool) /* Set underlying term, optionally preserving original splices */
```
7h3kk1d, line 29: Don't want to be finnicky but I think it would be nice to have a record here for the bool so we can see what it's for.

7h3kk1d: Are the original splices found from the term-level splices with the same id?

7h3kk1d: Followup, is there a mechanism for handling duplicating of a splice.

src/haz3lcore/projectors/ProjectorBase.re
```diff
@@ -136,10 +162,21 @@ module View = {
...
+  /* Thunk returning the view for a splice by its id. Thunked so that
+   * expensive splice rendering only happens for splices the projector
+   * actually places in its view. */
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type splice_view = Id.t => Node.t;
+
+  /* Intrinsic bounding box size of a splice's content, in the splice's
+   * own coordinate frame. */
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type splice_size = Id.t => Util.Point.t;
```
7h3kk1d, line 178: Also pedantic but I don' like using Point.t as a size as opposed to absolute position.

src/haz3lcore/projectors/ProjectorBase.re
```diff
@@ -237,12 +290,27 @@ module type Projector = {
...
-  /* The space left for the projector in the base editor */
-  let placeholder: (model, info) => ProjectorCore.Shape.t;
+  /* The space left for the projector in the base editor. The projector
+   * receives [splice_size] so that it may size itself around its splice
+   * children; splice sizes are intrinsic (independent of the layout of
+   * the parent projector) and may be looked up by splice id. */
+  let placeholder: (model, info, View.splice_size) => ProjectorCore.Shape.t;
+  /* The vertical position (row offset from the top of the placeholder
+   * block) at which each of the projector's direct splices is laid out
+   * on screen. Lets frame-level decorations (e.g. a probe's offside
```
7h3kk1d, line 300: Playing with this it seems like it doesn't let you update probes that are in different splices on the same row. It just doesn't update if you try to add a probe on a table cell when you already have one.

src/language/term/Pat.re
```diff
@@ -57,7 +57,8 @@ let cls_of_term: Grammar.pat_term('a) => cls =
   | TupLabel(_) => TupLabel
   | Tuple(_) => Tuple
   | Parens(_) => Parens
-  | Projector(_) => Projector
+  | Projector(_)
+  | Splice(_) => Projector
```
7h3kk1d, line 61: This isn't how Exp worked. I wonder if this should be cls of the underlying.

src/language/term/Typ.re
```diff
@@ -94,7 +94,8 @@ let cls_of_term: Grammar.typ_term('a) => cls =
   | Label(_) => Label
   | ExplicitNonlabel => ExplicitNonlabel
   | Parens(_) => Parens
-  | Projector(_) => Projector
+  | Projector(_)
+  | Splice(_) => Projector
```
7h3kk1d, line 98: Same question as Pat.

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -1129,37 +1132,41 @@ and typ_term: unsorted => (Typ.term, list(Id.t)) = {
...
-        | (["(", ")"], [Typ(body)]) => Parens(body)
-        | (["PROJ_WRAP", "PROJ_WRAP"], [Typ(body)]) => body.term
-        | (["[", "]"], [Typ(body)]) => List(body)
-        | ([t], []) when is_hole_label(t) => hole(tm)
-        | ([t], []) => Unknown(Hole(Invalid(t)))
-        | _ => hole(tm)
-        },
-      )
+      switch (tile) {
+      | (["SPLICE_WRAP", "SPLICE_WRAP"], [Typ(body)]) => ret(body.term)
```
7h3kk1d, line 1136: Can this be moved below next to the PROJ_WRAP?

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -3112,3 +3140,64 @@ let any_to_segment =
...
+      switch (p) {
+      | Splice(s) => collect_splices(Id.Map.add(s.id, p, acc), s.content)
+      | Tile(t) => List.fold_left(collect_splices, acc, t.children)
+      | Projector(pr) => collect_splices(acc, pr.syntax)
+      | Grout(_)
+      | Secondary(_) => acc
+      },
+    acc,
+    seg,
+  );
```
7h3kk1d, line 3156: I think Segment.splices already does this but returns a list. It would probably be better to consume that and build the map from it. I think it's just
```reason
let splice_map = (seg: Segment.t): Id.Map.t(Segment.t) =>
  Segment.splices(seg)
  |> List.map((s: Base.splice) => (s.id, s.content))
  |> Id.Map.of_list;
```
Similarly
ProjectorView.collect_splices and ProjectorView.has_splices are similar. If we don't want to allocate lists we can use Seq.

7h3kk1d: The accumulator in the existing function is also always empty.

src/haz3lcore/tiles/Base.re
7h3kk1d: This is a self-comment so feel free to ignore.

src/haz3lcore/lang/MakeTerm.re
disconcision: FYI this hacky wrap crap finally dies in the 'make tiles less illegal' PR

src/regen/regen_docs.ml
7h3kk1d, line 1: I think this can probably all go now that the docs are plaintext on dev

src/haz3lcore/zipper/Ancestor.re
```diff
@@ -15,74 +16,166 @@ type t = {
...
+let nibs = (a: t): (Nib.t, Nib.t) =>
+  switch (a) {
+  | Tile(a) => tile_nibs(a)
+  | Projector(_)
+  | Splice(_) =>
+    /* Projector and Splice ancestors are opaque Convex/Convex. */
+    Nib.(
+      {
+        shape: Convex,
+        sort: Any,
```
7h3kk1d, line 92: When do we care about sort for these fields?

src/haz3lcore/zipper/Ancestor.re
```diff
@@ -15,74 +16,166 @@ type t = {
...
-     * and previously fell back to Any. */
-    l.sort;
-  | None => raise(Empty_shard_affix)
+/* Zip a child segment back up into the parent piece.
+ * - Tile: wrap child in the tile at the caret position.
+ * - Splice: produce a Splice piece whose content is the child.
+ * - Projector: produce a Projector piece whose syntax is before ++ child ++ after.
+ *   (This is used when zipping a Projector frame whose child is the already-
+ *   zipped Splice segment; [before]/[after] surround it in projector syntax.) */
+let zip = (child: Segment.t, a: t): Base.piece =>
```
7h3kk1d, line 107: Not sure exactly what goes in Ancestor vs zipper but this seems more like a zipper function

src/haz3lcore/zipper/Ancestor.re
```diff
@@ -15,74 +16,166 @@ type t = {
...
+  switch (a) {
+  | Tile(a) =>
+    let (pre, suf) = a.shards;
+    switch (ListUtil.split_last_opt(pre), suf) {
+    | (Some((_, i)), [_, ..._]) =>
+      let (_, l) = Mold.nibs(~index=i, a.mold);
+      l.sort;
+    | _ => raise(Empty_shard_affix)
+    };
+  | Projector(_) => Sort.Any
```
7h3kk1d, line 134: Same sort question as above.

src/haz3lcore/zipper/action/Move.re
```diff
@@ -27,39 +78,48 @@ let by_char_right = (z: t): option(t) =>
...
   | ByChar
-  | BySmart => by_char(d, z)
+  | BySmart => by_char_with(mv, d, z)
   };
 };
 
+/* Confined movement: geometric scans, extremes, jumps. */
+let by_char = by_char_with((d, z) => move(d, z));
+let by_token = by_token_with((d, z) => move(d, z));
+let local = local_with((d, z) => move(d, z));
```
7h3kk1d, line 118: I'm confused as to why `mv` was parameterized

src/haz3lcore/zipper/action/Move.re
```diff
@@ -186,11 +246,170 @@ let vertical =
...
+        switch (Id.Map.find_opt(id, measured.splices)) {
+        | None => None
+        | Some(info) =>
+          let contains =
+            goal.row >= offset.row
+            && goal.row <= offset.row
+            + info.size.row
+            && goal.col >= offset.col
+            && goal.col <= offset.col
+            + info.size.col;
```
7h3kk1d, line 284: Without claude I definitely would think we need a bounding box type and a contains function. This is fine if you don't think it will make mistakes on it in the future.

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -19,6 +19,26 @@ let go =
       {zipper: z, col_target}: state,
     )
     : Action.Result.t(Zipper.t) => {
+  /* Splice switcheroo: when the caret is inside a splice, swap that
```
7h3kk1d, line 22: > Splice switcheroo

An Ocean's 11 level heist

src/haz3lcore/CachedSyntax.re
```diff
@@ -145,3 +204,69 @@ let calculate = (z: Zipper.t, info_map, dyn_map, ~elaborated=None, old: t) =>
...
+        ),
+    candidates,
+  );
+};
+
+/* Whether the frame identified by [frame] (None = the root editor,
+ * Some(sid) = splice sid's sub-editor) owns the on-screen position of
+ * [id]: an id inside a splice is positioned only in that splice's own
+ * frame, so other frames must not draw decorations anchored on it. */
+let frame_owns_id = (frame: option(Id.t), id: Id.t, syntax: t): bool =>
```
7h3kk1d, line 242: Is frame here synonymous with splice/root editor?

src/haz3lcore/projectors/ProjectorInit.re
```diff
@@ -19,20 +19,52 @@ let to_module = (kind: ProjectorCore.Kind.t): (module Cooked) =>
...
+ * module initialization. Resolving a Term override requires
+ * ExpToSegment, but this module cannot depend on ExpToSegment: it is
+ * reachable from ExpToSegment via MakeTerm -> ... -> Refractors ->
+ * ProjectorInit. ProjectorPerform sits above both and registers the
+ * real printer (the same conversion SetTerm uses, reusing splices
+ * from the original syntax by id). If unregistered, Term overrides
+ * degrade to keeping the selected syntax. */
+let term_printer:
+  ref((~original_syntax: Base.segment, Language.Any.t) => Base.segment) =
+  ref((~original_syntax, _term) => original_syntax);
```
7h3kk1d, line 32: Can this not just be dependency injected in?

src/haz3lcore/projectors/ProjectorPerform.re
```diff
@@ -98,6 +143,72 @@ let update =
     : ZipperBase.t =>
   ZipperBase.MapPiece.fast_local_seg(update_piece(f, id), id, z);
 
+let inside_projector = (id: Id.t, z: Zipper.t): bool =>
```
7h3kk1d, line 146: I wonder if some of these should be in zipper.

src/web/app/common/ProjectorView.re
```diff
@@ -154,21 +167,55 @@ module Model = {
     + offset
     - measurement.origin.col;
 
+  /* Whether a statics error is anchored in the projector's own hidden
+   * syntax — pieces between its splices, like a table row's comma —
+   * which is rendered as projector chrome rather than laid-out text.
+   * Such errors get no arms anywhere (see Arms.tiles_data), so the
+   * projector chrome reports them. Splice contents report errors in
+   * their own sub-editors, and nested projectors on their own chrome. */
+  let has_hidden_error =
```
7h3kk1d, line 176: nice, does this mean we get fold errors now?

src/haz3lcore/zipper/action/Move.re
7h3kk1d: robot is telling me this is dead code. I haven't verified.

matthewhammer:
@Negabinary — a heads-up that we're merging this branch, and a question about whether anything from you would help. **Nothing here is blocking on you**, and I'm not asking you to reprioritise: we're set up to proceed either way, and I'd rather work independently than add to your plate.

## What we're doing

User-defined livelits are our current critical path. There are two branches:

- `integration/livelits` (#2576) — Modules II parts 1-3 + the HTML substrate + the livelit stack. It adds a builtin module signature type `Livelit` and checks each livelit definition against it, so `expand : Model -> Expansion` is verified at the definition site rather than only per use.
- `integration/livelits-splicerefs` — the same, plus this branch, plus the paper's `SpliceRef`. Currently it's an empty placeholder: identical to the first, because the merge is where we stopped.

The goal on the second one is PLDI 2021's `expand : Model -> (Exp, List(SpliceRef))`. Worth stating plainly so we're not talking past each other: your splices and the paper's `SpliceRef` are different things. This branch gives `Splice` in the grammar (all three sorts), `SubEditor.re`, and the projector `splice_view`/`splice_size` API — the editable-region machinery. `SpliceRef` as a *type* that `expand` hands back appears nowhere in the repo (grepping `SpliceRef|splice_ref` here returns nothing). So for us this is a merge plus a build, and only the merge involves you.

## What we found trying the merge

Merging `splices-11` into our branch gives 70 conflicted files:

- 32 are generated B2T2 `.ml` slides that `dev` deleted in 53401e4158 ("Slides load from .hz text", 2026-08-08) — one day after this branch's merge-base. Those resolve cleanly by accepting the deletion.
- 38 are content conflicts, 87 regions. We resolved `Typ.re` and `Equality.re` by hand (keeping our physical-equality short-circuits, adding your `Splice` cases, keeping the `ignore_projectors`/`ignore_parens` split that post-dates this branch).

Of the 76 regions left, **only 3 are purely splice-related; 73 carry other new content**, in `Zipper`, `Piece`, `Segment`, `Perform`, `MakeTerm`, `Measured`, `CachedSyntax`, `ProbePerform` and the `Projector*` modules. For each we have to decide whether your version is splice design to keep or code `dev` has since moved past, and we often can't tell. Two examples that bracket the problem:

- `main_splice` — 10 uses here, zero on `dev` and zero on ours. Clearly yours and load-bearing; we keep it.
- `FocusEffect` — 2 uses on `dev` and ours, 8 here. Either you extended it or `dev` moved; from the outside we can't say.

The branch is 428 commits behind `dev` (merge-base 6b611ba3be, 2026-08-07), which is what makes 73 of these ambiguous rather than 3.

## What would help, and what isn't needed

**Necessary: nothing.** We can do the merge as it stands. It's slow and the failure mode is quiet — we just found a 14x performance regression on the other branch that all 4003 tests passed, so we're wary of judgment calls we can't verify — but it is doable, and we're not waiting on you to start.

**Most helpful, if it's something you wanted to do anyway:** rebase `splices-11` onto current `dev`. That collapses those 73 ambiguous regions into roughly the 3 real splices-vs-livelits conflicts, because you'd be resolving your own drift with knowledge we don't have. If the branch is going to land eventually it probably wants this regardless of us.

**Cheaper and nearly as useful:** don't rebase, just tell us which of your changes outside `SubEditor.re` and the projector API are load-bearing for splices versus incidental to the older `dev` you branched from. Even a rough list of files would turn most of our 73 guesses into lookups. `ProbePerform`, `CachedSyntax`, `Measured` and `Piece`/`Segment` are where we're least sure.

**Also useful and free:** if you have a view on how `SpliceRef` should relate to your `Splice` terms — whether `expand` returning splice references should produce your `Splice` wrappers directly, or something that elaborates into them — we'd rather build toward what you have in mind than guess and diverge.

No rush on any of it. If we don't hear back we'll proceed with the merge and flag anything where we clearly guessed.

PR #2417, opened 2026-07-31
Single Source of Truth For Undo

Negabinary:
Every Update module carried two encodings of "should this action be undoable": the `historic` field each update returns via `Updated.return`/`return_quiet`, and a static `can_undo(action)` predicate mirroring the exact same delegation tree. Only `can_undo `was ever consulted — so the historic flags had silently rotted. 

I [Claude] ran a branch-by-branch audit of all ~30 modules (four parallel agents) comparing the two, then fixed `historic` to return the correct flag everywhere before deleting the tree.

### DIFF
--- a/src/web/view/CodeExerciseMode.re
+++ b/src/web/view/CodeExerciseMode.re
@@ -314,20 +314,14 @@ module Update = {
       (~settings: Settings.t, action: instructor, model: Model.t)
       : Updated.t(Model.t) =>
     if (settings.instructor_mode) {
-      instructor_update(action, model);
+      {
+        /* Instructor form edits are not undoable */
+        ...instructor_update(action, model),
+        historic: false,
+      };
     } else {
       Updated.return_quiet(model);
     };
-
-  let can_undo = (action: t) => {
-    switch (action) {
-    | Editor(_, action) => CellEditor.Update.can_undo(action)
-    | RefreshStatics => false
-    | ResetEditor(_) => true
-    | ResetExercise => true
-    | Instructor(_) => false
-    };
-  };
 
   let update =
       (~settings: Settings.t, ~schedule_action as _, action, model: Model.t)
--- a/src/web/view/CodeExerciseMode.re
+++ b/src/web/view/CodeExerciseMode.re
@@ -314,11 +314,7 @@ module Update = {
       (~settings: Settings.t, action: instructor, model: Model.t)
       : Updated.t(Model.t) =>
     if (settings.instructor_mode) {
-      {
-        /* Instructor form edits are not undoable */
-        ...instructor_update(action, model),
-        historic: false,
-      };
+      instructor_update(action, model);
     } else {
       Updated.return_quiet(model);
     };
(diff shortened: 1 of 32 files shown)


Negabinary:
Tests

Negabinary:
@SirQuackyy Cyrus mentioned you're working on some undo-related stuff; feel free to have a look and let me know if this interacts with your work.

Negabinary:
> Tests

I'm not sure I'm convinced tests are actually worth adding to this pr - I can't see what they'd add other than duplication

src/web/view/CodeExerciseMode.re
```diff
@@ -314,21 +314,15 @@ module Update = {
       (~settings: Settings.t, action: instructor, model: Model.t)
       : Updated.t(Model.t) =>
     if (settings.instructor_mode) {
-      instructor_update(action, model);
+      {
+        /* Instructor form edits are not undoable */
+        ...instructor_update(action, model),
+        historic: false,
+      };
```
7h3kk1d: but whyyy

src/web/view/DerivationExerciseMode.re
```diff
@@ -280,7 +270,11 @@ module Update = {
       (~settings: Settings.t, action: instructor, model: Model.t)
       : Updated.t(Model.t) =>
     if (settings.instructor_mode) {
-      instructor_update(action, model);
+      {
+        /* Instructor form edits are not undoable */
+        ...instructor_update(action, model),
+        historic: false,
+      };
```
7h3kk1d: also why

src/web/view/TheoremExerciseMode.re
```diff
@@ -188,7 +188,11 @@ module Update = {
       (~settings: Settings.t, action: instructor, model: Model.t)
       : Updated.t(Model.t) =>
     if (settings.instructor_mode) {
-      instructor_update(action, model);
+      {
+        /* Instructor form edits are not undoable */
+        ...instructor_update(action, model),
+        historic: false,
```
7h3kk1d: tres why

7h3kk1d (changes requested):
left questions on instructor mode but otherwise lgtm

src/web/view/CodeExerciseMode.re
Negabinary: "Because that preserves the behavior the app already had"

src/web/view/DerivationExerciseMode.re
Negabinary: why not

src/web/view/TheoremExerciseMode.re
Negabinary: mais whyever not

src/web/view/CodeExerciseMode.re
7h3kk1d: If the previous version of the app jumped off a bridge...

7h3kk1d (approved):

PR #2419, opened 2026-08-02
negative number literals

### LINKED ISSUE #719: fix floating point / unary minus issue

### LINKED ISSUE #748: negative floats
negative floats for haz3l
(-.3.14) ?? currently does subtraction

disconcision:
closes #719
closes #748

specifically float negation and negative literal patterns. `-1.5` was a static type error ("Expecting type Int") and `case x | -1 => ...` didn't parse. statics now re-kinds unary minus the same way as. patterns now support prefix minus for both ints and floats and MakeTerm folds it into the literal atom.


7h3kk1d:
ExplainThis still calls the unary `-` Integer negation.

`use` is also behaving oddly (I mean it is always odd but y'know)

https://github.com/user-attachments/assets/27db6bb8-2d1d-4930-ae78-4e8d35be0670

src/haz3lcore/lang/MakeTerm.re
```diff
@@ -1101,6 +1101,24 @@ and pat_term: unsorted => (Pat.term, list(Id.t)) = {
...
+         and a matched value is a plain negative atom, so fold the minus
+         into the literal. The literal's ids are adopted, as in ListLit
+         absorption. Non-literal operands (`-x`) stay holes. */
+      switch (r) {
+      | {term: Atom(Int(n)), annotation: {ids, _}} =>
+        adopted_ids := ids @ adopted_ids^;
+        (Atom(Int(Bigint.neg(n))), ids);
+      | {term: Atom(Float(f)), annotation: {ids, _}} =>
+        adopted_ids := ids @ adopted_ids^;
+        (Atom(Float(-. f)), ids);
```
7h3kk1d, line 1120: This relies on the fact that float negation reassociates with parsing so `parse(-x) == -parse(x)`. I made a local pbt to confirm this is true but we may want to note it in the comment. Everything FP scares me.

7h3kk1d (changes requested):

src/haz3lcore/lang/MakeTerm.re
disconcision: noted

disconcision:
@7h3kk1d explainthis updated, `use` improved somewhat

7h3kk1d:
Trying to fix the builds so I can test this without building locally.

Negabinary:
If you type `-5` and then add `use Float in` before you get this
image

and if you click anywhere you get a Not_found crash

disconcision:
@Negabinary looks like a dev bug; fix here #2451

src/language/statics/Statics.re
```diff
@@ -711,6 +711,21 @@ and uexp_to_info_map =
       );
     | UnOp(op, e) =>
       let op = Operators.replace_un_op(op, ctx.use_mode); // Replace op if necessary due to `use`
+      /* Re-kind numeric negation by the analyzed class, else by the
+         operand's own class (peeked with a discarded pass) — mirrors
+         replace_literal, so `-1.5` and `let x: Float = -5` type as float
+         negation instead of locking the operand to Int. */
+      let op =
+        switch (Typ.is_ana_atom(ana)) {
+        | Some(_) as ana_cls => Operators.replace_un_op_cls(op, ana_cls, None)
```
Negabinary: I don't like the fact that this does two replacements, can we fold all the logic into just replace_un_op_cls?

src/language/term/Operators.re
```diff
@@ -157,6 +157,35 @@ let replace_un_op = (op: op_un, use_mode: option(mode)): op_un => {
   };
 };
 
+/* Re-kind a numeric unary op the way replace_literal re-kinds literals:
+   the analyzed class wins (so `let x: Float = -5` is float negation and
+   the literal beneath retypes); with no atom analysis, an operand of
+   definite class decides (bare `-1.5`, `-x`; the Int arm matters under
```
Negabinary: If we're doing this for unary operators, should we also do this for binary operators to be consistent?

disconcision:
@7h3kk1d issues addressed

src/language/statics/Statics.re
```diff
@@ -711,6 +711,21 @@ and uexp_to_info_map =
...
+        | None =>
+          let (peek, _, _) = go(~ana=syn, e, m);
+          Operators.replace_un_op_cls(
+            op,
+            None,
+            Typ.is_ana_atom(peek.elab_syn_ty),
+          );
```
7h3kk1d, line 727: I'm just staring at this right now trying to figure out why we need the syn as well. Will let you know when I'm done.

7h3kk1d: I think I'm close. The original Operators.replace_un_op replaces the negation operator with whatever the active `use` mode is. If there's a analyzed type the ana_cls is used to override the operand again (which I noticed Nat has unary negation for some reason prexisting?). If there is no ana it synthesizes the operand to pick the right negation for it.

So stuff like this blows up oddly 

Image

But the whole thing has become so complicated I'm not sure if it's worth worrying about. I could imagine always peeking, looking at the ana type, and the operand and doing one large pattern match on all of them and just picking a best-fit fall through. So I'm going to approve since I don't think there's necessarily a great answer here.

7h3kk1d (approved):

src/language/statics/Statics.re
disconcision: probably having the minus be part of the literal syntax rather than a separate operator would alleviate this and some related issues but it annoyingly breaks some current invariants in the insert logic so its not quick fix.

PR #2420, opened 2026-08-03
Add Reals

nskh:
adding a `use Reals in` form to support more arithmetic, especially fractions, in the Hazel Prover.


Negabinary (changes requested):
Looks like it supports rational numbers, and then just leaves a stuck expression if it needs to do anything irrational. Seems like a reasonable approach.

The one thing I'm a little scared of is the overloading approach, it looks like it could be a little ad-hoc. I don't know whether we've thought about whether we want different built-in names for reals and floats (which is what we do currently for int and sint).

nskh:
yeah agreed, removed the overloading. pi is still a Float and the Real version is now pi_real, like abs_int / abs_float. use Real in now just changes literals and the regular numeric operators. tests pass.

Issue #2204, opened 2026-03-31
make indexed db table name depend on git branch name


cyrus-:
- make sure hazel still builds when it isn't in a git repo as well

PR #2421, opened 2026-08-03
Add telemetry sections to the Debug sidebar (Evaluation, Statics, Editor & Memory, Frame Timing)

7h3kk1d:
Adds four Debug sidebar sections — **Evaluation**, **Statics**, **Editor &
Memory**, **Frame Timing** — showing where a keystroke's time goes and what the
editor currently holds. They share a table, a formatter and a heat map with
#2370's **Worker Messaging** panel, which moves onto the same pieces.

## Up front: refs and side effects, deliberately

The collectors are module-level refs, written by side-effecting calls at the
points being measured rather than values threaded through the component tree.
Deliberately: this is developer instrumentation — tangential to the codebase, off
a normal user's path, inert unless a panel is open — and doing it properly would
mean widening every `update` / `calculate` signature along the way to carry a
value only a debug panel reads. Worth revisiting if these panels ever become
load-bearing.

What a section touches is still closed: each collector and `PerfFormat` has a
`.rei`, each section is pinned to `DebugSection.S`, and the mutation stays inside
the collectors.

## What gets collected

**Stage timings, per keystroke.** The edit action's `perform` (update phase), then
each calculate-phase stage — statics + elaboration, the syntax-cache rebuild
(MakeTerm + Measured), cursor info, the ExplainThis color map — and the total. 30
frames kept; a stage that didn't run shows `—`, never a misleading `0`.

**The eval round trip, per request.** The latency the main thread sees
(`postMessage` → `onmessage`) beside the worker's own evaluator time, which it now
returns in a new `eval_time` on `ServerMessage.result` — so the gap between them
is queue + serialization + transfer. Plus the outcome, encoded payload sizes and a
restart count. 10 requests kept.

**Structural counts, per frame.** Info-map entries, error and warning counts, why
statics ran or didn't (recomputed / forced / deferred / cached), segment tokens,
tiles, rows, projectors, and live undo/redo/backpack depth.

Durations are `Core.Time_ns.Span`, sizes `Core.Byte_units`. Byte-exact heap sizes
are deliberately absent — walking the heap per frame is too expensive, and timings
plus counts answer the question.

## The changes, in reading order

1. **`util/Metrics.re`** (new) — the gate and bounded history every collector
   shares, as a functor; recorders run through `when_enabled`. Read first.
2. **`util/PerfMetrics.re`** — per-frame collector: the `frame` record, the
   `time_*` stage wrappers, the count recorders.
3. **`util/EvalMetrics.re`** — round-trip collector, keyed by request id so its
   rows line up with Worker Messaging's.
4. **`util/WorkerMetrics.re`** — pre-existing; moved onto the functor.
5. **`CodeEditable`, `CodeWithStatics`, `Page`, `History`** — where the timings
   are taken, and where gating syncs once per cycle.
6. **`util/WorkerServer.re`, `util/WorkerClient.re`** — evaluator time on the
   wire, and the recording around the request lifecycle.
7. **`sidebar/PerfFormat.re`** — the table vocabulary: a `column` carries header,
   tooltip and cell projection; `cell` is opaque; `view` assembles a body.
8. **`sidebar/*Section.re`** — one per panel: columns plus a `view` call.
9. **`sidebar/SidebarModel.re`, `Settings.re`** — sidebar settings store
   `debug_expanded`, so the whole Debug sidebar starts collapsed.
10. **`util/TimeUtil.re`** — `timed`, and the `span` alias that lets a deriving
    type carry a `Time_ns.Span`. The only file outside `src/web/`.

## Closed panels cost nothing

Each collector is gated on **its own** section being expanded, synced once per
cycle in `Page.Update.calculate`, and every section ships collapsed — so ordinary
editing pays nothing.

## Testing

`make dev` clean, `make test-quick` green (3311), and a manual pass over all five
sections in Chrome. `Test_WorkerServer` covers a `Time_ns.Span` surviving Marshal
+ `structuredClone` and its nanosecond json encoding.


src/web/util/PerfMetrics.re
```diff
@@ -0,0 +1,195 @@
...
+  mutable cursor_info: option(Core.Time_ns.Span.t),
+  mutable color_map: option(Core.Time_ns.Span.t),
+};
+let building: builder = {
+  perform: None,
+  statics: None,
+  syntax: None,
+  cursor_info: None,
+  color_map: None,
+};
```
7h3kk1d: Why mutable instead of just overriding fields. I can't imagine the copying costs that much.

src/web/util/PerfMetrics.re
```diff
@@ -0,0 +1,195 @@
...
+let history_limit = 30;
+let frames: ref(list(frame)) = ref([]); /* newest first */
+
+/* Cheap structural counts of the most recently calculated editor, the statics
+ * recompute mode, and undo/redo depth — not per-frame timeline data. */
+type counts = {
+  mutable info_map_entries: int,
+  mutable error_count: int,
+  mutable warning_count: int,
+  mutable statics_mode: string,
```
7h3kk1d: What is statics_mode.

src/web/util/PerfMetrics.re
7h3kk1d, line 1: Give this an rei file.

src/web/app/editors/code/CodeWithStatics.re
```diff
@@ -155,28 +155,55 @@ module Update = {
...
+      PerfMetrics.record_statics_counts(
+        ~info_map_entries=Id.Map.cardinal(statics.info_map),
+        ~error_count=List.length(statics.error_ids),
+        ~warning_count=List.length(statics.warning_ids),
+        ~statics_mode=
+          recompute
+            ? statics_mode == StaticsForce ? "forced" : "recompute"
+            : statics_mode == StaticsDefer ? "deferred" : "cached",
+      );
+    };
```
7h3kk1d: Can this logic be in PerfMetrics nad just pass in the necessary information?

src/web/app/editors/code/CodeWithStatics.re
```diff
@@ -155,28 +155,55 @@ module Update = {
...
+        )
+      );
+    if (PerfMetrics.enabled^) {
+      let syn = editor.syntax;
+      PerfMetrics.record_editor_counts(
+        ~segment_tokens=List.length(syn.segment),
+        ~tiles=Id.Map.cardinal(syn.measured.tiles),
+        ~rows=Measured.Rows.cardinal(syn.measured.rows),
+        ~projectors=List.length(syn.projector_list),
+        ~backpack=List.length(syn.cached_backpack),
```
7h3kk1d: Also can this be in PerfMetrics.

src/web/app/sidebar/EditorSection.re
```diff
@@ -0,0 +1,63 @@
...
+            "rebuild",
+            "Time to rebuild the syntax cache: MakeTerm (segment → term) + Measured (layout).",
+          ),
+          ("tokens", "Number of pieces (tokens) in the editor segment."),
+          ("tiles", "Number of distinct tiles in the measured layout."),
+          ("rows", "Number of rendered rows in the measured layout."),
+          ("proj", "Number of projectors in the segment."),
+        ]),
+        ...List.map(row(~max), frames),
+      ]),
```
7h3kk1d: Make a datatype that contains the column header, the tooltip description and the projection function from max/PerfMetrics.frame and use that data structure to build this table. Otherwise this list and the row function have to stay in sync which is fragile.

Then use that datastructure, max, and frames to build the table in it's own function.

7h3kk1d: Do something like that for all the tables we're building this way.

7h3kk1d:
Default the debug sidebar to everything collapsed. Otherwise there's too much navigation.

src/web/app/sidebar/EvaluationSection.re
7h3kk1d, line 1: Make sure all the sidebars are limited to the sidebar rei file.

src/web/app/sidebar/EvaluationSection.re
```diff
@@ -0,0 +1,81 @@
...
+            "Request id, shared with the Worker Messaging panel so rows correlate.",
+          ),
+          ("cells", "Number of cells (request entries) evaluated together."),
+          (
+            "status",
+            "Outcome: pending (awaiting response), ok, fail (evaluator error), or timeout.",
+          ),
+          (
+            "round-trip",
+            "Wall-clock from posting the request to receiving its result: worker queue + evaluation + result serialization + transfer. Dominated by evaluation, but not the worker's isolated evaluate() call.",
```
7h3kk1d: Is there a way to get the eval time separately from the full round trip.

src/web/app/sidebar/PerfFormat.re
7h3kk1d, line 1: Give this an rei and only include the external interface

src/web/app/sidebar/PerfFormat.re
```diff
@@ -0,0 +1,120 @@
...
+  | Some(x) => to_string(x)
+  };
+
+let span = (s: Core.Time_ns.Span.t): string =>
+  Core.Time_ns.Span.to_string_hum(~decimals=2, s);
+let span_str = opt_str(span);
+
+let bytes = (b: Core.Byte_units.t): string =>
+  Core.Byte_units.to_string_hum(b);
+let bytes_str = opt_str(bytes);
```
7h3kk1d: include fmt or something else indicative of it being stringified for display

src/web/app/sidebar/PerfFormat.re
```diff
@@ -0,0 +1,120 @@
...
+
+/* A header cell with hover text (title) explaining what the column measures. */
+let head = ((label, tooltip): (string, string)): Node.t =>
+  Node.td(
+    ~attrs=[clss(["perf-head"]), Attr.title(tooltip)],
+    [text(label)],
+  );
+
+let head_row = (cols: list((string, string))): Node.t =>
+  Node.tr(List.map(head, cols));
```
7h3kk1d: These are so specific I think the right answer is to have a data structure representation of the table defined in this file and have all of the formatting/building of the HTML handled inside this and not expose these.

src/web/app/sidebar/PerfFormat.re
```diff
@@ -0,0 +1,120 @@
...
+/* Truncate a possibly-long label; the full text goes on the cell's tooltip. */
+let truncate = (n: int, s: string): string =>
+  String.length(s) <= n ? s : String.sub(s, 0, n) ++ {|…|};
+
+/* The edit action label a frame's perform stage carries (— when none). */
+let action_of = (p: option((string, Core.Time_ns.Span.t))): string =>
+  switch (p) {
+  | None => {|—|}
+  | Some((a, _)) => a
+  };
```
7h3kk1d: Use opt_string.

src/web/util/EvalMetrics.re
7h3kk1d, line 1: rei file. And if there's anything we can do to enforce the similarities with PerfMetrics we should.

src/web/util/WorkerClient.re
```diff
@@ -58,15 +58,34 @@ let with_latest = (request_id, f) =>
...
+  worker##postMessage(encoded);
+};
 
 let fail_latest = latest => {
   clear_timeouts();
   latest_request := None;
+  switch (latest.metrics_id) {
+  | Some(id) when EvalMetrics.enabled^ => EvalMetrics.record_timeout(~id)
+  | _ => ()
+  };
```
7h3kk1d: Can the enabled checking happen inside EvalMetrics.

src/web/util/WorkerClient.re
```diff
@@ -58,15 +58,34 @@ let with_latest = (request_id, f) =>
...
+ * see WorkerServer.Active). Callers still deal in Request.t/Response.t.
+ *
+ * Encodes once: the Evaluation panel needs the posted byte length, and its
+ * latency clock must start as close to the post as possible (excluding any
+ * Worker Messaging benchmarking done earlier by record_request). */
+let post_evaluate =
+    (~metrics_id: option(int)=None, worker, request: Request.t) => {
+  let encoded = Active.encode_request(ClientMessage.Evaluate(request));
+  switch (metrics_id) {
+  | Some(id) when EvalMetrics.enabled^ =>
```
7h3kk1d: Can the enabled checking happen inside EvalMetrics.

src/web/util/WorkerClient.re
```diff
@@ -103,8 +124,29 @@ let setup_worker_message_handler = worker => {
              * can take tens of ms and must not delay evaluation latency. */
             latest.callbacks.on_result(response);
             switch (latest.metrics_id) {
-            | Some(id) => WorkerMetrics.record_response(id, msg)
             | None => ()
+            | Some(id) =>
+              if (WorkerMetrics.enabled^) {
```
7h3kk1d: Can the enabled checking happen inside EvalMetrics.

src/web/app/sidebar/PerfFormat.re
```diff
@@ -0,0 +1,192 @@
...
+    (spans: list(option(Core.Time_ns.Span.t))): Core.Time_ns.Span.t =>
+  List.fold_left(
+    (acc: Core.Time_ns.Span.t, s: option(Core.Time_ns.Span.t)) =>
+      switch (s) {
+      | None => acc
+      | Some(x) => Core.Time_ns.Span.compare(x, acc) > 0 ? x : acc
+      },
+    Core.Time_ns.Span.zero,
+    spans,
+  );
```
7h3kk1d: We need to start using Seq and fold for stuff like this with dropping the optionals. So much more composable and easier to read.

src/web/util/PerfMetrics.re
7h3kk1d: 🤖 Agreed. Both mutable records are gone: `builder` and `counts` collapsed into a single `ref(frame)` updated with `{...current^, …}`, plus a small `live` record for the values that only mean anything *now* (undo/redo/backpack) rather than per frame.

Removing the two-record split also deleted the 15-field hand-copy that `end_calc` used to spell out when committing a frame — it's `push({...frame, total: Some(total)})` now.

src/web/util/PerfMetrics.re
7h3kk1d: 🤖 It answered "did statics actually recompute this frame, and if not why not" — but as a string built by a nested ternary at the call site, which is why it read as nothing in the type. It's a documented variant in `PerfMetrics` now, `option`-wrapped so "no statics yet" renders `—` like every other absent metric:

```reason
/* Why statics did or didn't recompute on a frame — the Statics panel's `mode`
 * column. */
type statics_mode =
  | Recomputed  /* an edit landed and statics ran this frame */
  | Forced      /* the debounce timer fired and forced a run */
  | Deferred    /* an edit landed but the debounce postponed the run */
  | Cached;     /* nothing to redo — the cached statics were reused */
```

`PerfMetrics` derives which one happened; `StaticsSection` owns how it's labelled.

src/web/util/PerfMetrics.re
7h3kk1d: 🤖 Added `PerfMetrics.rei`: the `frame`/`live`/`statics_mode` types, `sync`, `history`, the `time_*` wrappers and the three count recorders. `current`, `empty_frame`, `add`, `time`, `push`, `update` and `enabled` are all private now.

src/web/app/editors/code/CodeWithStatics.re
7h3kk1d: 🤖 Moved. It's now `PerfMetrics.record_statics_counts(~recompute, ~mode, statics)`, taking the `CachedStatics.t` and deriving both the counts and the reported mode itself — the call site passes facts, not a label.

To hand over the mode by type rather than as two booleans I pulled the three-way throttle mode out of this file into `StaticsMode.t` (`Normal | Defer | Force`), so the collector can name the decision without depending on the editor. `StaticsDebounce` stays here; `CellEditor` was the only other place that named a constructor.

src/web/app/editors/code/CodeWithStatics.re
7h3kk1d: 🤖 Same treatment: `PerfMetrics.record_syntax_counts(editor.syntax)` takes the `CachedSyntax.t` and does its own `List.length` / `Id.Map.cardinal` / `Measured.Rows.cardinal`. The call site is one line, and the `if (PerfMetrics.enabled^)` wrapper around it is gone too — every recorder is self-gating now.

While moving it I noticed `frame.backpack` was written but never read (the section reads it from the live counts), so it moved to `live` and left the frame.

src/web/app/sidebar/EditorSection.re
7h3kk1d: 🤖 Done, and for all five tables. A column carries its header, the tooltip explaining what it measures, and the cell projection as one value, so the two can't drift apart:

```reason
type column('row) = {label: string, tooltip: string, cell: 'row => Node.t};

type row('data) =
  | Row('data)
  | Group(group);   /* a full-width label spanning every column */
```

`PerfFormat.table(~columns, rows)` emits the tooltipped header row itself, so a section supplies only its columns and its rows and builds no HTML at all. `Group` covers Worker Messaging's request/response separators — and its colspan now follows `List.length(columns)` rather than the hardcoded `"7"` it was.

The `action` column those three per-frame tables open with is `PerfFormat.action_column`, so its wording lives in one place instead of three copies.

src/web/app/sidebar/EvaluationSection.re
7h3kk1d: 🤖 Each of the five sections now has a `.rei` containing exactly `include DebugSection.S;`, so `title` + `view` are all that escapes — `row`, `status`, `columns`, `total_of`, `max_total` and the `wm_*` helpers are private.

Leaving this thread open on one question: I read "the sidebar rei" as `DebugSection.S`. `DebugSidebar`, `Sidebar` and `ProblemSidebar` are pre-existing and still have no interface of their own — say the word if that's what you meant and I'll constrain those too.

src/web/app/sidebar/EvaluationSection.re
7h3kk1d: 🤖 There is, and it's in. `ServerMessage.result` carries a new `eval_ms`, which the worker accumulates around `start_evaluation` and each `run_yielding_slice` and resets per request; the Evaluation table has an `eval` column beside `round-trip`, so the gap between them reads as queue + result serialization + transfer.

Slices are timed unconditionally — the worker can't see whether the panel is open, and two clock reads per 5000-step slice is noise against the slice — so there's no request flag for it. `round-trip` still means what it did: the latency the main thread (and the user) actually sees.

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 Added `PerfFormat.rei`, external surface only: the `column`/`row`/`group` types, `table`, the cell builders (`text_cell`, `int_cell`, `name_cell`, `label_cell`, `status_cell`, `heat_cell`), `action_column`, `max_span`, the `fmt_*` helpers and `note`/`empty`.

Private now: `head`, `head_row`, `heat_style`, `truncate`, `heat_floor_ms`, `heat_ceil_ms`.

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 Renamed: `span` → `fmt_span`, `bytes` → `fmt_bytes`, `opt_str` → `fmt_opt`. The `*_str` partial applications are gone, so a call site reads `fmt_opt(fmt_bytes, r.resp_bytes)` — the formatting is visible at the point of use.

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 That's what I did. The table is a data structure in this file now — a list of `column`s (header + tooltip + cell projection) plus `row('data) = Row('data) | Group(group)` — and `PerfFormat.table` does all the HTML building, including the header row and the group rows' colspan.

`head` and `head_row` are no longer exposed; nor are `heat_style` or `truncate`. Sections choose columns and hand over rows. The last raw `<td>` outside this file (Worker Messaging's encoding-name cell) is now `PerfFormat.name_cell`.

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 `action_of` is deleted. Its one use is `fmt_opt(fst, f.perform)` — passed through the shared `PerfFormat.action_column`, so the three per-frame tables express it once each rather than reimplementing the option handling.

src/web/util/EvalMetrics.re
7h3kk1d: 🤖 Both. `EvalMetrics.rei` is added, and the similarity is enforced by construction rather than by discipline: new `Metrics.Make` owns the `enabled` / `sync` / `when_enabled` / `history` / `push` / `update` machinery the collectors had each copied, and each one includes it:

```reason
type record = {...};
include Metrics.Make({
  type t = record;
  let limit = 10;
});
```

`WorkerMetrics` moved onto it too, since it's the module the other two were copied from — so all three share one shape, `enabled` stays out of every `.rei`, and `sync` is the only public gate.

src/web/util/WorkerClient.re
7h3kk1d: 🤖 It does now — every recorder in all three collectors runs through `Metrics.when_enabled`, so no `enabled^` read survives in this file. This site is just `EvalMetrics.record_timeout(~id=latest.request.request_id)`.

src/web/util/WorkerClient.re
7h3kk1d: 🤖 Done — the gate is inside `EvalMetrics` now, and this site reads:

```reason
EvalMetrics.record_sent(~request, ~encoded);
```

The id, cell count, clock read and byte length are all derived inside the gate, so nothing is computed here while the panel is closed either. One behavioural note: `record_sent` is insert-if-absent by id, which makes the ack-retry repost keep the row the first post opened — previously that path silently skipped recording because `post_evaluate`'s `~metrics_id` defaulted to `None`.

src/web/util/WorkerClient.re
7h3kk1d: 🤖 Done for both collectors — the branch is two unconditional, self-gating calls:

```reason
WorkerMetrics.record_response(request_id, msg);
EvalMetrics.record_done(~now, ~encoded=evt##.data, result);
```

The status derivation (the `List.for_all` over the response) moved into `EvalMetrics` as well, since it was panel vocabulary sitting in the client.

While here: the redundant metrics id is gone. `WorkerMetrics.next_id` duplicated the existing `next_request_id`, so the request id is now the row id — which deleted `latest.metrics_id` and the option plumbing around it, and means the two panels correlate by construction.

7h3kk1d:
🤖 > Default the debug sidebar to everything collapsed. Otherwise there's too much navigation.

Done. Rather than adding every title to the collapsed list, the sidebar settings now store the inverse — `debug_expanded` instead of `debug_collapsed` — so nothing is open until you open it: every section *and* every field, including `InfoExp`, `ctx`, `co_ctx`, `user_term` and `Segment`.

Inverting it also retired the hardcoded list of section titles in `Settings.re` that had to match `<X>Section.title` by hand, and it covers the per-field labels that list couldn't have reached anyway. Existing persisted settings still load rather than resetting — the sidebar settings record allows extra fields, so a stored `debug_collapsed` from an earlier build is ignored (persistence goes through sexp only, see `Store.F`).

The gating predicate reads positively now:

```reason
let panel_open = title =>
  settings.show_debug_panel && SidebarModel.Settings.is_debug_expanded(title, sidebar);
```

Replies to the individual review threads are inline; the two commits are 7ea565be6a (review pass) and cf90451076 (a follow-up moving the last pre-digested arguments into the collectors).

src/web/app/sidebar/PerfFormat.rei
```diff
@@ -0,0 +1,51 @@
...
+/* Cell builders, for a column's `cell`. */
+let text_cell: string => Util.WebUtil.Node.t;
+let int_cell: int => Util.WebUtil.Node.t;
+let name_cell: string => Util.WebUtil.Node.t;
+let label_cell: string => Util.WebUtil.Node.t;
+let status_cell:
+  (~cls: string, ~tooltip: option(string)=?, string) => Util.WebUtil.Node.t;
+let heat_cell:
+  (~max: Core.Time_ns.Span.t, ~total: bool=?, option(Core.Time_ns.Span.t)) =>
+  Util.WebUtil.Node.t;
```
7h3kk1d: Instead of exposing these to clients I think we should "defunctionalize" the interface and have these return a datatype (maybe private?) that represents the construction and then the single render call below that returns a table would give you the node. That way we don't have to worry about anyone controlling the styling separate from the overall description. 

This would also make testing easier in the future.

src/web/app/sidebar/SidebarModel.re
```diff
@@ -149,16 +149,25 @@ module Settings = {
     | ToggleFlat
     | ToggleExpanded(Id.t);
 
+  /* Extra fields are tolerated so a settings blob written by an older build
+     (which stored the inverse, `debug_collapsed`) still loads instead of
+     failing t_of_sexp and resetting every other setting. Persistence goes
+     through sexp only — see Store.F. */
   [@deriving (show({with_path: false}), sexp, yojson)]
+  [@sexp.allow_extra_fields]
```
7h3kk1d, line 157: We should really decide on https://github.com/hazelgrove/hazel/issues/2454

src/web/app/sidebar/StaticsSection.re
```diff
@@ -0,0 +1,70 @@
...
+let title = "Statics";
+
+/* How the panel names each outcome PerfMetrics reports. */
+let mode_label = (m: PerfMetrics.statics_mode): string =>
+  switch (m) {
+  | Recomputed => "recompute"
+  | Forced => "forced"
+  | Deferred => "deferred"
+  | Cached => "cached"
+  };
```
7h3kk1d: This feel ppx derivable

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 Done — `max_span` filters the optionals out and then the fold is just `max`:

```reason
let max_span =
    (spans: Seq.t(option(Core.Time_ns.Span.t))): Core.Time_ns.Span.t =>
  spans
  |> Seq.filter_map(Fun.id)
  |> Seq.fold_left(Core.Time_ns.Span.max, Core.Time_ns.Span.zero);
```

It takes a `Seq.t` now rather than a list, so the sections compose into it instead of each building an intermediate list:

```reason
let max =
  frames
  |> List.to_seq
  |> Seq.map((f: PerfMetrics.frame) => f.total)
  |> PerfFormat.max_span;
```

and Worker Messaging's `max_total`, which was a `concat_map` of a `map`, is one pipeline over both directions:

```reason
records
|> List.to_seq
|> Seq.concat_map((r: WorkerMetrics.record) =>
     Seq.append(List.to_seq(r.request), List.to_seq(r.response))
   )
|> Seq.map(total_of)
|> PerfFormat.max_span;
```

Pushed as 6244ff6e0e.

I left the other optional handling alone since it isn't the same shape — `fmt_opt` and `heat_cell` exist precisely to *render* an absent metric as `—`, and `PerfMetrics.add` is an accumulate-or-initialise on a single option rather than a fold over a collection. Happy to take those to `Option.fold` too if you want the style applied more broadly.

src/web/app/sidebar/WorkerMessagingSection.re
```diff
@@ -7,65 +7,86 @@ open Util.WebUtil;
...
+   encoding's cost, and the slower encodings read redder than the rest. */
+let max_total = (records: list(WorkerMetrics.record)): Core.Time_ns.Span.t =>
+  PerfFormat.max_span(
+    List.concat_map(
+      (r: WorkerMetrics.record) =>
+        List.map(total_of, r.request @ r.response),
+      records,
     ),
-  ]);
+  );
```
7h3kk1d: Seq avoids allocations

src/web/app/sidebar/StaticsSection.re
7h3kk1d: 🤖 Agreed — done in d0e833e7b5. `statics_mode` derives `show`, and the section lowercases the constructor:

```reason
let mode_label = (m: PerfMetrics.statics_mode): string =>
  String.lowercase_ascii(PerfMetrics.show_statics_mode(m));
```

The four display words can't fall out of step with the variant now. One visible consequence: `Recomputed` shows as "recomputed" rather than "recompute" — I took the derived word and updated the column tooltip to match, rather than renaming the constructor to fit the old string.

src/web/app/sidebar/SidebarModel.re
7h3kk1d: 🤖 Fair — this is another instance of exactly what #2454 catalogues, and I should flag what I did here rather than let it pass as a solution.

Renaming `debug_collapsed` → `debug_expanded` changes the persisted shape, so without something the whole `Settings.t` blob fails `t_of_sexp` and `Store.F.deserialize` silently swaps in `default()` — the failure mode #2454 opens with. What's in this PR is two more ad-hoc workarounds of the kind that issue lists: `[@sexp.allow_extra_fields]` on the record so a stale `debug_collapsed` is ignored, and `[@sexp.default []]` so the new field is absent-tolerant. Net effect today is that no setting is lost; the only thing that doesn't carry over is which debug sections you had expanded, which now start collapsed by design anyway.

But it is a workaround, not versioning: it can't distinguish "written by an older build" from "corrupt", and it silently accepts any other stray field forever after. Happy to (a) leave it as-is and note it on #2454 as another instance, or (b) drop the attributes and let this ride on the real versioned store once #2204 → #2454 land — at the cost of one settings reset for anyone with a stored blob. Your call; I'd lean (a) since the migration story shouldn't be invented inside a debug-panel PR.

src/web/app/sidebar/PerfFormat.rei
7h3kk1d: 🤖 I like it, and it also kills something I'm not happy with: every section's `columns` is currently `(~max) => [...]` purely to thread the heat scale into `heat_cell`, which is the styling leak you're describing in another form.

Concretely what I'd write, if this is what you meant:

```reason
/* What a cell shows — not how it's drawn. */
type cell =
  | Text(string)
  | Count(int)
  | Name(string)                 /* left-aligned identifier */
  | Label(string)                /* truncated, full text on the tooltip */
  | Outcome(outcome, string)     /* Ok | Failed | Pending picks the class */
  | Span(option(Core.Time_ns.Span.t))
  | Total(option(Core.Time_ns.Span.t));

type column('row) = {label: string, tooltip: string, cell: 'row => cell};

let table:
  (~columns: list(column('data)), ~scale: Core.Time_ns.Span.t, list(row('data))) =>
  Node.t;
```

That takes `Node.t` out of the interface entirely, so a section can't reach a class name or an inline style: `Outcome` replaces the `~cls` string the sections pass today, and the heat scale becomes one `~scale` argument to `table` instead of a closure parameter on all five `columns`. `columns` then becomes a plain constant per section, and a test can assert on a `list(cell)` without a DOM — which is the part I'd most want.

Two things I'd want your read on before I build it: whether `table` should derive the scale itself from its `Span`/`Total` cells (tidier, but the sections currently pick deliberately different scales — Frame Timing uses the peak *total*, Statics the peak statics time, Worker Messaging the peak per-encoding total), and whether `cell` should be private with constructor functions or just exposed as the variant. Leaving this open until you weigh in; it's a bigger change than the rest of the round and I'd rather not guess at it.

src/web/app/Page.re
```diff
@@ -394,80 +394,92 @@ module Update = {
...
   let calculate =
       (~schedule_action, ~is_edited, ~dynamics: bool, model: Model.t) => {
-    /* Sync worker-messaging benchmark gating here (settings aren't reachable at
-       the WorkerClient.request call sites); only run when the panel is open. */
-    WorkerMetrics.sync(
+    /* Sync debug-panel gating here (settings aren't reachable at the
+       WorkerClient.request call sites nor the per-frame instrumentation sites);
+       each collector only runs while its panel is open. */
+    let sidebar = model.globals.settings.sidebar;
+    let panel_open = title =>
```
7h3kk1d: probably rename this debug_panel_open?

src/web/app/sidebar/WorkerMessagingSection.re
7h3kk1d: 🤖 Agreed, and that's what it does as of 6244ff6e0e — this is a pipeline now, with `Seq.append` rather than `@`, so there's no intermediate list per record and no `list(option(span))` built just to fold over:

```reason
records
|> List.to_seq
|> Seq.concat_map((r: WorkerMetrics.record) =>
     Seq.append(List.to_seq(r.request), List.to_seq(r.response))
   )
|> Seq.map(total_of)
|> PerfFormat.max_span;
```

`max_span` takes a `Seq.t` and drops the absent metrics with `Seq.filter_map`, so the other four sections compose into it the same way.

src/web/app/Page.re
7h3kk1d: 🤖 Renamed in a06d6c7f65 — it only answers the question for debug sidebar sections, so `debug_panel_open` is the honest name.

src/web/util/PerfMetrics.re
```diff
@@ -0,0 +1,264 @@
...
+  errors: int,
+  warnings: int,
+  statics_mode: option(statics_mode),
+  segment_tokens: int,
+  tiles: int,
+  rows: int,
+  projectors: int,
+};
+
+let empty_frame = {
```
7h3kk1d, line 42: consider using ppx_factory

src/web/util/PerfMetrics.re
```diff
@@ -0,0 +1,264 @@
...
+
+/* Why statics did or didn't recompute on a frame — the Statics panel's `mode`
+ * column. Derived here from the throttle's decision so the panel's vocabulary
+ * lives with the panel, and `show`n rather than hand-mapped to strings. */
+[@deriving show({with_path: false})]
+type statics_mode =
+  | Recomputed /* an edit landed and statics ran this frame */
+  | Forced /* the debounce timer fired and forced a run */
+  | Deferred /* an edit landed but the debounce postponed the run */
+  | Cached; /* nothing to redo — the cached statics were reused */
```
7h3kk1d: It's confusing we have two notions of statics mode

src/web/app/sidebar/StaticsSection.re
```diff
@@ -0,0 +1,67 @@
...
+   resulting info-map size / error / warning counts, plus the triggering edit
+   action. Read from PerfMetrics.history, populated in CodeWithStatics.calculate
+   while this panel is open. Implements DebugSection.S. */
+
+let title = "Statics";
+
+/* The panel names each outcome by its constructor, lowercased — no hand-written
+   map to fall out of step with the variant. */
+let mode_label = (m: PerfMetrics.statics_mode): string =>
+  String.lowercase_ascii(PerfMetrics.show_statics_mode(m));
```
7h3kk1d: This isn't saying anything and can just be inlined

src/web/app/sidebar/PerfFormat.rei
7h3kk1d: 🤖 Built in e428675586, taking both of your calls.

**Scale derived.** `table` computes it from its own cells, and it's the same number the sections were computing by hand — every table here carries a total column that dominates its own row (a frame's total covers its stages, an encoding's total is the sum of its three phases, a request's round trip contains its evaluation), so the peak over all heat cells equals the peak over totals. No tint changes anywhere. The two sections that print the scale in a legend call `PerfFormat.scale(~columns, rows)` — the same function `table` uses, so the legend can't disagree with the tint — and that's the only reason it's public.

**`cell` is private.** It's an abstract type built with `text_cell` / `int_cell` / `name_cell` / `label_cell` / `status_cell` / `heat_cell` / `total_cell`, and PerfFormat alone turns one into markup. Two styling leaks went with it: `status_cell` now takes `~outcome=Good | Bad | Waiting` instead of a CSS class, and group rows take `kind: Primary | Secondary | Absent` instead of `cls: "wm-req"` — so the only place in the sidebar that names a `perf-*` class is PerfFormat itself (the `wm-*` group classes are now `perf-group-*`).

Two knock-on wins: with the scale out of the cells, every section's `columns` stops being a function of `~max` and becomes a plain constant; and an opaque `cell` is still structurally comparable, so a test can assert `col.cell(frame) == PerfFormat.text_cell("3")` with no DOM whenever we want it.

src/web/util/WorkerClient.re
```diff
@@ -93,19 +99,17 @@ let setup_worker_message_handler = worker => {
...
             /* Hand the result off first; benchmarking the other encodings
              * can take tens of ms and must not delay evaluation latency. */
-            latest.callbacks.on_result(response);
-            switch (latest.metrics_id) {
-            | Some(id) => WorkerMetrics.record_response(id, msg)
-            | None => ()
-            };
+            latest.callbacks.on_result(result.response);
+            WorkerMetrics.record_response(request_id, msg);
+            EvalMetrics.record_done(~now, ~encoded=evt##.data, result);
```
7h3kk1d: we should just use result.request_id and stop unpacking it if we're binding result.

src/web/app/sidebar/SidebarModel.re
7h3kk1d: 🤖 Noted on the issue: https://github.com/hazelgrove/hazel/issues/2454#issuecomment-5331564696 — recorded as another instance, with the two attributes called out as workarounds rather than a fix, and the concrete migration (bump, map `debug_collapsed` to its complement, drop the attributes) written down so it's not re-derived later.

Leaving the attributes in this PR as agreed.

src/web/app/sidebar/PerfFormat.re
7h3kk1d: 🤖 Swept the rest of the PR for this, in e428675586. Where a combinator reads better it's now a combinator:

- `PerfFormat.scale` (the old `max_span`) is the Seq pipeline from above, and it now traverses `columns × rows` with `Seq.concat_map` + `Seq.filter_map` rather than materialising a list of every cell.
- `fmt_opt` is `Option.fold(~none={|—|}, ~some=to_string)`.
- `title_attrs` is `tooltip |> Option.map(Attr.title) |> Option.to_list`.
- `PerfMetrics.add` folds the accumulator instead of matching both cases to re-wrap in `Some`.
- A frame's total is `frame.perform |> Option.map(snd) |> Option.fold(~none=calc, ~some=p => p + calc)`.

Deliberately left alone, since a Seq or a fold would read worse there rather than better:

- `List.for_all` in `EvalMetrics.status_of_response` and `List.exists` in `record_sent` — both short-circuit, and they're already combinators rather than ad hoc folds.
- The `List.map`s that build row lists, and `table`'s maps over columns and rows — they produce lists that `Node.table` requires and that `table` and `scale` each traverse; a Seq there would be single-use and would need `List.of_seq` at the end anyway.
- `heat_td`'s match on the option, which picks between two different node shapes rather than reducing to one value.

src/web/app/sidebar/PerfFormat.rei
```diff
@@ -0,0 +1,66 @@
+/* Shared formatting and table building for the instrumented debug sections; see
+ * PerfFormat.re. A section describes its table as columns and rows and never
+ * builds markup: `cell` is opaque, so how a cell is drawn — its classes, its
+ * tint, its heat scale — is decided here and nowhere else. */
+
+/* Renderings of the units the panels report; `fmt_opt` gives an em dash for a
+ * metric that didn't run rather than a misleading 0. */
+let fmt_opt: ('a => string, option('a)) => string;
+let fmt_span: Core.Time_ns.Span.t => string;
+let fmt_bytes: Core.Byte_units.t => string;
```
7h3kk1d: I kind of expected fmt_option to go away in place of an opt_cell that takes an option(cell) and returns a cell.

I wonder if span and bytes can themselves be cells or some other way of not requiring the client to do it.

src/web/util/PerfMetrics.re
7h3kk1d: 🤖 Agreed, and they were genuinely two different things wearing one name. Fixed in d9dfb6c59e by naming the second one for what it is:

- `StaticsMode.t` = what the throttle *asked* for this frame: `Normal | Defer | Force`.
- `PerfMetrics.statics_outcome` = what *became* of statics: `Recomputed | Forced | Deferred | Cached`, derived from the request plus whether the recompute actually ran.

The column is labelled "outcome" now rather than "mode", so the word appears in exactly one place in the tree — `grep statics_mode` is only the throttle's labelled argument.

src/web/app/sidebar/StaticsSection.re
7h3kk1d: 🤖 Right — inlined, and the comment went with it (it was restating the code). The cell reads:

```reason
cell: f =>
  PerfFormat.text_cell(
    PerfFormat.fmt_opt(
      o => String.lowercase_ascii(PerfMetrics.show_statics_outcome(o)),
      f.statics_outcome,
    ),
  ),
```

src/web/util/WorkerClient.re
7h3kk1d: 🤖 Done in d9dfb6c59e — `| ServerMessage.Result(result) as msg =>` and then `result.request_id` at both uses. Unpacking one field beside the whole record was just noise.

src/web/util/PerfMetrics.re
7h3kk1d: 🤖 Worth doing, but it needs a dependency decision that I don't think belongs to me: `ppx_factory` isn't in `hazel.opam` and isn't installed here, so this would add a new ppx to the `web` library (and to `hazel.opam.locked`) to derive one `default_frame` for a 14-field record — about 14 lines saved at the call site.

Happy either way; say the word and I'll add the dep in this PR, or I can file it so we pick it up when there's a second use for it. Leaving this thread open for your call.

(For what it's worth, the record is all `option(_)`/`int`, so it is exactly the case `default_` handles cleanly.)

src/web/app/sidebar/PerfFormat.rei
```diff
@@ -0,0 +1,66 @@
...
+ * legend. `table` derives its own from the same rows, so the two agree. */
+let scale:
+  (~columns: list(column('data)), list(row('data))) => Core.Time_ns.Span.t;
+
+let table:
+  (~columns: list(column('data)), list(row('data))) => Util.WebUtil.Node.t;
+
+/* Italic one-liners above a table. */
+let empty: string => Util.WebUtil.Node.t;
+let note: string => Util.WebUtil.Node.t;
```
7h3kk1d: Merge these into the interface for constructing a table or something and don't leave it to the clients.

src/web/app/sidebar/PerfFormat.rei
7h3kk1d: 🤖 Both, in 761f0519a9 — nothing outside PerfFormat formats a value now.

`fmt_opt` is gone, replaced by `opt_cell: option(cell) => cell`, which supplies the em dash:

```reason
cell: r => PerfFormat.opt_cell(Option.map(PerfFormat.bytes_cell, r.resp_bytes)),
```

and `action_column` takes the label as an `option(string)` for the same reason, so the three per-frame tables just pass `Option.map(fst, f.perform)`.

Bytes became a cell (`bytes_cell`), so `fmt_bytes` is private. Spans didn't need one — every span in a table is already a `heat_cell`/`total_cell` — but `fmt_span` was still escaping for the heat legend, which both heat tables were building with the same wording. That's now `PerfFormat.scale_note(~columns, rows)`, so `fmt_span` and `scale` are private too and the legend text lives in one place.

The public surface is down to: the cell constructors, `column` / `action_column`, `row` / `group`, `table`, `scale_note`, `note`, `empty` — no formatters at all.

src/web/util/WorkerServer.re
```diff
@@ -72,6 +72,11 @@ module ServerMessage = {
   type result = {
     request_id: int,
     response: Response.t,
+    /* Time the worker spent inside the evaluator for this batch, in ms, so the
+     * Evaluation panel can separate evaluation from the queue + result
+     * serialization + transfer that the client's round trip also covers. A
+     * float rather than a Time_ns.Span since this record derives yojson/show. */
+    eval_ms: float,
```
7h3kk1d: Maybe int in ns would be more consistent at least. Is there a truecost to yojson/show of Time_ns?

src/web/util/WorkerServer.re
```diff
@@ -299,6 +304,19 @@ let predict_reuse_for_request = ((key, req_value): (key, Request.value)) => {
   (key, stream);
 };
 
+/* Evaluator time accumulated for the request in flight, in ms, reported back in
+ * the result. Timed unconditionally — the worker cannot see whether the
+ * Evaluation panel is open, and two clock reads per slice (5000 trampoline
+ * steps) is noise against the slice itself. */
+let eval_ms_total = ref(0.0);
```
7h3kk1d: Should this start at 0 or should it be an optional that's set?

7h3kk1d: Either way, I think this can still be a Time_ns until we convert?

src/web/app/sidebar/PerfFormat.rei
7h3kk1d: 🤖 Merged in 783532caca. `PerfFormat.view` is the single entry point and a section no longer assembles anything:

```reason
let view = (~globals as _: Globals.t): list(Node.t) =>
  PerfFormat.view(
    ~columns,
    ~empty="No frames recorded yet — type in the editor.",
    ~legend=true,
    List.map(f => PerfFormat.Row(f), PerfMetrics.history^),
  );
```

It owns the branch on whether there are rows, the empty message in their place, and the order of the legend and the note above the table. `~live` covers the one line that should show either way (Editor & Memory's undo/redo/backpack), `~note` a line about the table, `~legend` the heat-scale line.

`empty`, `note`, `scale_note` and the node-building table are all private now, so a section can't forget the empty case or order the pieces differently from its neighbours. Output is byte-identical in all five.

The one thing still assembled by a client is Worker Messaging's `[encoding_toggles(~globals)] @ PerfFormat.view(...)` — those chips are an interactive control of that section's own, not table content, so I left them outside rather than inventing a hook for them. Say if you'd rather they went in too.

src/web/util/WorkerServer.re
7h3kk1d: 🤖 Checked rather than assumed, and you were right that the float was a workaround for a cost that mostly isn't there:

- `show` and `sexp` are fine — Core gives `Time_ns.Span` both a `pp` and sexp converters, so those derive with no help.
- `yojson` is the only gap; Core ships no converters for it.
- `int` in ns is the one option that doesn't work here: jsoo's `int` is 32-bit, so it overflows at ~1.07s of nanoseconds and our eval timeout is 20s.

So, taking your `[@to_yojson]` suggestion (e37dab6dc9): the field is `option(span)` where `span` is an alias for `Core.Time_ns.Span.t` carrying all four converters in one place, and json is integer nanoseconds — Time_ns's own representation — as a bigint literal rather than `Int`, for that 32-bit reason. Evaluator time is a Span like every other duration in the panels now, the sibling message types keep their yojson, and `EvalMetrics` takes the span straight off the wire with no conversion either side.

I'd briefly dropped the unused yojson deriving from `result`/`t` instead (bea6567481); this replaces that, so nothing loses a capability. Tests pin the hand-written pair — 12.5ms encodes as `12500000` and comes back equal — and the fixture now carries a span so Marshal, Direct and Sexp all drag an Int63-backed value through `structuredClone`, which nothing previously covered.

src/web/util/WorkerServer.re
7h3kk1d: 🤖 Optional — good catch. The accumulator starts as `None` and `timed_eval` fills it in, so a result that evaluated nothing reports absence instead of a zero that reads as "instant".

That also lines up the whole path: `EvalMetrics.record.eval` was already `option(Span.t)`, so it now takes `result.eval_time` verbatim, and the panel renders `—` through the same `opt_cell` every other absent metric uses.

src/web/util/WorkerServer.re
```diff
@@ -68,10 +68,35 @@ module ServerMessage = {
...
+  let span_of_sexp = Core.Time_ns.Span.t_of_sexp;
+  let yojson_of_span = (s: span): Yojson.Safe.t =>
+    `Intlit(Core.Int63.to_string(Core.Time_ns.Span.to_int63_ns(s)));
+  let span_of_yojson = (json: Yojson.Safe.t): span =>
+    switch (json) {
+    | `Intlit(ns) => Core.Time_ns.Span.of_int63_ns(Core.Int63.of_string(ns))
+    | `Int(ns) => Core.Time_ns.Span.of_int63_ns(Core.Int63.of_int(ns))
+    | _ =>
+      failwith("WorkerServer.span_of_yojson: expected integer nanoseconds")
+    };
```
7h3kk1d: This is a lot of noise. Can you move the helper functions to util and use the inline annotation for choosing how to convert from/to json using

[@to_yojson] and [@from_yojson]

src/web/util/WorkerServer.re
```diff
@@ -299,6 +324,26 @@ let predict_reuse_for_request = ((key, req_value): (key, Request.value)) => {
...
+    eval_total :=
+      Some(
+        Option.fold(
+          ~none=span,
+          ~some=s => Core.Time_ns.Span.(s + span),
+          eval_total^,
+        ),
+      );
+    x;
+  };
```
7h3kk1d, line 322: If this is going to accumulate over time there's no reason to use optional like i suggested rather than start at 0.

src/web/util/WorkerServer.re
7h3kk1d: 🤖 Moved in 46488d372f — `TimeUtil` owns a `span` alias and its converters, and the wire field is just `eval_time: TimeUtil.span`. Nothing about json lives in the message type any more.

One thing didn't work as suggested, though: `[@to_yojson]` / `[@of_yojson]` are `ppx_deriving_yojson` attributes, and this tree preprocesses with Jane Street's `ppx_yojson_conv`, which ignores them silently — it still went looking for `Core.Time_ns.Span.t_of_yojson` and failed to build. So the converters are reached the way the rest of the repo does it (`Environment`, `CachedSyntax`): by naming the type, which resolves `TimeUtil.{pp_span, sexp_of_span, span_of_sexp, yojson_of_span, span_of_yojson}`.

That does mean five one-liners rather than two, since naming the type also moves pp and sexp off `Core`'s versions — but they're in util now, reusable by anything else that wants a Span in a deriving type, instead of noise in the wire module. Happy to add `ppx_deriving_yojson` as a dep instead if you'd rather have the annotation form, though I'd expect two ppxs both claiming `yojson` to be a fight.

src/web/util/WorkerServer.re
7h3kk1d: 🤖 Fair — reverted to a zero start in 46488d372f. It only ever accumulates, so `None` was carrying information that a zero already carries, and it cost an `Option.fold` in the hot path plus an option all the way through to the panel row.

7h3kk1d:
Look into david's performance profiling stuff.

7h3kk1d:
🤖 > Look into david's performance profiling stuff.

There's prior art worth linking to. David's #2161 (`perf-bench`) is an offline harness measuring the syntax and statics pipeline across cached and uncached scenarios, comparing results across commits. Andrew's `hazel bench-eval`, which landed with #2404, times evaluation on its own. Both are profiling: a controlled run over a fixed program, producing a number you compare against another commit's.

This PR is telemetry rather than profiling. It instruments the shipped app and reports the work an actual editing session is doing — real programs, real keystrokes, including the worker round trip that neither of the others touches. That's a different axis from theirs, not a competing answer on the same one, which is also why it lives behind the debug sidebar and collects nothing while the panels are closed.

Two places the two do overlap, worth converging on before both land:

**The timing primitive, and the shape it records into.** `PhaseTiming.record(label, f)` on `perf-bench` and `TimeUtil.timed(f)` here both wrap a thunk and take wall-clock behind an enable flag, but they record differently. `PhaseTiming` appends to a flat global list of `(label, duration)` pairs that a harness drains after a run. `PerfMetrics` accumulates into a typed `frame` record, commits it once per update cycle, and keeps a bounded newest-first history for the panels. A label stream suits a harness; a per-frame schema suits a live table.

**Granularity.** His instrumentation names sub-phases inside the pipeline — the individual steps within statics and within syntax. This PR times statics and syntax as single units.

What I'd want out of that eventually: one set of instrumentation points, declared in a typed and structured way rather than each site side-effecting a global ref, with two consumers behind it — telemetry into the bounded per-frame history the production app shows, profiling into the labeled accumulator an offline script drains. Both approaches deserve to exist and neither should have to own the call sites. Between the two, `TimeUtil` is the better base to build that on: it returns a typed `Core.Time_ns.Span.t`, and it doesn't reach through `Js_of_ocaml.Js.Unsafe` from `src/util`, which matters for the de-webbing work.

Nothing needs to change in this PR for that. But it's worth settling before we end up with two sets of timing calls threaded through the same pipeline.

src/web/util/PerfMetrics.re
7h3kk1d: Decided: not adding the dependency. A new ppx plus the \`hazel.opam.locked\` churn is more than one 14-field default is worth, and not filing an issue for it either — if a second `default_` case shows up, that is when it earns the dep. Leaving the hand-written record and closing this out.

Issue #2203, opened 2026-03-31
set up merge queue


cyrus-:
we tried this before but there were some issues that need to be resolved

7h3kk1d:
# CI/CD Long-Term Plan


## Goals

- Separate workflows for build vs. deploy.
- Native GitHub merge queue on `dev`, gated on tests.
- Branch deploys remain fast — **not** gated on tests.
- Faster status-check feedback (fmt/build/quick-tests) before slow tests finish.
- Minimize total CI wall time. Opam/OCaml setup is the biggest current cost.
- Keep the public URL contract: `hazel.org/build/<branch_name>` must continue to work.
- Open to changing the deploy mechanism and/or the structure of `hazelgrove/build`.

## Current pain points

1. **Single monolithic job** — build, deploy, test, coverage all in one job. Test failure means the bad code already deployed; deploy failure means tests didn't run; any step's flake restarts everything.
2. **Opam/OCaml setup dominates wall time** — every run does opam install + zarith setup + release build cold. The `_opam` cache helps but only when the lockfile is unchanged; setup-ocaml's compiler restore is still ~1–2 min minimum.
3. **No artifact reuse** — every push to every branch rebuilds the world independently. Two branches sharing a base commit don't share build output.
4. **Push-to-build-repo is a serialization bottleneck** — even with the per-branch concurrency fix (#2272), all deploys contend on `hazelgrove/build`'s `main` ref. The retry-with-rebase loop handles it, but the build repo also keeps growing (~400 branch directories today).
5. **Build repo is approaching GitHub's repo size limit** — every branch's artifacts accumulate in one git history. We will eventually run out of space; pruning is manual and disruptive.
6. **No merge queue** — PRs can pass CI individually but break `dev` when merged together (semantic conflicts, dependency drift).
7. **Merge queue's temp branches would deploy.** GitHub creates `gh-readonly-queue/<base>/pr-<n>-<sha>` and pushes to them; the current `on: [push]` workflow would deploy these. This is what blocked previous merge-queue adoption.

## Proposed shape

### Workflow graph

Three workflows + one composite action.

**Composite action `.github/actions/setup-hazel-toolchain`**: encapsulates node setup, OCaml setup, opam cache restore, opam archive repo, deps install, zarith setup. Single source of truth for the slow part. Versioned, reusable across `ci.yml` and `deploy.yml`.

**`ci.yml`** — triggers: `pull_request`, `push` (scoped to `dev` only), `merge_group`.

[...]

PR #2422, opened 2026-08-03
ci: fix Actions cache eviction, split into independent jobs, enforce release warnings on test/

### LINKED ISSUE #2334: CI: in-place opam-repository edits can retroactively break our lock file
## What happened

On 2026-06-03 every CI run on every branch started failing at `opam install . --deps-only --with-test --locked`:

```
[ERROR] Package conflict!
  * No agreement on the version of conf-pkg-config:
    - deps-of-hazel → conf-libssl >= 4 → conf-pkg-config >= 5
    - deps-of-hazel → conf-pkg-config = 4
```

ocaml/opam-repository#29976 edited already-published package versions in place: `conf-libssl.4`, `conf-libffi.2.0.0`, and `conf-zlib.1` now require `conf-pkg-config >= 5`, conflicting with our locked `conf-pkg-config = 4`. Fixed by #2333, but nothing prevents a recurrence — opam package metadata is mutable, and upstream routinely edits existing versions (especially of virtual `conf-*` packages) rather than cutting new ones.

## Why this hurts

- It breaks **all** branches simultaneously, not just dependency-update PRs — unrelated work gets red CI overnight with no code change on our side.
- The breakage is external and time-triggered, so bisecting our history finds nothing.

## Proposal: pin the opam repository to a commit SHA in CI

`ocaml/setup-ocaml@v3` accepts an `opam-repositories:` input, so workflows could use `https://github.com/ocaml/opam-repository#<sha>`. The daily `update_deps` workflow would bump the SHA alongside the regenerated lock file, so upstream changes only ever land (and fail) in the `bot-update-deps` PR, never on existing branches.

Note on why we'd want both the SHA pin *and* the lock file: an opam lock file pins which versions the solver picks, but not their metadata — the repository can edit a published version's dependencies after the fact (no content hashes, unlike npm/cargo lock files). The repo SHA freezes the metadata; the lock file freezes the version selection and keeps installs reproducible for local development against the default repo. Each covers a hole the other leaves open.

Maintenance cost: the pin lives in several workflow files unless factored out (e.g. into a single env/var or composite action), and the `opam-repository-archive` repo we also add would ideally be pinned the same way.

Refs: #2333, ocaml/opam-repository#29976

### LINKED ISSUE #2456: CI: release-profile warnings aren't enforced on test/, so warning cruft accumulates there
## Problem

The `release` profile promotes OCaml warnings to errors, and that's the pass that gates PRs. But CI only ever compiles `test/` under the **dev** profile, which is deliberately lax. So warnings in `test/` print into the job log and the job still goes green.

### Concrete example

[Run 32063723487 / job 95490665795](https://github.com/hazelgrove/hazel/actions/runs/32063723487/job/95490665795?pr=2435) (PR #2435) concluded **success**. Its `Run Tests` step printed, twice:

```
Warning 33 [unused-open]: unused open Dune__exe.EditingPrelude.
```

(`test/Test_ReparseDocSlides.re:10`.) Compile the exact same tree under `release` and it's a hard error:

```console
$ dune build test/haz3ltest.bc.js --profile release
File "test/Test_ReparseDocSlides.re", line 10, characters 0-19:
10 | open EditingPrelude;
     ^^^^^^^^^^^^^^^^^^^
Error (warning 33 [unused-open]): unused open Dune__exe.EditingPrelude.
$ echo $?
1
```

## Why it happens

`.github/workflows/deploy_branches.yml` has two build steps, and neither compiles `test/` strictly:

| Step | Command | Profile | Covers |
| --- | --- | --- | --- |
| `Build Release` | `dune build @src/fmt --auto-promote src --profile release` | release | **`src` only** |
| `Run Tests` | `make ci` &rarr; `dune runtest --instrument-with bisect_ppx --force` | **dev** (no `--profile`, so the default) | `src` + `test` |

`test/` is the only source tree in the repo with a `dune` file that never sees `--profile release`.

To be clear, the dev/release split is working as designed and is worth keeping &mdash; dev's laxness is what makes local iteration fast (`src/util/dune` states it outright: dev `-warn-error -A`, release `-warn-error +A-58`; elsewhere dune's own per-profile defaults do the same job). The bug is only that the strict pass is *scoped* to `src`.

## Current debt

Exactly **one** error today, on a clean `test/` rebuild under release (dune keeps going past the first error by default &mdash; `--stop-on-first-error` is opt-in): the `Test_ReparseDocSlides.re:10` unused open above. That's small enough to fix in the same PR that adds the gate, which is the argument for doing it now rather than after it grows.

## Possible fixes

1. **Extend the release build to cover `test`** &mdash; `dune build @src/fmt --auto-promote src test --profile release`. One-line change, gates everything. Cost: `test/` gets compiled twice per run (once under release for the gate, once under dev + `bisect_ppx` for coverage).
2. **A dedicated strict step/target** &mdash; e.g. a `make ci-strict` running `dune build test --profile release`, kept separate from the instrumented coverage run.
3. **Not** making dev strict for `test/` &mdash; that would defeat the purpose of the dev profile locally.

No dune changes are needed either way: `test/dune` has no `(env)` stanza, so it already inherits dune's strict release defaults. Only CI's scoping has to change.

## Related

- **#2422** (`ci-cache-and-job-split`) is already reorganising these jobs into separate deploy/test steps and drops `--auto-promote` from the CI entry points &mdash; probably the natural place to land this.
- Same root cause, adjacent gap: the release step checks `@src/fmt` but not `@test/fmt`, *and* passes `--auto-promote`, so formatting violations in either tree get silently rewritten in a throwaway checkout instead of reported.

7h3kk1d:
Part of the CI/CD plan on #2203. Closes #2456.

## What changed

- **Replaced the one serial job with three independent ones.** `Build and Deploy Hazel` was a
  single 21–31.5 min job that published the preview and *then* ran the tests — one check covering
  everything, with the test result buried after a deploy that had already happened. Now
  `Fast checks`, `Full tests` and `Deploy branch preview` are three jobs with no `needs:` between
  them. Each re-runs alone, so a flaky suite costs one job rather than another deploy and another
  30 minutes, and the fast signal comes back in ~4 minutes instead of at the end of the full run.

- **Turned off `setup-ocaml`'s dune cache and made the opam switch cache single-writer.** Its
  entries are keyed on the run id, so they never hit, and 10.24 GiB of them were evicting the
  `_opam` switch from a 10 GiB budget. The switch cache is now `restore` + `save` with
  `dev`/`bot-update-deps` as the only writers, plus `restore-keys` for a warm start after a
  lockfile bump.

  Caching the switch ourselves is deliberate and unsupported: `setup-ocaml` declines to do it
  (ocaml/setup-ocaml#1098), on the grounds that cached dependencies drift from opam-repository.
  We take that trade knowingly — it moves `Install dependencies` from 3–7.5 min to ~6 s, which is
  what sets how long we wait for a deploy and for static-check feedback — and the lock file bounds
  the risk. It doesn't eliminate it: in-place edits to already-published packages have broken us
  before (#2334).
- **Added `make ci-check` to enforce release warnings on `test/`** (#2456). `dune build @check
  --profile release` type-checks both trees without linking — 21 s, no second build of the test
  bundle. Warnings in `test/` previously printed and passed.
- **Made test failures produce a non-zero exit code.** Alcotest is run with `and_exit:false` so
  the JUnit report can be written first, and junit_alcotest hands the outcome back as a function
  to call afterwards — the only thing that propagates a failure. `test/haz3ltest.re` discarded
  it, so the test binary exited 0 whatever happened, and every entry point keyed on that exit
  code passed on a red suite: `dune runtest`, `make ci`, `make test`, `make test-quick` and
  `./run_tests`.
- **Published the test reports as check runs** (`use-actions-summary: false`). The default writes
  results to the run's job summary and creates no check run, so `Test Report` never appeared in a
  PR's check list — nothing to click through to, no inline annotations on failing tests, and
  nothing that could be named as a required status on `dev`.

## What runs where

```
checks.yml   ── on: workflow_call                       the definitions, one copy
   ├── Fast checks       fmt → make ci-check → make ci-quick → Quick Test Report
   └── Full tests        make ci → Test Report → Codecov

ci.yml       ── on: push, except gh-readonly-queue/** | merge_group
   └── calls checks.yml

fork-ci.yml  ── on: pull_request, if the pull request comes from a fork
   └── calls checks.yml, with the report and coverage steps off

deploy.yml   ── on: push, except gh-readonly-queue/**
   └── Deploy branch preview    release build → hazel.org/build/<branch>
```

Every filter is a trigger rather than an `if:`, which leaves **two conditionals in the whole
system**: the fork test, and one input that turns off the report check runs and the Codecov
upload for forks, whose `GITHUB_TOKEN` is read-only and which get no secrets.

**`ci.yml` has no `pull_request` trigger on purpose.** A same-repo pull request fires both it and
`push` for the same commit, so the full suite would run twice. Forks are the only case `push`
cannot see — a fork's pushes happen in the fork — and `fork-ci.yml` covers them.

**Fork PRs gain CI.** Today's `deploy_branches.yml` is `on: [push]`, so fork PRs get no checks at
all; 3 of the last 200 PRs here came from forks. They now get `Fast checks` and `Full tests`,
without the report check runs their token can't create. They still get no branch preview.

**No merge queue exists yet**, but both `ci.yml` triggers are load-bearing once one does. A queue
pushes its `gh-readonly-queue/**` branch, yet Actions only reports status into the queue from a
`merge_group` run — so the trigger is what makes the queue's checks exist, and `branches-ignore`
is what stops the same commit being checked twice and a speculative merge being published as a
preview. Turning the queue on is #2203 phase 4.

**Why formatting, warnings and the quick suite share a job.** Every job installs OCaml and the
dependencies independently, ~2 min warm and 5–9 min cold, which is more than those three checks
cost to run. Splitting them would buy precise names and precise retry for roughly double the
wall clock. Worth revisiting after the build-once artifact work in the follow-ups, which is what
makes extra jobs cheap.

`Build release` is no longer a separate gate — the deploy job runs the same command. That leaves
`src/web`'s `main` and `worker` jsoo executables unlinked on fork PRs and in the merge queue,
which we're accepting until it causes a problem; `Full tests` still links the `web` library
through the test bundle.

## Notes for merge

- **`make test` and `./run_tests` now fail on a red suite.** That's the point, but it may surface
  failures that have been passing silently across the open PRs — #2128's evaluator/stepper QCheck
  flake most likely.
- The strict formatting gate wasn't in the original plan and can turn other branches red. Easy to
  drop.
- **The warm-cache win can't be demonstrated until this reaches `dev`.** Only `dev` and
  `bot-update-deps` write the switch cache, so the first push after merging populates
  `Linux-opam-*` and the *second* is where `Install dependencies` should fall to single digits.
- **Branch protection and the merge queue are both future work** (#2203 phase 4). When you set
  them up, the required checks are named `CI / checks / Fast checks` and `CI / checks / Full
  tests` — the reusable-workflow nesting makes those non-obvious, so it's noted in `ci.yml` too.
  The `merge_group` path is the one thing here that can't be exercised until a queue exists.
- ~~Also pins the opam repository, resolving #2334.~~ Reverted in bd8c01b3b2 — a separate concern
  that made this harder to review. #2334 stays open and gets its own PR.


7h3kk1d:
I'm going to remove the opam pinning from here. It makes it more complicated. We can do that in a followup.

disconcision (approved):
i dont feel fully qualified to review this but ive read through it and it seems conceptually sound. all of the changes seem decidedly for the better. mixed feelings about tripling our number of yaml files but it is what it is

Issue #1170, opened 2024-01-24
ExplainThis cleanup


disconcision:
ExplainThis could be generally cleaned up along the lines demonstrated in the PipelineExp module. Right now the infrastructure for (ExplainThisForm.Simple) that only supports forms with singleton groups, but it could be fairly straightforwardly generalized.

PR #2424, opened 2026-08-07
Simplify ExplainThis: forms own their colorings, one id enum, total model

7h3kk1d:
## What this does

`ExplainThis.get_doc` was a ~3100-line match in which every branch re-derived by hand which form of a doc group to show and which sub-terms it should highlight. The data a form needs was scattered: its explanation was a `%s` template in `data/`, filled by `Scanf.format_from_string` in `ExplainThis.re`; its colorings were a function in `data/` invoked at the call site; its specificity level was recomputed by comparing an explicitly named form id against the group's first form.

This branch moves all of it onto the form itself and deletes the scaffolding that existed to carry it around.

This is the direction issue #1170 asks for — generalize the `Simple` infrastructure so it isn't limited to singleton groups — reached by making the invariant enforceable rather than by rewriting all 154 forms.

| | dev | this branch |
|---|---:|---:|
| `ExplainThis.re` | 3117 | 1753 (−44%) |
| `ExplainThisForm.re` | 513 | 390 |
| `ExplainThisModel.re` | 225 | 174 |
| `format_from_string` sites | 95 | 0 |
| `get_specificity_level` references | 47 | 3 |
| `~colorings` arguments at call sites | 44 | 4 |
| id enums to keep in sync | 2 × 111 | 1 |
| raises in `ExplainThisModel` | 6 | 0 |
| test-only constructs in production types | — | 0 |

## How it's staged

Commits are written to be read in order; each says what it changes and what it deliberately leaves alone.

**Stage 0 — pin the behavior first.** `get_doc` gains a `Probe` mode that reports the decision it makes and renders nothing, so the golden can be written before anything moves. The color map could not serve as the fingerprint: it is re-derived from the explanation's markdown links and never reads `colorings` at all, so it is blind to the mapping under change. The golden canonicalizes each UUID to the printed text of the sub-term it belongs to, and sweeps every form of every group (by planting a selection), since `init` otherwise only ever reaches the most specific form. *(`Probe` is scaffolding and is deleted again in Stage 4 — see below.)*

**Stage 1 — remove the scaffolding.** Explanations are filled at their definition site so the compiler checks arity; the 44 specificity ladders collapse into one `leveled(~fallback, group)` helper; `group_id` becomes an alias of `form_id` (the two listed the *same 111 constructors*), which makes a `singleton` combinator expressible — 86 single-form groups now derive their id from their form.

**Stage 2 — colorings onto forms.** `form` gains a `colorings` field, filled per module. That the form-carried mapping really matches what the call sites were passing was established by temporarily inverting the precedence and rerunning the golden; exactly four cases differed, all fallback rows, all preserved deliberately.

**Stage 3 — totality and a wasted pass.** `ExplainThisModel` becomes raise-free, and `Page.re` stops computing a full `get_doc` pass it then discards.

**Stage 4 — split the decision from the rendering, and remove the seam.** `message_mode` was a tag telling one function which of three unrelated things to return — Vdom, a color map, or (Stage 0's addition) a test callback. `get_doc` becomes `decide`, returning a `decision` (`NoDoc` / `Prose` / `Markdown` / `DrvSyntax` / `Doc {group, form, options, explanation, colorings}`); `view_doc` and `color_map_of` are then ordinary functions over it. `message_mode`, the `probe` record and its test-only `sf_ids` field are gone, and the test calls `decide` directly. It's a small diff because the ~150 dispatch arms were already tail calls and are textually unchanged.

## User-visible changes

- **Type-function abstract form is now highlighted.** `get_doc` handed the `TypFunctionExp` group *`FunctionExp`'s* coloring function, so the `typfun a -> e` abstract in the sidebar was never coloured and hovering its links highlighted nothing. This is the concrete failure that motivated moving colorings onto the form. Verified in the browser.
- **Collapsing the sidebar no longer disables code highlighting.** `get_color_map_deduction` guarded on `settings.explainThis.show` while `get_color_map` used `settings.sidebar.show`; nothing toggles the former and it defaults to true, so collapsing the sidebar stopped highlighting for ordinary terms but not derivation ones. Both now read `sidebar.show`.
- **Two forms' persisted feedback keys move.** `AppPat.funapp_pat` claimed `ApConsPat` in group `ApFuncPat`; `TerminalPat.sintlit_pat` claimed `IntPat` in group `SIntPat`. Neither is reachable through the UI (both groups are singletons, so no expander renders), but a user who had rated either keeps an orphaned row that now reads as unrated.
- **Prose typos fixed**: "raising left ro the right", "This if functionally equivalent", "evalutes", "autmatically" (×2).
- **One less full `get_doc` pass per frame** when the cursor is in a derivation cell.

## Bugs fixed along the way

- Two `%s`/`%d` mismatches (`FunctionExp.function_sintlit_exp`, `LetExp.let_sint_exp`) that were guaranteed `Scan_failure` if reached — surfaced by making the formats compile-checked.
- `ExplainThisModel.get_options` raised for any form in a multi-form group lacking an `expandable_id`, and 47 of 154 forms have none — so adding a form to a two-form group would have crashed the sidebar. The invariant held by luck, not construction.
- Several dead branches that passed a mismatched group (`Prod([])` → `TupleTyp.tuple2`, the 0-tuple type case).

## Testing

- `make dev` clean and warning-free on a forced rebuild.
- `make test-quick`: 3255 tests pass.
- 35 characterization cases plus two structural assertions, in `test/Test_ExplainThis.re`:
  - the set of group ids the corpus reaches is asserted *exactly*, with a separate `known_unreachable` list asserted **not** reached, so losing coverage fails loudly instead of going quiet;
  - every coloring's piece id must occur in its own form's `syntactic_form` — naming a piece the form doesn't contain is always a bug and a silent one, since the pair simply never matches and the link renders unhighlighted.
- Zero golden diff at every refactoring step; the deliberate diffs are one line for the `ApFuncPat` fix and the coverage additions.

The corpus is hand-written because `AST.gen_exp_sized` cannot produce `Asc`, `Use`, `SInt`/`Nat`, module forms or pipelines, so the property test cannot reach those docs at all.

## Review notes

- **Locally-created pieces** (`cons`, `comma`, `ap`, Grout holes) were hoisted to module level, because forms are now rebuilt per call and `Piece.id`s must stay stable. This is the change with the most plausible visual failure mode.
- **Four `~colorings` overrides remain and are load-bearing**: `FunctionExp.function_exp`, `LetExp.let_base_exp`, `LetExp.let_tuple_exp`, `ListPat.cons_base_pat`. A least-specific form is shared across a whole family (one `function_exp` backs all 19 `FunctionExp` groups), so it is not built for any one call site. This is also why `FunctionExp(Base)` and `LetExp(Base)` never appear as a *group*.
- **`Fun`'s fallback links a different id than its specific forms**: for `fun (x) -> x` the `Base` form highlights `(x)` and `Var` highlights `x`. Preserved as-is with a comment rather than silently changed — each highlight matches the wording of the form doing the describing, but it's worth a deliberate decision.
- **Unused id constructors are kept on purpose** (`TrivExp` and others). They appear in persisted `ExplainThisModel` state, and removing a variant makes `t_of_sexp` fail on a stored model that names it. For the same reason `settings.explainThis.show` is left in place though now unread.
- **The 53 `*_id: form_id` constants in `data/`** existed to feed the specificity comparisons and now only label their own form. Left rather than inlined, since restating a `form_id` by hand is the same class of mistake as the two id mismatches fixed here.

## Known gaps / follow-ups

- The dispatch still runs twice per frame — once in `calculate` for the code highlights, once in `view` for the sidebar. `decide` now returns a plain value that *could* be computed once, but the two callers sit on opposite sides of the calculate/view boundary **and do not pass the same input**: `ExplainThis.view` applies `Info.pre_labeled_info` to the cursor info while `Page.re`'s `get_color_map` call does not. Sharing one decision between them is therefore a behavior change rather than a pure optimization, and that asymmetry deserves its own look — it may well be a bug.
- `SIntPat` docs appear unreachable from the editor: `MakeTerm` only ever builds `Atom(Int)` literal patterns from surface syntax. Asserted unreachable rather than fixed.
- The `SInt`/`Nat` binary operators are admitted by `Operators.op_bin` and `form_id` but `OpExp` defines no docs for them, so they fall through to "No docs available". Also asserted, not fixed.
- The golden records which *user-code* ids a doc links, not which *abstract-form piece* each is paired with. That's why the TypFun bug needed a browser check — the new piece-membership assertion now covers that class going forward.

## Stacking

The first commit (`ExplainThis: fill doc explanations at definition, drop runtime formats`) is the tip of `explainthis-compile-checked-docs`, which has no PR of its own. It is included here. `ExplainThis PBT` also appears in history; the equivalent landed on dev separately as #2414, and the merge takes dev's reviewed version.


src/web/app/explainthis/ExplainThisForm.re
```diff
@@ -303,139 +308,41 @@ type form_id =
...
-  | FilterSelector
-  | ModuleExp
-  | ModLetDecl
-  | ModTypeDecl
-  | SigTyp
-  | SigLetDecl
-  | SigTypeDecl
-  | ModuleKeywordExp
-  | ModuleKeywordDecl;
+/* Resolves the TODO that stood here: group_id and form_id listed the same 111
```
Negabinary, line 324: Comment reference a resolved todo

Negabinary (changes requested):
ExplainThis is still so huge I don't really know how to go about understanding and reviewing it all. 

I noticed one comment that describes a resolved TODO - we can probably get rid of that (idk if it's possible to get claude to check through its other megacomments)

Functionally seems to still be working, and I'm very encouraged by the dramatic line count drop; so after fixing comments I'd be in favor of merging

Negabinary (approved):

PR #2426, opened 2026-08-09
Fast parsing for complete terms

disconcision:
This introduces a fast path to all editor parse paths which attempts to use the Menhir parser, falling back to the per-character editor parser if it does not succeed. The new fast path lexes the source into tokens + whitespace gaps, parses with Menhir, print the term back to a segment via ExpToSegment, then weaves the original tokens and whitespace back in positionally. This is done so that we have string fidelity by construction (doesn't reply on completion provenance to e.g. preserve float rep), but more importantly so that MakeTerm's someone-edge-casey whitespace-attachment policy stays single-homed. A fall back to the slow path prints a `SLOW PARSE (context, N chars)` console line naming the caller. Parser agreement is pinned by differential suites (QCheck equivalence property, fuzz, per-file corpus term diff, and an empty-known-gaps ratchet that every `.hz` in the repo must fast-parse).

This involved making a bunch of patches and extensions to the menhir parser.

All docs slides, and basically all serialized hazel programs, are now just .hz text files. Exceptions include the in-built derivations (neither Claude nor I was inclined to touch this) and exercises mode (as it would remove our security by obscurity). All of these are parsed on-demand now, which makes editor startup time less than a second. Each current doc slides parses in less than 50ms.

The previous backup_text tests are replaced by tests that do both parse paths for each slide to see if they match up, which should force keeping both in sync as long as we add new syntax to the slides

Also:
- `Equality.pat_names_equal` reflexivity fix: identical literal module-let patterns (`let [] = ...`) compared unequal (found by the equivalence property).
- Persistence: backup text prints holes as `¿` (lossless), unpersist is fast-first, sexp fallback retired; b2t2 Datasheet now a a `^^text` slide.
- CLI parses fast-first and no longer loads the slide set holistically on invocation.


src/web/view/ScratchMode.re
```diff
@@ -6,6 +6,15 @@ open Util;
...
    last save. Eliminates expensive Zipper.zip + Base.equal_segment checks. */
 let dirty_slides: ref(Sets.StringSet.t) = ref(Sets.StringSet.empty);
+
+/* Lazy hydration: boot builds a full editor (parse + statics cache +
+   agent state) for the CURRENT slide only; every other slide sits in
+   the model as a cheap blank placeholder and is registered here
+   ("prefix:name"). Hydration replaces the placeholder on first switch
+   (Persist.hydrate_current). Dormant slides are never persisted over:
+   their persist_cache entries are seeded from storage at load. */
+let dormant_slides: Hashtbl.t(string, unit) = Hashtbl.create(64);
```
Negabinary: This probably doesn't have to be a global ref right? (I'm guessing this Hashtbl is nowhere near big enough that it's worth it for the lookup cost)

Negabinary (changes requested):
I found a Hashtbl

Actually I found several, but I'm not mad about the lexing ones. Very down for what this PR is trying to do; the description looks sound and it pleases me that ExpToSegment is now robust enough for us to parse the slides through ASTs!

src/web/view/ScratchMode.re
disconcision: fixed. dormant is now a flag on the slide not a side channel

.github/workflows/deploy_branches.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           eval $(opam env)
           make ci
         working-directory: ./source
-        timeout-minutes: 20
+        timeout-minutes: 60
```
7h3kk1d, line 111: I need to get back on github.com/hazelgrove/hazel/pull/2422

hazel-programs/docs/b2t2/errors-malformed-tables.hz
7h3kk1d, line 1: Self-reminder to make sure these files match the most recent versions that may have drifted.

hazel-programs/docs/reference/tuples.hz
7h3kk1d, line 1: @disconcision what's the workflow for updating these files. It seems like copying from the editor includes leading whitespace but this representation does not?

src/b2t2/Slides.re
```diff
@@ -1,42 +1,159 @@
+/* B2T2 benchmark slides: the committed .hz files in
+ * hazel-programs/docs/b2t2 ARE the slides — embedded at compile time,
+ * parsed at load. The Datasheet slide is markdown (ppx_blob via
+ * Datasheet.re), not a program, and keeps its own path. */
+let text_slides: list((string, Haz3lcore.PersistentSegment.t)) =
```
7h3kk1d: We should look and see if thers's a ppx that will do this without having to enumerate all of them. Or do a custom dune script.

src/CLI/Cli.re
```diff
@@ -10,9 +10,39 @@ let read_input = path => {
   );
 };
 
+/* Fast-first CLI parsing: FastParse (linear) with pin collection, then
+   the ¿-aware typing parser; the fallback names itself on stderr-ish
+   console so slow passes are visible. */
+let parse_to_zipper = (s: string): option(Haz3lcore.Zipper.t) => {
```
7h3kk1d: This code looks similar but subtly different from the code in PersistentZipper.

src/CLI/Cli.re
```diff
@@ -10,9 +10,39 @@ let read_input = path => {
...
+  Haz3lcore.(
+    switch (
+      FastParse.of_text(
+        ~materialize=Triggers.invoked_projector,
+        ~collect_refractors=true,
+        ~root=Exp,
+        String.trim(s),
```
7h3kk1d: Is there a reason we're trimming the text?

src/CLI/Cli.re
```diff
@@ -340,19 +366,22 @@ let probe_hazel = (auto: bool, many: bool, path: string): unit => {
...
+    /* Probe ids up front — statics needs them too, since instrumentation
+       (e.g. the livelit view fold-in) happens during elaboration */
+    let base_probe_ids = Haz3lcore.CachedStatics.probe_ids_of_zipper(zipper);
+
+    /* Run statics to get info_map and the (instrumented) elaboration */
+    let (info_map, elaborated) =
+      Statics.mk(
+        ~probe_ids=base_probe_ids,
+        CoreSettings.on,
+        Builtins.ctx_init(Some(Int)),
```
7h3kk1d: Operators.default_mode instead of Int

src/haz3lcore/lang/Form.re
```diff
@@ -473,6 +474,7 @@ let get: compound_form => t =
   | Not => mk_prefix("!", Exp, P.not_)
   | TypSumSingle => mk_prefix("+", Typ, P.or_)
   | UnaryMinus => mk_prefix("-", Exp, P.neg)
+  | UnaryMinusPat => mk_prefix("-", Pat, P.neg)
```
7h3kk1d: Was this from github.com/hazelgrove/hazel/pull/2419 ?

src/haz3lcore/lang/Form.re
```diff
@@ -571,7 +573,16 @@ let infix_delimiter_ops_prefixes: list(Token.t) =
...
+/* Hot predicate: runs per atomic-form candidate on every molding query
+   (so, superlinearly during text parsing), so membership is a hash set
+   rather than a List.mem scan with polymorphic compare. */
+let infix_delimiter_ops_prefix_tbl: Hashtbl.t(Token.t, unit) = {
+  let tbl = Hashtbl.create(64);
+  List.iter(t => Hashtbl.replace(tbl, t, ()), infix_delimiter_ops_prefixes);
+  tbl;
+};
+let is_infix_delimiter_op_prefix = t =>
+  Hashtbl.mem(infix_delimiter_ops_prefix_tbl, t);
```
7h3kk1d: If we're not walking over the keys at all I would just inline the hashtable into the mem call

Something like this. I'm especially trying to minimize the surface area of our "public" apis
```suggestion
let is_infix_delimiter_op_prefix = 
  Hashtbl.mem({
  let tbl : Hashtbl.t(Token.t, unit) = Hashtbl.create(64);
  List.iter(t => Hashtbl.replace(tbl, t, ()), infix_delimiter_ops_prefixes);
  tbl;
}, _);
```

src/haz3lcore/lang/Form.re
```diff
@@ -598,7 +609,12 @@ let amiguous_polymorphs: list(Token.t) = {
   single_token_labels |> List.filter(appears_in_other_forms);
 };
 
-let is_ambiguous_polymorph = List.mem(_, amiguous_polymorphs);
+let amiguous_polymorph_tbl: Hashtbl.t(Token.t, unit) = {
+  let tbl = Hashtbl.create(16);
+  List.iter(t => Hashtbl.replace(tbl, t, ()), amiguous_polymorphs);
+  tbl;
+};
+let is_ambiguous_polymorph = t => Hashtbl.mem(amiguous_polymorph_tbl, t);
```
7h3kk1d: Same note as above.

src/haz3lcore/projectors/implementations/TableRenderer.re
```diff
@@ -40,6 +40,7 @@ let parse = (_sort: Sort.t, exp: Exp.t) => parse_table(exp);
 
 /* Initialize table model from parsed value */
 let init = (_: v) => {menu_state: None};
+let empty = {menu_state: None};
```
7h3kk1d:
```suggestion
let empty = {menu_state: None};
let init = (_: v) => empty;
```

hazel-programs/docs/reference/tuples.hz
disconcision: @7h3kk1d  it'll strip leading whitespace on-parse. so it doesn't matter what's in the .hz. if you copy-paste from the editor that should round-trip, even if indentation is retained in the .hz. I didn't want to be too precious on this issue as it may change with completion-provenence, where indentation is now not obligate system-managed.

do you have preference for what an actual policy here should look like? after completion-provenence we'll now have the option of just importing exactly as-is. debatably though some kind of normalization (just indentation, or more) would be reasonable to do.

src/haz3lcore/lang/Token.re
```diff
@@ -300,6 +300,30 @@ let of_projector_invoke = (input: t): option(t) =>
...
   };
 
+/* A `_opt` suffix on the invoke body is a trigger OPTION (e.g. the
+   probe renderer in `^^probe_table`) — stripped for validity; Triggers
+   parses the option itself. `_` is safe: no kind name contains one,
+   and it merges into a single editor token. */
+let split_invoke_opt = (body: t): (t, option(t)) =>
+  switch (String.index_opt(body, '_')) {
+  | Some(i) => (
+      String.sub(body, 0, i),
```
7h3kk1d: Post ocaml 5.5 we should use https://ocaml.org/manual/5.5/api/String.html#VALsplit_first

It might be worth adding this to string_util and noting that or we can continue going to base

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -45,6 +45,10 @@ module Settings = {
     show_filters: bool,
     show_ascriptions: bool,
     show_unknown_as_hole: bool,
+    /* Emit EmptyHole as an explicit "?" TILE instead of Grout — used by
+       FastParse so a source `?` lands as the tile the typing parser
+       would make, keeping `?` (explicit) distinct from `¿` (Grout). */
```
7h3kk1d: Want to clarify here is `¿` signifying actual grout or the upside down question mark we use to represent grout.

src/haz3lcore/pretty/ExpToSegment.re
```diff
@@ -2622,14 +2638,16 @@ and typ_to_pretty = (~settings: Settings.t, typ: Typ.t): pretty => {
...
         p_just([
           Grout({
             id,
             shape: Convex,
           }),
         ]);
+      } else if (settings.hole_tiles) {
+        text_to_pretty(typ |> Typ.rep_id, Sort.Typ, "?");
       } else {
         text_to_pretty(typ |> Typ.rep_id, Sort.Typ, "?");
```
7h3kk1d: Why not just else

7h3kk1d: I guess I'm confused as to the difference between these settings in this case.

hazel-programs/docs/reference/tuples.hz
7h3kk1d: I believe there are slides that had additional left side padding to align stuff. I don't think we'd want to lose that unless we're saying indentation should be entirely system managed.

disconcision: @7h3kk1d ah, i've generally been proceeding under the assumption that indentation was entirely system managed, ie i've been stripping it for pasting etc... wasn't aware anyone was using it to align things. standard editor/IDE behavior seems quite variable here. i don't care at all myself, hence heading to the completion-provenance regime where it's all user managed with settings for various kinds of auto... right now we're in the awkward position where part of it is forced system managed

src/haz3lcore/lang/Form.re
disconcision: @7h3kk1d as written doesn't this rebuild the table per-call?

src/haz3lcore/lang/Form.re
disconcision: yes, snuck in, removed

src/CLI/Cli.re
disconcision: no, removed. well, there is an issue with trailing newlines getting autoinserted in the file but now we target that specifically

src/haz3lcore/lang/Form.re
7h3kk1d: It shouldn't since it's just constructing the function. If you reintroduce the lambda it would.

hazel-programs/docs/reference/tuples.hz
7h3kk1d: I'm fine with having entirely system-managed indentation, but it was being used. If you want to make a decision I support you.

src/b2t2/Slides.re
disconcision: @7h3kk1d  we need to associate the files with titles and with an order... title is fine, but if we put the order in the filename, reordering would cause diff chaos. so we need some kind of separate manifest... could do some separate text file system for that, but this seems 6 of one...

7h3kk1d: I'm fine keeping the order in here if you want. For the b2t2 stuff I'm also fine keeping the current names and just doing it alphabetically. Putting the order in the filename shouldn't mess up the diffs much though because git will just track them as moves.

src/CLI/Cli.re
disconcision: unified into PersistentZipper.parse_text. remaining difference is intentional: CLI reports an error, persistence loads an empty buffer so boot doesn't brick on a bad blob

src/haz3lcore/lang/Form.re
disconcision: i think with the underscore it's a syntactic lambda, not partial application so it will re-run every time... in any case, i'm putting up a different version, let's see if you're happy with it

src/haz3lcore/pretty/ExpToSegment.re
disconcision: its the marker: grout's text serialization. ¿ in source becomes real Grout on read. comment reworded to "Grout (whose text serialization is the ¿ marker)"

src/haz3lcore/pretty/ExpToSegment.re
disconcision: collapsed

src/haz3lcore/lang/Form.re
7h3kk1d: Ugh, I got partial application mixed up with currying again.
```reason
let x = (a, b) => print_endline(Int.to_string(a + b));

let y =
  x(
    {
      print_endline("Hello world");
      3;
    }, _
  );

y(4);
y(5);
```
outputs 
```
Hello world
7
Hello World
8
```

and
```reason
let x = (a, b) => print_endline(Int.to_string(a + b));

let y =
  x(
    {
      print_endline("Hello world");
      3;
    },
  );

y(4);
y(5);
```
just outputs hello world once

 Either way, if you don't want to do it, feel free to do whatever you're comfortable with.

src/haz3lcore/zipper/action/Perform.re
```diff
@@ -41,13 +41,65 @@ let go =
...
       (
-        Parser.can_fast_paste(clipboard, z, ~root)
-          ? Parser.fast_paste(clipboard, z, ~root)
-          : Parser.to_zipper(~root, ~zipper_init=z, clipboard)
+        switch (menhir_pasted) {
+        | Some(z) => Some(z)
+        | None => Parser.to_zipper(~root, ~zipper_init=z, clipboard)
+        }
       )
       |> Option.map(maybe_reassoc_thorough)
```
7h3kk1d: Just want to make sure I understand the new flow. 

`Parser.try_segment_paste` uses the stored segment from internal hazel copy to paste without parsing.
Otherwise `Parser.fast_paste_blocker` determines whether to try using the menhir version of the parser for pasting the clipboard into a segment or fallback to inserting character by character.

So the menhir_pasted block is the new `Parser.fast_paste`? I'm assuming there's some circular dependency which keeps us from putting it there.

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,536 @@
...
+  };
+
+let of_text =
+    (
+      ~materialize: materialize=(_, _) => None,
+      ~collect_refractors: bool=false,
+      ~root: Sort.t,
+      text: string,
+    )
+    : option(Segment.t) => {
```
7h3kk1d: Can this return the collected refractors or is there a reason it's a ref?

7h3kk1d: I just tried it with moving the code into fast_paste. Feel free to try the patch. If you prefer it the way it is no pressure.

[fast-paste-extract.patch](https://github.com/user-attachments/files/31000548/fast-paste-extract.patch)

7h3kk1d: I'm still reviewing the code and just trying to get it in a spot I can comprehend.

src/haz3lcore/zipper/action/Perform.re
disconcision: yeah there was a cycle Parser -> PersistentZipper -> MarkerParse -> Parser

src/haz3lcore/FastParse.re
disconcision: yeah that's better, merging it in

src/haz3lcore/zipper/action/Perform.re
disconcision: yeah i'm not satisfied by the complexity here... the issue is that the menhir fast path can only be used when we have a complete or almost-complete term (i've special-cased things that can be made terms by adding a hole to the end, though these incur a double parse). so there are still some things that the segment parser covers that the fast parse doesn't. not sure it's actually worth keeping the complexity here long term but wasn't inclined to throw it away immediately either

7h3kk1d: Yeah, that sounds good. Just getting my head around it with the naming

src/haz3lcore/zipper/action/Triggers.re
```diff
@@ -4,19 +4,59 @@ open OptUtil.Syntax;
 
 /* Syntax replacement operations to automatically run after insertion */
 
+/* `^^kind_opt` — a trigger may carry an option after `_`: the base name
+   picks the kind, the option a non-default model (`^^probe_table` places
+   a probe with the table renderer active). `_` is an identifier char, so
+   the whole trigger lexes as one editor token. */
+let split_trigger_opt = (s: string): (string, option(string)) =>
+  switch (Token.of_projector_invoke_base(s)) {
+  | Some(name) => ("^^" ++ name, Token.of_projector_invoke_opt(s))
```
7h3kk1d: I don't know how frequently this is called but it's splitting the string twice in the calls to `of_projector_...`. It seems like the work happenign inside `Token.split_invoke_opt` should somehow be kept through this. Currently the prefix handling seems to be the main difference.

src/haz3lcore/zipper/action/Triggers.re
```diff
@@ -4,19 +4,59 @@ open OptUtil.Syntax;
...
+
 /* Check if a string is a refractor trigger name (e.g., "^^type", "^^probe") */
-let is_refractor_trigger = (s: string): bool =>
+let is_refractor_trigger = (s: string): bool => {
+  let (s, _) = split_trigger_opt(s);
   String.length(s) > 2
   && String.sub(s, 0, 2) == "^^"
   && {
     let kind_name = String.sub(s, 2, String.length(s) - 2);
     ProjectorCore.Kind.is_name(kind_name)
```
7h3kk1d: This seems reminiscent of Token.of_projector_invoke.

7h3kk1d: I did another thing here. Where I just reused the existing Token functions (also added some slightly longer doc strings with examples).

[trigger-parsing-review.patch](https://github.com/user-attachments/files/31033012/trigger-parsing-review.patch)

The same parsing was happening in a few different ways:
* Triggers.split_trigger_opt re-attached the ^^ prefix that its callers immediately stripped back off
* refractor_of_invoke_token parsed the token three times once for the kind, again for the kind, again for the option.

The `^^kind_opt`` encoding would probably be easier to reason about if we had separate types for the full parse match as well as for `kind` and `opt` but I don't think you want to do that as part of this pull request.

If you think this is less readable or not helpful feel free to disregard.

src/haz3lcore/zipper/action/Triggers.re
```diff
@@ -81,10 +141,21 @@ let expand_projector = (z: t): option(t) => {
...
-  Piece.mk_tile(Form.mk_atom_op(Exp, Token.mk_projector_invoke(kind)), []),
-  Piece.mk_tile(Form.get(ApExp), [seg]),
-];
+    (~model: string="()", kind: ProjectorCore.Kind.t, seg: Segment.t)
+    : Segment.t => {
+  let opt_suffix =
+    switch (refractor_opt_of_model(kind, model)) {
+    | Some(opt) => "_" ++ opt
+    | None => ""
+    };
```
7h3kk1d, line 153: Maybe

```reason
let refractor_to_invoke =
    (~model: option(string)=?, kind: ProjectorCore.Kind.t, seg: Segment.t)
    : Segment.t => {
  let opt_suffix =
    switch (Option.bind(model, refractor_opt_of_model(kind))) {
    | Some(opt) => "_" ++ opt
    | None => ""
    };
```

src/haz3lcore/zipper/Parser.re
```diff
@@ -16,38 +16,42 @@ let set_segment_cache = (seg: option(Segment.t), str: string): unit =>
 /* Try pasting from segment cache. Returns Some if cache hits and
    guards pass (caret Outer, no token merging at boundaries).
    The segment gets fresh IDs to support multiple pastes. */
```
7h3kk1d: This comment got shifted up

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+   this; everything else takes of_text below. */
+let parsed_of_text =
+    (
+      ~materialize: materialize=(_, _) => None,
+      ~collect_refractors: bool=false,
+      ~root: Sort.t,
+      text: string,
+    )
+    : option(parsed) => {
+  bail_note := None;
```
7h3kk1d: This should just return a result

src/haz3lcore/zipper/MarkerParse.re
```diff
@@ -0,0 +1,84 @@
...
+      List.fold_left(
+        (acc, child) =>
+          switch (acc) {
+          | Some(_) => acc
+          | None => find_marker(~implicit_hole, child)
+          },
+        None,
+        t.children,
+      )
+    | _ => None
```
7h3kk1d, line 90: This doesn't descend into projectors. Not sure if that's concerning or not.

src/haz3lcore/zipper/PersistentSegment.re
```diff
@@ -41,27 +48,41 @@ let restore = (persisted: t): Zipper.t =>
...
-        );
-        restore_from_backup_text(persisted.backup_text, ~root);
-      }
-    )
-    |> Zipper.sexp_of_t
-    |> Sexplib.Sexp.to_string,
+/* A text-backed slide: parsing is deferred to PersistentZipper's text
+   path (FastParse first), so boot does no sexp round-trip for it. */
+let of_text = (text: string): t => {
+  segment: "",
```
7h3kk1d: Are we doing this instead of an option for persistence reasons?

src/haz3lcore/zipper/PersistentSegment.re
```diff
@@ -41,27 +48,41 @@ let restore = (persisted: t): Zipper.t =>
...
-    }
-  )
-  |> Zipper.unzip(~direction=Left);
+/* Committed .hz slide text keeps human indentation, but Hazel computes
+   indentation at layout time and renders literal leading spaces ON TOP
+   of it (doubled, drifting) — so text slides are flattened at load.
+   The strip is blind per-line (matching the old regen-slides.sh encode
+   step); a multi-line string literal would be altered, so slide sources
+   must not contain them. */
+let flatten_indentation = (text: string): string =>
```
7h3kk1d: I think this is already StringUtil.trim_leading

src/haz3lcore/zipper/PersistentSegment.re
7h3kk1d, line 1: I think PersistentSegment was just added to have a non-cursor version of the ml files that the slides could be stored in. Since we're not doing that anymore and we don't store it in the browser, these functions can probably be inlined as part of the conversion between the text and the PersistentZipper.

7h3kk1d:
There's a stray Livelits.ml file that needs to be deleted.

src/haz3lcore/zipper/PersistentZipper.re
```diff
@@ -6,23 +6,114 @@ type t = {
...
 
-let unpersist = (persisted: t, ~root) =>
-  try(Sexplib.Sexp.of_string(persisted.zipper) |> Zipper.t_of_sexp) {
-  | _ =>
+/* A text-backed slide: the program text (a committed .hz file) is the
+   only source. The empty zipper string marks the text path as
+   intentional, so unpersist takes it without the stale-serialization
+   warning. */
+let of_text = (text: string): t => {
+  zipper: "",
```
7h3kk1d, line 29: Same question as above. Are we not using an option type because of compatibility with the storage?

src/haz3lcore/zipper/PersistentZipper.re
```diff
@@ -6,23 +6,114 @@ type t = {
...
+  if (persisted.zipper == "") {
+    from_backup_text(persisted.backup_text, ~root);
+  } else {
+    try(Sexplib.Sexp.of_string(persisted.zipper) |> Zipper.t_of_sexp) {
+    | _ =>
+      print_endline(
+        "Warning: using backup text! Serialization may be for an older version of Hazel.",
+      );
+      from_backup_text(persisted.backup_text, ~root);
     };
```
7h3kk1d, line 121: I'm reading this as a migration story. Long term we should decide how important this is and if we're going to keep doing it we probably need versions.

src/haz3lcore/zipper/TextRoundtrip.re
```diff
@@ -14,10 +14,7 @@
...
  * round-trip distinct from implicit Grout. */
 
-open Util;
-open Base;
-
-let default_implicit_hole = "\xc2\xbf";
+let default_implicit_hole = Token.implicit_hole_marker;
 
 let to_text =
     (~implicit_hole=default_implicit_hole, persisted: PersistentSegment.t)
```
7h3kk1d: If we remove PersistentSegment and this just takes a segment maybe the MarkerParse and this file can be reunified.

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+     token is parsed back into a kind by Triggers.refractor_of_invoke_token,
+     keeping FastParse below the action layer. */
+  let refractors = ref([]);
+  /* PROVENANCE RETIREMENT LIST — these token equivalences (and the
+     hole_tiles printer setting) exist only because the menhir AST loses
+     concrete syntax: atom spellings, optional tokens, display-flag ids.
+     When completion-provenance (lexeme/shard retention) lands, each one
+     deletes; if this list grows past a handful, flip priorities and
+     land provenance first. Members: float spellings, label quoting,
+     optional leading sum +, nullary f() vs f(()). */
```
7h3kk1d: I don't know what this comment is in reference to

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+     MakeTerm re-reads the source spelling and meaning is preserved.
+     Both sides must be float-syntax (dot/exponent): an int/float pair
+     is a genuine sort difference and must still mismatch. */
+  let float_syntax = (s: string): bool =>
+    String.exists(c => c == '.' || c == 'e' || c == 'E', s)
+    && Option.is_some(float_of_string_opt(s));
+  let float_equal_toks = (a: string, b: string): bool =>
+    float_syntax(a)
+    && float_syntax(b)
+    && float_of_string(a) == float_of_string(b);
```
7h3kk1d: Probably easier to just have float_syntax return the optional and do the option zip or else false.

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+  /* The printer quotes labels only when necessary; a source label may
+     carry backticks the reprint drops. Source spelling lands. */
+  let unquote = (s: string): option(string) =>
+    String.length(s) >= 2 && s.[0] == '`' && s.[String.length(s) - 1] == '`'
+      ? Some(String.sub(s, 1, String.length(s) - 2)) : None;
+  let label_equal_toks = (src: string, printed: string): bool =>
+    switch (unquote(src), unquote(printed)) {
+    | (Some(s), None) => s == printed
+    | (None, Some(p)) => src == p
+    | _ => false
```
7h3kk1d: I can't tell but should this case just check `src == printed`?

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+    if (idx^ >= Array.length(toks)) {
+      note(
+        "segment expects '" ++ text ++ "' but source tokens are exhausted",
+      );
+      raise(Mismatch);
+    };
+    let t = toks[idx^];
+    if (t.text == text
+        || float_equal_toks(t.text, text)
+        || label_equal_toks(t.text, text)) {
```
7h3kk1d, line 191: I guess this covers the equal labels case.

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+    idx^ + k < Array.length(toks) ? Some(toks[idx^ + k].text) : None;
+  /* Projector trigger in the source (^^kind( ... )): the menhir parse
+     dropped it, so the segment holds the bare wrapped term. Consume the
+     trigger tokens around the corresponding pieces and materialize the
+     projector piece. The first unmatched `)` closes the trigger: inner
+     parens are consumed symmetrically by the wrapped pieces. */
+  let is_trigger_next = (): bool =>
+    switch (peek(0), peek(1)) {
+    | (Some(t), Some("(")) =>
+      String.length(t) > 2 && String.sub(t, 0, 2) == "^^"
```
7h3kk1d:
```suggestion
      Option.is_some(Token.of_projector_invoke(t))
```

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+          switch (Token.of_projector_invoke_base(trigger)) {
+          | Some(name) => name
+          | None => raise(Mismatch)
+          };
+        Language.ProjectorKind.of_name(name);
+      }
+    ) {
+    | kind => !Language.ProjectorKind.is_refractor(kind)
+    | exception _ => false
+    };
```
7h3kk1d:
```suggestion
    switch (trigger_kind(trigger)) {
    | Some(kind) => !Language.ProjectorKind.is_refractor(kind)
    | None => false
    };
```

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+          (List.rev(acc), ps);
+        } else {
+          switch (ps) {
+          | [] => raise(Mismatch)
+          | [q, ...qs] => grab(qs, List.rev(weave_piece(q)) @ acc)
+          };
+        };
+      let (wrapped, rest') = grab([p, ...rest], []);
+      let (close_gap, _) = expect(")");
+      let inner = wrapped @ gap_pieces(close_gap);
```
7h3kk1d: This is a duplicate of above.

I think you can just put it in a function and share it:

```reason
consume_trigger =
      (seg: Segment.t): (string, string, Segment.t, Segment.t) => {
    let trigger = toks[idx^].text;
    let (trig_gap, _) = expect(trigger);
    let (paren_gap, _) = expect("(");
    if (paren_gap != "") {
      raise(Mismatch);
    };
    let rec grab = (ps: Segment.t, acc: Segment.t) =>
      if (peek(0) == Some(")")) {
        (List.rev(acc), ps);
      } else {
        switch (ps) {
        | [] => raise(Mismatch)
        | [q, ...qs] => grab(qs, List.rev(weave_piece(q)) @ acc)
        };
      };
    let (wrapped, rest) = grab(seg, []);
    let (close_gap, _) = expect(")");
    (trigger, trig_gap, wrapped @ gap_pieces(close_gap), rest);
  }
```

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+               gap before the closing token belongs inside the child */
+            let child = weave_seg(List.nth(t.children, shard_i - 1));
+            let (gap, tok) = expect(label_tok);
+            (
+              children_acc @ [child @ gap_pieces(gap)],
+              [tok, ...label_acc],
+              shard_i + 1,
+            );
+          },
+          ([], [tok0], 1),
```
7h3kk1d: Any reason we're doing 1-based indexing?

Also I think this is probably easier with List.fold_left2 but if you prefer tracking the index that's fine.

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+    | Some({
+        segment: [Tile({label: ["{", "}"], children: [inner], _})],
+        refractors,
+      }) =>
+      Some({
+        segment: inner,
+        refractors,
+      })
+    | _ =>
+      note("mod-root wrap did not yield a single module body");
```
7h3kk1d: I think we're blowing away the attempt notes above

src/haz3lcore/FastParse.re
```diff
@@ -0,0 +1,576 @@
...
+      None;
+    };
+  } else if (root != Sort.Exp) {
+    None;
```
7h3kk1d: If this returns a result we can give the reason here.

src/menhirParser/AST.re
```diff
@@ -476,6 +481,41 @@ let rec gen_exp_sized = (~minimal_idents: bool, n: int): QCheck.Gen.t(exp) => {
...
+              );
+            let+ body = self((n - 1) / 2);
+            Let(VarPat(name), Module(Array.to_list(items)), body);
+          },
+          {
+            /* Builtin-livelit name in expression position: ^name —
+               combines with the existing Ap/Dot generators for uses. */
+
+            let+ name = gen_ident;
+            LivelitName("^" ++ name);
```
7h3kk1d, line 517: Do we want to generate actual livelit names if those are hardcoded in the system?

src/menhirParser/AST.re
```diff
@@ -476,6 +481,41 @@ let rec gen_exp_sized = (~minimal_idents: bool, n: int): QCheck.Gen.t(exp) => {
...
+                        let* tp = gen_tpat;
+                        let+ t = gen_typ_sized(size / 2);
+                        ModItemType(tp, t);
+                      },
+                    ]),
+                  sizes,
+                ),
+              );
+            let+ body = self((n - 1) / 2);
+            Let(VarPat(name), Module(Array.to_list(items)), body);
```
7h3kk1d, line 510: Do we need module shrinking?

src/menhirParser/Conversion.re
```diff
@@ -215,13 +220,19 @@ module rec Exp: {
...
     | ListExp(l) => list_lit(List.map(of_menhir_ast, l))
-    | TupleExp([TupLabel(_) as tl]) => parens(tuple([of_menhir_ast(tl)]))
-    | TupleExp([e]) => parens(of_menhir_ast(e))
-    | TupleExp(e) => parens(tuple(List.map(of_menhir_ast, e)))
+    | ParenExp(TupleExp([TupLabel(_) as tl])) =>
+      parens(tuple([of_menhir_ast(tl)]))
+    | ParenExp(e) => parens(of_menhir_ast(e))
+    | TupleExp([]) => tuple([])
+    | TupleExp([TupLabel(_) as tl]) => tuple([of_menhir_ast(tl)])
+    | TupleExp([e]) => of_menhir_ast(e)
```
7h3kk1d, line 237: This is my fault but I think since the comma was required we can drop this special casing https://github.com/hazelgrove/hazel/commit/7a644a8ef72fb5f904bce3b798d2e1c1bf2c2fc3

It would be great for all the sorts if parens matched parens and tuples matched tuples. I don't think we need to special case on tuplabel at all anymore.

src/menhirParser/Conversion.re
7h3kk1d: This no longer needs to wrap a singleton one.

src/web/view/ScratchMode.re
```diff
@@ -56,23 +47,39 @@ module Scratchpad = {
...
+      /* Text-backed originals (committed .hz, zipper == "") mint fresh
+         ids on every parse, so id-sensitive segment equality can never
+         match; compare by the text projection instead — FastParse loads
+         the text verbatim, so an unedited slide prints byte-identically
+         modulo the stored final newline (the writer's artifact, which
+         the print never carries). */
+      let unchanged =
+        switch (original) {
+        | None => false
+        | Some(pce) when pce.editor.zipper.zipper == "" =>
```
7h3kk1d: Isn't this always true for the builtin slides?

test/Test_FastParseCorpus.re
```diff
@@ -0,0 +1,104 @@
...
+
+let tests = (
+  "FastParseCorpus",
+  [
+    test_case(
+      "every corpus program fast-paths",
+      `Quick,
+      () => {
+        let files =
+          ["hazel-programs", "../hazel-programs"]
```
7h3kk1d, line 47: Why do we need both of these? Is it using the users working directory?

test/Test_MenhirCorpus.re
7h3kk1d, line 1: What's the real difference between the FastParse and Menhir corpuses? Aren't they both using the menhir parser?

7h3kk1d: Is it the additional logic in FastParse for maintaining the gaps?

test/Test_MenhirCorpus.re
```diff
@@ -0,0 +1,120 @@
...
+  let s = really_input_string(ic, n);
+  close_in(ic);
+  s;
+};
+
+let equal_terms =
+  Equality.(
+    equality({
+      ...syntactic_settings,
+      ignore_parens: true,
```
7h3kk1d: Why are we ignoring parens here?

test/Test_ReparseDocSlides.re
```diff
@@ -22,41 +22,65 @@ let doc_slide_reparses = ((name, slide: CellEditor.Model.persistent)) => {
...
-           ),
-      );
+      /* Text-backed slides (committed .hz, zipper == "") have no stored
+         sexp. Structural segment equality is also the wrong bar there:
+         projector MODELS embed freshly-minted ids per materialization.
+         The meaningful contract is text-level: the LOAD path (fast or
+         fallback) reproduces the committed text byte-for-byte, and when
+         the fast path succeeds it reads the same term as the typing
+         parse. */
+      if (slide.editor.zipper.zipper == "") {
```
7h3kk1d: This is always true right?

test/Test_ReparseDocSlides.re
7h3kk1d, line 1: I can't tell how these tests aren't duplicating what the corpus and fastparse tests already do.

test/Test_TextRoundtrip.re
```diff
@@ -66,9 +66,23 @@ let slide_roundtrip_case =
...
+    ? (
+      name,
+      PersistentZipper.from_backup_text(p.backup_text, ~root=Exp)
+      |> PersistentSegment.persist,
+    )
+    : (name, p);
+
 let doc_slide_cases =
   Web.Init.documentation_slides
   |> List.filter(s => !is_b2t2(s))
```
7h3kk1d: claude tells me this is so fast now we can do b2t2 slides.

test/Test_Editing.re
7h3kk1d, line 1: I don't think it would pass right now so not a worry. But a useful/cool pbt would be to cut random substrings and paste them back in.

7h3kk1d (changes requested):
I probably left too many comments. I did a functional review and everything seems to work great. I'm available if any of this would be easier to go over in person or on a call. Also feel free to push any cleaning tasks to a different pr or delay them.

src/haz3lcore/zipper/action/Triggers.re
disconcision: applied your trigger-parsing patch

src/haz3lcore/zipper/action/Triggers.re
disconcision: applied

src/haz3lcore/zipper/action/Triggers.re
disconcision: done

src/haz3lcore/zipper/PersistentSegment.re
disconcision: i guess... option would change the serialized shape of existing idb data, didn't want to deal with that

src/haz3lcore/zipper/PersistentZipper.re
disconcision: yes

src/haz3lcore/zipper/PersistentZipper.re
disconcision: maybe should make an issue for versioning

7h3kk1d: My preference is to just allow ourselves to blow away idb persistence as long as we're comfortable with it. But if we're avoiding migrating types then I think we should do it sooner than later.

src/menhirParser/AST.re
disconcision: the round-trip PBT is syntax-level and never consults the registry so i dont think this matters?

src/menhirParser/AST.re
disconcision: would be nice i suppose

src/menhirParser/Conversion.re
disconcision: agent tried it... supposedly they're load-bearing at the term level: MakeTerm reads (a=1,) as Tuple([TupLabel]) (Prod for types), so without the pre-empting case the singleton unwrap yields a bare TupLabel and some MenhirParser cases fail. left a comment at the site

test/Test_FastParseCorpus.re
7h3kk1d: I was mostly concerned about it being cwd relevant at all in case someone runs it from a different directory. You can just leave it and if it breaks someone can fix it then.

test/Test_MenhirCorpus.re
disconcision: yes. well, technically there's more... the fastparse path takes things like float literals verbatim from the source at the same time it weaves that gaps back it. the need for this part goes away with completion-provenence, but the gaps don't

src/menhirParser/AST.re
7h3kk1d: It's the same generator that gets used in all of the tests including statics/evaluator. I guess it's fine for it to just be free livelits. Making a better generator for those cases can be future work.

test/Test_ReparseDocSlides.re
disconcision: overlapping i guess... these go through the slide machinery, technically. can try to rejigger depending on how much you care

test/Test_Editing.re
disconcision: yeah, have dreamed of this. we're close (tm)

test/Test_MenhirCorpus.re
7h3kk1d: That's fine.

src/haz3lcore/zipper/MarkerParse.re
disconcision: documented as a gap. will address later

test/Test_MenhirCorpus.re
disconcision: we don't need to, and no longer are. that flag was also doubling as ignore_projectors, which is now a separate flag. we need to ignore projectors as long as they have a non-syntactic init step

src/menhirParser/AST.re
disconcision: plus all livelits are moving to user defined imminently

src/haz3lcore/FastParse.re
disconcision: just a note that when completion-provenance is merged in, the taking text token content verbatim part of this will no longer be necessary, as it'll all be in the term. I asked for this as a cue to re-evaluate this aspect of the parse approach when that lands

src/menhirParser/AST.re
7h3kk1d: soon :tm:

disconcision:
@7h3kk1d @Negabinary most issues dealt with. some replies above. still more that could be done but i'm inclined to leave the rest to follow-ups if yall are ok with the current state

7h3kk1d (approved):

Negabinary (approved):

PR #2429, opened 2026-08-11
Fix #2410

### LINKED ISSUE #2410: Probes on ap disappear when argument becomes a tuple
If you probe the expression `f(x)` and then delete the `x` so it becomes `f()`, the probe disappears.

Behaviour observed on dev and probes-iv

Negabinary:


src/haz3lcore/zipper/action/Insert.re
```diff
@@ -105,9 +105,16 @@ let insert_shard = (~id: Id.t, ~d: Direction.t, t: Token.t, z: t, ~root): t =>
...
+/* Replace `d`-neighbor shard with a new one based on token `t`.
+ * `~id` overrides the default id choice (adjacent monotile or fresh);
+ * used by merge_or_noop to keep a tile's id through a shard merge. */
+let replace_shard =
+    (~id=?, d: Direction.t, t: Token.t, z: t, ~root): option(t) => {
+  let id =
+    switch (id) {
+    | Some(id) => id
+    | None => Zipper.adjacent_monotile_or_new_id(d, z)
+    };
```
disconcision, line 117: this seems fine, but another option to consider is internalizing the logic to get the id at the callsite below into the adjacent_monotile_or_new_id function, likely renaming it. that function retains an existing tile id during merging of separate tiles; this new case seems analogous, except it's separate shards belonging to the same (conceptual) tile, so it would debatably be conceptually clean to combine them.

Either way though i'm not convinced it's useful to separate the fix in the way that's currently being done; replace_shard gets the zipper so i'm not sure why id calc would happen at the callsite

disconcision (changes requested):
minor adjustment for conceptual clarity. lmk if its not clear

PR #2430, opened 2026-08-11
Dynamic Errors

Negabinary:
image

Surface the dynamic errors that are already calculated during evaluation


Negabinary:
`1/0;1`

PR #2432, opened 2026-08-12
Add an ocaml-index-based dead-code detector

7h3kk1d:
Pure tooling — no source changes, nothing deleted. The deletions in the
follow-on branches are justified by what this reports, so it is worth reading
first even though it is the least interesting diff.

## Why

OCaml's warning 32 is per-compilation-unit. This repo has ~9 `.rei` files
against ~485 `.re`, so essentially every top-level binding is exported and the
compiler cannot see cross-module dead code. **No dune warning setting changes
this** — not dev, and not the strict release profile that CI gates on.

The previous approach was a whole-program text scan. It was wrong twice in ways
the compiler caught (trailing primes, ppx-synthesised call sites) and blind in
ways nothing caught (self-recursive-only functions, members of nested
`module X = {...}`).

## What

`dune` was already building 11 `.ocaml-index` files — the `@ocaml-index` alias
is wired into five Makefile targets and **nothing consumed the output**. Each
library's index records occurrences located in its own files but keyed by the
*defining* uid, so the union of all 11 is complete cross-library reference data.

The predicate: **a uid is dead iff every recorded occurrence lies within the
source span of its own definition.** Self-recursive calls fall inside the span
by construction, which was the text scan's largest blind spot. Runs in ~3s.

`make dead-code`, `dead-code-json`, `dead-code-summary` (markdown, for
`$GITHUB_STEP_SUMMARY`). If you have run any of
`make dev`/`watch`/`test`/`test-quick`/`watch-test`, the index build is a no-op.

## Reviewing the rules

Every rule in the script exists because it killed an *observed* false positive —
found by deleting what the tool reported and then building. The four in the
third commit each produced a build break or broken code first. The ones most
worth a reviewer's attention:

- **ocaml-index columns are byte offsets, not character offsets.** Decoding
  source as UTF-8 shifted every column past `Node.text("·")` in
  `TextAreaProj.re`.
- **ppx-generated call sites are absent from the index entirely.**
  `RichProbeRegistry.yojson_of_packed_action` has exactly one occurrence — its
  own definition — despite being resolved by a `[@deriving yojson]`. So a
  name-based suppression for deriver conventions is mandatory, not a heuristic.
  It is blunt and will hide a genuinely dead `pp_foo`.
- **Modules with a `.rei` are skipped**, deferring to warning 32, which is right
  where this tool is not.
- `{|...|}` literals break the indentation-based span, and this repo's tests
  embed whole Hazel programs at column 0.

Suppressed candidates are classified and reportable (`--show-suppressed`), not
silently dropped.

## Limits, documented in the script header

Will not catch unused constructors, record fields, types,
`module`/`exception`/`external` bindings, pattern bindings, or mutually
recursive dead groups. Above all it is a **"no recorded reference" tool, not a
reachability analysis** — delete, then build. Do not point it at a
`bisect_ppx`-instrumented build; uid numbering differs.

## Also here

- `.PHONY` was missing six real targets including `ci` — a stray file named `ci`
  would have silently broken the CI build.
- `.gitignore`: `*.log` (two 0-byte scratch logs were sitting untracked at the
  root) and `__pycache__/`/`*.pyc`.

## Verification

`dune build @src/fmt src --profile release` and `make dev` both clean. No source
files are touched, so behaviour is unchanged by construction.

`--baseline`/`--check` (a ratchet mode that fails only on *new* entries) is
implemented but deliberately not wired into CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,7 @@ AGENTS.md
 
 # Scratch build/test logs
 *.log
+
+# Python bytecode (scripts/ helpers get imported by ad-hoc tooling)
+__pycache__/
+*.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ plans/
 # Cursor IDE (local)
 .cursor/
 AGENTS.md
+
+# Scratch build/test logs
+*.log
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 HTML_DIR="$(shell pwd)/_build/default/src/web/www"
 SERVER="http://0.0.0.0:8000/"
 
-.PHONY: all deps change-deps setup-instructor setup-student dev dev-helper dev-student fmt watch watch-release release release-student echo-html-dir serve serve2 repl test clean setup-zarith
+.PHONY: all deps change-deps setup-instructor setup-student dev dev-helper dev-student fmt watch watch-release release release-student echo-html-dir serve serve2 hot repl test test-quick watch-test coverage generate-coverage-html ci dead-code dead-code-json dead-code-summary clean setup-zarith
 
 all: dev
 
@@ -87,10 +87,27 @@ coverage:
 	dune runtest --instrument-with bisect_ppx --force
 	bisect-ppx-report summary
 
+# Report definitions that nothing references, using the .ocaml-index files dune
+# already builds. See scripts/find_dead_code.py for the predicate and its limits.
+# Analyses whichever ExerciseSettings/TutorialSettings variant is currently in
+# place; deliberately does not run setup-instructor, which would flip a student
+# checkout. Do not point this at a bisect_ppx-instrumented build.
+dead-code:
+	dune build @ocaml-index --profile dev
+	python3 scripts/find_dead_code.py --no-build
+
+dead-code-json:
+	dune build @ocaml-index --profile dev
+	python3 scripts/find_dead_code.py --no-build --format=json
+
+dead-code-summary:
+	dune build @ocaml-index --profile dev
+	python3 scripts/find_dead_code.py --no-build --format=markdown
+
 ci: setup-zarith
 	dune build --profile dev
 	dune runtest --instrument-with bisect_ppx --force
-	
+
 generate-coverage-html:
 	bisect-ppx-report html
 
--- a/scripts/find_dead_code.py
+++ b/scripts/find_dead_code.py
@@ -1,0 +1,613 @@
+#!/usr/bin/env python3
+"""Find definitions in src/ and test/ that nothing references.
+
+OCaml's warning 32 ("unused value declaration") is per-compilation-unit. A
+top-level `let` in a .re file with no accompanying .rei is part of that module's
+public interface, so the compiler cannot conclude it is unused -- another module
+might call it. Hazel has ~9 .rei files against ~485 .re, so essentially every
+top-level binding is exported and invisible to warning 32. No dune warning
+setting changes this, including the strict release profile CI gates on.
+
+This script closes that gap using ocaml-index, which dune already builds (the
+`@ocaml-index` alias is wired into the dev-helper, watch, test, test-quick and
+watch-test Makefile targets). Each library's index records occurrences located
+in its own files but keyed by the *defining* uid, which may belong to another
+library, so the union of all indexes is complete cross-library reference data.
+
+    make dead-code
+
+THE PREDICATE
+
+    A uid is dead iff every recorded occurrence lies within the source span of
+    its own definition.
+
+The definition site and all self-recursive calls fall inside the span by
+construction, so a function whose only callers are itself is correctly reported.
+That was the largest blind spot of the text-scanning approach this replaces.
+
+WHAT THIS WILL NOT CATCH
+
+  - Unused variant constructors, record fields, and types. Constructor
+    declarations are indexed, but [@deriving] emits ghost occurrences for every
+    constructor at the type declaration's span, and the ghost filter discards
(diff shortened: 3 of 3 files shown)


7h3kk1d:
## Sample output

Run on this branch's tip (`2d800fa7c9`), against a freshly built index:

```
$ dune build @ocaml-index --profile dev      # 16s cold, no-op if you've run make dev
$ python3 scripts/find_dead_code.py --no-build --format=markdown
                                             # 2.4s
```

### One note before the numbers

The commit message says **301** dead definitions; running it here reports **400**.
Both are correct. In the branch history this commit came from, the detector landed
*after* two rounds of text-scan-driven deletions had already removed ~120
definitions. In this stack it comes first, so it still sees all of that. The
follow-on PR is what takes `src/` to zero.

`has-interface 50` is likewise higher than it will be later — `VarBstMap.rei` and
`src/pretty` still exist at this point, and both get deleted downstream.

### `--format=markdown`

Intended for `$GITHUB_STEP_SUMMARY`. Truncated here to 12 of 400 rows; the real
output caps the table at 40 and prints `_… and 360 more_`.

### DEAD IN src/ — 400 definitions, ~3300 LOC

| Location | Definition | LOC |
|---|---|--:|
| `src/language/derivation/RuleImage.re:572` | `of_sort` | 133 |
| `src/web/exercises/CodeExercise.re:682` | `append_exp` | 97 |
| `src/util/API.re:409` | `node_request_streaming` | 83 |
| `src/web/exercises/CodeExercise.re:279` | `transition` | 67 |
| `src/util/API.re:343` | `node_request` | 63 |
| `src/haz3lcore/tiles/Segment.re:684` | `remold_sig_uni` | 49 |
| `src/web/view/ScratchMode.re:175` | `Model.persist` | 44 |
| `src/web/exercises/Tutorial.re:162` | `eds_of_spec` | 40 |
| `src/haz3lcore/tiles/Segment.re:764` | `remold_mpat_uni` | 37 |
| `src/haz3lcore/CompositionCore/CompositionView.re:133` | `ContextUtils.Local.context_of` | 36 |
| `src/language/term/Typ.re:376` | `num_nodes` | 36 |
| `src/pretty/LayoutOfDoc.re:97` | `layout_of_doc` | 35 |
| _… and 388 more_ | | |

### USED ONLY BY test/ — 21 definitions, ~168 LOC

Reported separately and never mixed in with the dead list, because some of these
are legitimate test-only fixtures rather than rot.

| Location | Definition | LOC |
|---|---|--:|
| `src/haz3lcore/CompositionCore/GeneralTreeUtils.re:176` | `get_refs_to` | 42 |
| `src/language/dynamics/Sample.re:613` | `Selection.select` | 29 |
| `src/util/StringUtil.re:106` | `trim_trailing_whitespace` | 21 |
| `src/haz3lcore/.../AgentContext.re:27` | `Utils.freshen_paths` | 16 |
| `src/language/statics/ConstructorMap.re:18` | `empty_variant_ann` | 4 |

_Suppressed 484: scope:sig-module 153, scope:expression 147, ppx-name 108,
scope:functor-or-call 67, has-interface 50, contracted-module 29, gensym 23,
entry-point 6._

## `--format=text` (the default)

Shows the evidence inline — occurrence count and the span every occurrence fell
inside, which is what lets you check a verdict without rerunning anything:

```
DEAD IN src/  (400 definitions, ~3300 LOC)
  src/language/derivation/RuleImage.re:572    of_sort            ~ 133 LOC  (1 idx occ, all inside 572-704)
  src/web/exercises/CodeExercise.re:682       append_exp         ~  97 LOC  (8 idx occ, all inside 682-778)
  src/util/API.re:409                         node_request_streaming ~ 83 LOC (1 idx occ, all inside 409-491)
  src/haz3lcore/tiles/Segment.re:684          remold_sig_uni     ~  49 LOC  (7 idx occ, all inside 684-732)
```

`remold_sig_uni` is the case worth looking at: **7 recorded occurrences, all of
them inside its own definition.** It is self-recursive and nothing else calls it.
Any count-based scan calls that live.

## `--format=json`

One object per candidate. Unlike the text format, JSON **always includes
suppressed rows**, each carrying its reasons in `sup` — so you can audit the
suppression logic without a second run:

```json
{
  "uid": "Haz3lcore__Segment.94",
  "file": "src/haz3lcore/tiles/Segment.re",
  "line": 684,
  "name": "remold_sig_uni",
  "qual": "remold_sig_uni",
  "kind": "DEAD",
  "area": "src",
  "loc": 49,
  "span": [684, 732],
  "scope": "toplevel",
  "sup": [],
  "n_occurrences": 7,
  "n_test_uses": 0
}
```

A suppressed one — reported, not silently dropped:

```json
{
  "uid": "Haz3lcore__Action.1822",
  "file": "src/haz3lcore/zipper/action/Action.re",
  "line": 247,
  "name": "prevent_in_read_only_editor",
  "kind": "DEAD",
  "loc": 32,
  "span": [247, 278],
  "scope": "toplevel",
  "sup": ["contracted-module"],
  "n_occurrences": 1,
  "n_test_uses": 0
}
```

That last one is a good example of why the suppressions exist. It has exactly one
occurrence — its own definition — so the naive predicate calls it dead. It is
reached through a module packed as `(module ...)` against a signature, so the call
site is a module reference, not a name reference, and the index has nothing to
record. Deleting it breaks the build.

926 rows total: 400 dead in `src/`, 21 dead in `test/`, 21 `src/` definitions used
only by `test/`, 484 suppressed.

Negabinary (approved):
I always trust random agent-generated python scripts.

But yes I like the idea of being able to find dead code and this looks well thought out.

PR #2434, opened 2026-08-12
Repo hygiene: stale docs, a dead npm dep, and a dangerous leftover script

7h3kk1d:
Four unrelated small cleanups. None of them touch compiled behaviour.

## Drop `src/util`'s warning-policy override

`src/util/dune` carried the repo's only per-library OCaml flags override — every
other dune file overrides `js_of_ocaml` flags only — with no comment explaining
it:

- `(dev (flags :standard -warn-error -A))` disabled all warnings-as-errors for
  util in the profile `make dev`/`test`/`ci` use.
- `(release (flags :standard -warn-error +A-58))` replaced the workspace's tuned
  release set, dropping `-strict-sequence`, `-strict-formats`, `-short-paths`,
  `-keep-locs` for util.

It was hiding nothing: both profiles are clean with it gone. This is the change
most likely to have been load-bearing, so it is the one worth verifying
independently.

## Delete `scripts/migrate_precedence.{sh,md}`

The migration completed in `054e5df2b9`. Keeping it is **actively dangerous**:
its `needs_migration()` guard matches `(Concave 35|36|37|38|40|43|44|47)`, and
35, 37, 44 and 47 are also *post*-migration values, so all 9 files under
`src/web/init/docs/` still match. Re-running it would silently corrupt every one
of them.

## Drop `@esbuild-plugins/node-resolve`

Nothing imports it: not `prebundle.js`, not `vite.config.mts`, and
`src/web/www/dune` invokes bare `esbuild`. Removing it also drops 10
transitive-only packages and correctly re-marks `esbuild` as dev-only, since it
was a peer dependency of the removed package — hence the lockfile churn.

`@types/node` stays. It looks unreferenced but `vite.config.mts` imports `fs`
and `path`.

## Doc drift

| file | fix |
|---|---|
| `INSTRUCTORS.md` | `make student` / `student-release` → `dev-student` / `release-student` (neither old name is a target) |
| `CONTRIBUTING.md` | `src/hazelweb/www/` → `src/web/www/` |
| `INSTALL.md` | `Updating.md` → `Updating-OCaml-Version.md` |
| `docs/modules.md` | `statics/Elaborator.re` is gone (elaboration lives in `Statics.re`); `tiles/Arms.re` moved to `web/app/editors/decoration/Arms.re` |

Also removes the `Version_autogenerated.re` gitignore rule — nothing generates or
references that name.

## Verification

`dune build @src/fmt src --profile release` and `make dev` both clean.


docs/modules.md
Negabinary, line 1: I don't know how I feel about this file in general - it feels weird. Could be replaced with just a few comments around the codebase and the in-hazel documentation page; then we wouldn't need to keep this up to date.

Won't block this pr on doing that though.

Negabinary (approved):

docs/modules.md
7h3kk1d: I'm ngl I didn't realize this file existed. I'm not opposed to having meta-level documentation separate from the code itself. I do generally tend away from these huge lists of file names. Since this is pre-existing I'm not going to change it here. I mostly feel like if this is agent/ai oriented we should maybe put it in a folder signaling such.

PR #2436, opened 2026-08-12
Delete dead code, to a fixpoint

7h3kk1d:
176 files, **−5,589 lines**, +16. **Worth reading commit-by-commit rather than as
a single diff** — the commits are deliberately ordered so each is independently
checkable, and squashing them together makes the reasoning impossible to follow.

| area | files | removed |
|---|--:|--:|
| `src/web` | 52 | −1,678 |
| `src/language` | 38 | −1,213 |
| `src/haz3lcore` | 35 | −1,058 |
| `src/pretty` | 14 | −880 |
| `src/util` | 20 | −607 |
| `test` | 13 | −119 |
| `src/CLI`, `src/menhirParser`, dune/opam | 4 | −34 |

## Kept at reviewer request

Review round 1 did exactly what the PR asked for — people recognised things they
wanted. **19 definitions across 12 files are restored** in "Restore definitions
reviewers asked to keep". They are listed here rather than tracked by the
detector, so this table is the record: `make dead-code` reports every one of them
again, by design.

| restored | file | asked by | reason |
|---|---|---|---|
| `err_msg` | `language/term/InvalidOperationError.re` | @Negabinary | "wanna use it shortly" |
| `fresh` | `language/term/MPat.re` | @Negabinary | `MPat` was the only term module missing one |
| `temp` ×3 | `language/term/{Sig,Mod,MPat}.re` | @Negabinary | same — three had lost it, not one |
| `print_timestamp` | `util/JsUtil.re` | @Negabinary | "don't want to look up how to do this again" |
| `print_time_prefix`, `print_time`, `print_times`, `measure_time` | `util/TimeUtil.re` | @Negabinary | "looks useful" |
| `Debug` | `haz3lcore/projectors/implementations/ProbeProj.re` | @disconcision | "would prefer to keep this" |
| `empty`, `listlit_lbl`, `mod_lbl` | `haz3lcore/lang/Token.re` | @disconcision | used in #2389 |
| `sorted_children` | `haz3lcore/tiles/Tile.re` | @disconcision | used in #2374 |
| `abbreviate_pat` | `language/term/Abbreviate.re` | @disconcision | used in #2115 |
| `shard_index` | `haz3lcore/zipper/action/Indicated.re` | @disconcision | used in #1959 |
| `to_actions` | `web/app/Log.re` | @Negabinary | wanted by the log-replay work |
| `exercise` ×3 | `web/exercises/examples/Blank*Exercise.ml` | @Negabinary | **these are templates** — see below |

The `Blank*Exercise.ml` files are the one place the sweep was simply wrong. They
were hollowed out rather than deleted, leaving a nine-line header that ends
*"…and edit the arguments below"* with nothing below it. They are unreferenced
**by design**, in the same way as the entry-point files the detector already
exempts — a template nobody has instantiated yet has no callers and never will.

## If you wanted something here, say so

The detector finds definitions with **no recorded reference**. It cannot tell rot
from:

- an API deliberately built out ahead of its callers,
- an escape hatch kept around for debugging,
- work someone has in progress on another branch and hasn't wired up yet.

All three look identical to "dead". Round 1 caught the items above; if there is
more, a comment on the relevant commit is enough, and everything here is
recoverable from git history.

## The commits

1. **Delete the dead `src/pretty` library.** 866 lines, 14 modules, zero
   references anywhere — no `Pretty.*` use, no `open Pretty`. The apparent hits
   are `PrettySegment`, an unrelated module in `haz3lcore`. It stayed compiled
   only because `src/web/dune` listed it in `(libraries)`. Removing it also drops
   **the 12 `Obj.magic` calls in `MeasuredLayout`/`Box`/`LayoutOfDoc` — the only
   ones in the repo outside js_of_ocaml DOM interop** — and
   `src/pretty/{ListUtil,Unicode}.re`, which duplicated `src/util` verbatim
   because the library declared no `(libraries)` and so could not see it.

2. **Empty files, orphan modules, commented-out code.** Two 1-byte files;
   `ReadTools.re` (115 lines, never wired into `CompositionUtils.re`'s tool list
   while its six siblings have 3-18 references each); `LogEntry`, `FailedInput`,
   `ExerciseUtil`; and 57 consecutive commented-out lines in `CodeGrading.re`.

3. **111 unused top-level definitions.** Largest: `RuleImage.of_sort` (133
   lines), `API.node_request_streaming` (81), `Indentation.go'` (46). Densest:
   `Icons.re` (11 unused SVG icons, including `add_file`/`add_folder`/
   `add_project` from the removed projects UI).

4. **Dead code exposed by the previous sweep**, to a fixpoint: 8 more in round
   two, 1 in round three, 0 in round four. Notably `Sidebar.switch_assistant`,
   whose only caller was the old assistant tab handler.

5. **Dead definitions found by the index**, to a fixpoint: 3,769 lines across 134
   files. This is the category the text scan structurally could not see — 58
   members nested inside `module X = {...}` blocks, and functions whose only
   callers are their own recursive calls (`Segment.remold_sig_uni`,
   `CodeExercise.append_exp`, `ValueChecker.check_value_mod_ctx`).

6. **Dead types, constructors and write-only record fields** — the category the
   index cannot see *at all*, because `[@deriving]` emits ghost occurrences for
   every constructor at the type declaration. Found by hand, each checked
   individually. Includes `SafeTriangle.active_state`'s
   `exit_point`/`menu_direction`, which are written in `activate` and never read;
   warning 69 cannot see this because they *are* written.

7. **`OptUtil.filter`**, restored — it became live on `dev` while this branch was
   open.

8. **Restore definitions reviewers asked to keep** — the 19 above, plus four
   dangling comments left behind by earlier deletions (`Typ.re`, `API.re`,
   `TutorialGrading.re`, and an over-long rationale in
   `Test_Evaluator_Properties.re`).

## What was deliberately not deleted

The part I would most like scrutinised, because it is where judgement replaced
mechanism.

- **Anything a ppx can synthesise a call to** without it appearing in source:
  `Sample.{show,sexp_of,…}_targets`, `RichProbe.show_packed_*`,
  `RichProbeRegistry.yojson_of_packed_action`, `ConstructorMap.equal_constructor`,
  `Tree.equal_struct`. `Sample.re`'s are reached from `EvalInfo.re`'s
  `[@deriving sexp]` on a `Sample.targets` field.
- **Comment blocks that carry knowledge rather than rot**: the `pp` overrides in
  `Annotated.re`/`IdTagged.re` are documented debugging affordances;
  `ProofRule.re`'s disabled `BinOp` rule carries a TODO about negation and
  implication plus a type-level sketch of an in-progress feature; `Statics.re`'s
  `NOTE(yuchen)` documents a correctness caveat.
- **`Test_Statics_Sums`' `badTypeName`**, reported by an earlier scan — it is
  Hazel source inside a `{|...|}` literal, not OCaml.
- **Seven reported definitions were false positives the build caught.** Each
  became a suppression rule in the detector rather than a one-off exclusion.

## Two things that grew coverage instead of shrinking it

- `Test_Coverage`'s `nested_constructors_exhaustive` and
  `type_function_scrutinee` are complete tests that were never added to the suite
  list. Both pass; now registered (62 -> 64 in that suite).
- `Test_Evaluator_Properties`' `qcheck_preservation_test` was unreferenced
  because its registration was commented out with "Disabled due to known issues
  with preservation". It is now registered as an explicit `Alcotest.skip()` so the
  gap appears in the report instead of hiding as dead code.

## One deletion worth naming

**`VarBstMap`** — a 237-line module plus a 187-line `.rei` — is deleted entirely.
Its only reference in the whole repo was
`module M = { include VarBstMap.Ordered; };` inside a match arm of
`Test_Grammar.re`, where `M` is then never used: a no-op that made the module look
reachable. Flagging it because it is a documented, odoc-commented data structure,
not obviously junk. If anyone wants it kept as future API, this is the commit to
object to.

## Verification

`dune build @src/fmt src --profile release` — the gating CI build, where warnings
32/33/34/37 are errors — is clean. `make test-quick`: **3,268 passing**.

`make dead-code` reports **23** dead definitions in `src/`, not zero:

- **19** are the restorations above, kept at reviewer request.
- **3** arrived dead from the `dev` merge and are not this branch's to judge —
  `CLI/Run.re`'s `evaluate_with_probe_map`, `Refractors.persist`, and
  `Token.of_projector_invoke_opt`, all landed by the fast-parse work and not yet
  wired up.
- **1**, `TutorialGrading.percentage_view`, is removed by #2437's
  `include Grading;` dedup, one branch up the stack.

So the detector-zero claim from round 1 no longer holds, deliberately. Anyone
re-running the sweep should read the table above first.


src/language/term/InvalidOperationError.re
```diff
@@ -7,14 +7,3 @@ type t =
   | NegativeNat
   | IntegerTooBig
   | Incomparable;
-
-let err_msg = (err: t): string =>
```
Negabinary, line 11: We should probably keep this one; I wanna use it shortly

src/language/term/MPat.re
```diff
@@ -8,8 +8,6 @@ type cls =
 
 include TermBase.MPat;
 
-let fresh: term => t = IdTagged.fresh;
```
Negabinary, line 11: I think it's nice for completeness to have the freshes around

src/language/term/Sig.re
```diff
@@ -33,9 +33,3 @@ let show_cls: cls => string =
   | EmptyHole => "Signature hole"
   | SigLet => "Let declaration"
   | SigType => "Type declaration";
-
-let temp: term => t =
```
Negabinary, line 37: Also nice for completeness to have the temps around; though they should really all be implemented with IdTagged.temp probably

src/language/term/Typ.re
```diff
@@ -373,42 +275,6 @@ let fresh_var = (var_name: string) => {
 
 /* Calculates the total number of nodes (compound
```
Negabinary: comment should go

src/language/term/Var.re
```diff
@@ -5,26 +5,6 @@ type t = string;
 
 let equal = String.equal;
 
-let length = String.length;
```
Negabinary, line 8: Fine to delete since this reflects the actual state of the codebase, but we should probably be pushing the other direction to start using these instead of strings

src/pretty/Box.re
Negabinary, line 1: yeah we haven't used pretty for a while now; it's nice code though; I'm wondering if we should spin it out into another repo as we delete it just for archival purposes; maybe that's not useful.

src/util/API.re
```diff
@@ -340,151 +328,6 @@ let request_streaming =
...
-        Firebug.console##log(error);
-        handler(None);
-      }),
-    ),
-  );
-  ignore(req##write(Js.string(Json.to_string(body))));
-  ignore(req##end_());
-};
-
 /** Streaming counterpart to [node_request]. Mirrors [request_streaming] on the
```
Negabinary: hanging comment?

src/util/JsUtil.re
```diff
@@ -48,13 +48,6 @@ let timestamp = () => date_now()##valueOf;
 
 let precise_timestamp = () => Js.Unsafe.global##.performance##now()##valueOf;
 
-let print_timestamp = (ts: float): string => {
```
Negabinary, line 51: vote keep; I don't want to look up how to do this again

src/util/TimeUtil.re
```diff
@@ -1,46 +1,3 @@
-let print_time_prefix =
```
Negabinary, line 1: looks useful

src/web/app/common/Icons.re
Negabinary, line 1: weak vote to keep all the icons

src/web/app/editors/cell/CellCommon.re
```diff
@@ -3,12 +3,6 @@ open Node;
 
 /* Helpers for creating cell ui components - mostly used by exercise mode */
```
Negabinary, line 4: dangling

src/web/app/Log.re
```diff
@@ -86,29 +78,6 @@ let update = (action: Page.Update.t, result: Updated.t('a)): unit =>
     Entry.save(Entry.mk(action));
   };
 
-let to_actions = () => {
```
Negabinary, line 89: can't remember where the log replay pr got to but I think this is useful for that

src/web/exercises/examples/BlankCodeExercise.ml
Negabinary, line 1: I assume we have blanks for a reason

src/web/exercises/examples/BlankDerivationExercise.ml
Negabinary, line 1: I assume we have blanks for a reason

src/web/exercises/examples/BlankTheoremExercise.ml
Negabinary, line 1: I assume we have blanks for a reason

src/web/exercises/TutorialGrading.re
```diff
@@ -46,377 +46,19 @@ module TestValidationReport = {
...
-               };
-             };
-           });
-      switch (found) {
-      | None => Fail
-      | Some(_) => Pass
-      };
-    };
-  }; // for each hidden bug
   //   in the test results data, find a test ID that passes test validation but fails against
```
Negabinary: dangling

src/web/init/Init.re
Negabinary: dangling (also possibly important function?)

test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -676,6 +676,17 @@ let tests = (
     QCheck_alcotest.to_alcotest(qcheck_stepper_confluence),
     QCheck_alcotest.to_alcotest(qcheck_pattern_equivalence_test),
     QCheck_alcotest.to_alcotest(qcheck_incremental_matches_fresh_after_edit),
-    // QCheck_alcotest.to_alcotest(qcheck_preservation_test), // Disabled due to known issues with preservation
+    /* Registered as an explicit skip rather than a commented-out line, so the
```
Negabinary: too much comment

Negabinary (commented):
I love the smell of deleted code

src/util/API.re
7h3kk1d: I'll get an agent to go back through and make sure it addresses all the comments.

src/haz3lcore/projectors/implementations/ProbeProj.re
```diff
@@ -387,31 +387,6 @@ let cursor_clss =
   @ depth_clss(~settings, ~ap_id, dynamics, sample);
 };
 
-module Debug = {
```
disconcision, line 406: would prefer to keep this... could stick a flag in to make it used

src/language/term/Typ.re
```diff
@@ -154,29 +154,6 @@ let rec is_arrow = (typ: t) => {
   };
 };
 
-let is_atom = (ty: t): bool =>
```
disconcision, line 157: mixed feelings about deletions of things like this... debatably an argument for interfaces. dont strongly care tho

src/language/term/Var.re
disconcision: also most of this kind of thing should go through Token rather than string at the bottom for unicode reasons. my unicode PR likewise guts this module

src/web/app/common/Icons.re
disconcision: mixed feelings. i'm not totally sure what the initial motivation was to include svgs in the code like this, but i've now found workarounds for any reason i would want to. so i wouldn't mind a purge

src/haz3lcore/lang/Token.re
```diff
@@ -40,8 +40,6 @@ let of_list = Unicode.of_list;
 
 /* Token Recognition Predicates */
 
-/* A. Secondary Notation (Comments, Whitespace, etc.)  */
-let empty = ""; /* This is invalid for view */
```
disconcision, line 43: Used in #2389

src/haz3lcore/lang/Token.re
```diff
@@ -243,7 +240,6 @@ let is_typ_var = str => is_var(str) || match(capitalized_name_regexp, str);
 /* List literals */
 let list_start = "[";
 let list_end = "]";
-let listlit_lbl = [list_start, list_end];
```
disconcision, line 309: Used in #2389

src/haz3lcore/lang/Token.re
```diff
@@ -257,7 +253,6 @@ let is_empty_tuple = equal(empty_tuple);
 /* Modules */
 let mod_start = "{";
 let mod_end = "}";
-let mod_lbl = [mod_start, mod_end];
```
disconcision, line 323: Used in #2389

src/haz3lcore/zipper/PersistentSegment.re
```diff
@@ -9,9 +9,6 @@ type t = {
 
 let to_string = Printer.of_segment(~holes="", ~indent="");
 
-let refractors_init_str =
```
disconcision: used in #2426

src/haz3lcore/tiles/Tile.re
```diff
@@ -36,15 +34,6 @@ let shapes = (t: t) => {
 
 let to_piece = t => Tile(t);
 
-let sorted_children = ({mold, shards, children, _}: t) =>
```
disconcision, line 39: used in #2374 and descendants

src/language/term/Abbreviate.re
```diff
@@ -1860,14 +1831,3 @@ let abbreviate_exp = (~available as a=12, exp: Exp.t): (Exp.t, int) => {
     (exp, length_exp);
   };
 };
-
```
disconcision, line 1843: used in #2115 and descendants

src/haz3lcore/zipper/action/Indicated.re
```diff
@@ -184,35 +184,6 @@ let for_decoration = indicated(~no_ws=true, ~ign=Piece.is_secondary);
    whitespace. Ignores secondary but returns them as fallback. */
 let for_index = indicated(~no_ws=false, ~ign=Piece.is_secondary);
 
-let shard_index = (z: ZipperBase.t): option(int) =>
```
disconcision, line 187: Used in #1959

src/web/exercises/CodeExercise.re
```diff
@@ -232,117 +219,6 @@ let editor_positions = (eds: p('a)) =>
 let positioned_editors = state =>
   List.combine(editor_positions(state), editors(state));
 
-let idx_of_pos = (pos, p: p('code)) =>
```
disconcision, line 235: i would have low confidence that unused exercise mode code isn't artifactual of weird functional regressions but not terribly personally concerned

disconcision (changes requested):
I did a manual skim of the diff and marked a few things. I also got an agent to compare this to all my feature branches, and marked several functions that would be restored verbatim with those branches, so i'd be inclined to keep those. i filtered this list a bit, omitting stuff that could be structurally better; i'll just let the diff agents regenerate hopefully-better versions of those.

I definitely spotted a few core things that I was surprised arent used, though not disbelieving. this could probably use some functional review sanity checking.

all in all a lot of stuff i'm glad is on the way out though

src/web/exercises/CodeExercise.re
7h3kk1d: Yeah, I half put this PR up with the intent for people to know what's unused rather than to blanket merge.

src/language/term/InvalidOperationError.re
7h3kk1d: 🤖 Restored.

src/language/term/MPat.re
7h3kk1d: 🤖 Restored. For the record `MPat` was the only one that lost it — `Exp`, `Pat`, `Typ`, `TPat`, `Mod` and `Sig` all still have theirs, so the family is complete again.

src/language/term/Sig.re
7h3kk1d: 🤖 Restored — and it was three, not one: `Sig`, `Mod` and `MPat` had all lost `temp`. All three are back.

On "they should really all be implemented with `IdTagged.temp`": there is no `IdTagged.temp` today. The only `temp` is `IdTagged.IdTag.temp()`, which returns the annotation, not a wrapped term — which is why all six call sites hand-write the same `{term, annotation: IdTagged.IdTag.temp()}` record. A top-level `IdTagged.temp` mirroring `IdTagged.fresh` would be about four lines plus six call-site edits. Happy to do it, but it is new code rather than a restore, so I have left it out of a deletion PR. Say the word and I will open a follow-up.

src/language/term/Typ.re
7h3kk1d: 🤖 Gone. It documented `num_nodes`. The line below it — "Number of Unknown constructors in type AST" — is a separate comment belonging to `count_unknowns`, which is still live, so that one stays.

src/language/term/Var.re
7h3kk1d: 🤖 Leaving this deleted, since you both landed in the same place — and @disconcision's point that this should go through `Token` rather than `String` for unicode reasons is the stronger version of the argument for pushing the other direction. Happy to defer to the unicode PR here.

src/pretty/Box.re
7h3kk1d: 🤖 I'll leave it, on the grounds of your own "maybe that's not useful" — the history isn't going anywhere, and `git show 14806b4ef8^:src/pretty/...` gets any of it back. If you do want an archival repo later, that commit is the one to export.

src/util/API.re
7h3kk1d: 🤖 Yes, and doubly so — it documented `node_request_streaming`, and it cross-references `[node_request]`, which the next commit also deleted. Both referents were gone. Removed.

(It survived because `refmt` turned the stranded `/** */` into a floating doc-comment structure item with a trailing `;`, so it still compiled. Worth knowing as a failure mode.)

src/util/JsUtil.re
7h3kk1d: 🤖 Restored.

src/util/TimeUtil.re
7h3kk1d: 🤖 Restored — `print_time_prefix`, `print_time`, `print_times` and `measure_time`. (`format_time_diff`, the one the file kept, was never in question.)

src/web/app/common/Icons.re
7h3kk1d: 🤖 I've gone with the purge

src/web/app/editors/cell/CellCommon.re
7h3kk1d: 🤖 I think this one's a false positive of how GitHub anchored the comment — it's the first context line of the hunk, immediately above the deletion, rather than part of it. The header is a file-level section header and five helpers still follow it (`simple_cell_item`, `caption`, `simple_cell_view`, `report_footer_view`, `panel`); only `narrative_cell` was deleted. Its claim also still holds — the consumers are `CodeGrading`, `TutorialGrading`, `CodeExerciseMode`, `DerivationExerciseMode` and `InstructorEditViews`, plus one use in `EvalResult.re`, which the "mostly" covers. Left as-is; say if you disagree.

src/web/app/Log.re
7h3kk1d: 🤖 Restored.

One caveat, since it will bite whoever picks up log replay: `to_actions` does not work as written. `DB.get_all` is an async IndexedDB callback, so `actions^` is read — and the final `print_endline` fires — before the callback ever runs; it returns `[]` every time. It also still has a `print_endline("HELLO??")` in it. I have put it back verbatim rather than "fixing" it silently, but it wants a rewrite to callback style like `get_and` before it is relied on.

src/web/exercises/examples/BlankCodeExercise.ml
7h3kk1d: 🤖 You're right, and this is the one place the sweep was straightforwardly wrong. These weren't deleted, they were hollowed out — the nine-line header survived, ending "…and edit the arguments below", with nothing below it.

They're unreferenced by design: a template nobody has instantiated has no callers and never will, which is the same category as the entry-point files the detector already exempts. All three `let exercise` bindings are restored, and I verified the three `blank_spec` signatures they call still match. `ExerciseSettings_base.re` never listed them, so nothing else changes.

src/web/exercises/examples/BlankDerivationExercise.ml
7h3kk1d: 🤖 Restored — see the note on `BlankCodeExercise.ml`. All three were hollowed out rather than deleted, and `DerivationExercise.blank_spec`'s signature still matches.

src/web/exercises/examples/BlankTheoremExercise.ml
7h3kk1d: 🤖 Restored — see the note on `BlankCodeExercise.ml`. All three were hollowed out rather than deleted, and `TheoremExercise.blank_spec`'s signature still matches.

src/web/exercises/TutorialGrading.re
7h3kk1d: 🤖 Removed. It was the second half of a two-line trailing comment on `hidden_bug_status`; the first line went with the function body and this one, being standalone, survived as a sentence with its subject clause amputated.

Separately, while in there: `MutationTestingReport` in this file is now a type-only module with no references anywhere — the `MutationTestingReport` that `CodeGrading.re`, `CodeExerciseMode.re` and `CLI/Grade.re` use is the one defined in `CodeGrading.re:230`. I've left it alone rather than grow this PR, but it looks deletable if you agree.

src/web/init/Init.re
7h3kk1d: 🤖 This one has aged out — `Init.re` isn't in this branch's changed-file set at all any more.

To answer the parenthetical, since it's a reasonable thing to want to know: at the commit you reviewed, the deleted function was `get_original_doc_segment`, a memoised cache of unpersisted documentation-slide segments, plus its `original_doc_segments` ref. It existed purely so `ScratchMode.persist`'s "don't save unchanged slides" check didn't re-parse every original slide on every autosave. The lazy-startup rework on `dev` (`53401e4158`) replaced that mechanism wholesale, so both the cache and the comment are gone upstream rather than by this sweep.

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: 🤖 Fair — the first two sentences were commit-message content justifying the choice against the alternative. Trimmed to the part that tells a future reader something: that preservation does not currently hold. The `Alcotest.skip()` stanza is unchanged.

src/haz3lcore/projectors/implementations/ProbeProj.re
7h3kk1d: 🤖 `Debug` is restored as-is. I've left the "stick a flag in" part alone — wiring up a new debug flag is new behaviour, and I'd rather not add it to a PR that is meant to be subtractive. Happy to do it as a follow-up if you want it actually reachable.

src/language/term/Typ.re
7h3kk1d: 🤖 Leaving it deleted on the strength of "dont strongly care" — but it's 23 lines and one `git revert` away, so if you'd rather have it, say so and it's back.

On the "debatably an argument for interfaces" point: agreed, and that's what #2439/#2440 are for. A `.rei` makes warning 32 do this work at compile time, per-module, instead of a whole-repo sweep — the index and the interface have complementary blind spots.

src/haz3lcore/lang/Token.re
7h3kk1d: 🤖 Restored, along with the `/* A. Secondary Notation (Comments, Whitespace, etc.) */` section header that went with it — that header labels the whole predicates section, so losing it was collateral.

src/haz3lcore/lang/Token.re
7h3kk1d: 🤖 Restored. Worth noting `tuple_lbl` survived the sweep, so the three-way symmetry with `listlit_lbl`/`mod_lbl` was broken; it isn't now.

src/haz3lcore/lang/Token.re
7h3kk1d: 🤖 Restored, alongside `listlit_lbl`.

src/haz3lcore/zipper/PersistentSegment.re
7h3kk1d: 🤖 This one's out of scope now — #2426 merged, and its follow-up `b1009cf26a` removed `PersistentSegment` entirely on `dev`. The module isn't absent because of this sweep; it was already gone at this branch's base. Restoring `refractors_init_str` would mean resurrecting the module and reverting a merged design change. `PersistentZipper.re` / `MarkerParse.re` are the surviving path.

src/haz3lcore/tiles/Tile.re
7h3kk1d: 🤖 Restored.

src/language/term/Abbreviate.re
7h3kk1d: 🤖 Restored — the public `abbreviate_pat` wrapper at end of file, the one mirroring `abbreviate_exp`. (The internal recursive `abbreviate_pat` it shadows was never touched.)

If you also want the other four `Abbreviate.re` deletions back — `is_flat_ellipses`, `AbbrevBudget.label_cap`, `AbbrevBudget.label_min_budget`, `indet_term_rul` — say so and I'll add them.

src/haz3lcore/zipper/action/Indicated.re
7h3kk1d: 🤖 Restored.

src/web/exercises/CodeExercise.re
7h3kk1d: 🤖 Leaving `idx_of_pos` deleted — recoverable from history if the exercise-mode work turns out to want it.

7h3kk1d:
🤖 Everything restored is now listed in a table in the PR description, because with no allowlist in the detector that table is the only thing stopping the next sweep deleting it all again.

src/language/term/Typ.re
7h3kk1d: 👨: I don't think we're going to start making the adts for our grammar private and expose everything through a public interface so unless these are used I think I'd rather delete them. Also `is_*` functions in general are often a code smell to me since they don't give the underlying payload on a match. Obviously you don't always need it but it encourages validate instead of parse behavior (https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

src/util/Aba.re
7h3kk1d, line 1: If we're going to keep this file as is I think it neeeds a doc string describing what it is all about.

src/util/ListUtil.re
```diff
@@ -543,25 +519,6 @@ let map_with_history = (f: (list('y), 'x) => 'y, xs: list('x)): list('y) => {
   aux([], xs);
 };
 
-let rec fold_left2_opt =
```
7h3kk1d, line 546: If we want this I think we should just use fold_left2 and catch the exn

src/web/exercises/CodeExercise.re
```diff
@@ -36,11 +36,6 @@ type point_distribution = {
   impl_grading: int,
 };
 
-let validate_point_distribution =
```
7h3kk1d, line 39: @disconcision @thomasporter522 is there another repo for the TA work that uses this stuff somehow?

russell-rozenbaum (approved):

Negabinary (approved):

src/web/exercises/CodeExercise.re
disconcision: there's an outdated copy of hazel in the 490 org that has the 490 exercises. i don't think (but am not sure) that that touches any more implementation code than the sample exercises in the main repo. however there is also autograder code there that might. i looked briefly but i either have forgotten or am not familiar with the current structure. i wouldn't worry about it that much, i doubt it'll be hard to get an agent to rebuilt the 490 version

disconcision (approved):

PR #2437, opened 2026-08-12
Deduplicate three copy-pasted implementations

7h3kk1d:
Eight files, +145/−308. Three independent de-duplications, one per commit. Unlike
the rest of the stack these *change* code rather than delete it, so each commit
carries an argument for why the surviving version is equivalent.

## 1. Grading primitives and the SVG icon helper

`Grading.re` documents itself as the shared primitives module and
`CodeGrading.re` does `include Grading;`. **`TutorialGrading.re` did not** — it
re-declared `type percentage`/`points`/`score`, `score_of_percent`, `score_view`
and `percentage_view` verbatim, down to the `Printf` format strings. Replaced with
`include Grading;`, which also picks up `pending_score_view`, the one primitive it
had not copied.

Separately, `icon_size`/`simple_icon` were byte-identical in `Icons.re` (web) and
`TableRenderer.re` (haz3lcore). `haz3lcore` cannot depend on `web`, so it had been
retyped. Hoisted into `Util.SvgUtil`, which both libraries reach and which already
opens `Virtual_dom`.

## 2. Point `Tutorial.re` at `EditorUtil.append_exp`

`append_exp` existed three times. `EditorUtil.re` is canonical; `CodeExercise`'s
copy was already dead and went in the preceding branch; `Tutorial`'s was the last
live duplicate.

**Verified equivalent rather than assumed**, because the two built annotations
differently:

| | Tutorial | EditorUtil |
|---|---|---|
| base case | `IdTag.fresh()` | `IdTag.mk([Haz3lcore.Id.mk()], empty_secondary)` |
| recursive | `IdTag.mk_internal(ids(e1))` | `IdTag.mk(ids(e1), empty_secondary)` |

`IdTagged.re:36-47` makes `mk_internal(ids)` structurally equal to
`mk(ids, empty_secondary)`, and `fresh()` equal to
`{ids: [Id.mk()], secondary: empty_secondary}`.

The remaining question was whether `Haz3lcore.Id.mk` and the `Id.mk` resolved
inside `IdTagged.re` are the same generator — **there are three modules named `Id`
in this repo.** Both `src/language/Id.re` and `src/haz3lcore/Id.re` are
`include Util.Id;`, so they share the one `Uuidm.v4_gen` closure.

Constructor coverage is also identical: both dispatch recursively on exactly
`Seq`, `Filter`, `Let`, `Theorem`, `TyAlias` and `Use`, and both put the same 41
constructors in the catch-all — including `ModuleExp`, which an earlier copy
handled recursively.

## 3. Functorize `SliderProj` / `SliderFProj`

63 lines each, 52 identical — the entire difference was `Bigint` vs `float`: which
atom they read, which atom they build, how they render. `SliderCore.Make` now holds
the shared body and each variant supplies only that difference, 14-16 lines apiece.

`Atom.t` is the natural seam: both were matching `Exp({term: Atom(Int(i)), _})` /
`Atom(Float(f))` and rebuilding the same way, so `PARAMS` asks for
`of_atom`/`to_atom`/`to_string` and nothing else.

Net −17 lines, which undersells it — the point is that a change to slider
behaviour is now one edit rather than two that can drift apart.

Failure message wording is preserved (`"Slider: Get: not integer literal"`,
`"SliderF: Put: not float literal"`) via a separate `literal` label rather than
being degraded to the module name. `ProjectorInit.re:12-13` references
`SliderProj.M`/`SliderFProj.M` and is unchanged.

## Verification

`make test-quick`: 3,225 passing. dev, test and release builds clean.

Both sliders were also exercised by hand in the browser — dragged each, and the
cursor inspector read `Projector (Slider) : Int` and `Projector (SliderF) : Float`.
That check matters here because a functor that typechecks can still be wired to the
wrong literal, and no test covers projector interaction.

### DIFF
--- a/src/haz3lcore/projectors/implementations/SliderCore.re
+++ b/src/haz3lcore/projectors/implementations/SliderCore.re
@@ -1,0 +1,79 @@
+open Util;
+open Virtual_dom.Vdom;
+open ProjectorBase;
+
+/* Shared implementation of the numeric slider projectors. SliderProj and
+   SliderFProj were byte-identical apart from the literal they edit, so the
+   only thing a variant supplies is how to read that literal out of an atom,
+   how to build one back from the slider's string value, and how to render it. */
+module type PARAMS = {
+  /* Prefixes the failure messages, e.g. "Slider: Get: ...". */
+  let name: string;
+  /* Names the literal in those messages, e.g. "integer". */
+  let literal: string;
+  /* The OCaml type of the literal being edited. */
+  type t;
+  let of_atom: Language.Atom.t => option(t);
+  let to_atom: string => Language.Atom.t;
+  let to_string: t => string;
+};
+
+module Make = (P: PARAMS) : Projector => {
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type model = unit;
+  [@deriving (show({with_path: false}), sexp, yojson)]
+  type action = unit;
+
+  let value_of = (any: Language.Any.t): option(P.t) =>
+    switch (any) {
+    | Exp({term: Atom(a), _}) => P.of_atom(a)
+    | _ => None
+    };
+
+  let init = (any: Language.Any.t) =>
+    switch (value_of(any)) {
+    | Some(_) => Some()
+    | None => None
+    };
+
+  let get = (info: info): P.t =>
+    switch (
+      info.syntax |> info.utility.seg_to_term |> OptUtil.and_then(value_of)
+    ) {
+    | Some(v) => v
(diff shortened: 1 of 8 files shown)


Negabinary (commented):
weak reject

- first two commits look like a very half-hearted attempt to solve some of the duplication between Tutorial mode and Exercise mode. Not sure it's worth just trimming this around the edges, I'd leave this work to be picked up when someone wants to commit to combining them
- I don't know how much value we get from functorizing slider + sliderf; they're both pretty small, and we might actually want them to have different (e.g. snapping) behaviour later.

PR #2438, opened 2026-08-12
Delete dead stylesheets, CSS rules, and unreferenced image assets

7h3kk1d:
**No compiler can check any of this**, so the liveness argument matters more than
the diff. Each commit uses a progressively stronger test.

## 1. Whole stylesheets

- **`assistant.css`** — every rule is scoped under `#assistant`, and
  nothing renders `id="assistant"`. The sidebar's HelpfulAssistant panel renders
  `AgentView.view` (in `Sidebar.re`), styled by `style/agent/*.css`, which is
  where every class the agent UI actually uses (`send-button`,
  `message-container`, `message-identifier`, `llm-identifier`, `dot1`, …) is
  defined. The only unscoped rules — `.llm-selector`/`.llm-label`/`.llm-dropdown`
  — are unreferenced in source.
- **`project.css`** — styles a projects/file-system UI that no longer exists. Its
  ids and nearly all of its classes have zero source references; the rest are
  generic (`.cell`, `.icon`) matched only as descendants of the dead ids.
- **`agent.css`** — empty file, imported nowhere.

## 2. Individual rules, judged by source grep

Rule blocks across several stylesheets, unreferenced icons, and
`style/palette.html` (a design swatch page nothing links to and no dune rule
copies into the build).

The test is deliberately over-cautious. A selector counts as **live** if any of:

- its name appears as a string literal anywhere in `src/` or `test/`;
- **some literal is a prefix of it**, since classes get concatenated —
  `"depth-" ++ string_of_int n`, `"agent-md-h" ++ string_of_int level`,
  `"highlight-code-" ++ c`, `"slash-card-" ++ kind`;
- **its name appears as a bare OCaml token.** `Sort` and `cls` constructors reach
  the DOM through `show`/`to_string` and never appear as a literal at all — that
  is `.Mod`, `.Sig`, `.MPat`, `.Rul`, `.EmptyHole`, `.Single`. A naive
  literals-only scan called all of those dead.

That cut the candidate list sharply; each remaining name was then re-checked by hand for
both hyphenated and underscored spellings.

A rule is removed only when **every** comma-separated selector is dead, judged by
the key (rightmost) compound: `.projector.probe .window-toggle` can only match an
element carrying `.window-toggle`, so it goes even though `.projector` is live.
Rules whose key compound is an element selector are left alone
(`.step-back:hover span` survives) — which leaves a few orphans but cannot break
anything.

## 3. The stronger test: grep the compiled bundle

If a class name appears nowhere in the compiled `hazel.js`, no code path can put it
on an element, so any rule naming it is unreachable. Controls confirm the test
discriminates rather than trivially passing: `projector`, `probe`, `hsod`,
`depth-`, `highlight-code-` and `agent-md-h` all appear in it.

These classes, still defined in CSS, can never be emitted, and their rules go:
`developer-notes-display`, `view-icon-button`, `deselected`, `call-cursor`,
`click-kbd-pill`, `is-index`, `qr-when-unfocused`, `window-toggle`, `step-back`,
`step-expand`. All but `call-cursor` are ones commit 2 deliberately spared for having an element
key selector — unreachable all the same, since nothing carries `.step-back`.

**`call-cursor` is a correction.** Commit 2 kept it because `call_cursor` exists in
source — but that is a function name; the hyphenated class is never emitted. The
source-grep test was wrong there and the bundle test is right. That is the honest
summary of why the bundle test is a commit of its own.

Where a rule mixed dead and live selectors, only the dead branches are dropped:
`.prompt-display, .developer-notes-display` keeps `.prompt-display`. One such trim
initially took the rule's leading comment with it, because the dead branch was
listed first and carried the preceding trivia — restored.

## Browser verification

Done, and it covers every commit: probe sidebar (including
`#sample-focus-bar` with a focused sample, and probe hover cards with kbd badges),
agent panel, agent main menu, nut menu, editor context menu, stepper (took a
step), tutorial mode (completed a step, checkmark renders), exercises mode with
grading reports, and the projectors slide. No console errors, no unstyled surfaces.

If you re-verify: **restart the browser rather than hard-reloading** — a stale Web
Worker survives a reload and reports `EXN:TypeError reading '1'` from eval, which
looks like a real regression and is not.

`licenses/Icons.md` is a blanket Noun Project attribution rather than per-icon, so
removing unused icons does not affect it.


src/web/www/style/palette.html
7h3kk1d, line 1: this looks like it might just be a useful thing

7h3kk1d: @Negabinary  @disconcision is this palette page something either of you use?

PR #2439, opened 2026-08-12
Add .rei interfaces for src/util modules

7h3kk1d:
Without an interface, every top-level binding in a module is exported, so the
compiler can never tell you one is unused. Adding a `.rei` turns warning 32 back
on — and it is a hard error under `--profile release`, which
`deploy_branches.yml` gates on. So for these modules, unused code now fails
CI instead of accumulating.

Each interface declares only the module's real external surface, so anything used
solely inside its own file becomes private. `JsUtil` alone hides a long tail of helpers
(`date_now`, `find_scroll_container`, `QueryParams.get_arguments`, …).

## Which modules

Picked by how many lines of type declaration you would have to maintain in two
places — not by file size. In that order, most first:

`ListUtil` · `JsUtil` · `StringUtil` · `Unicode` · `Tree` · `OptUtil` · `Aba` ·
`VarMap` · `StructureShareSexp` · `Direction` · `Either` · `FloatingElement`

## What's excluded

- **`Menu` and `OpenRouter`**, which restate more type lines than any of the above. Inline
  `module M: { … } = { … }` sealing gets the same coverage with the signature
  written once. `Id.re` already does it that way.
- **`include <third-party>` modules** — `BigInt`, `Sets`, `Maps`, `MapUtil`,
  `IntMap`, `Result`. `BigInt` extracts to a signature many times its source; `refmt`
  cannot parse it. Such an interface is a hand-copy of someone else's API that
  breaks on their next release.
- **`Util.re`**, which dune treats as the library interface module.

## It found dead functions immediately

`Tree.equal_struct`, and then `Tree.equal` once that went. **The index-based
detector could not see either** — both match its `^equal` ppx-name suppression,
which exists because ppx-generated call sites do not appear in the index at all.
Worth knowing that the two tools miss different things.

## Tooling

`scripts/gen_rei.sh` / `make rei FILE=...` generates a starter interface:
`ocamlc -i` over the ppx-expanded source dune already built, piped through
`refmt`. Output is refmt-clean so `@src/fmt` accepts it. It strips the library's
own alias prefix, since `-short-paths` renders siblings as `Util.Either.t`, which
is a dependency cycle inside `src/util`.

One gotcha if you write one by hand: **a `.rei` can declare fewer derivers than
the `.re` and still compile**, because the ppx wraps generated values in
`[@@@ocaml.warning "-32"]`. That drift only surfaces as `Unbound value` at a
consumer. Structural drift is caught immediately.

## Verification

`make test-quick` green. dev, test and release builds clean — release is
the one that matters, since dev will not fail on the unused values a new interface
exposes.


Negabinary:
Not sure I agree with the way Claude picked which files; I think we should be adding reis to any file that should have private methods (of which there are many). ListUtil I don't think is a good example because pretty all of the methods there are public; there's no hiding warranted at all, all the rei does is slow us down.

Honestly I think I'm more in favour of adding reis eveywhere but util.

7h3kk1d:
> Not sure I agree with the way Claude picked which files; I think we should be adding reis to any file that should have private methods (of which there are many). ListUtil I don't think is a good example because pretty all of the methods there are public; there's no hiding warranted at all, all the rei does is slow us down.
> 
> Honestly I think I'm more in favour of adding reis eveywhere but util.

I like having rei files on utils because having a strong standard library is useful for functions being available regardless of whether or not there's any current callers. Similar to your comment on the timestamp logic in JsUtil. The signatures in a bunch of our other modules were really long and non-intentional. I would like to spread the reis to those files as  well but I think that needs to be gradual adoption and thinking about what needs to change in the interfaces. I do think some of these util files are going to go away once I put up the pr to adopt janestreet base in more totality.

src/util/Aba.rei
7h3kk1d, line 1: This file deserves a comment talking about what aba is

src/util/Either.rei
7h3kk1d, line 1: Isn't this builtin to ocaml?

src/util/FloatingElement.rei
```diff
@@ -0,0 +1,2 @@
+let update_all: unit => unit;
+let setup_scroll_listener: unit => unit;
```
7h3kk1d, line 21: doc comments

src/util/Unicode.rei
7h3kk1d, line 1: A bit of documentation on this file would be  good.

src/util/VarMap.rei
7h3kk1d, line 1: Docs for what this is for.

src/util/Either.rei
7h3kk1d: Yes — both stdlib (4.12+, `Left`/`Right`) and Core (`First`/`Second`) ship one, and ours adds nothing: the `show`/`sexp`/`yojson` derivations are dead, since no `[@deriving]` type in the tree embeds an `Either.t`. Worse, every library that uses it has `-open Core` in its dune and every file that uses it also does `open Util`, which lands second — so `Util.Either` silently shadows `Core.Either` everywhere.

Following the 24 call sites, though, most of them do not want an `Either` at all:

- **`Result`-shaped (20 refs)** — `Operators.re` (13) and `Atom.re` (3) are `Either.t(_, InvalidOperationError.t)` with `L` = value and `R` = error; `Transition.re` (4) is literally `Either.L(return_value)` / `Either.R(error)`. `MatchedTyp.args` is the same shape, returning the actual arity when it disagrees with the requested one.
- **A genuine partition (4 refs)** — only `Aba.split` and its one caller `Segment.split_by_matching`, classifying each element into the A-stream or the B-stream, neither side an error. That one wants `Core.Either`.

So the end state is `Result` in four files, `Core.Either` in `Aba`, and `src/util/Either.{re,rei}` deleted. That is a behavior-affecting refactor across `language/` and `haz3lcore/` rather than an interface addition, so it is going on its own branch at the top of this stack instead of growing this PR. Resolving here.

PR #2440, opened 2026-08-12
Add .rei interfaces for leaf modules

7h3kk1d:
Same mechanism as the previous PR, applied to leaf modules — ones with very few
callers. These had no interface, so every internal helper was exported to every
module in the build. Declaring just the surface used across files takes **most of
each module's values out of the public API for a handful of declarations**.

The public set came from the `ocaml-index` occurrence data — a value is public iff
some occurrence lies in another file — not from grep, so it is the same data
`scripts/find_dead_code.py` already reasons about.

The modules: `DebugSidebar`, `TyDiForms`, `ContextMenu`, `CursorInspector`,
`Abbreviate`, `CustomStatics`, `SampleFocusBar`, `Sidebar`,
`SlashCommandOutputView`, `CompositionPrompt`, `WorkerMessagingSection`,
`ProbeText` and `EdgeScroll`.

`DebugSidebar.rei` is a single declaration that seals everything else, which is the general shape
here: leaf modules expose one `view` and everything else was incidental.

`WorkerMessagingSection.rei` is just `include DebugSection.S;` — that moves its
contract check from the `(module ...)` pack site in `DebugSidebar.re` to the
definition itself.

## No dead code turned up, and that's the expected result

Warning 32 is a hard error under `--profile release`, so a clean build is positive
confirmation that every newly private value is used internally. `src/` remains at the
detector's fixpoint. Unlike the util PR — where `Tree.equal_struct` fell out —
there was nothing left to find here.

## What the interfaces surfaced, documented rather than fixed

- **`CustomStatics.custom_statics_deferred_ap` discards `~ancestors`**, and its one
  caller in `Statics.re` passes a bare pun that resolves to `unit`, where every
  other `ancestors` call site in that file passes `ancestors_inclusive`. Harmless while the parameter is discarded; a trap if anyone ever
  reads it. Deciding what it *should* receive is a semantic call, so the `.rei`
  records the situation instead.
- **`Abbreviate.AbbrevBudget.split_evenly` is exported only because
  `Test_Abbreviate` covers it directly** — no `src/` caller. Noted in the interface
  so it is not mistaken for production API.

## Verification

`dune build @src/fmt src --profile release` and `make dev` both clean.

The seals were checked rather than assumed: referencing `DebugSidebar.toggle_bar`
and `Abbreviate.AbbrevBudget.with_budget` from outside now fails with
`Unbound value`, covering both the top-level and nested-module cases.


Negabinary (approved):
Yesss these are good rei candidates

src/haz3lcore/CompositionCore/prompt_factory/CompositionPrompt.rei
```diff
@@ -0,0 +1,4 @@
+/* The composition agent's system prompt, assembled from the private
+   sections below. */
```
7h3kk1d: Bad comment

src/haz3lcore/projectors/ProbeText.rei
```diff
@@ -0,0 +1,22 @@
+/* Plain-text rendering of a segment or zipper with probe values inlined, for
+   the CLI, the composition agent and the debug console. The walk itself is
+   private. */
```
7h3kk1d: Improve comment. "The walk itself is private" doesn't seem very relevant.

src/haz3lcore/TyDi/TyDiForms.rei
```diff
@@ -0,0 +1,8 @@
+/* Type-directed suggestion sets, keyed by what the cursor can accept. The
```
7h3kk1d: Stop mentioning what's private in rei files.

src/language/statics/CustomStatics.rei
```diff
@@ -0,0 +1,34 @@
...
+   `custom_statics_deferred_ap` ignores `~ancestors`, and its one caller
+   (Statics.re:1751) passes a unit-typed pun rather than the
+   `ancestors_inclusive` every other call in that file uses -- hence the
+   unconstrained type here. Worth reconciling; annotating it list(Id.t)
+   does not compile today. */
+
+let custom_statics_deferred_ap:
+  (
+    ~ctx: Ctx.t,
+    ~ancestors: 'a,
```
7h3kk1d: Can we put the actual type for ancestors?

src/language/statics/CustomStatics.rei
```diff
@@ -0,0 +1,34 @@
+/* Statics for custom (livelit-provided) application forms. The per-kind
```
7h3kk1d: What's livelit about these?

src/web/app/sidebar/DebugSidebar.rei
```diff
@@ -0,0 +1,5 @@
+/* The debug sidebar's only entry point: Sidebar renders this. Every field
+   builder and section driver below it is private. */
+
+let view:
+  (~globals: Globals.t, ~cursor: Cursor.cursor('a)) => Virtual_dom.Vdom.Node.t;
```
7h3kk1d: Is 'a a specific type or generic?

src/language/statics/CustomStatics.rei
7h3kk1d: Done, and it turned up why it had no type. `uexp_to_info_map` computes `ancestors_inclusive` and then does `let ancestors = ();` — a deliberate shadow so the exclusive list cannot be used by mistake — and this call site was passing that shadow. It compiled only because the parameter was untyped and unused, which is exactly the accident the shadow exists to catch. Now `list(Id.t)` in both the `.re` and the `.rei`, and the call passes `ancestors_inclusive` like every other call in the file.

src/language/statics/CustomStatics.rei
7h3kk1d: Nothing — the comment was wrong. All six `Ctx.custom_statics` tags are attached to builtins in `BuiltinsTupleOperations.re`; no livelit is involved. What they actually have in common is a result type that depends on the labels written at the call site (`project_labels(tup, "a", "b")` returns a tuple of just those labels), so they cannot be given an ordinary arrow type and get a bespoke rule here instead. Header says that now.

src/web/app/sidebar/DebugSidebar.rei
7h3kk1d: Genuinely generic, and unread. `Cursor.cursor('update)` is parametric in the host's action type — `editor_action`, `undo_action`, `redo_action` carry it, and `Cursor.map` exists to change it — but this panel only displays, so it never touches those fields. Renamed the variable to `'update` so the signature says which type it is rather than leaving a bare `'a`.

PR #2448, opened 2026-08-14
FastParse: weave nested triggers

### LINKED ISSUE #2447: FastParse: nested triggers bail the fast path
`^^probe(^^slider(50))` — a refractor pinned on a projector invoke — bails the fast parse with `token 5: segment has '50', source has '^^slider'` and falls back to the quadratic typing parser. Inside `consume_trigger`, `grab` weaves the wrapped pieces with `weave_piece`, which is trigger-blind, so any trigger nested inside another trigger's parens mismatches. Loading still works via the fallback, so this stays latent until a large paste/slide with a nested pin gets slow, or a committed .hz containing a nest hits the `Test_FastParseCorpus` zero-bail ratchet — which is how it surfaced (the strudelize branch's sound.hz).

disconcision:
`^^probe(^^slider(50))` bailed the fast path (`token 5: segment has '50', source has '^^slider'`): inside `consume_trigger`, `grab` wove the wrapped pieces with `weave_piece`, which is trigger-blind. This factors the two trigger arms of `weave_seg` into a shared `weave_trigger` that `grab` now tries first, so triggers nest. The regression test asserts the fast path itself handles the nest (no silent fallback), that the pin is collected and the wrapped projector materializes, and that the load path reprints the source verbatim.

Fixes #2447


disconcision:
@cyrus- minor fix... the new menhir fast parser path didn't handle nested projector invocations, now it does

PR #2449, opened 2026-08-15
Fix: menhir treats () as an atomic pattern (editor parity)

disconcision:
`fun () : T -> e`, `fun x :: () -> e`, and `fun (), x -> x` were menhir parse errors the editor parser accepts. The editor treats `()` as an atomic pattern; the menhir grammar listed UNIT at hand-picked parameter positions and every unlisted one was a gap. Final commit makes UNIT an atomic pattern (nonAscriptingPat) and deletes the positional special cases. 8 pins in Test_Menhir; MenhirParser/MenhirFuzz/FastParse/FastParseCorpus/MenhirCorpus green.

Pinning surfaced a pre-existing divergence parked as a skip case: on `fun x :: y : [T] ->` MakeTerm ascribes the last element, menhir the whole chain, needs its own decision.


src/menhirParser/Parser.mly
```diff
@@ -332,6 +333,9 @@ funAscElem:
       shift continues the inner exp, which is MakeTerm parity.
    3. One r/r between the two no-leading-plus sum productions (bare-ctor
       head vs general head) — identical semantic actions, either wins.
+   4. Two r/r states from funConsPat/funConsTail sharing nonAscriptingPat
+      and UNIT completions across head/tail contexts — identical
+      semantic actions, either wins.
```
7h3kk1d, line 338: Are we just planning on keeping these conflicts forever?

7h3kk1d (approved):

src/menhirParser/Parser.mly
disconcision: yes 😔

PR #2451, opened 2026-08-15
Fix regrout crash for use-in and minus operator

### LINKED ISSUE #2446: Stale grout after remold: prepending use-in to -5 gives 'in <grout> -5'; caret there crashes the view (Not_found)
Type `-5`, move the caret to the start, type `use Float in `. Result: `use Float in ⬡-5`, with a convex grout between `in` and `-5`. Clicking on the `-5` region — or moving the caret into it (e.g. cmd-right) — kills the page with "Exception during View: Not_found". Found by @Negabinary during review of #2419; reproduces on dev (6b611ba3be).

The final segment is `Grout(Convex)` directly followed by `-` molded *prefix* (convex left nib) — a malformed operand-operand junction. How it gets there: at the intermediate state `use Float i¦ -5` the `-` molds infix and regrouting inserts a convex grout as its left operand; when `in` completes the use form, `-` remolds back to prefix, but the grout survives — `Relatives.regrout`'s right-trim branch keeps a grout when `Grout.fits_shape(g, s_l)` holds, which only consults the left neighbor's shape (`in`'s concave nib), never the grout's right neighbor (now convex).

The crash is downstream: MakeTerm absorbs the `-` and `5` tiles into a hole without registering their ids in the terms map, and `Arms.term_range` does an unchecked `Id.Map.find(Piece.id(p), syntax.terms)` (Arms.re:370), so the view raises Not_found whenever the caret indicates either tile.

It needs `-`'s infix/prefix duality plus `use`'s intermediate states: `use Float in ` before `!true` is clean, and `let`/`if`/`fun` wrappers before `-5` are clean. Test-harness repro: `mk({|¦-5|}) @ string_to_ltr_actions("use Float in ")` ends with two segment tiles missing from `CachedSyntax.terms`.

Two fix loci: regrout should drop trim grout that doesn't fit both neighbors, and separately Arms.re could use `find_opt` so a piece missing from the terms map degrades gracefully instead of killing the view.

Matt's screenshot

disconcision:
closes #2446


disconcision:
@Negabinary tagging u as u discovered this

Negabinary (approved):

PR #2458, opened 2026-08-18
Split the B2T2 slides per operation, in line with the current benchmark

7h3kk1d:
> [!WARNING]
> **Needs a second opinion on one thing:** commit 1 switches `MenhirCorpus` from
> `Parser.to_zipper` to `Parser.to_segment`, and I am **not confident that is safe
> in general** — only that it holds across today's corpus. See
> "[The part I'm unsure about](#the-part-im-unsure-about)". The slide split
> (commit 2) is independent and can land without it.

## What

**Splits the B2T2 slides per operation, bringing them in line with the current
version of the benchmark.** The committed slides grouped several categories into
one file each — `Aggregate`, `Subtable`, `Properties`, `Ordering`,
`MissingValues`, `DataCleaning`, `AccessSubcomponents`, and the three Errors
"parts" — while the artifact's lab sources keep one file per operation. This
regenerates `hazel-programs/docs/b2t2` from those sources: **39 files → 62**, one
slide per operation, and picks up the upstream slide edits along with the split.

The menu therefore goes from 39 to 62 entries. That's the user-visible part of
this PR and the bit worth an opinion.

Names and menu order come from the manifest in `sync-slides.sh` (B2T2 operation
order, not alphabetical). The 28 slides that were already one-per-operation keep
their existing filenames, so the churn is limited to the 10 grouped files and the
33 replacing them.

## Why it also makes the suite much faster

The grouped files were the largest programs in the corpus — `table-api-aggregate.hz`
alone was 10.5 KB — and the editor parser is superlinear: `to_segment` only splits
at top-level boundaries, so within a segment cost still grows faster than linearly.
Re-splitting the same content (largest file now 6.7 KB, total bytes essentially
unchanged) cut `MarkerParse.of_text` over the corpus by **39%** and
`Parser.to_segment` by **32%**.

So this is faster while testing *more*: the suite gains 69 cases, because 62
slides generate more per-slide checks than 39 did.

| group | before | after |
|---|---:|---:|
| `MenhirCorpus` | 90.4 s | 21.2 s |
| `DocSlides.ReparseBackuptext` | 52.7 s | 11.7 s |
| `TextRoundtrip.DocSlides` | 103.4 s | 78.0 s |
| **full suite (local)** | **345.5 s** | **211.0 s** (−38.9%) |
| test cases | 3496 | 3565 |

Measured locally with a `junit_alcotest` patched to record real per-case times —
upstream hardcodes `~time:0.`, which is why none of this was visible from the
existing reports.

## Commit 1: the test-side parse changes

Two changes, independent of the slide split:

**`MenhirCorpus` parses via `to_segment` instead of `to_zipper`** (90.4 s → 37.2 s
on the old corpus). See the warning below.

**`MenhirCorpus` and `DocSlides.ReparseBackuptext` share the parse.** They turn out
to check the same programs — the corpus files *are* the slide sources — and both
need the typing parser's segment, which each was paying for separately.
`test/ParsedCorpus.re` memoizes it, taking `ReparseBackuptext` from 51.7 s to
11.5 s. Making the cache hit required `MenhirCorpus` to normalize the way the load
path does (dedent every line, strip one trailing newline) rather than
`String.trim`, since otherwise the two callers key different strings for the same
program. That drops the indented-form differential — narrow, and a shape
production never produces, since every slide is dedented on load.

Worth recording that the original hypothesis here was wrong: the two *slide*
groups looked redundant, but `DocSlides` spends its time in `MarkerParse.of_text`
and `ReparseBackuptext` in `Parser.to_segment`, and the one call they share costs
about a tenth of a second. The real duplication was with `MenhirCorpus`.

## The part I'm unsure about

`Parser.to_zipper` is the **definitional** editor parse — it replays what the
editor does per keystroke. `Parser.to_segment` is an *optimization* over it: it
splits the input at points it believes are safe, strips the convex grout artifact
`Zipper.init()` leaves behind, then does a final top-level regrout to reconcile
shapes across boundaries. Any of those three steps could plausibly differ from
to_zipper for a program shape not in today's corpus.

**What I did verify**, because it's the failure mode that would hide itself:
`check_file` silently passes when the editor parser returns `None`, so a file that
`to_segment` rejected would have dropped out of coverage with the suite still
green. That does not happen — across all corpus files both entry points accept
every file and produce structurally equal terms.

**What I did not verify** is that the two agree in general. The consequence is
that `MenhirCorpus` now pins menhir against the optimized parser rather than the
ground-truth one, so if `to_segment` ever drifts from `to_zipper`, this test would
compare against the wrong baseline and not notice.

The invariant "`to_segment` ≡ `to_zipper` corpus-wide" is worth asserting, but
doing so costs about what using `to_zipper` here would, which is the entire saving
— so it belongs in a periodic run rather than per-PR. That's noted in the file's
header. **If reviewers would rather not take this on faith, commit 1's swap can be
dropped and the shared-parse half kept; the slide split carries most of the win
regardless.**

## `known_gaps` fallout, which is an improvement

The single `table-api-properties.hz` gap entry was tolerating three operations at
once. Split, the `[()]` tokenizer quirk turns out to affect `header` and `ncols`
but **not** `nrows` — so the ledger now names the two files that actually carry it,
and `nrows` is genuinely checked for the first time. No new gaps.

## Verified

- `make dev` builds the web bundle against the regenerated `Slides.re`
- full suite green
- `MenhirCorpus` green with the re-attributed ledger

## Not addressed here

`TextRoundtrip.DocSlides` is now the largest remaining group at 78 s, all of it in
`MarkerParse.of_text` → `to_zipper`. Reimplementing that over `to_segment` is worth
roughly 40 s more, but it is production code (`PersistentZipper`'s fallback path),
needs a zipper rather than a segment, and also strips implicit holes and preserves
refractors — a real refactor rather than a substitution, so it is deliberately out
of scope.

Separately: the `unused open EditingPrelude` warning in `Test_ReparseDocSlides.re`
is untouched here. Its removal belongs to #2456, on #2422.


test/Test_MenhirCorpus.re
```diff
@@ -77,10 +89,13 @@ let equal_terms =
...
-  let txt = read_file(path) |> String.trim;
+  /* Normalize as the load path does, rather than String.trim: that is the
+     form the editor ever sees, and it lets ParsedCorpus share the parse with
+     DocSlides.ReparseBackuptext, which checks the same programs. */
+  let txt = read_file(path) |> ParsedCorpus.normalize;
   let mk =
-    switch (Parser.to_zipper(txt, ~root=Exp)) {
-    | Some(z) => Some(MakeTerm.from_zip_for_sem(z, ~root=Exp).term)
+    switch (ParsedCorpus.to_segment(~root=Exp, txt)) {
+    | Some(seg) => Some(MakeTerm.go(seg).term)
```
7h3kk1d, line 98: @disconcision this is the part I'm not sure if it's safe to do. It's a lot quicker and this shares it between both but if it's not safe we can either skip some b2t2 slides or the splitting of the slides I did might make it quick enough.

disconcision: I think that it is safe ish, in the sense that I intend to use the fast parser for this kind of thing going forward, and feel that although there are probably bugs, they'll come out in the wash as we use it in more places. this is probably a pretty low grade notion of safety though so ymmv

7h3kk1d: This is using the same parser I think it's just switching from Haz3lcore.Parser.to_zipper to Haz3lcore.Parser.to_segment

PR #2460, opened 2026-08-19
Fix: transitory grout entering fun -> due to unary minus pattern

disconcision:
Typing `fun x -` molds the `-` as unary minus (#2419's `UnaryMinusPat`), interposing junction grout: `fun x~-`. It should read as a pending `->`.

Fix: the existing `InfixDelimiterPrefix` backup-infix mechanism (which holds e.g. the `i` of a nascent `in` infix) was alphanumeric-only; this adds proper prefixes of non-leading *symbolic* delimiters (`-` from `->`, `=` from `=>`).


disconcision:
@cyrus-  fix to minor but annoying bug i introduced with negative float lits

PR #2465, opened 2026-08-21
Fix cross-term id collisions in ExpToSegment pad_ids

disconcision:
In a pretty-printed program with two multi-case `case` expressions, the k-th rule tile of each case ends up with the SAME id: several `pad_ids` call sites pass the `List.tl` of a term's ids without `~base`, so for freshly-parsed single-id terms the base falls back to `Id.invalid`, and the deterministic `Id.derive("pad<k>", invalid)` sequence repeats across terms. Anything grouping shards by tile id then misbehaves — observed as a `Highlight.of_tile: shard mismatch` crash (three 2-shard rule tiles under one id). Same latent collision for list-lit commas, deferred-ap commas, module semicolons, pat list-lits, and derivation rules.

Fix: pass `~base` = the term's head id at all seven sites, as pad_ids' own doc comment prescribes.

### DIFF
--- a/src/haz3lcore/pretty/ExpToSegment.re
+++ b/src/haz3lcore/pretty/ExpToSegment.re
@@ -1270,7 +1270,11 @@ let rec drv_exp_to_pretty =
   | Ctx([x, ...xs]) =>
     let* x = go(x, ~sort=Prop)
     and* xs = xs |> List.map(go(~sort=Prop)) |> all;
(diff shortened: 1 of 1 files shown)


disconcision:
Folded into completion-provenance (#2374) — the fix targets its deterministic pad_ids, which doesn't exist on dev.

PR #2469, opened 2026-08-27
Modular editors

disconcision:
## Goals

Two thrusts, deliberately on one branch because they feed each other:

1. **Outline mode + stacked subeditors** as a UI feature: navigate a large program by its structure (modules / definitions / types / tests as a collapsible outline) and edit any definition in a small focused cell while whole-program semantics (results, tests, probes) stay live.
2. **Editor performance**, aimed at concrete editing metrics — per-keystroke latency, held-key throughput, worst-case action spikes, slide switching, reload, and autosave cost — across several modes and use cases: flat editing of large programs, stacked/outline editing, and documentation-style slide decks.

The perf work is not incidental: the stacked UI is only viable if editing a 1k–4k-line program is responsive, so most of the pipeline (statics, parse, remold/regrout, view, persistence) was made incremental along the way.

## Try it

Open Documentation → Mega 1k (a ~1,000-line, 10-module program), click `running_sum` in the outline sidebar, and just that definition opens as an editor: typing costs 40–120ms per keystroke while the whole program's result, tests, and probes stay live below the stack. The same program is unusable to edit on dev. For raw-scale editing, the Documentation → "Perf /" folder has Mega 1k/2k/4k flat slides.

## Outline view

- **Outline sidebar**: modules/definitions/types as a collapsible, resizable tree; error badges attach to the deepest visible row and roll up onto collapsed ancestors; erroring names tint red.
- **Stacked per-definition editors**: clicking ⊙ opens a definition as a header cell (`name : type`, pat-rooted) plus a body cell, stacked in program order with module-path qualifier chips. Contexts are frozen at the definition and refreshed when upstream exports change, so cross-definition renames error their dependents live.
- **Whole-program semantics while stacked**: the master keeps evaluating the spliced program — result strip and test results below the stack, and probes sample even when every call site is outside the open cell (samples appear on placement, no follow-up edit needed).
- **Navigation**: jump-to-definition works across cells (opens the target and aligns it only if it isn't already visible); arrow keys walk the caret seamlessly across header/body/cell boundaries and between stacked panes.
- **Structural ops**: right-click an outline row for new definition / new type / new module below, duplicate, move up/down, delete — segment surgery on the live program, so unchanged definitions keep their ids (open cells and probes stay pinned). Ops cost the changed item plus its downstream users, not the program (mega-2k: ~3s → ~220–260ms steady-state; the term itself is parsed per item and grafted — `MakeTerm.Incr`, exact-parity test-gated). With a cell open at mega-2k, an idle 8s window has zero long tasks.
- **Tests, statements, and trailing bodies in the outline — in every block**: `test …;` statements get rows with live ✓/✗ icons at any depth (module-member tests included), other `…;` statements get `;` rows, and trailing bodies get `⇒` rows — each opens as a headerless cell.
- **Pins and collapse persist** per slide, across slide switches and reloads (name-anchored via outline label paths); collapse is modeled state, so folding no longer bleeds between slides or resets on structural edits.
- **Structural ops at every level**: the row context menu works on module members and function-body items too — inserts mint proper member forms inside modules — and moves stay within their block.
- **Headers got real**: cell statics are projected from the whole-program per-item analysis (one statics run, cells read it) — unannotated headers show their inferred types, module headers have live statics, unused-binder warnings reach headers.
- **Statics in every cell kind**: expression bodies, type-alias bodies, and pat/tpat headers all get cursor-inspector info and sort-correct coloring.
- **Persistence**: autosave is per-item serialized (`ItemPersist`): each top-level item's segment is stored as its own sexp under an ordered roster — the edited item's write is ~3–10ms — and reload restores the segment exactly, incomplete tiles included, with no text parse. A content-changing pause additionally rewrites the whole-document text blob (schema-drift fallback), which keeps an O(document) print on the autosave path (~11ms at 1k / ~27ms at 4k in the headless SerBench harness); making that blob per-item too is an open follow-up. Caret survives reloads via a side key.

## Performance

- **Compositional statics** (`DefStatics`): whole-program statics computed per top-level item with chained contexts; a typical edit re-analyzes the dirty set — the edited item plus export-affected dependents, e.g. 16 of 891 items (56ms headless) on a surgical 4k edit vs 872ms for the cold pass. Member-granular items inside module literals analyze per member. Fixes the browser stack-overflow crash that made mega-2k unloadable. Parity with monolithic statics is test-gated (errors, grafted-elaboration eval, probe capture). Per-document LRU slots make returning to a slide a warm lookup instead of a full cold re-analysis.
- **Sparse remold/regrout** (`Segment`/`Zipper`): remold short-circuits on same-mold tiles; regrout runs only on a caret window whose extent is derived from a read-only normal-form scan, memoized by tile identity — the remold/regrout phase drops from ~1.2s per 60 keystrokes at 4k lines to ~1ms/action. Parity-gated: the test runner executes both the sparse and the global pipeline on every call and asserts equivalence across the whole suite.
- **Read-free caret reveal** (`CaretReveal`): scroll-into-view no longer forces layout on every keystroke — the caret decoration publishes its model-side row geometry; during typing bursts the reveal is pure arithmetic against mirrored scroll state, with a same-frame rAF anchor read only on cold (post-pause) reveals and a throttled heal for drift. Held-key benchmarks at 4k: enter p95/max 550/936 → 133/448, letter 367/960 → 200/360.
- **Identity-preserving splices**: stacked-mode edits re-mint only the changed item's pieces, which is what makes the per-item dirty tracking (statics, persistence, vdom) actually fire; plus cached stacked parse and outline vdom memo for the stacked-mode edit spikes.
- **Incremental view**: chunked measurement and per-chunk content-keyed vdom — 4k flat edit ~1.6s → ~100ms attributable; 2k edit 690 → ~53ms.
- **Text round-trip for operator holes**: the printer now spells concave grout ⧖ (convex stays ¿), so documents containing infix holes survive text persistence without hitting the quadratic recovering parser (previously a multi-minute reload at ~27k chars). Lexer/parser support landed separately on dev (#2479); this branch carries the integration.
- **Worker-resident incremental eval cache / streaming overhaul / memory fixes**: unchanged from the earlier description — request payloads no longer ship the previous cache; incremental stream collector (parity-gated); exponential co_ctx leak fixed; undo snapshots compacted; render-generation closure chain broken. Mega-2k slide open: ~47s main-thread busy → ~3s.

## Review notes

The design docs live in a gitignored `plans/` tree, so this description is the context. Recent, least-reviewed range: `9c24363bcc..8e0c9d7dbc` (sparse regrout, caret reveal, per-item persistence, ⧖ integration); older subsystems have had more soak time.

Codex review findings 1–3 are addressed in `29b726d3c8` (occurrence-qualified sexp-encoded outline paths, content-key-tagged pending restore state, CaretReveal single listener + caret-identity anchoring), with gates in `Test_OutlinePaths`.

Where extra scrutiny would help most:
- `src/haz3lcore/tiles/Segment.re` — the normal-form scan (`stale_in_seg`/`stale_affix_ids`) and its identity memo (global id-keyed table, pointer-validated, 200k cap): memo soundness and multi-editor sharing.
- `src/haz3lcore/zipper/Zipper.re` — `split_window`/`window_bound`: the caret-window bounds for sparse regrout (the window bound is deliberately the boundary solid's shape *complement*; threading the true shape double-decides out-of-scope runs).
- `src/haz3lcore/zipper/ItemPersist.re` + `src/web/view/ScratchPersist.re` — write ordering (items → roster → GC), load fallback on any inconsistency, and the root-repair path.
- `src/util/CaretReveal.re` — mirrored scroll geometry and the WAAPI-vs-CSS animation filtering (the caret's CSS blink must not pin reveals cold).
- `src/haz3lcore/derived/DefStatics.re` — `head_equal` uses polymorphic compare on term heads; derived stable ids for mod-item wrappers; LRU keying by whole-term rep id.

Deliberate decisions (please don't re-litigate without a failure case): the whole-text autosave blob is kept as schema-drift fallback alongside the per-item sexps; parity gates for sparse regrout stay enabled in the test runner; DOM virtualization is planned but explicitly not built (the remaining 4k spikiness is GC + per-key view floor, semantics-independent).

Known follow-ups: stepper while stacked, projector sidebar apps while stacked, worker-resident program segments (stacked PR: w2-worker-residency), hierarchical (sub-item) chunking and persistence for huge single items, episodic eval frame at 4k.

---
### Fix brought over from the aria-demo branch (2026-09-09)
- 46bdb6b16e Spine roots carry the whole-suffix elab_term. The incremental evaluator gates reuse on a node's elab being unchanged, and a spine root's hollow elab read as unchanged whatever happened below it, so an edit to any definition but the first was never re-evaluated. New Test_ModRoot case fails without the fix.


disconcision:
## Codex review — correctness, performance, and structure

I reviewed this at `8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0`. Overall, the incremental core has a better safety story than I expected: the sparse normalization path is parity-run against the global path throughout the editing suite, the item persistence ordering/fallback tests are substantive, and identity-preserving splices plus compositional statics are coherent optimizations. The branch builds cleanly and the complete local `@runtest` suite passes.

I did find several correctness issues in the newer UI/persistence plumbing, plus one performance claim that needs narrowing.

### Correctness findings

1. **[P1] Persisted outline paths are not unique, and some valid labels cannot round-trip through the ad-hoc encoding.** [`label_path`/`resolve_path`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/app/sidebar/OutlineTree.re#L340-L376) use only display labels and `resolve_path` selects the first match. Two separated test runs both create a container labeled `tests`; a saved pin under the second group resolves into the first group and then fails. Duplicate definitions have the same problem. Collapse is worse: [`my_path = path @ [n.o_label]`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/app/sidebar/OutlineSidebar.re#L372-L380) means collapsing one top-level `tests` group collapses every top-level `tests` group. In addition, [`write_pins`/`read_pins`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchPersist.re#L69-L122) use spaces, `/`, and newlines as unescaped delimiters; a backticked definition name containing a space is silently discarded on reload. Use a derived sexp/JSON value and a durable component such as `{kind, label, same-label occurrence}` (or a source/item anchor that survives text reload). Please test duplicate names, two disjoint test groups, and a backticked name with spaces.

2. **[P1] Caret/pin/collapse restoration state is module-global and not tagged with the document it belongs to.** [`pending_caret` and `pending_pins`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchPersist.re#L65-L88) are singleton refs. `read_caret` leaves the previous value intact for a missing or malformed key ([lines 124–143](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchPersist.re#L124-L143)), while `RestoreCaret` applies the point to whichever slide is current when the queued action executes ([ScratchMode lines 773–803](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchMode.re#L773-L803)). A hydration/mode-switch race can therefore move the wrong editor. Collapse and transient pins are keyed only by slide name, not `(prefix, name)`, and [`read_collapse(None)` does not clear the old same-name entry](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchPersist.re#L90-L103), so Scratch and Documentation slides with the same name bleed state. Put this state in the slide/workspace model, or at minimum key/tag every value and restore action with a document key and verify it against the current document before applying. Clear pending state before every read.

3. **[P1] `CaretReveal` leaks one scroll handler per cold reveal and can reuse another editor's geometry during a burst.** Every cold anchor calls [`attach_scroll_listener`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/util/CaretReveal.re#L92-L104), but no old handler is removed. The browser retains every old `geom` closure and invokes all of them on each scroll. Separately, `published` contains only `(row, row_height)`, and the hot path accepts any connected cached container ([lines 269–289](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/util/CaretReveal.re#L269-L289)). Moving from one stacked cell/header to another within 500 ms can combine the new row with the old editor's `editor_top`, causing a wrong scroll until verification heals it. Maintain one removable listener per active container and include an editor/cell identity in the published geometry; identity change should force a cold anchor. This module currently has no test coverage, so a small DOM-level listener-count and cross-cell reanchor test would pay for itself.

### Performance findings and measurements

4. **[P2] Per-item persistence pays off, but the PR description overstates the end-to-end autosave result.** [`save_current`](https://github.com/hazelgrove/hazel/blob/8e0c9d7dbc7323e5edee6ea00c0f91c5db7f4ff0/src/web/view/ScratchPersist.re#L288-L389) writes changed items and then serializes/writes the whole text-backed slide on every content-changing pause. On this machine, the branch's `SerBench` harness measured text printing at **11 ms for Mega 1k** and **27 ms for Mega 4k**; by comparison, whole-segment sexp+string at 4k was **335/351/353 ms**, so the move away from whole-zipper sexp is clearly worthwhile. But autosave still has an O(document) ~27 ms main-thread floor in addition to the claimed 3–10 ms item write. I would keep the item store primary and move the schema-insurance text blob to an idle/rate-limited policy (plus a page-hide flush), or serialize it in a worker. At minimum, revise “writes only the edited item” in the PR description.

5. Current headless statics figures also support the broad direction but show why the exact workload matters: `LabelBench` measured Mega 4k cold `DefStatics.calc` at **872 ms / 891 items**, then a surgical one-item edit at **56 ms / 16 items analyzed**; the full incremental `init_compositional_term` path was **58 ms**. That is a large improvement over a cold pass, but it is not literally one analyzed item for this dependency shape. These are Node/headless numbers, not browser interaction latency; the browser keystroke/view/GC claims still need a reproducible UI harness.

### Structure recommendation

The split from the old ~4.7k-line file is useful, but `ScratchMode.re` is still 2,752 lines and the feature now depends on shared globals in `ScratchPersist`, `ScratchMode`, `CaretReveal`, `CodeWithStatics`, and `DefStatics`. The state-leak findings above are symptoms of an unclear ownership boundary.

I would make the core abstraction a **multi-slide editor workspace**, not “scratch mode”:

- `SlideDeck`: document identity, names/current slide, hydration, persistence namespace, dormant state.
- `EditorWorkspace`: master editor + optional stack, focus/navigation, splicing/restructure, projected statics.
- `OutlineController`: outline tree plus modeled pins/collapse/menu state, keyed by document identity.
- Thin Scratch and Documentation adapters: defaults, labels, persistence prefix, and whether derivation slides are offered.

Outline/stack is then an editor capability consumed by both modes; Scratch and Documentation stop being the place where core editor lifetime is defined. I would also move the remaining module-global mutable refs into explicit non-historic workspace/controller state or lifecycle-owned services. That is a larger cleanup, but it directly prevents the prefix/slide/race classes above and makes future Project/editor reuse much less risky.

### Suggested merge posture

I do not see a reason to discard the main incremental work: sparse regrout parity, compositional statics, item storage, chunked measurement/view, and identity-preserving splices are individually defensible. I would fix findings 1–3 before merging, narrow or change the autosave claim/policy, and keep #2480's consumer flip separate until its protocol recovery issues are resolved.

cyrus-:
If I click z here it crashes with `Failure("Skel.push_output: split_kids: index out of bounds")`

image

PR #2471, opened 2026-08-28
Replace hand-rolled util helpers with their Core equivalents

7h3kk1d:
Now that `src/util` is Core-opened, replaces hand-rolled helpers with the Core functions that already do the same job. Still `src/util`-only.

| hand-rolled | Core |
| --- | --- |
| `ListUtil.take` / `truncate` | `List.take` — *these two were the same function* |
| `ListUtil.remove_first_n` | `List.drop` |
| `ListUtil.join` / `intersperse` | `List.intersperse` — *byte-identical to each other* |
| `ListUtil.hd_opt` / `last_opt` | `List.hd` / `List.last` |
| `ListUtil.count_pred` / `findi_opt` | `List.count` / `List.findi` |
| `OptUtil.get` / `zip` / `sequence` | `Option.value_or_thunk` / `Option.both` / `Option.all` |
| `StringUtil.split_first` | `String.lsplit2` |

Two stale TODOs die here: `take` was waiting on "ocaml 5.3" and `split_first` on "5.5". Core had both already.

**Deliberately not swapped**, because the semantics differ:
- `group_consecutive` vs `List.group` — ours compares each item against the group's *first* element, `List.group` compares *adjacent* ones. Only equivalent when the predicate is a true equivalence.
- `dedup` / `dedup_f` vs `List.dedup_and_sort` — Core's sorts, ours preserves order.
- `ListUtil.slice` vs `List.sub` — `sub` raises out of range, `slice` clamps.

## Verification
`make dev` clean · `make test-quick` green · release-clean.


src/util/ListUtil.re
```diff
@@ -372,57 +349,19 @@ let first_and_last = (xss: list(list('a))): list(('a, 'a)) =>
...
-    | _ => acc
-    };
-  loop(List.rev(a), List.rev(b), []);
-};
+let intersperse = (sep, xs) => List.intersperse(xs, ~sep);
 
-let common_suffix_length = (~eq: ('a, 'a) => bool=Poly.equal, s1, s2) =>
-  List.length(max_common_suffix(~eq, s1, s2));
+let flat_intersperse = (sep, xss) =>
+  List.concat(List.intersperse(xss, ~sep=[sep]));
```
7h3kk1d, line 361: This seems slower with allocations. Look at current use sites and evaluate whether or not it matters and if it does use Seq.

src/util/OptUtil.re
```diff
@@ -1,40 +1,21 @@
...
-    ~f=(x, acc) => map2((y, ys) => [y, ...ys], f(x), acc),
-    ~init=Some([]),
-    l,
-  );
+let sequence = (l: list(option('a))): option(list('a)) => Option.all(l);
 
-let sequence = (l: list(option('a))): option(list('a)) =>
-  traverse(Fun.id, l);
+let traverse = (f: 'a => option('b), l: list('a)): option(list('b)) =>
+  sequence(List.map(l, ~f));
```
7h3kk1d, line 18: consider the allocation consequences.

7h3kk1d:
If we're not doing it here we should make sure we inline the functions that are direct replacements instead of keeping them around forever.

PR #2476, opened 2026-08-28
Migrate test/ to Jane Street Core

7h3kk1d:
Migrates `test/` to Jane Street Core, completing the migration. Typed equality first, then the flip, then a follow-up for tests dev added after it.

**The failure the type checker cannot catch.** A name-based sweep prefixing deprecated primitives with `Stdlib.` rewrote them **inside string literals** — which in this repo are Hazel program text and expected values. `Test_Atom` asserts on a builtin named `"string_of_int"`; `Test_TableTransforms` embeds Hazel source. The result is a green build and a *failing test*. Any future sweep over `test/` must exclude string literals, handle `{|...|}` quoting, and be checked by diffing the per-file `Stdlib.` count against the base.

**Verification caveat worth flagging to reviewers:** `make test-quick` skips the QCheck property suites. They compile and load but do not execute, so a behavioural break there survives a green quick run. A full `make test` is warranted before this lands.

`test/` is not release-clean, and was not before this PR — `Test_ReparseDocSlides` has an unused `open EditingPrelude` present unchanged on `dev`. Left alone deliberately.

## Verification
`make dev` clean · `make test-quick` green · full `make test` still recommended before merge.


test/evaluator/Test_Evaluator_ProbeCallStack.re
```diff
@@ -118,7 +119,7 @@ in ^^probe(celsius(72.)); ^^probe(celsius(100.))|},
       | _ =>
         fail(
           "Expected exactly 2 samples, got "
-          ++ string_of_int(List.length(samples)),
+          ++ Stdlib.string_of_int(List.length(samples)),
```
7h3kk1d, line 122: Should we be using Int.to_string

PR #2477, opened 2026-08-28
Use Core's list and option combinators where we hand-rolled them

7h3kk1d:
With every library on Core, replaces the remaining hand-rolled list and option combinators with Core's — the cross-cutting pass that the earlier per-library PRs deliberately left alone.

Touches `src/web` (9), `test/` (7), `src/util` (6), `src/haz3lcore` (6) and `src/language` (4). It could not have been done earlier: the call sites live in libraries that were not Core-opened until the PRs below this one, so referring to Core would have meant verbose `Core.…` prefixes everywhere.

**The test count drops.** That is intentional — `Test_ListUtil.re` loses the cases covering hand-rolled helpers that this PR deletes in favour of their Core equivalents.

## Verification
`make dev` clean · `make test-quick` green (see above) · `dune build --profile release src` warning-free.


src/haz3lcore/projectors/implementations/CardProj.re
```diff
@@ -247,11 +247,7 @@ module SyntaxTerm = {
       info.syntax,
     );
 
-  let get_opt = (any: Any.t): option(state) =>
-    switch (any |> any_to_syntax) {
-    | Some(syntax) => Some(syntax)
-    | None => None
-    };
```
7h3kk1d: incredible code

PR #2480, opened 2026-08-31
W2: worker-resident program segments (stacked on modular-editors)

disconcision:
## Goals

Stacked on `modular-editors` (#2469) — review the diff against that base, not dev. This is the experimental next stage of the same perf program: move semantic computation (statics, elaboration, eval bookkeeping) off the UI thread entirely, so the main thread retains only what keystroke latency needs. Everything here is flag-gated and off by default — including shadow mode itself, as of the review fixes (it duplicates statics derivation and worker messaging, so it is a development diagnostic: `__w2shadowToggle(true)` enables it, `__w2flip(true)` implies it). The branch exists to de-risk the architecture with parity evidence before any flip.

## What's here

- **Segment-residency protocol**: the worker holds its own copy of the program segment and applies the same edit deltas the main thread does, so semantic requests ship deltas rather than documents. A headless loopback harness parity-gates worker-side state against the main thread's.
- **Shadow mode**: the app runs the worker pipeline alongside the main-thread one and compares summaries, counting mismatches (`SUMMARY MISMATCH` console reports). This is diagnostic scaffolding, deliberately noisy.
- **Budgeted synchronous statics**: a small time-budgeted synchronous statics pass for the caret's item only, so type-directed feedback at the caret stays immediate even when full statics moves to the worker.
- **Worker summary payload + consumer flip**: the view-side consumers can read statics summaries from the worker instead of the main thread, behind a runtime toggle (`__w2flip`).
- **Instrumentation**: `__staticsProf` per-phase statics-tick diagnostics (gated, zero-cost off); `Test_SerBench` serialization-cost harness (temporary, for the persistence sizing decisions).

## Status / caveats for review

- Codex review findings 1–6 are addressed in `d59762623c`: Resident evals are generation-gated (a rejected delta can no longer evaluate the old program under the new generation's label), recovery is immediate and restart-proof (cached authoritative Full resync on NeedResync and on worker restart, before any retried eval posts), flip toggles bust every DefStatics cache (`reset_caches`), summary grafts update the keyed cache entry and schedule a recalculate (durable + observable), and shadow mode defaults off. Gates in `Test_W2Protocol`.
- Shadow mode (when enabled) reports summary mismatches on some documents; under investigation — treat the comparison logic itself as review-worthy (false positives are as plausible as real divergence).
- The residency protocol's delta-application path remains the highest-value review target: it must reproduce main-thread edit semantics exactly (the loopback parity gate covers what it covers, no more).

---
2026-09-09: merged modular-editors forward for the spine elab fix (46bdb6b16e).


disconcision:
## Codex review — correctness, performance, and structure

I reviewed this at `53523318354af9c8678ed20b1e3d4494a6d53be5`, as a delta from `modular-editors`. My recommendation is **do not fold the W2 consumer flip into #2469 yet**. The resident-program core and parity harness are good groundwork, but the recovery/consumption path has several correctness holes that can make a stale worker generation appear current.

### Blocking correctness findings

1. **[P0] `Resident.generation` is ignored, so an eval can run a stale program after a rejected delta.** [`resolve_payload`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/WorkerServer.re#L411-L435) matches only the key and discards the requested generation. [`handle_sync`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/WorkerServer.re#L856-L882) deliberately retains the old resident on `RosterMismatch`, `UnknownItem`, or a missing resident. Because Sync(g) and Evaluate(Resident(g)) are FIFO, the eval immediately following a rejected Sync(g) resolves against the previous resident and returns its result as generation g. The generation needs to be part of the resident slot match (`rp.generation == requested_generation`), with evaluation gated until recovery succeeds. Please add the production-shaped test: Full(g1), bad Items(g2) -> NeedResync, Resident(g2) must not evaluate g1.

2. **[P0] Worker restart drops all residency, then the ACK retry reposts only the resident eval.** [`restart_worker`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/WorkerClient.re#L152-L158) creates a blank worker; the retry path then calls only [`post_evaluate`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/WorkerClient.re#L182-L202). A flipped `Resident` request therefore fails with “no resident program”. Worse, `ShadowResidency.mirror` and `last_ds` survive the restart, so an unchanged document need not emit another sync. The residency owner should cache the latest authoritative Full payload, invalidate its mirror on restart, Full-sync the new worker, and only then retry the eval. This needs an ACK-timeout/restart test with a resident request.

3. **[P1] `NeedResync` does not actually resync until a later semantic edit.** [`on_summary`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/ShadowResidency.re#L151-L159) only sets `mirror := None`. If the user stops typing, the worker stays stale indefinitely; the already-queued Resident eval is also unsafe because of finding 1. Cache the latest authoritative root/settings/segment/probes and immediately send Full on a current-generation `NeedResync`; gate eval for that generation until `SyncOk`.

4. **[P1] Disabling the flip does not clear the cache `calc_auto` actually reads.** [`set_flip`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/ShadowResidency.re#L65-L71) clears `DefStatics.slot`, but [`calc_auto`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/haz3lcore/derived/DefStatics.re#L1362-L1385) obtains its previous value from the separate `slots` table. A clamped chain can therefore be reused after toggle-off even though the comments and `Test_PropagateClamp` say it cannot be caught up incrementally. Add a public reset API that clears/evicts the relevant per-document slot and its MRU entry, and test the real `calc_auto` toggle-on/edit/toggle-off path.

5. **[P1] A worker summary is neither made observable to the view nor durable in the actual statics cache.** [`graft_summary`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/ShadowResidency.re#L110-L149) mutates only the module-global `DefStatics.slot`. It dispatches no model action, while [`CodeWithStatics.calculate`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/app/editors/code/CodeWithStatics.re#L178-L223) retains the existing `CachedStatics` on an ordinary post-result calculation. The per-document `DefStatics.slots` table is not updated either, so a later `calc_auto` can overwrite the graft from its ungrafted entry. In effect the promised “next calculate pass” does not rebuild the displayed error/warning IDs. Prefer carrying a generation-tagged resident summary in modeled editor state and dispatching an explicit `ApplyResidentSummary`; if the global cache remains, update the keyed entry atomically too. Add a test that a cross-item error arriving only in a summary changes the rendered/model error IDs without another edit.

6. **[P2] Grafting replaces the full ID lists with only piece-anchored IDs.** The summary retains synthetic errors/warnings only as counts, but lines 135–139 replace `d_error_ids`/`d_warning_ids` outright. That contradicts the comment that stale items keep their synthetic IDs and can change aggregate problem counts/other consumers. Either preserve the old synthetic subset during the graft or make the worker payload sufficient for every flipped consumer.

### Default behavior and architecture

- The PR description says the experiment is off by default, but [`ShadowResidency.enabled`](https://github.com/hazelgrove/hazel/blob/53523318354af9c8678ed20b1e3d4494a6d53be5/src/web/util/ShadowResidency.re#L17) is `true`. Thus every ordinary statics update duplicates derivation and worker messaging even with `__w2flip(false)`. Before folding anything into #2469, default shadow mode off or make it an explicit development/benchmark flag and measure its steady-state cost.
- The recurring failure mode is that ownership is split across `ScratchMode` (sync ordering), `EvalResult` (generation selection), `WorkerClient` (restart), `ShadowResidency` (mirror/recovery), and `DefStatics` (two caches). I would introduce one `ResidentWorker` state machine owning `{worker_epoch, key, synced_generation, pending_full, recovery_state}`. It should expose “sync authoritative program” and “evaluate generation” operations; no caller should be able to post a Resident eval unless that state machine has acknowledged the exact generation. That would eliminate several comments-as-invariants and make restart/resync tests headless.

### What looks solid

The `ResidentProgram` full/delta engine, roster verification, summary parity representation, and loopback tests are a sensible base. I would be comfortable merging that core and its tests independently, but not the consumer flip/recovery wiring above. The missing test matrix is specifically: generation mismatch, rejected-delta eval, worker restart, immediate resync, real `calc_auto` toggle-off, and summary-to-view application.

disconcision:
Validation update: `dune build @src/fmt src --profile dev` passed, and the complete local `@runtest` run passed **3,861 tests in 843.536s** at `5352331`. The findings above are gaps in scenarios the current suite does not exercise, not existing test failures.

PR #2481, opened 2026-09-01
Colors: one semantic vocabulary, from the config slide to the stylesheets

7h3kk1d:
Stacked on #1591 (`configuration`).

The Colors slide's type used to *be* the CSS variable list with Hazel syntax on
it: `expected_type` was generated from lists of raw custom-property names, and
the header said so outright — "The label IS the CSS custom-property name,
verbatim." So a themer edited CSS internals: `BR2`, `T3`, `shard-caret-tpat`,
`GB1`, and a misspelling (`backback-targets`) that was load-bearing because it
was a CSS name.

This gives the theme an explicit, semantic representation and maps it onto CSS,
then collapses the three stacked CSS vocabularies into one.

## What changed

**The shape of a theme is a type.** `BuiltinsColorScheme.re` holds it as Hazel
types in the builtin context — `ColorSeeds`, `ColorPalette`, `ColorRoles`,
`ColorScheme` — and the slide writes them in its own signatures:
`let theme_of(p: ColorPalette, role: ColorOverrides): ColorScheme`.
`ColorRamp`/`ColorOverrides` are declared in the slide, being seams between its
own two functions. That puts each check on the record a mistake is in; without
it the only analysis is the whole-program one, which lands on the final `case`
as a single inconsistency between two enormous products.

**The slide names roles**, grouped by the part of the UI they dress (`menu`,
`editor`, `cursor`, `hole`, `probe`, …), with a fan-out table in
`ColorConfiguration.re` expanding each field to the CSS properties that carry
it — one field per *decision*, several targets where several properties carry
one decision. ~~`expected_type` and
`all_targets` both derive from that table, so they cannot drift apart.~~
`all_targets` derives from that table; `expected_type` is the named
`ColorScheme`, and a test pins that the slide fills it exactly.

The two layers differ by **reuse, not derivation**: a palette color is one the
theme is built out of, fanned out to several properties at once; a role is one
job a color is put to, usually carrying one. Roles that only forwarded a
stated color are palette entries now, renamed for what they are (`exp-base` →
`probe-value`, `main-base` → `statics-background`).

**~~The palette took the slide's names: `--Y1` → `--attention-2`, `--T1` →
`--surface-1`, `--SAND` → `--code-background`.~~** The rename landed, and then
the names went: nothing read `--ink`, `--frame-1` or the rest — no stylesheet,
no OCaml, no script, and the only references left in the tree were commented
out. A palette color now reaches CSS only under the semantic names the fan-out
gives it, which makes "components consume roles, not the palette" structural
rather than lint-enforced.

**`roles.css` is deleted.** It was a third vocabulary: hand-written aliases
carrying nearly half the component `var()` sites, and its names resolved to far
fewer palette colors than there were names, so `--text-faint`,
`--border-soft` and `--shadow-soft` were one color with no way to separate
them. Every entry was a bare alias, so they fold into the fan-out table and the
theme writes them itself. ~~**No component stylesheet changed** — every site
reads the same names as before.~~ None changed for *that*; elsewhere on the
branch some sites were renamed off the raw palette names.

**`variables.css` has no hand-written color left**: type, timing and the
z-index ladder. ~~Type, a generated block (`make update-css-defaults`, guarded
by a test), and the z-index ladder.~~ The generated defaults are their own
labelled file, `style/theme-generated.css`, byte-identical to what the theme
sets so nothing shifts when it lands.

**The gap is closed.** The colors the theme did not own were dead, pure
indirection, or derived by CSS itself (the `doc` hue rotations, the
`incremental-pending` fades); they are decided in the slide now.

**Two flags, and they compose.** `--hazel-color-scheme` feeds the standard
`color-scheme` property; `--hazel-contrast` is new, so a stylesheet can answer
high contrast with a style query. Four schemes from two booleans, one `case`.

**Also fixed:** `color-scheme` was set nowhere, so native scrollbars rendered
light on a dark editor; the two flags overrode each other instead of composing;
`proj-fold.css` had an inert `color: --var(BLACK)`; and table rows and headers
were near-illegible in high contrast, because a row highlight was placed on the
page's contrast axis when table rows sit on the code background.

**The slide reads better.** It opens with every scheme but the first folded;
tuple extension instead of restating records, direct function definitions, and
a smaller `ColorOverrides` — several of its fields turned out to be one
decision spelled a few ways. `style/README.md` is new, and documents the
dataflow, file ownership and both layers.

## How it is checked

- **Golden** (`test/goldens/colors.tsv`): every property × every scheme.
  Written sorted by name, so a pure restructuring leaves it and
  `theme-generated.css` byte-identical — which is how every refactor after the
  vocabulary work was verified.
- **Reverse substitution** for renames: mapping the new names back reproduces
  every touched file byte for byte.
- **Chrome**, at each stage: every variable resolves to the golden in all four
  schemes, and a stylesheet-only page with no JavaScript resolves them all to
  the light column.
- **`./hazel analyze hazel-programs/config/colors.hz`** names the offending
  term when an annotation stops holding.
- `lint_css_roles.py` rewritten for the two-layer shape: components read roles
  not the palette, only the generated file declares a theme-owned color on
  `:root`, and the dangling-reference ratchet stays.

## Known gaps

- Folding kept "one field per decision", so `--text-faint` and `--border-soft`
  still come off one `palette.frame-1` row. Splitting a case is a two-file
  change; a field per property was rejected as worse to read.
- Two further override merges would each move a color by one ramp step, so
  they want an eye rather than a green test: collapsing the `frame-*` dials,
  and merging `projector.textarea-text` into `inspector.text`.
- **TODO:** we still need some way of enforcing that newly added colors stay
  themeable. The lint catches a component declaring a theme-owned color on
  `:root`, but not a new hardcoded literal in a rule.


Makefile
```diff
@@ -95,6 +95,12 @@ coverage:
 lint-css:
 	python3 scripts/lint_css_roles.py
 
+# Rewrites the generated block of variables.css from the Colors slide's light
+# scheme, so a default is byte-identical to the value the theme will set. Read
+# the diff: an unexplained line in it is the bug.
+update-css-defaults:
+	UPDATE_CSS_DEFAULTS=1 ./run_tests test 'ColorConfiguration' -q
```
7h3kk1d, line 102: Think more about if this is how we want golden tests to look.

src/web/www/style/variables.css
```diff
@@ -1,251 +1,302 @@
...
-  --G0: oklch(70% 0.15 150); /* page title, passing tests */
-  --G1: oklch(85% 0.15 150); /* ! passing tests hover */
-  --G2: oklch(80% 0.05 150); /* ! comments */
-  --GB0: oklch(70% 0.05 120); /* ! nut menu active */
-  --GB1: oklch(45% 0.05 120); /* ! nut menu fill */
-
-  /* FREE PARKING - ??? */
-
-  --textarea-v-stripe: oklch(78% 0.14 6 / 55%);
+  /* BEGIN GENERATED DEFAULTS -- do not edit; run `make update-css-defaults` */
```
7h3kk1d: Should we put this generated stuff in a separate file so it's easier to deal with?

Issue #2457, opened 2026-08-18
Incremental eval leaves a binding unsubstituted when evaluation gets stuck after an edited literal


7h3kk1d:
## Summary

The QCheck property **"Incremental eval agrees with fresh eval after a literal edit"** (`test/evaluator/Test_Evaluator_Properties.re:307`, `~count=2000`) fails for some seeds. It presents as a flake, but it isn't: given a program of the right shape it fails **every time**. Only whether QCheck generates that shape is random.

Minimal repro, no seed needed:

```hazel
{ let x = (); let _ = 4; let true = A }
```

Edit the `4` to `5` (which is what the property does — a type-preserving `+1` on an int literal), then evaluate the edited program two ways:

| | result |
|---|---|
| fresh (no `prev`) | `let true = A in (x=(()))` |
| incremental (`prev` = original's `incr_eval`) | `let true = A in (x=x)` |

Both get stuck on the failing pattern match `let true = A`, as expected. The disagreement is in the residual environment: **fresh has substituted `x`'s value, incremental has left `x` as a variable.** `Equality.semantic.exp` then reports them unequal, which is what fails the property.

## Required ingredients

Removing or reordering any of these makes it pass, which should localize it:

- **A named binding before the literal.** `let x = ()` fails; `let _ = ()` passes — the name has to survive into the residual.
- **The literal bound to `_`.** `let _ = 4` fails; `let y = 4` passes.
- **Evaluation must get stuck *after* the literal.** `{ let true = A; let _ = 4 }` (stuck first) passes.
- **Stuck via a pattern-match failure, not an unbound variable.** `let true = q` passes; `let true = A`, `let true = ()`, `let false = A`, `let 7 = A` all fail.

Irrelevant: the literal's value (`0` and `4` both fail), and the number of preceding bindings — `{ let x = (); let z = (); let _ = 4; let true = A }` fails with `(x=x, z=z)` against `(x=(()), z=(()))`, so every named binding is affected, not just the first.

## Seeds

`QCHECK_SEED=3` and `QCHECK_SEED=468045375` both reproduce via the property itself; `QCHECK_SEED=1` and `2` pass. Seed 3 shrinks to

```
(0, { let x = (); let _ = 4; let true = A; type y = y; type ? = Bool; type x = Void })
```

and the three trailing type aliases are incidental — the minimal case above is what's left after removing them.

## Why this matters now

[...]

PR #2482, opened 2026-09-01
test: disable two seed-flaky evaluator properties, pin them as expected failures

7h3kk1d:
Two `Evaluator.Properties` QCheck properties fail occasionally.

This disables both and pins each bug deterministically instead. **It fixes neither bug** — #2128, #2372 and #2457 stay open.

| property | issue | reddened |
|---|---|---|
| `Evaluator and stepper are consistent` (case 5) | #2128, dup #2372 | a random local seed while preparing this PR |
| `Incremental eval agrees with fresh eval after a literal edit` (case 7) | #2457 | [33513972427](https://github.com/hazelgrove/hazel/actions/runs/33513972427), [33513747011](https://github.com/hazelgrove/hazel/actions/runs/33513747011), [33412342909](https://github.com/hazelgrove/hazel/actions/runs/33412342909) — three unrelated branches, each dying on case 7 alone |

Disabled the way `Preservation of types (disabled)` already is here: a `test_case` that skips but still `ignore`s the QCheck test, so it keeps compiling and re-enabling is one line.

### The pins

Each disabled property is replaced by a unit test on its issue's counterexample:

- **#2128/#2372** — `fun (x::x) -> x`: one engine freshens the duplicate binder, the other doesn't, and `Equality.semantic.exp` doesn't equate the alpha-variants.
- **#2457** — `{ let x = (); let _ = 4; let true = A }` with the literal bumped: fresh eval substitutes `x` into the residual environment, incremental leaves it a variable.

Both assert the **current, wrong** behaviour, so they go red the day the bug is fixed and name the property to re-enable. Hence not `skip_known_bug` (`test/Test_Elaboration.re:456`, `test/statics/Test_Statics_Prelude.re:269`) — a skipped test can never fail, so it stays silent forever.

So pin and property can't drift, the shared work is extracted and called by both: `reduce_both_engines` and `eval_after_literal_edit`, each returning `None` for vacuous cases (step limit, known-benign raise).

### Notes

**#2408** fixes a different confluence cause — uncollapsed casts in indeterminate `if` branches, not binder freshening — so the #2128 pin should still be red after it lands. Both PRs edit this file; happy to rebase on it, or to drop the confluence half here and keep only the #2457 disable.

**#2397** is untouched: "Pattern equivalence" exhausts node's 4 GB heap on ~1 seed in 6, and an OOM kills the process rather than failing a test, so it can't be pinned this way. Wants a lower `~count`/step limit, separately.

### Verification

`make ci-check` (release, warnings-as-errors over `test/`) and `dune fmt` clean. `Evaluator.Properties` green, including #2457's seeds 3 and 468045375 which previously failed. Both pins confirmed to fail when the disagreement they assert goes away.

### DIFF
--- a/test/evaluator/Test_Evaluator_Properties.re
+++ b/test/evaluator/Test_Evaluator_Properties.re
@@ -1,4 +1,3 @@
-open Haz3lcore;
 open Language;
 open Test_Evaluator_Prelude;
 open Alcotest;
@@ -47,19 +46,13 @@ let qcheck_evaluator_does_not_crash_test =
     }
   });
 
-let show_core_exp = exp =>
-  exp
-  |> ExpToSegment.exp_to_segment(
-       ~settings=
-         ExpToSegment.Settings.of_core(~inline=true, CoreSettings.off),
-       _,
-     )
-  |> Printer.of_segment(~holes="?", _);
-
-/* Output is easier to view through ExpToSegment. This may result in a loss of
-   information. */
+/* Rendered by the same function that prints QCheck counterexamples, so a
+   failure here and a failure from the generator read alike. */
 let testable_core_exp =
-  testable(Fmt.using(show_core_exp, Fmt.string), Equality.semantic.exp);
+  testable(
+    Fmt.using(QCheck_Util.show_core_exp, Fmt.string),
+    Equality.semantic.exp,
+  );
 
 /* Elaborate `uexp`, reduce it with the evaluator and with the stepper, and
    `check` the two results against each other with `testable`: pass
@@ -193,8 +186,12 @@ let qcheck_pattern_equivalence_test =
             LimitedCompleted((first_exp, _)),
             LimitedCompleted((second_exp, _)),
           ) =>
-          print_endline("First expression: " ++ show_core_exp(first));
-          print_endline("Second expression: " ++ show_core_exp(second));
+          print_endline(
+            "First expression: " ++ QCheck_Util.show_core_exp(first),
+          );
+          print_endline(
+            "Second expression: " ++ QCheck_Util.show_core_exp(second),
+          );
           Alcotest.check(
             dhexp_typ,
             "Evaluated expressions are equal",
--- a/test/evaluator/Test_Evaluator_Properties.re
+++ b/test/evaluator/Test_Evaluator_Properties.re
@@ -46,19 +46,16 @@ let qcheck_evaluator_does_not_crash_test =
     }
   });
 
-/* Rendered by the same function that prints QCheck counterexamples, so a
-   failure here and a failure from the generator read alike. */
 let testable_core_exp =
   testable(
     Fmt.using(QCheck_Util.show_core_exp, Fmt.string),
     Equality.semantic.exp,
   );
 
-/* Elaborate `uexp`, reduce it with the evaluator and with the stepper, and
-   `check` the two results against each other with `testable`: pass
-   `testable_core_exp` to assert they agree, or `neg(testable_core_exp)` to
-   assert they still differ. Returns false when the case was filtered out — a
-   step limit, or a raise these properties skip — and nothing was compared. */
+/* Reduce `uexp` with the evaluator and with the stepper and `check` the two
+   against each other: pass `testable_core_exp` to assert they agree, or
+   `neg(testable_core_exp)` to assert they still differ. False when the case was
+   filtered out and nothing was compared. */
 let check_evaluator_against_stepper =
     (~step_limit: int, ~testable: testable(Exp.t), ~msg: string, uexp: Exp.t)
     : bool =>
@@ -102,8 +99,7 @@ let qcheck_stepper_confluence =
     ~count=1000,
     QCheck_Util.arb_exp(~minimal_idents=true, 10),
     uexp => {
-      /* `check` raises on a disagreement; a filtered-out case is not a
-         failure, so the compared/not-compared answer is discarded here. */
+      /* Failure arrives as a raise from `check`; the bool is only vacuity. */
       ignore(
         check_evaluator_against_stepper(
           ~step_limit=100,
@@ -116,11 +112,9 @@ let qcheck_stepper_confluence =
     },
   );
 
-/* Disabled: fails on any seed that draws a duplicate binder, where one of the
-   two freshens it and the other does not. Known bug
-   https://github.com/hazelgrove/hazel/issues/2128 (dup #2372). The test after
-   this one pins that counterexample and goes red once the bug is fixed, which
-   is the signal to re-enable this property. */
+/* Disabled: fails on any seed that draws a duplicate binder — known bug
+   https://github.com/hazelgrove/hazel/issues/2128 (dup #2372). The test below
+   pins that counterexample and goes red once it is fixed. */
 let qcheck_stepper_confluence_disabled =
   test_case(
     "Evaluator and stepper are consistent (disabled, #2128)", `Quick, () => {
@@ -285,8 +279,8 @@ let qcheck_preservation_test =
  * reuse_check path at all; if we regenerated the tree from source the
  * id spaces would be disjoint and nothing would match prev.
  *
- * Known skips below (return `true`): expressions with no int literal to
- * edit (nothing to test), anything that hits the step limit, and anything
+ * Known skips (the property still passes): expressions with no int literal
+ * to edit (nothing to test), anything that hits the step limit, and anything
  * that raises from statics/evaluation (filtered the same way as the
  * other evaluator QCheck tests in this file). */
 
@@ -339,13 +333,10 @@ let eval_limited =
     elab,
   );
 
-/* Bump the int literal at `target_id` by one, evaluate the edited program two
-   ways — incrementally, reusing the unedited run's cache, and from scratch —
-   and `check` the two against each other with `testable`: pass
-   `testable_core_exp` to assert they agree, or `neg(testable_core_exp)` to
-   assert they still differ. Both sides share one elaboration, so a
-   disagreement is a reuse / dirty-propagation bug rather than a semantic one.
-   Returns false when the case was filtered out and nothing was compared. */
+/* Bump the int literal at `target_id` by one and `check` the edited program
+   evaluated incrementally against it evaluated fresh: pass `testable_core_exp`
+   to assert they agree, or `neg(testable_core_exp)` to assert they still
+   differ. False when the case was filtered out and nothing was compared. */
 let check_incremental_against_fresh_after_edit =
     (
       ~target_id: Id.t,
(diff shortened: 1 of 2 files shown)


test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -119,6 +56,105 @@ let show_core_exp = exp =>
...
   |> Printer.of_segment(~holes="?", _);
 
+/* Elaborate `uexp` and reduce it with both engines, returning
+ * `Some((bigstep, smallstep))` when both land on a result. `None` means the
+ * case is vacuous — either engine hit the step limit, or statics/evaluation
+ * raised (filtered the same way as the other evaluator QCheck tests here).
+ *
+ * Shared by the confluence property and by the deterministic #2128 pin, so
+ * the two cannot drift apart. */
+let reduce_both_engines =
```
7h3kk1d: We don't use the word engine anywhere before so I'd rather you not invent terminology. I also think you can write this function in a way that returns either an test result or an assertion that can be checked as successful in the pbt or asserted failure in the counter example tests.

test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -304,6 +340,75 @@ let eval_limited =
...
   );
 
+/* Run the one-literal-edit experiment on `exp`: bump the int literal at
+ * `target_id` by one, then evaluate the edited program both incrementally
+ * (against the cache from the unedited run) and from scratch. `None` means
+ * the case is vacuous — a known-benign statics/dynamics failure or the step
+ * limit — and callers should treat it as no evidence either way.
+ *
+ * Shared by the property below and by the deterministic #2457 pin, so the
+ * two cannot drift apart. */
```
7h3kk1d: Improve this comment. Don't make it more verbose just make itmore clear what's being tested. Also it would be good not to return an option but actually do some form of assertion. It could return a test result or a result or something.

test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -313,75 +418,59 @@ let qcheck_incremental_matches_fresh_after_edit =
...
+ * for from the generator. Every ingredient matters: a *named* binding that
+ * survives into the residual, the edited literal bound to `_`, and
+ * evaluation getting stuck *after* the literal via a failing pattern match.
+ *
+ * This asserts the CURRENT, WRONG behaviour — fresh eval substitutes `x`'s
+ * value into the residual environment, incremental eval leaves `x` as a
+ * variable. It is deliberately NOT an `Alcotest.skip()` like
+ * `skip_known_bug` elsewhere in test/: a skipped test stays silent forever,
+ * whereas this one goes red the day #2457 is fixed and tells us to re-enable
+ * `qcheck_incremental_matches_fresh_after_edit` above. */
```
7h3kk1d: This comment is too long. Just say it's a known bug with a link to the github issue and say it's why the pbt is disabled.

test/evaluator/Test_Evaluator_Properties.re
```diff
@@ -673,9 +762,38 @@ let tests = (
     yielding_streaming_current_state_test,
     yielding_streaming_current_id_invalid_race_test,
     QCheck_alcotest.to_alcotest(qcheck_evaluator_does_not_crash_test),
-    QCheck_alcotest.to_alcotest(qcheck_stepper_confluence),
+    /* Disabled: fails for any seed that generates a duplicate binder
```
7h3kk1d: Put these comments/information in one place. Don't put it on the definition and on the usages.

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: Both dropped. `reduce_both_engines` is now `evaluator_and_stepper_agree`, which is just the property's own name — and `eval_after_literal_edit` became `incremental_agrees_with_fresh_after_edit`.

On the return type: both helpers now produce a `verdict` (`Holds` / `Fails(rendered)` / `Vacuous`) defined at the top of the file, and the `Equality.semantic.exp` comparison moved inside them — previously each caller redid it, which was the actual duplication. Two readers of a verdict:

- `holds` — QCheck predicate; prints the rendered disagreement and returns false only on `Fails`.
- `check_still_fails(~issue, ~property, verdict)` — the counterexample; passes on `Fails`, and on `Holds` fails with `<issue url> looks fixed: re-enable the `\`'`X`' property, delete this`.

`Vacuous` is its own case rather than folded into either, so a filtered case can't be read as evidence in either direction — and a counterexample that stops reaching a result gets told to re-minimize instead of silently passing.

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: Same `verdict` treatment. Comment is now four lines and says what's under test rather than how the plumbing behaves:

```
/* Bump the int literal at `target_id` by one, then evaluate the edited program
   two ways: incrementally, reusing the unedited run's cache, and from scratch.
   Both must land on the same result. Both sides share one elaboration, so a
   disagreement is a reuse / dirty-propagation bug rather than a semantic one. */
```

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: Cut. The whole thing is now one comment on the disabled property, which the pin sits directly beneath:

```
/* Disabled: fails on any seed that draws the shape below. Known bug
   https://github.com/hazelgrove/hazel/issues/2457. The test after this one
   pins that counterexample and goes red once the bug is fixed, which is the
   signal to re-enable this property. */
```

The pin itself has no comment. Its name and its failure message carry the rest.

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: Done — the info lives at the definitions only. Each property now has a named `..._disabled` wrapper beside it holding the single comment, so the list is bare:

```
QCheck_alcotest.to_alcotest(qcheck_evaluator_does_not_crash_test),
qcheck_stepper_confluence_disabled,
stepper_confluence_known_bug_test,
QCheck_alcotest.to_alcotest(qcheck_pattern_equivalence_test),
qcheck_incremental_matches_fresh_after_edit_disabled,
incremental_literal_edit_known_bug_test,
```

I also dropped the #2408 note that was in the list — it's PR context, not something the code needs to say.

test/evaluator/Test_Evaluator_Properties.re
7h3kk1d: Right — Alcotest already has it: `neg : 'a testable -> 'a testable`, a testable with negated equality. The `verdict` type, `holds` and `check_still_fails` are all gone.

There's now one parameterized check per property, taking the `testable` and `msg` that `Alcotest.check` takes anyway:

```reason
let check_evaluator_against_stepper =
    (~step_limit: int, ~testable: testable(Exp.t), ~msg: string, uexp: Exp.t): bool
```

The property passes `testable_core_exp`; the counterexample passes `neg(testable_core_exp)` plus a message naming the property to re-enable. Same function, same comparison, opposite expectation. `testable_core_exp` came back because Alcotest renders both expressions itself on failure, which is better than the string I was building by hand.

The `bool` is "did it actually compare anything" — the properties discard it, and the counterexamples assert it, so one that stops reaching a result gets told to re-minimize rather than silently passing. Verified both directions: pointing the pin at `fun (x::y) -> x` (no duplicate binder, so the two agree) fails with `#2128 looks fixed: re-enable ...`, and forcing the case vacuous fails with `counterexample still reduces under both; re-minimize it otherwise`.

disconcision:
lmao at you talking to the agent wearing your face here

7h3kk1d:
> lmao at you talking to the agent wearing your face here

Am I not allowed to talk to myself now???

matthewhammer:
Went through this as a pair with Claude Code, and checked it out locally. It helps us, concretely.

### Why it matters on our side

The #2457 property is the one that actually cost us something. On 2026-09-04 we reverted a good commit on `fumola-livelit-mvp` (#2492) because this group went red — the revert was unnecessary, the failure was the flake. I filed #2499 the next day: same property, seed 955690280, [run 33941773846](https://github.com/hazelgrove/hazel/actions/runs/33941773846), on a branch whose diff is livelit/projector/rendering code and touches no evaluator code at all. That's a fourth CI data point for your #2457 row, from a fourth unrelated branch — worth adding to the table. #2499 is probably worth closing as a dup of #2457 so it doesn't get triaged twice.

### Verified locally, on 98721df

- `Evaluator.Properties` green in 21s. Cases 5 and 8 skip, 6 and 9 pass.
- Both pins have teeth. Dropping `neg` makes both fail, with exactly the disagreements you describe:
  - #2128 — `fun (x:: x):([?]) -> x` vs `fun (x:: x'):([?]) -> x'`
  - #2457 — `let true = A in (x=(()))` vs `let true = A in (x=x)`
- The drift protection is real, which is the part I most wanted to confirm. Injecting `~count="not-an-int"` inside a disabled QCheck body fails `dune build @check`. So a disabled property can't quietly rot against the code it calls — that's the whole argument for this over deleting them, and it holds.

### Two things worth folding in

**1. Case numbers shift, and the new case 7 is the OOM one.** Today case 7 is the incremental property — that's the number in #2499 and in our own triage notes. After this lands, 5 and 8 are the disabled pair and case 7 becomes "Pattern equivalence", i.e. #2397. The case index is what the checks API surfaces, so anyone following an existing note lands on the wrong property. Worth a line in the PR body.

**2. This doesn't get us to a quiet group.** Across the life of `fumola-livelit-mvp` we saw four distinct properties in here fail: #2372 (= #2128), #2457/#2499, #2287, and #2397. You take two. #2397 is the one that hurts most in triage, because the OOM aborts the run — the test count comes up short and later groups never execute, so it doesn't present as a flaky property, it presents as a crash. You're right that it can't be pinned this way, but the `~count`/step-limit trim you mention is small and would actually finish the job. I'd take it here if you're up for it; otherwise please open it so it doesn't get lost. #2287 is still live too.

### Smaller

- "re-enabling is one line" is more like eight — the `test_case`/`skip`/`ignore` wrapper, the list entry, and deleting the pin. Changes nothing, but I don't want someone planning around one line.
- The pins assert *that* the two differ, not *how*. `neg` is satisfied by any disagreement, so if #2457 gets fixed and something unrelated later makes those two results differ, the pin stays green and the property stays disabled — failing open in the direction that's hardest to notice. The failure output already renders the exact pair, so asserting the concrete wrong result would close it. Low priority, your call.
- codecov's -0.46% / 239 lines is real and worth naming in the body. A property that reddens at random wasn't giving usable signal either, so I think the trade is right — I'd just rather it be an explicit trade than a surprise.

Approach is right, and pinning the wrong behaviour so it goes red on fix is a genuine improvement over `skip_known_bug`.

Issue #1077, opened 2023-08-07
Whitespace/grout interaction roundup


disconcision:
This is intended to be a running list of problematic grout/whitespace interactions, intended to document an existing hack and to inform decisions about the new syntax model.

Ideally, we want to conform to text expectations by maintaining two properties:

1. When a user presses a key to insert a single character to the left, the caret moves one space to the right.
2. When a user presses backspace to delete a single character to the left, the caret moves one space to the left.

Note that violating these maybe isn't necessarily a deal-breaker. One situation where violations don't feel so bad is when we are filling a hole when the caret is on the right side of the grout; here, behaving identically to if we were on the left side feels okay I think:

1. Leftwards insert "+" on `2 ><| 1` results in `2 +| 1` (caret stays in place)
2. Leftwards insert "1" in `<>| + 2` results in  `1| + 2` (caret stays in place)

But often violation of either of these properties causes jank. Currently the base syntax model violates both these properties, as both insertion and deletion can result in grout being inserted or removed left of the caret. Consider the following examples:

I.  Leftwards insert "i" on `let a = 1 |` results in  `let a = 1 ><i|` (extra space; caret moves two characters to the right)
II.  Leftwards insert  " " on `let a = 1><in|` results in `let a = 1in |` (caret stays in the same place)
III.  Leftwards Insert " " on `[|]`  results in `[ <>|]` (extra space; caret moves two characters to the right)

These and directly analogous cases are the only examples I can remember; `If-then-else` has the same issue as let, `()` has the same issue as `[]`.

To mitigate these, we sometimes convert spaces to grout, and grouts to spaces when regrouting. The code to do so is in Segment.regrout. While attempting to mitigate these, I've often ended up breaking other examples. Here are some situations where we don't want to mess with the base behavior:

A. Rightwards delete on `1 +| 1` want `1 |>< 1`, currently correct in dev
B. Rightwards delete on `1| + 2` want `|<> + 2`, currently incorrect in dev `|<>+ 2`

I'm going to put up a PR which fixes B and adds the above examples as comments to justify the current hacks. Basically, the current approach is to selectively covert between spaces/grout depending on both the direction of regrouting and the shape of the grout being inserted. I have no theoretical justification using these criteria; it might be reasonable to use whether an insertion or deletion is happening instead, though that information is not currently piped to the regrouting level.

[...]

PR #2485, opened 2026-09-02
Author tutorial lessons as .hzt text

7h3kk1d:
Stacked on #2250, with #2513 on top. **All three PRs in this stack are meant to be merged together** — they are split for review only.

## Summary

Adopts the text-authored tutorial system from `probes-study-rebuilt` and converts all 25 lessons to it.

Each lesson used to be an `.ml` file holding a serialized zipper alongside its prose — ~26,000 lines of generated data that had to be regenerated by hand for any edit. Lessons are now plain-text `.hzt` files in `hazel-programs/tutorial/`, embedded with `ppx_blob` via a new `src/tutorialslides` library (a clone of `src/docslides`) and parsed into `Tutorial.spec` at startup by `TutorialText`:

```
@title      the lesson name
@flags      wrapper show_report version=N id=<uuid>
@prompt     markdown for the instructions panel
@hint       one-liner
@reference  markdown for the Task Reference sidebar
@hints      one per line
@code       editor contents (holes are ¿, or ? to survive a container)
@test       hidden tests
```

Net: **+2,855 / −26,285**.

The conversion was mechanical. `TutorialDecode` reads whatever lesson list is compiled in and emits `.hzt` via `MarkerParse.to_text`, so the CLI landed first and generated all 25 from the existing `Tu_*.ml`. `./hazel tutorial-verify` checks that each slide's impl and tests are a fixed point of `to_text ∘ of_text ∘ to_text`; all 25 are. `TaskRefDocs` is gone — decode inlines each lesson's composed `task_reference` into its own `@reference`.

One marker is ours rather than upstream's: **`@title`**. Upstream derives the title from the filename, which would renumber every lesson and mangle four of them (`Reverse Function Application (Pipelining)` losing its parens; `and`/`of` capitalising). `@title` pins the exact title and falls back to the derived one.

## Bugs the round-trip check surfaced

- **`$==` in the Case Expressions and Folding Lists hidden tests.** `$` does not lex in Hazel, so those six tests were unsatisfiable and always reported indeterminate — a live bug the serialized form hid. Now `==`, and both suites pass.
- **`[¿]` in List Literals** collapsed to `[]` on reparse, losing the fillable hole. Explicit `[?]` round-trips.
- Stray trailing grouts in five hidden-test cells.

## Tutorial-mode chrome

Also ports the styling/UX work from that branch:

- a collapsible **Tasks** section, split from the prompt at its `Task` header (all 25 lessons now have one)
- the hint as a keyed `<details>`, so its open state no longer leaks across slides
- first/last slide nav gating — **required**, since `.hzt` `version` defaults to 1 and the old `version > 1` / `version < 10` gates would break
- SAND background, focus-driven cell accent, `{{video:}}` and `{{no_editor}}` prompt markers
- per-slide scroll memory, and driving the slide `<select>` from the model so it follows keyboard navigation
- `Zipper.caret_to_start`, so `@code` loads with the caret at the start of the program

`show_report` changes meaning under the new `show_task_ui` gate — it now marks a slide as a *task* (grading face + report icon) rather than "start with the report panel expanded" — so it is set on the six converted lessons that have real hidden tests. Only Probes, whose sole test is `test true end`, leaves it off.

Four general CSS fixes come along: a broken pin-icon URL in `style.css` (which sits in `www/`, so `../img/` resolved outside it), the test sigil's `z-index` (it painted above projector overlays), `nowrap` on context-menu items, and overscroll containment so a nested horizontal scroll does not fire the browser back-gesture.

## Deliberately left behind

`TutorialProbeStrip` and `TutorialSlideInit` — both are exhaustive hardcoded tables over that branch's own slide ids and need `AutoProbe.t` plus new `ProbeProj.Settings` variants — and the probe/evaluator/Card rework they depend on.

## Test plan

- [x] `make dev` builds clean
- [x] `make test-quick` — 3465/3465
- [x] `./hazel tutorial-verify` — 25/25 slides round-trip cleanly
- [x] All 25 lessons in the dropdown, in order, with their exact original titles
- [x] Tables driven to 3/3 passing; Folding Lists to 3/3 (previously impossible)
- [x] Tasks + Hint collapsibles, grading face and panel, prev/next gating, Task Reference sidebar


7h3kk1d:
@disconcision — flagging where this diverges from `probes-study-rebuilt`, since we both touch the `.hzt` pipeline.

**Divergences**

- **`@title`** — new marker pinning the exact lesson title; filename-derived titles renumber everything and mangle punctuation. One-way compatible: your slides load fine here, mine break under your parser — an unrecognized `@title` falls through `parse_sections`' `| _ =>` arm into the `code` accumulator, so the title text lands in the program with no error. ~6 lines in `TutorialText` + 1 in `TutorialDecode` if you want it.
- **`is_tasks_header` also accepts singular `Task`** — all 25 of my lessons write `# Task`, so the collapsible never fired.
- **Kept `EvalResults`, skipped `TestSigilsOnly`** — several lessons teach through the result row (Tables shows the output table, Table Column Projection shows `["Herbert", "Gibson"]`).

**Fixes to your files**

- `tutorial/README.md` points at `src/haz3lcore/zipper/TextRoundtrip.re`, which exists on neither branch — it's `MarkerParse.to_text`/`of_text` over `PersistentZipper`. Two stale `gen-tutorial` mentions too.
- Added `hazel-programs/tutorial-imported/` to `.gitignore`; `tutorial-decode` writes there by default.

One thing worth knowing: `show_report` quietly changed meaning under `show_task_ui` — was "start with the report panel expanded", now "this is a task slide". Six converted lessons had it off and silently lost their grading face and panel.

Also — not sold on the code-block background in prompts (`.prompt-content .code`), so I may change that if it is alright with you.

7h3kk1d:
Consider changing color on the code background in descriptions

hazel-programs/tutorial/README.md
```diff
@@ -0,0 +1,78 @@
...
+| `@test` | `hidden_tests.tests` | defaults to `test true end` |
+| `@hint` | `display_hint` | short one-liner |
+| `@reference` | `task_reference` | markdown for the Task Reference sidebar |
+| `@hints` | `hidden_tests.hints` | one hint per non-empty line |
+| `@flags` | misc | space-separated: `wrapper`, `show_report`, `version=N`, `id=<uuid>` |
+
+- **No markers** → the whole file is `@code`.
+- Holes are written as `¿` (the implicit-hole marker). For a hole you want to
+  *survive* re-parsing inside a container (e.g. a fillable list element), prefer
+  the explicit hole token `?` — `[?]` round-trips, whereas implicit `[¿]` may
```
7h3kk1d: Does `?` get replaced with grout or is it an explicit hole?

hazel-programs/tutorial/14-list-literals.hzt
```diff
@@ -0,0 +1,36 @@
...
+```
+
+### List Type
+`[Int]`, `[String]`, `[Bool]`
+
+@hints
+first_four should be [1, 2, 3, 4]
+
+@code
+let first_four : [Int] = [?] in
```
7h3kk1d, line 32: This needs to be fixed with the implicit hole and grout. Figure out why it wasn't round tripping I guess.

hazel-programs/tutorial/README.md
```diff
@@ -0,0 +1,78 @@
...
+  collapse to `[]`. Probes/projectors round-trip as `^^probe(...)`.
+- `wrapper` wraps the impl as `let answer = <impl> in …` so the hidden tests
```
7h3kk1d: does it wrap the full impl or just the end? How do we reference other variables.

hazel-programs/tutorial/README.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+./hazel tutorial-verify            # per-slide OK/MISMATCH + summary
+./hazel tutorial-verify --verbose  # also print before/after text for mismatches
+```
+
+`tutorial-verify` checks that each slide's impl/tests text is a **fixed point**
+of the text round-trip (`to_text` == `to_text ∘ of_text ∘ to_text`). A clean
+slide is reproduced faithfully by decode→encode (IDs aside). The known
+non-fixed-point class is grout-placement quirks (e.g. `[¿]`→`[]`) — see the
+`?`-vs-`¿` note above.
```
7h3kk1d: Should this be a test as well

hazel-programs/tutorial/README.md
7h3kk1d: 🤖 An explicit hole tile. `?` is the parser's empty-hole token — real syntax that survives any round-trip. `¿` is only a *marker* for Grout, which has no textual form of its own: `of_text` deletes the marker and the load-time regrout puts Grout back. `MarkerParse`'s header defines both.

You were also right that this was the wrong place for it — the rule is about Hazel's text format, not tutorials. The README now says just what a lesson author needs (write `?` for a student blank) and leaves the format to `MarkerParse`.

hazel-programs/tutorial/README.md
7h3kk1d: 🤖 The **full** impl: `let answer = <impl> in <tests>`, so the tests see `answer` and nothing else it defined.

Other variables only work *without* `wrapper`. `EditorUtil.append_exp` recurses into `Let` bodies, so the tests get appended **inside** the impl's own let-chain and every binding it introduces is in scope — that is how a lesson whose tests reference `first_four` works, and why a lesson needing more than one name has to leave `wrapper` off. Both halves are in the README now.

hazel-programs/tutorial/README.md
7h3kk1d: 🤖 Yes — added a `TutorialLessons` group to `Test_TextRoundtrip.re`: 50 cases, every lesson's impl and hidden tests. That file rather than a new one because it is already `MarkerParse`'s round-trip suite and already had an equivalent per-slide `DocSlides` group, so a separate suite would have overlapped it.

Confirmed it is a real tripwire, not a rubber stamp: switching the list lesson to `[¿]` fails exactly one case, `List Literals (impl)`. So `tutorial-verify` is now for *diagnosing* a mismatch (`--verbose` prints before/after), not for catching one.

hazel-programs/tutorial/14-list-literals.hzt
7h3kk1d: 🤖 Figured it out, and it is not fixable here.

`strip_implicit_holes` recovers Grout by *deleting* the `¿` marker tile and letting the load-time regrout put Grout back. So a marker survives only where its removal leaves the term **incomplete**. Measured:

| authored | after one round-trip | |
|---|---|---|
| `[¿]` | `[]` | ✗ dropped |
| `[¿, 1]` | `[¿, 1]` | ✓ |
| `[1, ¿]` | `[1, ¿]` | ✓ |
| `(¿, 1)` | `(¿, 1)` | ✓ |
| `let x = ¿ in x` | `let x = ¿ in x` | ✓ |

Delete the marker in `[¿]` and `[]` is already a well-formed empty list, so regrout has no reason to insert anything.

That makes it a `MarkerParse`/regrout bug rather than a lesson bug — and a persistence one, since grout inside `[]` is reachable by typing (#1077 example III) and so a real document can lose it on reload. Filed as #2518 with a fix direction (swap the marker tile for a Grout piece directly instead of destruct-then-regrout). Worth knowing: #2479 adds exactly the right gate for this class (`TextRoundtrip.EditedGrout`) but its corpus does not cover the sole-element case, and #2406 would obsolete the whole thing.

Keeping `[?]` here regardless — an explicit hole tile is the right thing for a blank the student fills, and stays right after #2406 lands.

7h3kk1d: 🤖 Correction to my last paragraph — I had the intent backwards. Grout is what we *want* the student to fill; `[?]` is only a workaround for #2518, not the right pattern.

The README now says so: holes in lessons are `¿`, with the explicit-tile spelling flagged as a workaround to undo when the bug is fixed, and `14-list-literals.hzt` named as the only lesson doing it today. The difference is user-visible — with `?` the student replaces a piece of syntax rather than filling a gap — so #2518 has a real cost, which I have added to the issue.

Everything above the correction still holds: the mechanism, the table, and that it cannot be fixed from this PR.

hazel-programs/tutorial/README.md
7h3kk1d: 🤖 Recording an idea from discussion, **not** doing it here: the flag could go away entirely, with `append_exp` capturing the trailing value into `answer` while leaving the impl's own bindings in scope.

All 25 lessons already end in the expression that is the answer — the 20 plain ones as a let-chain ending in `first_four` / `triple(10)` / `table`, the 5 wrapper ones as a bare expression like `2 + ¿`. So the flag's only job is choosing which of two things the tests see: the impl's names *or* `answer`. Nothing requires that to be either/or:

```
<bindings> in let answer = <trailing expr> in <tests>
```

gives both. The wrapper lessons' tests keep using `answer`, the others keep using their own names, and no lesson binds `answer` itself so there is no shadowing. It is statics-neutral: today's join is `Seq(trailing_expr, tests)`, and `Seq` synthesizes its first operand with no expectation and adds no marks, as does an unannotated `let` pattern — and there is no unused-binding mark to trip on. It would also fix something small: `Seq` evaluates that trailing expression and discards the value, so today the last line of those 20 lessons is decorative as far as the tests are concerned. Implementation is a generalization of the walk `append_exp` already does through `Let`/`Seq`/`Filter`/`Theorem` bodies.

Two caveats for whoever picks it up: `answer` would mean "value of the trailing expression", which is not always the graded artifact (for the function lessons the last line is a demo call, so `answer` is `30`, not `triple`); and `wrapper: bool` is a field of `p('code)`, which derives sexp/yojson, so dropping it changes a serialized shape that student progress may be stored in.

**Why not now:** it changes `Tutorial.re`'s semantics and every lesson file, which would make integrating @disconcision's tutorials harder later. Not worth the divergence for a cleanup. Leaving the flag as-is and documented.

PR #2488, opened 2026-09-02
non-`dev` (`user-livelits`): Type-check the expansion of each user-defined livelit

matthewhammer:
`user-livelits` does the heavy lifting. This is a small modification on top of it.

Before:

- No `Expansion` type member being looked at.
- No check against this absent member.

Now:

- The type member `Expansion` is required, and so are `Model` and `Action`.
- The check is performed for each expansion, on its output.

The tuple definition form goes with this, having nowhere to declare the types.

### Coming into alignment with livelits in PLDI 2021

**Part 1 — this PR.** The check against the statically named type, performed at the dynamic sites. That is the paper's own strategy rather than an approximation of it: §3.2.5 records that Hazel does not statically check `expand`'s definition, and that the parameterized expansion is instead "validated at each livelit invocation site, with errors reported to the client." With no splices that expansion takes no arguments, so validating it degenerates to checking the expansion at the expansion type.

**Parts 2+ — outside the scope here, and absent from `dev` too.** These introduce:

- the `SpliceRef` primitive type and its operations — `new_splice`, `set_splice` and `eval_splice`, in the paper's `UpdateCmd` and `ViewCmd` monads;
- the quoted AST — `Exp`, an encoding of Hazel's syntax, whose introduction form is quasiquotation, `` `0` `` (§3.2.1);
- `expand` extended to return a pair with that list of `SpliceRef`s, as in Fig. 3 of the [paper](https://hazel.org/papers/livelits-pldi2021.pdf): `expand : Model -> (Exp, List(SpliceRef))`.

The pair is there because "the expansion must treat splices parametrically" (§3.2.5): the first component takes one argument per listed `SpliceRef` and returns the expansion type. Note that `expand` itself stays pure — splices bring the monads, but not to `expand`.

The quotation is entailed by the splices rather than an independent choice. We can do without `Exp` here only because the splice list is empty: a closed expansion need not be code at all, so `expand` returns an ordinary value of the expansion type and application does the work quotation would have done. Splices force the issue, since a parameterized expansion embedding client-site expressions cannot be an ordinary value. (§3 flags the paper's quasiquotation and do notation as presentational sugar rather than implemented Hazel, so both are new surface syntax, not just new library types.)

The paper names one further alternative we are not pursuing: definition-site verification via "a typed quotation system as in, e.g., MetaOCaml." That is a distinct thing from the quasiquotation above — untyped quotation merely introduces `Exp`, whereas typed quotation is what would move Part 1's check from the invocation site to the definition site. The paper notes it is awkward because "the type of the quotation depends on the type of each splice in the splice list."

### DIFF
--- a/docs/livelits.md
+++ b/docs/livelits.md
@@ -23,6 +23,7 @@ A Hazel program can define a livelit by binding a livelit name to a module:
 let ^pct = {
   type Model = Int;
   type Action = Int;
+  type Expansion = Int;
   let init : Model = 50;
   let update = fun (m, a) : (Model, Action) -> a;
   let view = fun m : Model -> ...;
@@ -35,10 +36,34 @@ with `update: (Model, Action) => Model`, `view: Model => HTML` (handlers emit
 Actions, as in the MVU apps — see mvu.md), and `expand: Model => Expansion`.
 An optional member `shape = Inline(width) | Block(width, height) |
 Tab(width, height)` (a `LivelitShape`) sets the projector's footprint in
-character cells. Type members are accepted but not yet semantically
-load-bearing. Helpers are ordinary additional members. Since modules are
-sugar for labeled tuples, a positional `(init, update, view, expand[, shape])`
-tuple is accepted as the equivalent form.
+character cells. Helpers are ordinary additional members.
+
+All three type members are required, and they are the livelit's interface.
+`Model` types each use's argument, `Action` types what the view emits, and
+`Expansion` is what clients type against. There is no tuple form: a tuple
+has nowhere to declare the types.
+
+## Type-Checking the Expansion
+
+A use of `^pct` synthesizes the declared `Expansion` — not the type of
+whatever code `expand` produced. That is the abstract reasoning principle
+of the paper: a client reads the livelit's declared type and never the
+expansion. The obligation it creates is discharged at each use, where
+statics types the expansion (in synthetic mode, on a throwaway info map)
+and compares the result with the declaration. An inconsistency is reported
+on the use as `BadLivelitExpansion` — "Livelit expands to type Int, but
+declares Expansion = String" — locating the fault in the livelit rather
+than in the client's surrounding code.
+
+Consistency, not equality, is the test, as everywhere else in the language:
+an expansion that synthesizes `Unknown` (an unannotated `expand` whose body
+gives statics nothing to go on, or a builtin livelit generating a hole)
+stays gradual and is not marked. What the declaration buys in that case is
+still real — clients type against a known type instead of `Unknown`.
+
+The expansion typed is the one built from the *surface* model, since statics
+traverses surface syntax only; the elaborated model, which a user-defined
+livelit's expansion embeds, is what actually evaluates.
 
 Each use elaborates to `^name.expand(model)` through the runtime `^name`
 binding, so shadowing and scoping behave like ordinary lets, and each use's
--- a/docs/livelits.md
+++ b/docs/livelits.md
@@ -7,6 +7,21 @@ Hazel implements a version of the live literals (livelits) mechanism described i
 - No parameters
 - No splices
 
+Splices are the remaining gap, and they will arrive as a `SpliceRef` primitive type with
+operations over it (`new_splice`, `set_splice`, `eval_splice`, in the paper's `UpdateCmd`
+and `ViewCmd` monads). When they land, `expand` extends to return a pair with the list of
+`SpliceRef`s, as in Fig. 3 of the paper:
+
+```
+expand : Model -> (Exp, List(SpliceRef))
+```
+
+The pair is there because the expansion must treat splices parametrically: the first
+component takes one argument per listed `SpliceRef` and returns the expansion type.
+`expand` itself stays pure — splices bring the monads, but not to `expand`. Until then the
+splice list is empty, so `expand : Model -> Expansion` is an equivalent encoding of the
+paper's closed expansion rather than a deviation from it.
+
 ## Overview
 
 A livelit is a live GUI widget which can be inserted into expressions and generates code by expansion to an expression of some given type. To invoke a livelit, insert the name of the livelit (always prefixed with ^) then press space.
@@ -45,21 +60,30 @@ has nowhere to declare the types.
 
 ## Type-Checking the Expansion
 
-A use of `^pct` synthesizes the declared `Expansion` — not the type of
-whatever code `expand` produced. That is the abstract reasoning principle
-of the paper: a client reads the livelit's declared type and never the
-expansion. The obligation it creates is discharged at each use, where
-statics types the expansion (in synthetic mode, on a throwaway info map)
-and compares the result with the declaration. An inconsistency is reported
-on the use as `BadLivelitExpansion` — "Livelit expands to type Int, but
-declares Expansion = String" — locating the fault in the livelit rather
+A use of `^pct` synthesizes the declared `Expansion`, not the type of whatever
+code `expand` produced. The check that earns this is performed for each
+expansion, on its output: statics types the expansion (in synthetic mode, on a
+throwaway info map) and compares the result with the declaration, marking the
+use `BadLivelitExpansion` on an inconsistency — "Livelit expands to type Int,
+but declares Expansion = String". The fault is located in the livelit rather
 than in the client's surrounding code.
 
-Consistency, not equality, is the test, as everywhere else in the language:
-an expansion that synthesizes `Unknown` (an unannotated `expand` whose body
-gives statics nothing to go on, or a builtin livelit generating a hole)
-stays gradual and is not marked. What the declaration buys in that case is
-still real — clients type against a known type instead of `Unknown`.
+Checking at the use site is the PLDI 2021 paper's own strategy rather than a
+deviation from it. §3.2.5 records that Hazel does not statically check
+`expand`'s definition, and that the parameterized expansion is instead
+"validated at each livelit invocation site, with errors reported to the
+client". With no splices that expansion takes no arguments, so validating it
+degenerates to checking the expansion at the expansion type, which is what
+happens here. The paper names definition-site verification via "a typed
+quotation system as in, e.g., MetaOCaml" as the alternative, awkward because
+"the type of the quotation depends on the type of each splice in the splice
+list".
+
+Consistency, not equality, is the test, as everywhere else in the language: an
+expansion that synthesizes `Unknown` (an unannotated `expand` whose body gives
+statics nothing to go on, or a builtin livelit generating a hole) stays gradual
+and is not marked. What the declaration buys in that case is still real —
+clients type against a known type instead of `Unknown`.
 
 The expansion typed is the one built from the *surface* model, since statics
 traverses surface syntax only; the elaborated model, which a user-defined
--- a/hazel-programs/docs/livelits/README.md
+++ b/hazel-programs/docs/livelits/README.md
@@ -12,6 +12,7 @@ A livelit definition binds a livelit name to a module:
 let ^name = {
   type Model = ...;
   type Action = ...;
+  type Expansion = ...;
   let init : Model = ...;
   let update = fun (m, a) : (Model, Action) -> ...;
   let view = fun m : Model -> ...;
@@ -19,18 +20,28 @@ let ^name = {
 } in ...
 ```
 
+All three type members are required — they are the livelit's interface:
+
+- `type Model`: the state a use carries, in its own argument
+- `type Action`: what the view's handlers emit
+- `type Expansion`: what a use means to the program. This is the type
+  clients see, so a use of `^name` has type `Expansion` however `expand`
+  is written; statics checks each use's expansion against the declaration
+  and marks the use when the two are inconsistent.
+
+and the four value members:
+
 - `init`: the model a fresh use starts with (`^name` + space inserts it)
 - `update: (Model, Action) => Model`
 - `view: Model => HTML` — handlers emit Actions (same HTML API as the MVU
   apps, see ../mvu/README.md)
-- `expand: Model => Expansion` — what a use means to the program
+- `expand: Model => Expansion`
 - optional member `shape`: `Inline(width)`, `Block(width, height)`, or
   `Tab(width, height)` — the widget's footprint in character cells
 
-Type members are accepted but not yet load-bearing; helpers (like the color
-picker's `css` and `pick`) are ordinary members. A positional
-`(init, update, view, expand[, shape])` tuple is the desugared equivalent.
-The definition must be closed — its functions evaluate in the builtin
+Helpers (like the color picker's `css` and `pick`) are ordinary extra
+members. There is no tuple form: a tuple has nowhere to declare the three
+types. The definition must be closed — its functions evaluate in the builtin
 environment — so helpers belong among the members. Each use's model is
 stored in its own argument syntax, so state survives in the program text.
 
--- a/hazel-programs/docs/livelits/defined-slider.hz
+++ b/hazel-programs/docs/livelits/defined-slider.hz
@@ -3,14 +3,22 @@
 # A livelit is a live GUI for a value in your program. Define one #
 # with a livelit name (^name) bound to a module: #
 
+# Three types declare the interface. Expansion is the one clients #
+# see: a use of ^pct has type Expansion, whatever expand returns. #
+
+# Model:     the state a use carries, stored in its argument #
+# Action:    what the GUI emits #
+# Expansion: what a use MEANS to the program #
+
 # init:   the model a fresh use starts with #
 # update: (Model, Action) -> Model, run when the GUI acts #
 # view:   Model -> HTML, whose handlers emit Actions #
-# expand: Model -> value, what a use means to the program #
+# expand: Model -> Expansion, checked against the declaration #
 
 let ^pct = {
   type Model = Int;
   type Action = Int;
+  type Expansion = Int;
 
   let init : Model = 50;
 
--- a/src/haz3lcore/CompositionCore/prompt_factory/DocPacks.re
+++ b/src/haz3lcore/CompositionCore/prompt_factory/DocPacks.re
@@ -131,6 +131,7 @@ module bound to a livelit name (`^` prefix):
 let ^pct = {
 type Model = Int;
 type Action = Int;
+type Expansion = Int;
 let init : Model = 50;
 let update(m: Model, a: Action) = a;
 let view(m: Model) =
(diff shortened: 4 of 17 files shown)


matthewhammer:
# hand-written by @matthewhammer 

Hey reviewers, as I'm playing with examples that have errors in the `expand` function, I see duplicated error messages for the same line.

image



I asked Claude about it, and got some insights; I asked Claude to post them here, and eventually got the text below, after some massaging.  I thought I'd add this preface to explain it with more context.

# Generated through @matthewhammer + Claude

A livelit use is always wrapped in a projector, and transparent wrappers restate the marks of the term they wrap, so every expansion error gets reported twice — the mechanism isn't livelit-specific (`(y)` reports its free variable twice, `((y))` three times), it's just guaranteed to bite for livelits. The cheap fix is at the collection layer: collapse each chain of ids sharing a mark list into a single row, reported at the innermost id, promoted to the innermost *visible* wrapper when the marked term sits under a projector and so has no measurement of its own. The alternative is to stop propagating marks through wrappers in `Statics`, which is cleaner in principle but would take the projector's error decoration with it, so it's worth deciding which layer this belongs in.

matthewhammer:
To keep this PR's scope on expansion typing, the dedup fix lives in #2489, which branches off this one rather than being folded in — parked there until reviewers have a view on which layer it belongs in (collection vs. `Statics`). Nothing here changes.

One clarification on the screenshot above: it's a local edit to the emotion slide's `expand` (returning `(word(m), 1)` against `type Expansion = String`), not the slide as committed — the committed slide is clean, so reproducing the duplication needs that edit, or any error inside parens.

hazel-programs/docs/livelits/defined-slider.hz
```diff
@@ -3,14 +3,22 @@
 # A livelit is a live GUI for a value in your program. Define one #
 # with a livelit name (^name) bound to a module: #
 
+# Three types declare the interface. Expansion is the one clients #
```
disconcision: Minor point but I tend to either tell claude to use simple english (no aphorisms, no poetic turns of phrase), or rephrase the output myself. After you use it a bit sentence structures like the first one here are glaring signs of AI writing. I've been guilty of this myself, and I'm not sure there's anything wrong with it per se, but I tend to find it grating, as do many others, so I try to reduce or eliminate it, especially in user-facing text

hazel-programs/docs/livelits/README.md
```diff
@@ -12,25 +12,36 @@ A livelit definition binds a livelit name to a module:
...
+- `expand: Model => Expansion`
 - optional member `shape`: `Inline(width)`, `Block(width, height)`, or
   `Tab(width, height)` — the widget's footprint in character cells
 
-Type members are accepted but not yet load-bearing; helpers (like the color
-picker's `css` and `pick`) are ordinary members. A positional
-`(init, update, view, expand[, shape])` tuple is the desugared equivalent.
-The definition must be closed — its functions evaluate in the builtin
+Helpers (like the color picker's `css` and `pick`) are ordinary extra
+members. There is no tuple form: a tuple has nowhere to declare the three
```
disconcision, line 43: A thing claude tends to do, especially in code comments... talk about how something used to be, or how it was changed, where no longer relevant (i'm assuming the tuple part is no longer relevant here, but may be wrong). i tend to do a holistic pass before putting a PR up for comment improvement, asking it to reevaluate comments/docs wrt utility to someone reading the code, erring on the side of minimalism.

That said, i probably haven't done such a pass on user-defined livelits yet, so i can't claim innocence here!

src/haz3lcore/CompositionCore/prompt_factory/DocPacks.re
```diff
@@ -144,20 +145,33 @@ let expand(m: Model) = m
...
 - Each use `^name(model)` carries its own model in its own argument.
   Wrap uses in `^^livelit(...)` so the GUI shows; a bare `^name(model)`
   is still a valid expression, just without the widget.
-- Helpers and type members are ordinary module members; keep the
-  definition self-contained (helpers inside the module).
+- Helpers are ordinary extra module members; keep the definition
+  self-contained (helpers inside the module).
+- A definition missing any of the three types or four members is
+  rejected, and the livelit name stays unbound. There is no tuple
+  form: it has nowhere to declare the types.
```
disconcision: similar

src/haz3lcore/TyDi/ErrorPrint.re
```diff
@@ -140,17 +140,23 @@ let exp_mark_to_string = (ctx: Ctx.t, ana: Typ.t, m: Mark.t): string => {
...
-  | InvalidLivelitDef(DefBadArity(n)) =>
+  | BadLivelitExpansion({declared, actual}) =>
     prn(
-      "Livelit definition should have fields (init, update, view, expand), got %d",
-      n,
+      "Livelit expands to type %s, but declares Expansion = %s",
+      Print.typ(actual),
+      Print.typ(declared),
     )
+  | InvalidLivelitDef(DefNotModule) => "Livelit definition should be a module declaring types Model, Action, Expansion and members init, update, view, expand"
```
disconcision: The need for this will hopefully be subsumed by a more advanced module system, but reasonable for now

docs/livelits.md
```diff
@@ -7,6 +7,21 @@ Hazel implements a version of the live literals (livelits) mechanism described i
...
+
+```
+expand : Model -> (Exp, List(SpliceRef))
+```
+
+The pair is there because the expansion must treat splices parametrically: the first
+component takes one argument per listed `SpliceRef` and returns the expansion type.
+`expand` itself stays pure — splices bring the monads, but not to `expand`. Until then the
+splice list is empty, so `expand : Model -> Expansion` is an equivalent encoding of the
+paper's closed expansion rather than a deviation from it.
```
7h3kk1d, line 23: I don't think we need to mention this until we actually start working on it.

7h3kk1d: If anything we should mention what we're doing different than the paper or what we're holding off on just so claude doesn't get confused.

docs/livelits.md
```diff
@@ -35,10 +51,43 @@ with `update: (Model, Action) => Model`, `view: Model => HTML` (handlers emit
...
-character cells. Type members are accepted but not yet semantically
-load-bearing. Helpers are ordinary additional members. Since modules are
-sugar for labeled tuples, a positional `(init, update, view, expand[, shape])`
-tuple is accepted as the equivalent form.
+character cells. Helpers are ordinary additional members.
+
+All three type members are required, and they are the livelit's interface.
+`Model` types each use's argument, `Action` types what the view emits, and
+`Expansion` is what clients type against. There is no tuple form: a tuple
+has nowhere to declare the types.
```
7h3kk1d, line 59:
```suggestion
`Expansion` is what clients type against.
```

Claude is just storing random facts as a chain of it's reasoning

docs/livelits.md
```diff
@@ -35,10 +51,43 @@ with `update: (Model, Action) => Model`, `view: Model => HTML` (handlers emit
...
 Tab(width, height)` (a `LivelitShape`) sets the projector's footprint in
-character cells. Type members are accepted but not yet semantically
-load-bearing. Helpers are ordinary additional members. Since modules are
-sugar for labeled tuples, a positional `(init, update, view, expand[, shape])`
-tuple is accepted as the equivalent form.
+character cells. Helpers are ordinary additional members.
+
+All three type members are required, and they are the livelit's interface.
+`Model` types each use's argument, `Action` types what the view emits, and
+`Expansion` is what clients type against. There is no tuple form: a tuple
```
7h3kk1d, line 58: > `Expansion` is what clients type against

I feel like this could be clearer. I think it's what the expansion is type checked against on invocation.

src/haz3lcore/projectors/implementations/LivelitProj.re
```diff
@@ -186,33 +186,21 @@ module M: Projector = {
...
-  /* Extract a member from the evaluated definition record: by label when
-     labeled (modules desugar to labeled tuples, any size); positional only
-     for a plain (init, update, view, expand) tuple */
+  /* Extract a member from the evaluated definition record. A definition is
+     always a module, and modules desugar to LABELED tuples, so members are
+     found by name and member order and helper count don't matter. */
   let record_field =
-      (record: TermBase.Exp.t, label: string, index: int)
-      : option(TermBase.Exp.t) =>
+      (record: TermBase.Exp.t, label: string): option(TermBase.Exp.t) =>
```
7h3kk1d, line 193: Unrelated to this or but I find modules labeled tuples sketchy and then transitively records extra sketchy

src/haz3lcore/projectors/implementations/LivelitProj.re
```diff
@@ -186,33 +186,21 @@ module M: Projector = {
...
-          f =>
-            switch (MvuShape.of_field(f)) {
-            | Some((l, v)) when l == label => Some(v)
-            | _ => None
-            },
-          fs,
-        )
-      ) {
-      | Some(v) => Some(v)
-      | None when List.length(fs) >= 4 =>
```
7h3kk1d, line 208: Curious what this code was doing

src/language/statics/Mark.re
```diff
@@ -30,9 +30,9 @@ type error_builtin =
 
 [@deriving (show({with_path: false}), sexp, yojson, eq)]
 type livelit_def_error =
-  | DefNotTuple
-  | DefBadArity(int)
-  | DefMissingMembers(list(string));
+  | DefNotModule
+  | DefMissingMembers(list(string))
+  | DefMissingTypes(list(string));
```
7h3kk1d, line 35: It's not wrong but it's surprising to me we aren't reusing the module error messages.

7h3kk1d (commented):
I'm still reading the PR and just got to the main Statics part. Mostly surface-level stuff so far, but I'm still somewhat confused why we validate the expansion type at invocation if it's a value expand is returning and it's a fixed type. It seems like expand being typed `Model -> Expand` could just be typed at livelit definition. I understand why this isn't the case in the papers quoted version.

src/language/statics/Mark.re
disconcision: part of this is artifactual of the fact that modules are currently fake. type members are ignored by module statics so if we want to provide feedback we'd need an additional error case.

disconcision: similarly I dont think we're analyzing livelit defs against a (module/tuple) type, as that would restrict having helper members in the module, so absent the ability to do that kind of thing at the module type level, I can see having livelit-def-specific member errors

disconcision (commented):
No big concerns though I also agree with most of alexander's comments... A challenge here though is that you're building on top of user-livelits, which is mostly slop I haven't reviewed yet, so definitely don't take the existing internals as principled.

I guess one thing I haven't fully thought out is how much we should be special-casing livelit def stuff as opposed to extending module features to support the required functionality. I'm fine doing the former for now but might be worth thinking about what we'd need from modules to avoid it.

I'll do a functional review tomorrow

hazel-programs/docs/livelits/README.md
matthewhammer: Makes sense.  I'll do a pass like that.

To do so, I was actually wondering about that language around "tuples" myself, and how modules are being faked now in a different way compared to before.

Claude kept inserting these distinctions because I said I wanted type members now that were actually "load bearing" now, unlike before (language about not being "load bearing"  was there before that it asked me to confirm was changing).  It kept putting talk about "tuples" into the comments and the PR description and I kept rewriting the PR description to minimize that aspect.

It would be helpful to have a discussion somewhere (maybe a doc?) about how we are implementing modules over time, including now.  They are part of the livelit spec as written by the paper, but I don't know if they've ever existed in any real form in Hazel before.  I'd be curious to know the past and future plans there.  I apologize if that doc already exists and I missed it.

matthewhammer:
> It seems like expand being typed `Model -> Expand` could just be typed at livelit definition. I understand why this isn't the case in the papers quoted version.

Yeah, I guess without splices, there's a chance to do something really different from the paper here and more eager/local, in terms of typing.

But to be most like the paper, for when we do have splices, the current PR is delaying the check.

I think that's what Cyrus is asking for, but I'm not certain.

7h3kk1d:
Now that I'm further reading the code I have a slightly more nuanced take.
Screenshot 2026-09-04 at 9 10 26 AM


We're using the static type of the expand function to determine whether or not the expansion type is correct at invocation. The livelit paper since it returns the quoted version is relying on the output of the dynamics of the expansion function to do the invocation typechecking. I think this is probably more consequential to this design being reused for the splice version. In the screenshot above I changed the static type of the expand function to be `?` then returned an incompatible type. Because statically it's unknown it's not giving an error even though the value is `8` dynamically and clearly incompatible with the expansion type.

src/language/statics/Statics.re
```diff
@@ -1674,14 +1674,45 @@ and uexp_to_info_map =
...
+
           // try to expand
           switch (expand(model_for_expand)) {
           | Some(expanded) =>
             let (info, elab, m) =
               add(
                 ~elab_term=expanded,
                 ~elab_syn_ty=expansion_t,
-                ~marks=[],
+                ~marks=expansion_marks(expanded),
```
7h3kk1d, line 1715: Actually I'm not sure since this is running on the expanded code Idk why it's not getting marks in the version I put.

7h3kk1d: Okay, so andrew's pr makes the expansion literally `^livelit.expand(model)` so whether or not this is desirable is entirely in his PR.

src/language/statics/Statics.re
```diff
@@ -1674,14 +1674,45 @@ and uexp_to_info_map =
...
+             inconsistent with the declaration. Statics traverses surface
+             syntax only, so what gets typed is the expansion of the SURFACE
+             model even where the elaborated one is what gets evaluated.
+             That re-traverses the model, so a use costs twice its model
+             subtree — small in practice, since a model is a literal or a
+             committed transition over one. */
+          let expansion_marks = (expanded: Exp.t) => {
+            let to_check =
+              Option.is_some(user_def)
+                ? expand(arg.user_term) : Some(expanded);
```
7h3kk1d, line 1692: I don't understand why we need to reexpand this

src/language/statics/UserLivelit.re
```diff
@@ -370,63 +315,52 @@ let instrument_view =
...
+   expansion type, so statics owes a check that the expansion actually has
+   that type. `actual` is the type the expansion synthesizes on its own; a
+   mark is due when the two are inconsistent. Consistency, not equality, is
+   the test: an expansion that synthesizes Unknown (an unannotated `expand`,
+   or a builtin livelit generating a hole) stays gradual, exactly as it would
+   anywhere else in the language. */
+let expansion_mark =
+    (ctx: Ctx.t, ~declared: TermBase.Typ.t, ~actual: TermBase.Typ.t)
+    : list(Mark.t) =>
+  switch (Typ.meet(ctx, declared, actual)) {
```
7h3kk1d, line 358: We should just use Typ.is_consistent which already does this

src/language/statics/Statics.re
disconcision: how do you feel about it? in the current barely-metaprogram version just writing the ap in and deferring any evaluation to runtime seemed to make sense

disconcision (approved):
Functionally it looks good to me; i'm content to merge this in to my branch. I think there are a lot of things conceptually unsettled but not obvious whether it makes sense to deal with them all here

src/language/statics/Statics.re
7h3kk1d: I'm fine with it. It's just a very different design. We don't have the toggle from the paper so there's no way in this scheme to switch to the expansion. The paper didn't allow it because of the fixed expansion type but I could imagine a version that left the expansion type unknown and the model could change the expressions type resulting in a different static synthetic type. 

I don't have any issue with merging all of this but we should be okay with either changing it when splices come or having multiple concurrent designs.

matthewhammer:

matthewhammer:
Closing this, with an accounting of the review notes — several of which the work since has answered, and two of which are still open.

## Where the work lives now

This branch is fully contained in **`integration/livelits`** (#2576): its head `da22f4ae76` is an ancestor, with zero commits not carried forward. Nothing is stranded. The companion branch is **`integration/livelits-splicerefs`** (#2580), for when splices land. #2526, the earlier integration, is closed in favour of those two.

Importantly, this PR's feature is **not** redone there. The use-site check still marks `BadLivelitExpansion` exactly as it does here. What changed is that a second check was added beside it.

## @7h3kk1d — "why validate at invocation ... expand could just be typed at livelit definition"

**You were right, and this is now implemented.** At the time the answer was that per-use validation is the paper's own strategy (PLDI 2021, §3.2.5), which is true but was not a reason to stop there. Cyrus asked for the definition-site check independently about six weeks later, and `integration/livelits` now does it: a definition is checked against a builtin module signature `Livelit`, so `expand : Model -> Expansion` is verified where it is written, with a new `DefMemberMismatch` mark.

Both checks are kept, and neither is redundant. The definition-site check cannot see an expansion whose type depends on the model *value*; that is what the use-site check is for. Your instinct that the fixed type made definition-site typing possible was the correct one.

## @disconcision — "special-casing livelit def stuff vs. extending module features"

Also now answered, in the direction you suggested. The livelit interface is a real module signature rather than hand-rolled member checks:

```
type Livelit = {
  type Model; type Action; type Expansion;
  let init   : Model;
  let update : (Model, Action) -> Model;
  let view   : Model -> Html.T;
  let expand : Model -> Expansion
}
```

Its three type members are **abstract** — Modules II part 3 (#2495) — which is what lets each livelit choose its own, and the check realizes them from the definition's own manifest types. Your two notes on `Mark.re` are resolved by the same thing: type members are no longer ignored by module statics, and helper members are fine because extra members are permitted by width subtyping (#2494). So the reason the analysis could not be done at the module-type level has gone away.

Nothing seals a livelit with this signature, deliberately: sealing would hide `Expansion`, which clients must see.

## @7h3kk1d — `Typ.is_consistent`

Fixed just now on `integration/livelits`. The use-site check was still doing `Typ.meet(...)` and matching `Some`/`None`, which is precisely what `is_consistent` wraps.

## Still open

Two notes I have **not** addressed, recorded so they are not lost:

1. **@7h3kk1d: "surprising we aren't reusing the module error messages."** Still true. The new definition-site mark is `DefMemberMismatch`, livelit-specific, sitting beside the existing `ModuleTypeMemberMismatch`. Now that livelit definitions really are checked against a real signature, reusing the module marks is more plausible than it was — but it is not done.

2. **@disconcision: comments describing how things used to be.** Also still true. `hazel-programs/docs/livelits/README.md` and `docs/livelits.md` still say "There is no tuple form: a tuple has nowhere to declare the three types," which is history, not information for a reader. @7h3kk1d's related suggestion on `docs/livelits.md` — that "`Expansion` is what clients type against" should say what is actually checked and when — is now doubly out of date, since there are two checks. The new deck slides say it properly; `docs/livelits.md` has not had that pass.

## And one note that is about to become live again

@7h3kk1d, on the expansion being written as `^livelit.expand(model)`: *"we should be okay with either changing it when splices come or having multiple concurrent designs."* Splices are the next piece of work (#2580), so that is the design question about to reopen rather than a settled one.

PR #2489, opened 2026-09-03
non-`dev` (`livelit-expansion-typing`): Report each error once, not once per wrapper

matthewhammer:
Follows #2488, and parked here rather than folded into it until reviewers weigh in on which layer the fix belongs in (see https://github.com/hazelgrove/hazel/pull/2488#issuecomment-5529767668).

## The problem

A transparent wrapper restates the marks of the term it wraps — `Parens` and `Projector` in `Statics` pass `~marks` straight through — so one error occupies a chain of nested ids and the problems sidebar lists every one of them:

| program | static errors reported |
| --- | --- |
| `y` | 1 |
| `(y)` | 2 |
| `((y))` | 3 |
| `^s(1)` (mistyped `expand`) | 1 |
| `^^livelit(^s(1))` (same) | 2 |

Not livelit-specific, but guaranteed to bite for livelits: a use is always inside a projector, so a `BadLivelitExpansion` was always duplicated.

## The fix

Collapse each chain at the collection layer (`ProblemCollection.make_problem_context`). `Statics` hands a wrapper the child's mark list itself, so physical equality identifies exactly the ids that inherited a mark — never two equal-looking marks on unrelated terms — and the innermost id of a chain is the one that owns it.

Ids under a projector have no measurement of their own, so when the owner of a mark is hidden the row is reported at the innermost wrapper the user can actually see: the projector. That keeps the row clickable and lets it highlight with the caret.

`statics.error_ids` is untouched, so code decorations — including the projector's own error styling — are unchanged.

## The open question

The alternative is to stop propagating marks through wrappers in `Statics` at all. Cleaner in principle, but the projector's error decoration reads the projector node's info, so it would go with it. Happy to move the fix there instead if that's the preference.

## Testing

Four cases added to `Test_ProblemCollection`: parens don't multiply an error, a collapsed chain reports the marked term, a projected livelit use reports one error, a hidden marked term reports at its projector. `ProblemCollection` (19), `UserLivelits` (34), `AgentTools` (174) and `ErrorPrint` (5) pass.

### DIFF
--- a/src/haz3lcore/ProblemCollection.re
+++ b/src/haz3lcore/ProblemCollection.re
@@ -96,17 +96,70 @@ let make_problem_context =
      own term so problems don't leak across groups. */
   let id_in_this_editor = id =>
     TermData.root_piece(id, syntax.term_data) != None;
+  /* One error, one row. A transparent wrapper re-states the marks of the
+     term it wraps (`Parens` and `Projector` in Statics pass `~marks`
+     straight through), so a single error occupies a chain of nested ids:
+     `(y)` carries the free-variable mark on the variable AND on the
+     parens, and a projected livelit use carries the expansion mark on
+     `^name(model)` AND on the projector around it. Statics hands the
+     wrapper the child's mark list itself, so physical equality picks out
+     exactly the ids that inherited a mark and never two equal-looking
+     marks on unrelated terms. */
+  let inherits_marks = (marks: list(Mark.t), id: Id.t): bool =>
+    switch (Statics.Map.lookup(id, info_map)) {
+    | Some(anc) => Info.marks_of(anc) === marks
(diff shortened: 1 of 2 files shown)


7h3kk1d (commented):
I filed https://github.com/hazelgrove/hazel/issues/2561 for a second source of duplicate errors, and I think we need to address that before we can really test this PR.

The two sit at different layers. This PR removes duplication downstream, in ProblemCollection and the problems sidebar, collapsing marks that wrappers inherited. #2561 is upstream: each enclosing form is generating distinct marks, even when it built its type out of an already-marked child. Those are distinct marks, so the physical-equality keying here won't collapse them.

matthewhammer:

Issue #1131, opened 2023-11-02
Port to wasm_of_ocaml


cyrus-:
see https://tarides.com/blog/2023-11-01-webassembly-support-for-ocaml-introducing-wasm-of-ocaml/ and https://github.com/ocaml-wasm/wasm_of_ocaml

cyrus-:
Looks like wasm_of_ocaml and js_of_ocaml have been officially merged and it should now be possible to start experimenting in earnest here.

Negabinary:
Notes for future travellers:

wasm_of_ocaml is part of newer versions of js_of_ocaml - but our version of Bonsai requires an older version of js_of_ocaml

The newer version of Bonsai looks like it has only been released on OxCaml, not for straight OCaml.

cyrus-:
does js_of_ocaml work for oxcaml? we could switch over to that...

Negabinary:
currently asking an agent that question

Negabinary:
ok so the answer is - ReasonML doesn't work with OxCaml, because OxCaml has a different ADT

cyrus-:
We could make an OxReason with Claude... (or bite the bullet and convert Hazel to base OCaml notation). 

@dm0n3y any idea if they plan to release Bonsai for regular OCaml some time soon?

dm0n3y:
I'll ask

rickyvetter:
Hi folks! Bonsai continues to be released for both OxCaml and OCaml. The new api does rely on [local](https://oxcaml.org/documentation/stack-allocation/reference/), but only for compile time checks of correctness. We erase this mode annotation in the OCaml release. We use `local` so that we can have `Bonsai_web.Start.start` provide `graph @ local`, which allows us to guarantee that `graph` is never used outside of construction time. This means that the new continuation-passing api no longer requires use of `let%sub` and we find it quite a bit easier to teach people. You could move to this api and continue using OCaml without the local check, but your app will have a runtime exception if you attempt to use `graph` outside of the function where it is provided.

If you want to continue using the `Proc` api that Hazel's client currently uses (which is totally reasonable and we plan to continue to support), you can find it in the [latest release here](https://github.com/janestreet/bonsai_web/tree/master/web/proc). Use that module instead of the main `Bonsai_web` module.

[...]

PR #2490, opened 2026-09-04
SPIKE: compile the Hazel evaluator with wasm_of_ocaml and benchmark it

matthewhammer:
## What this is

A **spike**, not a merge candidate. It answers one question: is Hazel's
evaluator meaningfully faster compiled by `wasm_of_ocaml` than by
`js_of_ocaml`? The evaluator is compute-bound and already runs off the UI
thread in a Web Worker, which makes it the part of Hazel most likely to
benefit from Wasm GC — boundary-crossing costs, which is where Wasm usually
loses, barely apply there.

## The constraint that shaped the whole branch

`virtual_dom` v0.16 requires `js_of_ocaml < 6.0.0`. `wasm_of_ocaml` is only
released from 6.0.1 onward. **They cannot coexist in one opam switch.**

Everything else here follows from that:

- the benchmark links only `language` + `menhirParser`, never `haz3lcore`
  or `web`;
- parsing goes through `menhirParser` rather than the zipper path the CLI
  uses, since `PersistentZipper`/`MakeTerm` live in `haz3lcore`;
- `language` had to be made Bonsai-free, which is what the `util` split and
  the livelit removal are for.

## Changes

- **Split `util`** into `util` (pure + `js_of_ocaml` only) and `util_web`
  (Bonsai/Virtual_dom). `Util_web` re-exports all of `Util`, so most
  consumers only changed `open Util` to `open Util_web`. `JsUtil` and `Key`
  were each *split* rather than moved, keeping their `js_of_ocaml`-only
  majority in `util` — that alone avoided ~48 files of churn.
- **Dropped livelits from the language core.** The three builtin livelits
  construct vdom directly and were the last thing holding `Virtual_dom`.
  `LivelitCtx` keeps its types so `Ctx.LivelitEntry` still typechecks, but
  nothing populates it and `LivelitProj` renders a placeholder.
- **Added `bench/wasm/`**: the benchmark, a switch-setup script, and a
  runner that builds both backends from the same compiler version.
- Bumped `(lang dune 3.16)` to `3.17` for `(modes wasm)`.

Resulting dependency closure is
`language → util → {ptmap, bignum, uuidm, unionFind, csv}` — verifiably
Bonsai-free.

## What it gives up

The livelit deletion is the part that must not merge. The principled version
registers livelit views from the web layer rather than deleting them; the
`util` split, by contrast, is worth keeping on its own merits and is the
same prerequisite an Internet Computer port would need.

Two livelit tests were removed (`^slider(50)` elaboration, `^emotion(50)`
statics-map ids) because they assert behaviour the builtins provided. The
`^^slider` tests exercise the projector, a separate mechanism, and are
untouched.

## Verification

| Check | Result |
|---|---|
| `dune build @check` (dev) | clean |
| `dune build src` (release, `-warn-error +A`) | clean |
| `@test-quick` | 3457 tests, 0 failures |
| benchmark runs end-to-end on node 22 | yes |

## Numbers

Not yet — the comparison needs the second opam switch, which is still being
built. A `js_of_ocaml` 5.9.1 smoke run gives 607 ms plain / 596 ms incr on
`bench/eval-bench.hz`, but that is *not* the baseline: `run.sh` rebuilds
both backends from the same 6.x compiler so the delta is attributable to the
backend rather than the toolchain. Results to follow in a comment.

Caveat worth stating up front: `eval-bench.hz` is a single workload. Under
`wasm_of_ocaml` the OCaml heap is managed by the browser's collector rather
than OCaml's, so allocation-heavy and allocation-light programs may not move
in the same direction — one number is a signal, not a characterization.

## Running it

```
bench/wasm/setup-switch.sh
git worktree add --detach ../hazel-wasm-bench wasm-eval-bench-spike
cd ../hazel-wasm-bench && bench/wasm/run.sh hazel-wasm 20
```

Node 22+ is required; Wasm GC postdates the node 18 this repo otherwise
builds against.


matthewhammer:
## Result: blocked on missing C stubs, not on the evaluator

The Wasm module **builds and loads**, then fails at link with **53 unimplemented primitives**:

| group | count | what it is |
|---|---|---|
| `ml_z_*` | 27 | zarith (arbitrary-precision integers) |
| `Base_*` | 19 | Jane Street Base: hashing, int-math intrinsics |
| `bin_prot_*` | 2 | Bin_prot blit stubs |
| `expect_test_collector_*` | 2 | Jane Street test collector |
| other | 3 | `caml_csel_value`, `caml_out_channel_pos_fd`, `time_now_*` |

Root cause: Hazel's `Int` and `Nat` are arbitrary-precision — `language` reaches `Bigint` from six files, including `BuiltinsBase.re` where the integer builtins are defined — so `bignum` is load-bearing for the *semantics*, not an incidental dependency that could be dropped to get a benchmark number. And `bignum` drags in zarith *plus* the whole Jane Street `Core`/`Base` C-stub surface. Under `js_of_ocaml` all of it is satisfied by hand-written JS runtimes (`zarith_stubs_js`, which this repo further patches — see `make setup-zarith`). Those do not port to `wasm_of_ocaml`.

Passing the existing JS runtimes to the wasm build via `(wasm_of_ocaml (javascript_files ...))` made essentially no difference (54 → 53). Dune accepts the field; whether the files are honoured at all is unconfirmed. The copy rules are left in because they're the right shape for a fix, not because they worked.

**No performance number was obtained, and none is obtainable without clearing this first.**

### Ways forward

1. Wasm-compatible runtimes for all 53 — most plausibly porting the existing JS stubs to wasm_of_ocaml's calling convention.
2. Build real zarith/GMP for Wasm — addresses the 27 `ml_z_*`, but not the 26 Jane Street ones.
3. **Drop `bignum` from the evaluator's cone**: a `Bigint` backend over JS `BigInt`. This removes zarith *and* Core in one move, and is the only option that also helps an Internet Computer port, where the problem is strictly worse — no JS host to fall back on at all.

Option 3 looks the most tractable, and is worth considering on its own merits regardless of Wasm.

### What is nonetheless usable here

- **The `util` split is real and green.** `language`'s dependency closure is now `util → {ptmap, bignum, uuidm, unionFind, csv}` — no Bonsai, no `virtual_dom`. Web-side directories uniformly `open Util_web` (which re-exports `Util`), so moving a module between the two libraries no longer causes call-site churn.
- **js_of_ocaml 5→6 portability debt is now mapped.** The two versions disagree in ways that bite: `Firebug` exists only in 5.x and `Console` only in 6.x; `setTimeout` changed from `(Js.callback, Js.number_t)` to `(Js.meth_callback, float)`; `scrollTop`/`Date`/`performance` return `Js.number_t` where they returned `float`/`int`. Surviving uses go through `Js.Unsafe` so one source compiles under both. This is what a future jsoo upgrade will cost.
- **`uuidm 0.9.8` transitively caps how new a js_of_ocaml Hazel can use** (via `topkg → cmdliner < 2.0`), which is why the switch pins 6.2.0 rather than 6.4.1.
- **A reproducible Bonsai-free switch recipe**, in `bench/wasm/setup-switch.sh`. Getting it to solve required the `archive` opam repo, pinning every shared package to what the main switch resolved to, and dropping `bisect_ppx`.

### Verification

| check | result |
|---|---|
| `dune build src bench --profile release` (ordinary switch) | clean |
| `@test-quick` | 3457 tests, 0 failures |
| `language` closure Bonsai-free | yes |
| Wasm module builds and loads | yes |
| Wasm module links | **no** — 53 missing primitives |

Still **not for merge**: the livelit deletion is a deliberate shortcut, and the principled version registers livelit views from the web layer instead.

matthewhammer:
## Update: ~4× faster, on fixed-precision workloads

Following the earlier bignum wall — the way through was to lean on Hazel's *fixed-precision* types. `Int`/`Nat` are `Bigint`-backed, but `SInt` (machine int) and `Float` are not, so workloads written in those avoid zarith entirely.

Both backends built from **js_of_ocaml 6.2.0**, node 22, 20 iterations. **Result checksums match across backends**, so this measures the same computation and not a silently-stuck one.

| workload | js_of_ocaml | wasm_of_ocaml | speedup |
|---|---|---|---|
| `sint-fib` (recursion) | 1050 ms | 285 ms | **3.7×** |
| `sint-fib` incr | 1053 ms | 276 ms | **3.8×** |
| `float-numeric` (alloc-light) | 290 ms | 74 ms | **3.9×** |
| `float-numeric` incr | 303 ms | 71 ms | **4.3×** |
| `sint-list` (alloc-heavy) | 168 ms | 42 ms | **4.0×** |
| `sint-list` incr | 162 ms | 41 ms | **3.9×** |

The differences *between* workloads are **not** resolvable from this data — across two runs the ordering flipped (`sint-list` lowest in one, highest in the other). The ~4× headline is solid; please don't read a profile effect into the breakdown.

### What getting bignum out of the way exposed

- `Statics.re` had **exactly one** use of Jane Street Core (`Core.Memo.general`). Replacing it with a local `Util.Memo` removed Core from the evaluator's cone entirely.
- `Sets`, `AST`, `Parser.mly` and `Lexer.mll` reached the `bignum` package directly; they now go through `util`'s `BigInt` wrapper, which is portable across both builds.
- The stub keeps construction/comparison total (`Builtins.ctx_init` needs them to boot) and makes **arithmetic raise**, so a workload reaching `Int`/`Nat` fails loudly rather than quietly benchmarking 63-bit math.

### Relationship to #1131

#1131 (*Port to wasm_of_ocaml*, open since Nov 2023) hit the same wall we did: Bonsai pins an older `js_of_ocaml`, and `wasm_of_ocaml` needs a newer one. That thread pursues **moving Bonsai forward** — newer Bonsai is OxCaml-first, ReasonML doesn't work with OxCaml, then bleeding-edge Bonsai with `basement` C-stub failures on Mac.

This PR takes the other route: **cut the evaluator out of the Bonsai cone and port only the compute core.** Bonsai never has to move. That's what the `util`/`util_web` split and the livelit removal buy.

Three findings from here that may be useful over there:

1. **Bonsai is not the last blocker, and zarith is worse.** Behind Bonsai sit 27 `ml_z_*` primitives plus the Core C-stub surface. A whole-app port solves Bonsai and then hits this immediately; an evaluator-only port only has to solve this one.
2. **`uuidm 0.9.8 → topkg → cmdliner < 2.0` caps `js_of_ocaml` at 6.2.x**, independently of Bonsai.
3. **Fixed-precision Hazel already runs in Wasm, at ~4×.** That number exists today without resolving either blocker.

### Arbitrary precision: the remaining work

Full-fidelity Hazel in Wasm still needs a bignum. The most promising route is the one Motoko took — a bignum compiled to Wasm, linked against the compiler output. The wrinkle is that `wasm_of_ocaml` represents OCaml values as **Wasm GC** structs while a C bignum lives in **linear memory**; unlike Motoko, which controlled both sides, this needs a copy-in/copy-out shim at the boundary. `wasm_of_ocaml`'s `.wat` runtime hook looks like the seam to try.

### Correction to an earlier note

I previously suggested the evaluator lacked TCO. That was wrong. [`Trampoline.re`](../blob/wasm-eval-bench-spike/src/language/dynamics/evaluation/Trampoline.re) is an explicit stack machine — object-language call depth lives in a heap-allocated callstack — and measured under js_of_ocaml, both tail *and* non-tail Hazel recursion reach depth 20000+. Deep recursion is not a backend risk. The stack overflow that prompted the claim was a mis-typed workload (an `SInt` counter inside `use Float in`), not a depth limit.

### Verification

`dune build src bench --profile release` clean on the ordinary switch; `@test-quick` 3457 tests passing. Still **not for merge** — the livelit deletion remains a deliberate shortcut.

matthewhammer:
## Revised: settings axis added, and a caveat on what the number means

A fair challenge was raised on the numbers above: benchmarking headless Hazel may mislead, because a lot of what makes Hazel slow is the extra per-expression information statics and evaluation record for the UI. That's right, and it prompted two changes plus a sharper claim.

Two questions were tangled together, and they have different answers.

### Is the comparison between backends fair? Yes.

This is what the harness is built to guarantee: **one source tree, one settings record, one opam switch, one compiler version** — the only thing that differs is the backend flag. The result checksums confirm both sides computed the same answer. A lean path doesn't threaten this; it just has to be the *same* lean path on both sides, and it is.

### Does ~3.5× predict how much faster the app would feel? No.

This measures the evaluator alone — no zipper, no vdom, no editor. End-to-end speedup would be diluted by whatever share of real time is spent outside this path. Rather than assume that away, two probes were added:

**A settings axis.** `probe_all` governs how much per-expression information statics and evaluation record for the UI. It is **off even in `CoreSettings.on`**, so the earlier numbers measured a leaner path than the editor actually runs. Both profiles are now reported — and they are **indistinguishable on these workloads**. So the probe machinery is not what's slow here. That doesn't refute the concern; it narrows where the cost can be hiding.

**A cold statics sample.** `Statics.mk` is memoized (via `Util.Memo`), so it can only be timed cold — a loop measures cache hits, which is what a first attempt did, producing meaningless 6–12× "speedups" on ~0.0 ms. The harness now emits `statics-cold-1shot`, but it is a single unrepeated sample visibly dominated by JIT warm-up: the *same program* measured 16.6 ms under one profile and 3.2 ms under another. **No conclusions should be drawn from it.** Qualitatively it supports only that statics is single-digit ms here while evaluation is hundreds.

### Updated results

Both backends from js_of_ocaml 6.2.0, node 22, 20 iterations, checksums matching.

| workload | profile | js_of_ocaml | wasm_of_ocaml | speedup |
|---|---|---|---|---|
| `sint-fib` plain | probes-off | 1020 ms | 341 ms | 3.0× |
| `sint-fib` plain | probes-on | 1016 ms | 273 ms | 3.7× |
| `float-numeric` plain | probes-off | 269 ms | 78 ms | 3.5× |
| `float-numeric` plain | probes-on | 277 ms | 75 ms | 3.7× |
| `sint-list` plain | probes-off | 151 ms | 43 ms | 3.5× |
| `sint-list` plain | probes-on | 174 ms | 42 ms | 4.1× |

Range across runs is **3.0×–4.1×**; call it ~3.5×. Between-workload differences are not resolvable — the ordering flips run to run.

### The limitation worth stating plainly

These workloads are **compute-heavy programs that are textually tiny** — the opposite shape from the editor, where programs are large and each evaluation is short. That shape is precisely why evaluation dominates statics here.

So the defensible claim is **"the evaluator is ~3.5× faster under wasm_of_ocaml"**, not "Hazel is ~3.5× faster". Establishing the latter needs a benchmark shaped like real editing — large programs, short evaluations, statics on the hot path — which this harness deliberately isn't. Worth building separately if the app-level number is what matters for the decision.

matthewhammer:

PR #2492, opened 2026-09-04
Fumola livelits: running an incremental Fumola runtime from Hazel

matthewhammer:
An experimental integration between Hazel and **Fumola**, a language whose execution model uses Adapton / demanded computation graphs, so that editing and re-running a program can reuse prior computation rather than starting over.

**Try it:** `hazel.org/build/fumola-livelit-mvp`, or `make && make serve`. No setup — the Fumola runtime is fetched from its published build. Type `^fumola_thunk` or `^fumola_editor` then space.

**Depends on** [Adapton/fumola#45](https://github.com/Adapton/fumola/pull/45) and [#46](https://github.com/Adapton/fumola/pull/46), both merged. Nothing binary is tracked here.

# Purpose of the MVP

Hazel and Fumola are each exploring live, responsive evaluation, and each relates itself to spreadsheets. Hazel's livelits are one such semantics; Fumola's demanded computation graph, where editing a program repairs prior computation rather than redoing it, is another.

This branch puts the second one *inside* the first: a Fumola runtime becomes something a Hazel program can hold and name, so two live evaluation models sit in one document, side by side, where they can be compared and composed.

Two livelits are the ways of reaching across. A **thunk** livelit holds a program whose value you are editing: it is wrapped in a named thunk, so re-forcing that name on each edit is what lets Fumola repair rather than recompute. An **editor** livelit holds a program run at the top level, for setting things up and for the operations a thunk cannot host — writing cells that thunks will read, making bindings that outlive the program, and reaching Adapton's own introspection.

**Instance ids are what keep Fumolas apart.** An id names a runtime, so a document can hold several independent ones, and livelits sharing an id share a store, a set of bindings and a demanded computation graph. Isolation and communication are both just a question of which id you write.

That framing is what makes the rest of this interesting. Every livelit shipped today is self-contained: `^slider`'s model *is* its value. These are the first whose model holds a **name for state that lives outside Hazel**, so the interesting question is not the widget but what it means for a Hazel value to denote something a foreign runtime owns.

A livelit's model names an entry of an external store

```text
Σ : FumolaInstanceId -> FumolaRuntimeState
```

held by the Fumola wasm module and reached through the `window.fumola` shim. The Fumola interpreter and its demanded computation graph are deliberately *not* serialized into the model; only the name is. Three consequences, each implemented explicitly rather than inherited:

- **Editing preserves the name**, so an edit re-evaluates against the same live runtime and state written in one edit is there in the next.
- **Expansion is an observation of Σ**, never a second source of truth in the model.
- **The runtime does the work.** There is no repair API in Fumola and none was needed: a thunk livelit wraps its program as ``force(`name := thunk { ... })``, and re-assigning that name and forcing it again is what makes the edit incremental. `:=` and `thunk` do the rest.

## Two livelits and a value

| | |
| --- | --- |
| `^fumola_thunk(instance, name, program)` | wraps its program in a named thunk; incremental across edits |
| `^fumola_editor(instance, program)` | evaluates at the top level; no reuse, but bindings outlive it |
| `FumolaPeek` | not a livelit — what a returned *pointer* becomes |

`^fumola_editor` is strictly more expressive: you can write the ``force(... := thunk { ... })`` yourself. What the thunk livelit adds is generating that wrapper with a distinct name. The editor is also the only way to reach what the wrapper forbids — `Adapton.reset()` clears the store the enclosing force is still inside, and `Adapton.peekForce` asserts there.

The **thunk name** is Fumola source for a symbol, parsed by Fumola itself, so ``​`myThunk``, `7` and ``​`a(`b)`` all work and Hazel needs to know about none of them. It is written rather than derived on purpose: a name taken from a Hazel id would start a *new* thunk whenever that id changed, quietly losing the history the thunk exists to keep.

Both name an instance, so **two livelits carrying the same id share one runtime**.

# Simple programs to try

Both livelits require an annotation — they only expand in checking mode, since the result's shape depends on the program.

**State surviving an edit.** Run this, then edit the program to just `get(1)`; it still reads `2`.

```
let n : Int = ^fumola_thunk(0, "`a", "1 := 2; get(1)") in n
```

**The worked example**, also in Documentation → Livelits → builtins. Three livelits over one runtime:

```
let x  : Int = ^fumola_thunk(0, "`t", "`x := 137 + 137; 137") in
let x_ : Int = ^fumola_editor(0, "pointer(`x)") in
let y  : Int = ^fumola_editor(0, "force(pointer(`t))") in
(x, x_, y)                    ->  (137, peek(`x)! = 274, 137)
```

The thunk writes a cell and returns 137. The first editor asks for a pointer to that cell, which arrives as a peek value showing **both the reference and what it holds**, typing as `Int` so the ascription is satisfied and evaluation continues through it. The second editor forces the thunk the *first livelit* made, by its name ``​`t`` — which is why explicit thunk names matter.

**Values that cross the boundary**, each verified against the runtime:

```
1 + 2                       ->  3
`counter := 41; get(`counter) + 1
                            ->  42
(1 := 10, 2 := 20)          ->  a pair of peek values
{x = 1; y = true}           ->  a Hazel record
peek(1)                     ->  Some(10)
peek(404)                   ->  None
`hello                      ->  "hello"
Gcd.gcd(12, 18)             ->  6     (module library, no import)
```

**Things needing `^fumola_editor`**, since they cannot run inside a thunk:

```
Adapton.reset()             ->  ()
Adapton.peekEvents()        ->  a list of records carrying events, spaces and times
```

**Always in scope:** `pointer(s)`, `get(s)`, `peek(s)`. They exist because Hazel string literals admit no escapes — `Token.is_string` permits at most two quote characters in a token — so a livelit's program can never contain a `"`, and every adapton primitive needs one. Note that `:=` coerces its left side into a pointer but `@` does not, so a cell is read back with `get(s)`.

**The module library is compiled into the runtime** and bound at the top level: `Adapton`, `List`, `LazyList`, `PureList`, `LevelTree`, `HashMap`, `Counters`, `RandomInput`, `MergeSort`, `Gcd`, `DelayedPut`. Anything else is reachable by path.

# Future work to explore

**Expansions are not typed.** Statics splices a livelit's expansion in as an already-elaborated term and never traverses it ([Statics.re](src/language/statics/Statics.re)). Three things follow: a livelit nested in an expansion is never expanded, so `FumolaPeek` cannot render as a widget; constructors inside an expansion are never resolved; and no annotation anywhere in the program can reach inside one. This branch works around it with type-directed expansion, but making expansions real syntax is the change that would dissolve several problems at once.

**Sum types across the boundary.** Fumola tags are structural and live in one flat namespace: `#Bar(3)` is a value whether or not anyone declared `Bar`. Hazel constructors are declaration-bound. So the boundary is lossy in the direction that matters — Fumola hands you a tag, Hazel needs a tag *plus* a home type. Expansion now resolves tags against the expected type, but round-tripping Hazel → Fumola → Hazel can still land in a different sum type.

**`Space` and `Time` both declare a `Symbol` constructor.** With an expected type it resolves correctly; without one it falls back to a context lookup that could pick either.

**Generative copy is not implemented.** A copied livelit inherits its original's instance id and shares its execution history. Doing it properly needs a projector identity that survives a model edit, which this branch does not establish.

**Instance lifetime across sessions.** A saved program naming instance 17 in a session with no `Σ(17)` realizes a fresh runtime, which then has no history. What *should* happen on reload is unsettled.

**The peek snapshot is load-bearing.** `FumolaPeek` carries a value that is sound only because it is regenerated on every expansion. Persisting one into a saved model would silently disagree with the runtime.

**Hazel has no parameterized types.** There is no type-level application, so `Option` is monomorphic with payload `?`. That is why `Option(SomeThing)` cannot be written, and why an option's payload type is lost on the way across.

**Pre-existing livelit issues found along the way**, two fixed here and two not — see the section below.

**Smaller ones.** Instance creation costs ~250ms and the wasm is 5.5MB, both from compiling in the module library; `Dot` symbols have no Fumola parser support; and `Adapton.peekForce` still fails inside a thunk, so introspection wants the editor.

# Pre-existing issues found along the way

Four problems in the livelit machinery with nothing to do with Fumola. Two were in the way and are fixed here; two are recorded but untouched.

**Fixed — clicking any livelit crashed.** `JsUtil.get_elem_by_id` ends in `assert(false)`, and `LivelitProj.focus_pointer` calls it on every pointer click — but the projector attaches its id only where `get_model` succeeds, so the `No livelit found` fallback asserted. Confirmed against deployed `dev`: `^slider` and `^emotion` crash on click there today.

**Fixed — a livelit with a text field could not be typed into.** Clicking the field focused it, then `Focus` refocused the projector's wrapper `<div>`, which has no `tabindex`; focusing an unfocusable element drops focus to `<body>`. So the field lost focus the instant it was clicked, and the only way to edit was to unproject the livelit. **This affects `^js` too.**

**Not fixed — livelits do not render in the Documentation slides.** All of them show `No livelit found`; verified on deployed `dev`. This is also why the two above went unnoticed: the widgets are invisible in the one place the docs point you at.

**Not fixed — livelits have no keyboard handoff.** `focusable.keyboard = None`, so arrow-key navigation into and out of one does not work. `TextAreaProj` is the model to follow: it puts the projector id directly on its textarea, which would subsume the focus fix above.

# Reviewer notes

- **The livelit interface changed**: `view` now takes `~id`, and `expand` takes `~id`, `~ana` and a `type_tools` record. Closures rather than a `Ctx.t`, because `Ctx` already depends on `LivelitCtx` and naming it there would be circular. The other three livelits ignore the new arguments.
- **One new `exp_term` constructor**, `FumolaPeek`, which is the only change under `term/`. The sweep touched ~19 files; the parts worth reading are dynamics, statics, unboxing, equality and `ExpToSegment`.
- **Nothing binary is tracked.** The runtime is published to GitHub Pages from the Fumola repo, deliberately rather than as a release asset: release downloads carry no `Access-Control-Allow-Origin` header, so a browser can never fetch one. An earlier revision committed the wasm into `src/web/www`, which put it in the deployed preview directory and made Pages fail to build the entire `hazelgrove/build` site.
- **A red `Full tests` run on this branch is probably #2499, not this PR.** `Evaluator.Properties` checks randomly generated programs, and a different sample is drawn each run, so a case that fails shows up as intermittent. It failed once here on `05be2f0276` -- a commit whose full suite passes locally, on a branch that does not touch the evaluator. The shrunk counterexample and seed in #2499 suggest a real disagreement between incremental and fresh evaluation rather than harness noise.


matthewhammer:

Issue #2243, opened 2026-04-30
Modules don't export datatype constructors


cyrus-:
<img width="307" height="214" alt="Image" src="https://github.com/user-attachments/assets/3fec4390-c5a4-41a6-a45e-ed034d3ac7d7" />

Issue #2242, opened 2026-04-30
module signatures should not allow complex patterns


cyrus-:
<img width="215" height="83" alt="Image" src="https://github.com/user-attachments/assets/2be693c5-fd49-4ede-bbc8-bcbde7db7810" />

Should only support variables and type-annotated variables, not other patterns.

PR #2493, opened 2026-09-05
Modules II, part 1: first-class signature types and module values

### LINKED ISSUE #2295: function definition shorthand not exporting correctly from modules
Image

### LINKED ISSUE #2467: Funlet-style module members don't statics-bind (member and params unbound)
In a module body, a funlet-sugar member leaves both the member name and its parameters unbound in statics:

```
module M = {
  let f(x) = x + 1;
  let g : Int -> Int = fun y -> f(y)
} in
M.g(1)
```

`x` in the body of `f` and the reference to `f` in `g` mark as unbound. The same definitions written fun-style (`let f = fun x -> x + 1;`) bind fine, and top-level funlets (outside modules) also bind fine — the gap is specific to the module-member statics path.

Found while generating large module-based test programs (hazel-programs/mega on the modular-editors branch encodes the fun-style workaround throughout).

7h3kk1d:
Modules become a real part of the language rather than sugar over labeled tuples. A module is a block of items, `{ let x = 1; type T = Int }`, and evaluates to a module value whose members are read with `m.x`. Its type is a signature, `{ let x : Int; type T = Int }`: a type in its own right, usable wherever a type goes, whose type members are types elsewhere (`M.T`, `M.Inner.U`). Sub-modules, `module Inner = …`, nest in both. Annotating a module with a signature checks each member against it, and a missing, extra or unknown member or a mismatched type member is reported at the item concerned. A hole among the items stands in for members still to come, completions offer a module's members, and `let f(x) = …` in a module body defines `f`. Sealing away extra members and abstract type members follow in the next two parts.

Bottom of stack #2496 (~~width subtyping~~ sealing at coercion sites, abstract type members, module-typed functions and the rest follow). This one replaces the labeled-tuple gloss described in `docs/modules.md` with real module statics and dynamics; the next two add ~~width subtyping~~ sealing and abstract type members on top.

## Statics

- `Sig` is a distinct, first-class type. `Typ.desugar_sig` is gone and a signature is inconsistent with a labeled tuple type in both directions. Signatures are dependent records: `free_vars`, `subst`, `normalize` and `meet` treat type members as sequential binders, and `meet` requires the same member names (exact; ~~width arrives in the next PR~~ sealing at coercion sites arrives in the next PR). `Sig.re` exposes the member view (`Val`, `TypeManifest`). A substitution whose type variable shares a name with a type member of the signature is reported (`TypeMemberCapture`, below) rather than carried out: a member name is a label and cannot be renamed away.
- A module synthesizes its signature in source order (`ModuleHelpers.module_sig_type`). Against an annotated signature, manifest type members are checked (~~`ModuleTypeMemberMismatch` on the `type` item~~ `ModuleTypeMemberMismatch` on the member's definition type, so the editor highlights `Bool` in `type T = Bool` rather than the whole item) and substituted into member expectations; missing members mark the module (`ModuleMissingMembers`), unless one of the items is a hole where a member could still be bound, in which case the missing members are assumed (see below).
- `M.T` resolves through the module variable's signature (`Typ.path_sig`, including `M.P.T`, module aliasing and signature aliases). The `TVarEntry` injection in the `Let` case is removed. `Dot` projects signature members. A type path rooted at a value of unknown type (`let n : ? = … in let y : n.T = …`) is accepted, since `n` may be a module; a root of a known non-signature type is still `TypWantModule`.
- `module M : S` is a signature item too (`SigModule`), so capitalized sub-modules can be declared inline. Module items report the member they declare in the cursor inspector.

## Dynamics

- Modules evaluate as modules. `Transition.re` steps `Module(items)` item by item with sequential scoping; an evaluated binding becomes the dynamics-only item `ModVal(x, v)`, and a module of `ModVal`s is a value. `Dot` reads bindings by name; ascribing a module value to a signature seals it. New `EvalCtx.ModuleItem` / `ModuleVal` contexts.
- Statics still lowers a module body to a `Let`/`TyAlias` chain for type checking (reusing the `Let` and pattern machinery); elaboration refolds the checked chain back into a `Module`.

## Also

TyDi qualified completions read signature members; the cursor inspector and `ErrorPrint` render the new marks; `docs/modules.md` and the Modules doc slide are rewritten; tests migrated (module types are signatures, module values are modules). A skipped test pins a known gap: `{ let ? : ? }` should eventually match any member once the hole is filled.

## Fixes to `dev` code, carried at the bottom of this PR

Three bugs in code that is on `dev` today, surfaced while building this stack. They were briefly a separate PR (#2528, closed) and now sit as the first three commits here, each with a test that fails without its fix on `dev`.

- **Incremental eval: a binding's ascription is part of its provenance.** Editing a binding's ascription in place left its dependents reusing values computed under the old type: the reuse provenance recorded only *that* the binding went through an ascription, not which type. An ascription changes the bound value (a cast; here, sealing a module), so the type now travels in the provenance as `Ascribed(Typ.t)`, compared with `Typ.fast_equal`. Pinned by retyping `let x : Int = 1 in x` as `x : Bool` in place — the one ascription edit whose stale value is observable on `dev` (retyping to `?` or relabeling a tuple is not: casts are stripped by the comparison and a projection's index comes from the new static type). Part 2 keeps the signature form of the test, where dropping a member from `module Stack : { ... }` must invalidate the cached `Stack.default`.
- **Keep a printed `:` apart from a following operator.** `:` is one of `Token.ascii_operator_chars` and `Token.is_potential_operator` matches any run of them, so a `:` printed flush against a leading-plus sum lexed as the single token `:+`, which is no form at all; Menhir has its own lexer and split them, so the two parsers disagreed (the Slow `Menhir and maketerm are equivalent` fuzz property found it). `should_add_space` kept `:` apart from `$` and `!` as a list of two characters; it now asks `Token.begins_with_potential_operator` for the property they share (`$` stays listed: it is a name character). Pinned at the `should_add_space` level here and on `{ let _ : + Adid }` in part 4. The other half of the divergence is the editor's lexer, which still reads a typed `:+` as one token on `dev`; `Test_Menhir` carries the two programs that show it as skipped cases, next to their spaced forms, which pass.
- **Stop re-parenthesizing the content of a Parens type.** Printing a parenthesized min-precedence type was not idempotent: `type T = + Adid` came out as `type T = (+ Adid)` and each further trip added a layer. `parenthesize_typ`'s `Parens` case ran its content through `paren_typ_at(Precedence.min)`, which wraps exactly the types whose precedence *is* min — a bare sum, a multihole, a `Sig` — on every pass. The wrapper being emitted is already the protection, so the content needs none; `already_paren` is still passed down. Pinned by printing `type T = + Adid in ?` twice; part 4 keeps the `Sig` form.

## Known gaps

Module structure, roughly by how much each blocks real programs. None is a
regression; they are what this construct does not reach yet.

- **Constructors declared inside a module are not members of it** (#2243).
  Qualified access fails and the bare name leaks into the enclosing scope, so
  two modules declaring the same constructor make that name unusable. This
  breaks the ordinary idiom of a type and its operations in one module.
- **Signature items accept arbitrary patterns and then bind nothing** (#2242).
  A tuple-pattern member is accepted, declares no members, and member access
  on the binder then reports that the module has none.
- **No `include` and no `open`** (#2260, #2261). Reusing a signature or
  bringing members into scope is copy-paste, which makes larger signatures
  repetitive.
- **No module destructuring in patterns.** Only a name, optionally
  signature-annotated, can bind a module; there is no way to pull several
  members out at once.
- **No mutually recursive modules**, and modules are values inside one
  program rather than compilation units, so there is no import story.
- **Module equality is not rejected statically**, only incomparable at
  runtime.

## Verification

`make dev` clean · `make test-quick` green (~~3665~~ 3752 tests) · `dune build @check --profile release` clean · the editor checked by hand on the two slides.

## Follow-up commits

- **Module-aware projection errors, one error per differing type member.** `m.y` with no member `y` now marks the label (`ModuleMemberNotFound`, which also says when `y` is a type member used as a value) and gives the dot only a message, the design the type-level `M.Fake` already had. `M.T` with no type member `T`, a missing sub-module in `M.P.T`, and a value used as the root of a type path get module-aware marks and wording; the dot marks that said "tuple" say "module or tuple". A manifest type member that differs from the signature is reported once, ~~on its `type` item~~ on its definition type: members are checked against the module's own definition. Slide items 11a to 11d show the messages.
- **The function shorthand binds a member.** `let f(x) = ...` in a module body exported `x` rather than `f` (#2467, #2295): binders were read with `Pat.bound_vars`, which treats an applied pattern as a constructor application. The lowering reads a let item's binder through `FunctionSugar.binder` and annotates a shorthand at its name with the type its signature expects; the `Let` case of statics then desugars it exactly as it does a top-level shorthand.
- **A cyclic member path is a type error, not a crash.** A signature that names a same-named outer binding, `let m = { type T = Int } in let m : { type T = m.T } = ...`, has no weak head normal form: normalizing `m.T` asks `path_sig` what `m` denotes, finds the inner `m` (`path_sig` looks in the value namespace, new with module paths), whose signature defines `T` as `m.T`, and is back where it started. Running out of fuel raised `Failure("weak_head_normalize exceeded 1000 recursive calls")` and aborted the whole analysis. Out of fuel the type now comes back unreduced — stuck, which is what the function already returns for anything else it cannot reduce — so the member that fails to match it gets an ordinary `ModuleTypeMemberMismatch`; anything that reduced within the fuel budget reduces identically. Pinned at the `Typ` level and on both the `let m : S` and `module M : S` programs. (The cyclic *alias* form, `type x = x.a`, cannot reach this until part 4's `path_sig` catch-all arm; it is pinned there.)
- **Expectations reach sub-modules and destructured binders.** `modlet_pat` annotated only a bare variable binder with its expected member type. It now annotates every variable inside a destructuring pattern (`let (a, b) = …` is checked as `(a : Int, b : Int)`, so a wrong component is reported on that component) and `module Inner = …` items, which receive the sub-signature the outer signature declares for them, so a member the sub-module lacks, a differing type member or a wrong definition is reported inside the sub-module rather than as one mismatch on the outer module. A binder the user annotated keeps its annotation; the refold strips the synthetic ascriptions by walking the user's pattern and its elaboration together. Slide items 20a and 20b. The form of the kept-annotation test that needs ~~width subtyping~~ sealing lives in part 6.
- **The Modules doc slide is split: Modules (construction, qualified type access) and Module Signatures (signatures, signature errors).** Evaluating the combined slide needed about 750KB of stack and the Web Worker that evaluates programs has about half the main thread's stack, so the first edit on the slide overflowed it. Section letters and item numbers are unchanged. Later parts add Module Sealing, Module Abstract Types and Module Implicits the same way.
- **A type path may be rooted at a value of unknown type.** `let n : ? = 1 in let y : n.T = …` reported "n is a value of type ?, not a module"; a root of unknown type may be a module, so the path is accepted and its type is unknown.
- **A differing type member is marked on its definition type**, not the whole `type` item; the CLI and the cursor inspector share one message for the mark on the type side.
- **A hole among the items stands in for missing members.** A module checked against a signature reported every member it lacked even when one of its items was a hole that could still be filled with them: a hole item, or a hole in a binding position of a `let` pattern or module or type name. Such a module now assumes the missing members, so nothing is reported and the module has the signature's type. Only binding positions count (not labels, constructors, definitions or annotations), and a wildcard binder still reports. Slide item 20c.
- **Editing: an incomplete tile takes the form its shards spell in its sort.** Typing a second module item before the `;` of the first made it an expression `let` owed an `in`; adding the `;` then left it incomplete in the module body, with the `in` stuck in the backpack. When a tile's label has no form in the sort at its position and its present shards spell a complete compound form of that sort, remolding gives it that form, so the item completes as `let y = 2` (`type` and `module` items likewise). Compound forms only: a lone keyword shard must not become a variable, which the agent-tools reserved-keyword paste test guards.
- **A module wins over a constructor of the same name in dot position.** A capitalized module name parses as a constructor, and a known constructor took precedence, so a module named like one could not be projected at all; the JSON type contributes `List`, `Int`, `String`, `Float` and `Bool` to every program's constructor namespace, which made those five names unusable as modules. The `Dot` arm now reads a bare constructor as a variable when a module of that name is in scope, the rewrite the `Constructor` arm already did.
- **A member name is a label, so substitution shadows it.** The `Module` case of `Substitution.in_exp` already refused to rename the names bound by evaluated bindings and module declarations, but a pending `ModLet` went through `in_pat`, which renames a binder away from capture, so the member label changed and `M.x` stopped finding it. The stepper substitutes the whole builtin environment, so any module with a member named after a builtin lost that member.
- **Module step kinds are classified in `provenance_of_kind`.** Probes IV added it over `step_kind`; `ModuleBind`, `ModuleDiscardExp` and `ModuleDiscardType` are proper reduction steps, grouped with `LetBind` and `Seq`.
- **The #2457 pin is flipped.** `dev` pins #2457 with a program that expects incremental and fresh evaluation to disagree; from this branch they agree, so the assertion expects agreement. The QCheck property stays disabled.
- **A captured substitution into a signature is reported, not degraded** (spot 1.1 of #2544). Substituting a type into a signature whose manifest type member is named like one of the type's free variables is captured by that member, and member names cannot be alpha-renamed; `Typ.subst` degraded the later items to `?`, so `f@<T>` against `{ type T = Int; let v : a -> Int }` gave `v : ? -> Int` and `m.v(1)` was accepted. `TypAp` now substitutes the normalized argument, which removes the collision for aliases; a type variable that cannot be normalized away (a `typfun` parameter, or an alias definition leaving a `type … in`) gets a `TypeMemberCapture` mark, placed on the type variable's pattern for a `typfun`. Every statics substitution goes through `StaticsBase.subst_marked`, and `make ci-check` rejects a bare `Typ.subst` under statics. Pinned in `Typ.SubstCaptures` and `Test_Statics_Modules`; slide item 26 shows the message.
- **Exported bindings come from the pattern.** `Transition.re` sorted a match's bindings by their index among the pattern's variables, with a variable not found there silently placed first; the exported list is now the pattern's variables looked up in the match.
- **A shadowed member is checked against its signature at its last definition only.** The lowering annotated every binder of a name with the type the signature expects, so a member defined twice was checked twice, and a shadowed definition was marked although the exported one was fine. Only the binders an item exports are annotated; a definition a later item shadows is synthesized. Also pinned, both ways, that the last definition wins at its own position: `{ let x = 1; let y = 2; let x = 3 }` gives `{ y; x }` as the synthesized signature and as the evaluated module, the rule `Mod.add_modval` and `Sig.dedup_last` share.
- **Dynamics substitution reaches a signature's item types** (spot 1.5 of #2544). `Substitution.in_typ` returned a `Sig` untouched because the generic traversal would also visit the item patterns; it now maps the item types through `Sig.map_typ`. `ProofOf` is the only type position that carries an expression, so a `proof_of x` inside a member's type was the case left unsubstituted. Pinned in `Test_Substitution`.

Closes #2467. Closes #2295.

https://claude.ai/code/session_01UQDwH6tMjSMGSSykhjgQUP


7h3kk1d:
Logging a divergence this PR's syntax makes reachable, found while chasing a red `Full tests` on the branches above. It is a **dev bug**, not this PR's, so it is only pinned here rather than fixed.

An ascription immediately followed by a sum prints with no space, and the editor lexes `:+` as one operator token, so the two parsers disagree:

```
let x :+ To = To in x
-- menhir:   Asc(Wild/Var, Sum [Variant "To"])
-- MakeTerm: MultiHole [Pat Wild; Exp (Constructor "To")]
```

Spacing it as `let x : + To = To in x` parses the same both ways, and reprinting the spaced form is not stable either, so the printer's spacing is the trigger rather than either parser. Verified on `origin/dev` at 0943461829, the base of this stack, with no part of the stack applied.

Why it shows up here: nothing about the shape needs signatures, but a signature member is how the random program generator reaches it, so `Menhir and maketerm are equivalent` began failing per-seed once part 4's generator started emitting signature types. It reproduces on demand:

```
QCHECK_SEED=803422318 ./run_tests test 'Menhir*'
```

That seed passes on part 3 and fails on part 4 and above, and without a fixed seed it hits roughly one run in three on part 8. Shrunk counterexamples all look like `(()) @< ({ let _:+ To }) >`.

This PR now carries four cases in `Test_Menhir.re`: the two diverging programs, skipped with the reason, and their spaced forms, which pass and are what the printer should have produced. The fix belongs in the printer's spacing rule, or in the lexer's token rule, and should be a separate change against dev.

matthewhammer (commented):
Starting a review pass over parts 1–7 — none of the nine had a review submission, so this is a first read rather than a verdict. I've put the whole pass in **#2544** so it stays in one place; per-part detail and reasoning are in that issue's comments.

Since a lot of the stack is agent-generated and reading ~9,000 lines end-to-end isn't practical, I went looking for the places where a human's attention actually pays off rather than producing an exhaustive defect list.

**Part 1 spots** ([detail](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653039367)):

| | Spot | Status after running the composed tree |
|---|---|---|
| 1.1 | `Typ.subst` capture fallback substitutes `Unknown` into remaining items | open — design question |
| 1.2 | `path_sig` tries the type namespace before the value namespace | ✅ verified handled, with a purpose-built diagnostic — I withdraw this one |
| 1.3 | `weak_head_normalize` returns unreduced rather than `failwith` at 1000 | ✅ verified: a cyclic alias reports a type error, no crash |
| 1.4 | Runtime `add_modval` appends; static `dedup_last` preserves order | ✅ the obvious case agrees (statics and runtime both give last-wins) |
| 1.5 | Statics descends into `Sig`; dynamics `Substitution` no longer does | open — the comment describes a traversal limitation, not a semantic argument |

The one I'd most like your read on is **1.1**: on capture the substitution degrades the remaining items to `?`, which is consistent with everything and so never errors. Part 4 later argues explicitly against exactly this pattern (`Typ.Escaped` exists because "unknown is consistent with everything, so the loss was silent"), so it's worth deciding whether parts 1–3 should follow part 4's answer.

Nothing here blocks anything — this is the foundation the other six sit on, so it seemed the right place to spend the attention.

src/language/term/Typ.re
```diff
@@ -361,7 +393,29 @@ let rec subst = (s: t, x: TPat.t, ty: t): t => {
...
+         back to substituting Unknown into the remaining items. */
+      let fv_s = free_vars(s);
+      let rec go = (items: list(Sig.t)) =>
+        switch (items) {
+        | [] => []
+        | [item, ...rest] =>
+          let item' = Sig.map_typ(subst(s, x), item);
+          switch (Sig.member_of_item(item)) {
+          | Some(TypeManifest(n, _)) when n == str => [item', ...rest]
+          | Some(TypeManifest(n, _)) when List.mem(n, fv_s) => [
```
matthewhammer, line 408: **Spot 1.1** (#2544) — on capture this substitutes `Unknown` into every remaining item. `?` is consistent with everything, so the program keeps type-checking and nothing tells the author. Should capture here be an error instead? Part 4 later argues explicitly against this pattern (`Typ.Escaped` exists because "unknown is consistent with everything, so the loss was silent"), so it's worth settling whether parts 1-3 follow that answer.

src/language/term/Typ.re
```diff
@@ -470,59 +589,137 @@ let remove_duplicate_labels =
...
+   namespace. */
+and path_sig = (~rec_counter=0, ctx: Ctx.t, t: t): option(list(Sig.t)) =>
+  if (rec_counter > 1000) {
+    None;
+  } else {
+    switch (term_of(t)) {
+    | Parens(t)
+    | Projector(_, t) => path_sig(~rec_counter=rec_counter + 1, ctx, t)
+    | Var(n) =>
+      switch (Ctx.lookup_tvar(ctx, n)) {
```
matthewhammer, line 676: **Spot 1.2** (#2544) — type namespace before value namespace. ✅ **Verified handled**: I ran a module shadowed by a same-named type on the composed tree and got a purpose-built diagnostic ("a module with the name of a type cannot start a type path"), not a silent wrong answer. Withdrawing my concern; the only remaining question is whether that restriction is the one you want.

src/language/term/Typ.re
```diff
@@ -470,59 +589,137 @@ let remove_duplicate_labels =
...
 };
 
-let rec weak_head_normalize = (~rec_counter=0, ctx: Ctx.t, ty: t): t => {
+/* Out of fuel the type comes back unreduced -- stuck, which is what this
+   function already returns for anything else it cannot reduce -- so a type
+   with no weak head normal form is a type error where it is used, not a crash
+   of the whole analysis. A signature can name itself through a same-named
+   outer binding (`let m : { type T = m.T } = ...` with an `m` already in
+   scope), and path_sig hands `m.T` straight back here. */
+let rec weak_head_normalize = (~rec_counter=0, ctx: Ctx.t, ty: t): t =>
```
matthewhammer, line 598: **Spot 1.3** (#2544) — `failwith` became "return unreduced". ✅ **Verified**: `type x = x.a in let z : x = 1 in z` reports a type error rather than crashing or hanging, so the comment's argument holds in practice. Noting only that 1000 is now a silent cliff rather than a loud one.

src/language/term/Mod.re
```diff
@@ -41,10 +43,116 @@ let show_cls: cls => string =
...
+      | ModVal(x, d) when x == name => Some(d)
+      | _ => acc
+      },
+    None,
+    items,
+  );
+
+/* Record an evaluated binding, replacing an earlier binding of the same
+   name so a module value has one entry per exported name. */
+let add_modval = (items: list(t), name: Var.t, d: TermBase.Exp.t): list(t) =>
```
matthewhammer, line 149: **Spot 1.4** (#2544) — `add_modval` *removes* an earlier binding and *appends*, so shadowing moves a member to the end of the module. The static side (`Sig.dedup_last`) keeps the last of each name but preserves the survivors' original order. ✅ The obvious case agrees (`{let x = 1; let x = 2}` → `2`, no static error), but member order becomes load-bearing once part 2's width subtyping lands. See spot 7.2 — this is one of four places that independently decide which duplicate wins.

src/language/dynamics/transition/Transition.re
```diff
@@ -1311,11 +1340,115 @@ module Transition = (EV: EV_MODE) => {
...
+          let.wrap_closure _ = (env, rebuild(Mod.with_def(item, d1')));
+          let {matches, samples} = matches(targets, dp, d1');
+          switch (matches) {
+          | IndetMatch
+          | DoesNotMatch => Indet
+          | Matches(env') =>
+            /* Export the bindings in pattern order. */
+            let order = Pat.bound_vars(dp);
+            let index = x =>
+              List.find_index((==)(x), order) |> Option.value(~default=0);
```
matthewhammer, line 1412: **Spot 1.4** (cont.) — `Option.value(~default=0)` silently sorts any variable not found in `order` to position 0. Worth confirming that can't happen, or making it explicit.

src/language/dynamics/Substitution.re
```diff
@@ -223,8 +288,9 @@ and in_typ = (env: Environment.t(Exp.t), typ: Typ.t) =>
         | ProdProjection(_, _)
         | ProdExtension(_, _)
         | ProofOf(_)
-        | Sig(_)
         | DrvQuoteTy(_) => cont(t)
+        // Signature items carry patterns, which this traversal cannot visit
```
matthewhammer, line 292: **Spot 1.5** (#2544) — this now stops at `Sig(_)` entirely, where it previously fell through to `cont(t)`. Meanwhile the statics-side `Typ.subst` *does* descend into signatures (see spot 1.1). So a `Sig` type mentioning a substituted variable goes stale in the dynamics but not in the statics. The comment describes a traversal limitation rather than a semantic argument, which is the part worth chasing.

matthewhammer (commented):

7h3kk1d:
**Spot 1.1.** Thanks — this one was worth chasing. On the composed tree, `f@<T>` with `type T = Bool` against a signature declaring a type member also named `T` gave `v : ? -> Int`, so `m.v(1)` was accepted; rename the member to `U` and it correctly errors. Two changes, both in this PR (1593f640af):

- **Alias arguments.** `TypAp` already normalizes the type argument for `elab_term` (`Statics.re:1826`) but passed the raw one to `subst` a few lines later. Substituting the normalized type removes the collision for aliases — the member's name no longer affects the result, and the runtime, which used the normalized type all along, is unchanged.
- **Abstract type variables** — a `typfun` parameter, or an alias definition being substituted out of a `type … in` — can't be normalized away, so those now get a mark (`TypeMemberCapture`: *"Type variable X collides with a type member named X in this signature, so X cannot be mentioned inside it; rename the type variable"*). All three statics substitution sites go through one `StaticsBase.subst_marked`, and `make ci-check` rejects a bare `Typ.subst` under statics, so the `?` fallback in `Typ.subst` is only reachable after a mark has been reported.

Not done here: giving members an identity separate from their label (stamps on `Typ.Var`, or self-paths — part 3's `~self` threading and `strengthen` are the start of that). That's the same gap as #2545, so I'd rather address it in part 3.

Pinned in `Typ.SubstCaptures` and `Test_Statics_Modules`, checked on the CLI, and example 26 on the Module Signatures slide shows the error.

src/language/term/Typ.re
7h3kk1d: Checked this one further. The self-cycle is fine, as you saw — `type x = x.a` is `Rec`-guarded. A **mutual** cycle isn't, and on this branch it still crashes statics: `type y = x in type x = y in (1 : y)` dies with an uncaught stack overflow, and the same aliases reached through a type application (`f@<y>`) hit `normalize`'s `failwith` at 1000 (`Typ.re:758`), which this PR left in place.

That's #2570, a dev bug: dev fails the same programs with `weak_head_normalize`'s `failwith`, so it isn't from this PR — though the PR does change *how* the ascription case fails (stack overflow rather than the fuel message). "Return unreduced" doesn't resolve it either; the cycle has to be caught at `TyAlias` through other aliases, or a stuck `Var` handled wherever it's recursed on. Leaving it to #2570 rather than fixing it here.

src/language/term/Typ.re
7h3kk1d: Addressed; the write-up is in [this comment](https://github.com/hazelgrove/hazel/pull/2493#issuecomment-5681549503). After the restack the commits are `ad501836eb` and `2b79291935`.

In short: `TypAp` substitutes the normalized argument, which removes the collision for aliases. A type variable that can't be normalized away (a `typfun` parameter, or an alias definition leaving a `type … in`) gets a `TypeMemberCapture` mark instead of a silent `?`. All statics substitution goes through `StaticsBase.subst_marked`, and `make ci-check` rejects a bare `Typ.subst` under `src/language/statics`, so the `?` fallback in `Typ.subst` is only reached after the mark has been reported. Pinned in `Typ.SubstCaptures` and `Test_Statics_Modules`; example 26 on the Module Signatures slide shows the message.

Deferred: giving members an identity separate from their label. Same gap as #2545, so it goes with part 3.

src/language/term/Typ.re
7h3kk1d: Keeping the type-namespace-first precedence in this part. https://github.com/hazelgrove/hazel/pull/2506 is moving capitalized identifiers toward most-recent-binding resolution, and that is where the namespace question should be settled once, so I'll revisit this after it merges rather than change the rule here.

src/language/dynamics/transition/Transition.re
7h3kk1d: Removed in `a44525861b`. Every binding a match produces is bound by the pattern, so the fallback was unreachable. The exported list is now built from `Pat.bound_vars(dp)` directly, looking each variable up in the match, so there is no index and no default.

src/language/term/Mod.re
7h3kk1d: The two orders agree: `add_modval` drops the earlier binding and appends, so the survivor sits at its last definition's position, which is also where `Sig.dedup_last` leaves it. Pinned both ways in `2b4d3f8ffe`: `{ let x = 1; let y = 2; let x = 3 }` evaluates to `{ let y = 2; let x = 3 }` (`Test_Evaluator_Modules`, "Shadowed binding is exported at the position of its last definition") and synthesizes `{ let y : Int; let x : Int }` (`Test_Statics_Modules`, "A shadowed member's type is at the position of its last definition"). Flipping either expectation fails, so the tests are order-sensitive. `Mod.add_modval`'s comment now names `Sig.dedup_last` as the rule it follows.

Chasing this also turned up a related mistake in the analyzed case: `ModuleHelpers.lower` annotated every definition of a shadowed member with the signature's type, so `module M : { let x : Bool } = { let x = 1; let x = true }` reported the shadowed `let x = 1`. Same commit: a member is checked against the signature at its last definition only, pinned in `Test_Statics_Modules` and stated in `docs/modules.md`.

The single documented rule for the four sites (7.2) is still on part 7.

src/language/dynamics/Substitution.re
7h3kk1d: Changed: the arm now maps each item's type through `Sig.map_typ(in_typ(env))` and rebuilds the `Sig`, so an expression a member's type mentions is substituted the same way it is in any other type. The generic continuation was skipped because it would also visit the item patterns, which this traversal refuses; `Sig.map_typ` reaches only the types, so that reason is gone along with the comment. Pinned in `Test_Substitution`: `1 : { let p : proof_of x end }` under `x ↦ 42` comes back with `proof_of 42`, and the case fails against the old arm.

For the record on what was reachable: this traversal substitutes expression variables into the expression-bearing positions of a type, and `ProofOf` is the only such position; a type path like `M.T` is a `Var` resolved through the context, so it was never touched by either version. `Typ.subst` on the statics side substitutes type variables, a different operation, which is why the two never had to agree on more than the traversal itself.

hazel-programs/docs/reference/module-signatures.hz
```diff
@@ -0,0 +1,114 @@
+# MODULE SIGNATURES #
```
disconcision, line 1: wondering why have this as a separate slide

src/haz3lcore/tiles/Segment.re
```diff
@@ -111,8 +111,31 @@ let rec remold = (~shape=Nib.Shape.concave(), seg: t, s: Sort.t) =>
   }
 and remold_tile = (s: Sort.t, shape, t: Tile.t): option(Tile.t) => {
   open OptUtil.Syntax;
-  let+ remolded =
+  /* A label with no form in this sort: an incomplete tile whose present
```
disconcision, line 114: curious if this addresses #2575 and other related module syntax issues... putting an agent on it

disconcision (approved):
I just did a brief architectural sanity check; I did not look deeply at the semantics. I think the design looks solid and favor getting this in soon to avoid the existing tuple gloss becoming load bearing in weird ways.

hazel-programs/docs/reference/module-signatures.hz
7h3kk1d: @disconcision I got some stack overflow issues when it was one large slide. I think related to your similar too many lets stack overflow. Didn't necessarily happen from just having the slide but while I was editing it it was happening.

PR #2494, opened 2026-09-05
Modules II, part 2: sealing at explicit coercion sites

7h3kk1d:
Second of the stack, on top of #2493. Where a module is ascribed a signature, bound with one, or passed to a function expecting one, it may have more members than the signature declares; the extra members are sealed away. Everywhere else signatures match exactly.

> **Revised 2026-09-11.** The first version of this PR checked width subtyping at every analysis position. After discussing it with Cyrus it now seals only at explicit coercion sites; the struck lines below are the claims that changed.
>
> ~~Where an expression is analyzed against a signature, the module may have more members than the signature declares. Only at **analysis positions**: places where the program states the type an expression must have.~~ Replacement: only at coercion sites (an ascription, an annotated binder, an application argument), found through parentheses, tuple-literal components and a function literal's body.
> ~~Annotated pattern: `fun (m : { let x : Int }) -> …` binds `m` at the narrower type.~~ Replacement: patterns are exact; the function's argument is sealed at the application instead.
> ~~Function domains, contravariantly: a `{ let x : Int } -> Int` fits where a `{ let x : Int; let y : Int } -> Int` is required.~~ Replacement: function types are not contravariant; eta-expand.
> ~~The check is `Typ.ana_meet(ctx, ~ana, ~syn)`.~~ Replacement: `Typ.coercion(ctx, ~from, ~to_)`.
> ~~Known gap: analysis-position subtyping has no list arm.~~ Replacement: a wider module as a list element is a mismatch by design, with a message that says an ascription seals it.

## Motivation

Part 1 made modules first-class values whose types are signatures, with an exact relation between them: `{ let x = 1; let y = 2 }` had type `{ let x : Int; let y : Int }` and nothing else. That rules out the things modules exist for: interfaces (`fun (m : { let name : String }) -> …` accepting any module with a `name`), information hiding (a `module Stack : S = { let default = 0; … }` keeping `default` private), and evolution (adding a member without breaking code written against the signature).

This part adds exactly the coercion ML has, at exactly the places ML has it.

### Where it applies

| Site | Example |
| --- | --- |
| Ascription | `({ let x = 1; let y = 2 } : { let x : Int })` |
| Annotated binder | `module M : { let x : Int } = { let x = 1; let y = 2 }`, `let m : S = wide`, a module item under a signature |
| Application argument | `greet({ let name = "Alice"; let age = 30 })`, `greet(Alice)`, `f(wide, 1)` |

The operand of a site is found through parentheses (and the other transparent forms: projectors, filters, closures), through the components of a tuple literal, since an n-ary application argument is a tuple, and through a function literal's body: under an expected arrow type the body is sealed to the codomain, as a functor body is to its result signature, so `let f : S -> S = fun m -> m` checks; the parameter is matched exactly. Sub-module members may themselves be wider than declared, and the coercion is structural through tuples, so a variable of tuple type is coerced componentwise at a site too.

The members the signature omits are sealed away: the binder's type is the signature as written, `m.y` on an omitted member is an error, and at runtime the module value keeps only the signature's members (`Ascriptions.re`).

### Everywhere else, signatures match exactly

A wider module as a list element, in an `if` or `case` branch, under a pattern annotation, or reached through a variable of list or function type is an `ExpectationMismatch`, and the message says so: "Expecting type S but got the wider type T; an ascription seals the extra members". Function types are not contravariant; eta-expand instead. Consistency (`Typ.meet`) is unchanged and remains the relation for joins, `==`, `is_consistent` and incremental-evaluation reuse.

### Why coercion sites rather than subsumption everywhere

The first version checked the directional relation at every analysis position, from the subsumption step. Cyrus pointed out that general width subtyping under bidirectional gradual typing belongs in one subsumption rule with a relation total over the type grammar, and the half-way point was neither: the relation had only signature and arrow arms, so `(1, m) : (Int, S)` and `[m] : [S]` were rejected while `f(m)` was accepted, and `if`/`case` joins rejected sealed branches. Rather than grow the relation and change the joins, this part takes the ML design: coercion at named sites, exact consistency elsewhere.

- **OCaml**: signature matching happens at `module M : S = …`, at functor application, at `(module M : S)`, and a functor body is sealed to its result signature; nowhere else is there subtyping between module types, and module-typed function types are not contravariant.
- **1ML** (Rossberg, ICFP 2015, Fig. 4): the subtyping judgment with a coercion appears in exactly three expression rules, EAPP (`E : T` is application sugar), ESEAL, and EIF, whose conditional is annotated in the core syntax so each branch is checked against the annotation with no join. There is no standalone subsumption rule. This part has the first two; forwarding into `if`/`case` branches is a follow-up that depends on a fix to how those joins report in analysis mode.

## What changed

- `Typ.coercion(ctx, ~from, ~to_)` (was `ana_meet`): coercive subtyping, consistency or, structurally through tuple components, a signature with more members than declared (`Typ.sig_sub`). The arrow arm is gone.
- `StaticsBase.subsume(~coercible, …)` picks `Typ.coercion` or `Typ.meet` for `fixed_typ`, `expectation_mismatch_mark` and `syn_ana_ok_common`; the flipped `_pat` variants are gone and patterns are exact.
- `Statics.uexp_to_info_map` and its `go` take `~coercible=false`. `Asc`, `Let` (all analyses of the definition), `ModuleExp` and `Ap` set it on their operand; `Parens`, `Projector`, `DynamicErrorHole`, `Closure`, `Filter`, `TupLabel`, tuple components and a `Fun`'s body pass it on; nothing else does. A sealed function body carries the sealing cast and the function's codomain is the body's checked type.
- `ErrorPrint` and the cursor inspector say "wider type … an ascription seals the extra members" when the coercion would have succeeded.
- `ModuleExtraMembers` (part 1) stays retired; hole-named signature members are not required members.
- Dynamics: the last binding step yields the module value marked final (re-traversal double-sampled probed literals). Incremental evaluation: a binding's ascription type is part of its reuse provenance (`Ascribed(Typ.t)`).
- Doc slide: section E becomes its own Module Sealing slide (evaluating the combined slide overflowed the Web Worker's stack); its contravariance item is replaced by the list example with its ascribed fix.

Tests: the sealing group in `Test_Statics_Modules.re` (sites, depth, hidden helpers, the `if` boundary with the mark on the branch, the sealed function body) plus a non-sites group (list element, module literal in a list, tuple argument, annotated tuple literal, labeled component, tuple variable, case pattern, ascribed scrutinee, parentheses, let body, `if` branches, constructor argument, sum variable, no contravariance); `Typ.Coercion` in `Test_Typ.re`; runtime sealing and probes unchanged.

## Verification

`make test-quick` green (3718 tests) · `dune build @check --profile release` clean · CLI probes for each row above.


matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**. Per-part detail: [comment](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653047619).

**Part 2 spots:**

| | Spot | Status |
|---|---|---|
| 2.1 | `is_value: suffix == []` vs the sibling path's `req_all_final` guard | **open — the one I'd chase first** |
| 2.2 | `~coercible` threaded by hand through ~26 sites, 8 set it true | open |
| 2.3 | `meet` used as symmetric, but its two `Sig` args take different dedup paths | open |
| 2.4 | `sig_sub` calls `fresh_var`, a global mutable counter, from statics | open |
| 2.5 | `sig_sub` binds the target's definition, using `Id.invalid` | open |

On **2.1**: the `Module` case now has one path that *checks* the already-evaluated bindings are final (`req_all_final` on `Mod.modval_defs(prefix)`) and another that *assumes* it (`is_value: suffix == []`). Both can't be right — either `ModVal` definitions are final by construction and the guard is dead weight, or they aren't and the new flag is wrong. The probe double-sampling symptom that motivated the change is real; it's the invariant underneath that isn't written down anywhere. I couldn't reach this from the CLI, so it stays open.

On **2.2**: both failure directions are silent — a missed propagation loses sealing (spurious error), a spurious one permits width subtyping where it shouldn't. Two I'd look at: the `Filter` body propagates `coercible`, and `Statics.re:2185` passes the outer `~ana` with `coercible=true`. Worth noting part 4 then makes *avoidance* depend on this same flag, so the two features aren't independent.

Sealing itself I verified works on the composed tree: a module with an extra member correctly rejects access to it, while the same module still passes to a function expecting the narrower signature.

src/language/dynamics/transition/Transition.re
```diff
@@ -1369,7 +1369,9 @@ module Transition = (EV: EV_MODE) => {
             expr: Module(prefix @ suffix) |> rewrap,
             side_effects: [],
             kind: ModuleDiscardType,
-            is_value: false,
+            /* Evaluated bindings are never re-traversed: once nothing is
+               pending, the module is a value. */
+            is_value: suffix == [],
```
matthewhammer, line 1374: **Spot 2.1** (#2544) — **the one I'd chase first in this part.** This asserts the module is a value once nothing is pending, but the sibling branch (nothing pending at all) still guards with `req_all_final` on `Mod.modval_defs(prefix)`. Both can't be right: either `ModVal` definitions are final by construction and that guard is dead weight, or they aren't and this flag is wrong. The probe double-sampling symptom that motivated the change is real; it's the invariant underneath that isn't written down. I couldn't reach this from the CLI, so it stays open.

src/language/statics/StaticsBase.re
```diff
@@ -262,11 +262,20 @@ let map_m2 = (f, xs, ys, m: Map.t) =>
 
 let syn = Unknown(SynSwitch) |> Typ.temp;
 
+/* The relation the subsumption step checks between the expectation and the
+   synthesized type: consistency, or at a coercion site (`coercible`, set by
+   an ascription, an annotated binder or an application argument on its
+   operand) coercive subtyping, which seals a wider module. */
+let subsume = (~coercible, ctx: Ctx.t, ana: Typ.t, syn: Typ.t): option(Typ.t) =>
```
matthewhammer, line 269: **Spot 2.2** (#2544) — `~coercible` is threaded by hand through ~26 sites, 8 of which set it true. Both failure directions are silent: a missed propagation loses sealing (spurious error), a spurious one permits width subtyping where it shouldn't. Two worth a look: the `Filter` body propagates it, and `Statics.re:2185` passes the outer `~ana` with `coercible=true`. Note part 4 then makes *avoidance* depend on this same flag, so the two features aren't independent.

src/language/term/Typ.re
```diff
@@ -1064,7 +1064,7 @@ let rec meet = (ctx: Ctx.t, ty1: t, ty2: t): option(t) => {
   | (Sig(xs), Sig(ys)) =>
     /* Exact consistency: the same value-member names and the same
        type-member names (order-insensitive), members pairwise consistent.
-       Subtyping between signatures lives in ana_meet, not here. */
+       Sealing between signatures lives in `coercion`, not here. */
```
matthewhammer, line 1067: **Spot 2.3** (#2544) — `meet` is now used as if symmetric (`subsume` calls `meet(ctx, ana, syn)`, `coercion` internally calls `meet(ctx, to_, from)`), but its two `Sig` arguments take different code paths: `xs` goes through `dedup_last_items` then `members`, `ys` through `members` then `dedup_last`. Those should agree — two implementations of one policy is the risk, not the current behaviour.

matthewhammer (commented):

src/language/term/Typ.re
```diff
@@ -1102,6 +1102,85 @@ let rec meet = (ctx: Ctx.t, ty1: t, ty2: t): option(t) => {
...
+and sig_sub = (ctx: Ctx.t, ~from: list(Sig.t), ~to_: list(Sig.t)): bool => {
+  let from_members = Sig.members(from) |> Sig.dedup_last;
+  /* Open `from`: its type members get fresh names so they cannot collide
+     with `to_`'s; references inside its member types follow the renaming. */
+  let (ctx, sigma) =
+    List.fold_left(
+      ((ctx, sigma), m: Sig.member) =>
+        switch (m) {
+        | TypeManifest(name, def) =>
+          let f = fresh_var(name);
```
matthewhammer, line 1149: **Spot 2.4** (#2544) — `fresh_var` draws from a global mutable `var_count` ref, and `sig_sub` freshens every type member of `from` on every check, with statics re-running on each edit. So the generated name depends on how many times statics has run. Worth confirming these can't reach anything compared, cached or serialized (incremental-eval reuse keys, golden tests, the inspector). The non-ASCII `α` in the generated name is a second reason not to want them in printed output.

src/language/term/Typ.re
```diff
@@ -1102,6 +1102,85 @@ let rec meet = (ctx: Ctx.t, ty1: t, ty2: t): option(t) => {
...
+      | Some(t_from) =>
+        Option.is_some(coercion(ctx, ~from=t_from, ~to_=t_to))
+        && go(ctx, ms)
+      | None => false
+      }
+    | [Sig.TypeManifest(t, d_to), ...ms] =>
+      switch (from_type(t)) {
+      | Some(d_from) =>
+        Option.is_some(meet(ctx, d_to, d_from))
+        && go(Ctx.extend_alias(ctx, t, Id.invalid, d_to), ms)
```
matthewhammer, line 1177: **Spot 2.5** (#2544) — subsequent members are checked against the *target's* definition, not the source's. Defensible since the two just had to `meet`. Flagging mainly because `Id.invalid` is overloaded in this stack — it's also the builtin-module marker in #2526, and part 4 makes it a wildcard in escaped-type equality (spot 4.1). Worth confirming these synthetic aliases can't be mistaken for builtin-module paths.

PR #2495, opened 2026-09-05
Modules II, part 3: abstract type members, sealing, path types

7h3kk1d:
Third of the stack, on top of #2494. A signature may declare `type T` with no definition; a module sealed by it must define `T`, but outside the module `T` is known only as the path `M.T`.

- **Syntax.** `type T` is a signature item (`SigTypeAbstract`) in the grammar, the tile forms (listed before the manifest form so that `type` typed in a signature expands to it), `MakeTerm`, the Menhir `sigItem` rule, the printers, equality, abbreviation and ExplainThis. In the editor, typing `=` after a bare `type T` upgrades the tile to `type T = …` (`Insert.upgrade_bare_sig_type`); there is no downgrade.
- **Sealing.** A module analyzed against `{ type T; let x : T }` must define `T` (a missing definition is a `ModuleMissingMembers`; a value named `T` does not count), and its members are checked against its own definition. `Typ.sig_sub` realizes an abstract member of the required signature by whatever the module provides; a manifest member is not satisfied by an abstract one. `Typ.meet` matches abstract members only with abstract members. Since #2494 the directional relation is `Typ.coercion` and runs only at coercion sites (an ascription, an annotated binder, an application argument, and a function literal's body under an expected arrow type); elsewhere signatures match exactly.
- **Path types.** An abstract member projected from a module path is the stuck normal form `M.T`, consistent with itself and `?` only. `Typ.path_sig` now returns the path along with the items; `weak_head_normalize`, `normalize` and `meet` stop at a stuck path. A module variable's type is strengthened (`Typ.strengthen`, at `Var`, the capitalized-constructor fallback and `Dot`), so `M.x : M.T` and `module N = M` shares `M.T`, while two modules sealed by the same signature have distinct types. Only a path can name an abstract type: projecting from any other expression, or through a signature alias, gives `?`. The cursor inspector reports `M.T` as an abstract type (`Message.PathAbstract`).
- **Dynamics.** Abstract members have no runtime content. Runtime sealing now ascribes each kept member to its closed type, so an abstract member casts to `?` (previously a bare `T` left stuck casts such as `0:T`).

Tests: eighteen `Statics.Modules` cases (sealing, path types, aliasing, distinct sealings, transparency, missing and kind mismatches, forward references), `Typ.SigPaths`, `Evaluator.Modules`, `Editing.SigAbstract` (bare form, `=` upgrade, Cmd+D), plus MakeTerm, Menhir, round-trip, TyDi and Equality cases. Slide section F (a `Counter` module) and `docs/modules.md` describe the design.

Known, planned for the next PR: `fun (m : { type T; let x : T }) -> m.x` currently types as `S -> m.T` with `m` escaping its scope; the generative `Typ.avoid` handles that.

## Known gaps

- **No sharing constraint.** There is no `with type` spelling, so a signature
  cannot be refined by fixing one member, and nothing says the equivalent of
  `SHOW with T = Int`. The workaround is a hand-written signature literal
  carrying a manifest member, which is verbose and easy to get wrong; the
  implicits part (#2508) documents it because resolution needs it often.
- **No opaque-versus-transparent ascription.** One colon seals, and a sealed
  module's type members can never be selected by an argument type, so a
  sealed implicit instance is unusable. That is documentation today rather
  than a language distinction; a second spelling that checks against a
  signature while keeping manifest types would remove the footgun.
- **Type members cannot take parameters.** The type-pattern grammar admits a
  bare name, so nothing abstracting over a type constructor is expressible:
  no collection, monad or functor signature. Value members may be
  polymorphic, so signatures in the style of show, equality and monoid are
  unaffected.

## Verification

`make dev` clean · `make test-quick` green (~~3600~~ 3777 tests) · `dune build @check --profile release` clean.

## Follow-up commits

- **Report a missing member once, and let signatures name sibling members.** The synthetic `x : T` expectation checked inside a module body no longer shares ids with the signature's type nodes, and is skipped when it mentions a name the module does not define, so a missing `type T` is reported only on the module. Signature items can now see earlier value and module members, so `{ module Inner : { type T }; let y : Inner.T }` is well-formed; in subtyping a sibling is bound to the type the module provides.
- ~~**The Modules slide is split here** rather than in part 4: sections E and F move to `module-signatures.hz`, because statics recursion depth grows with the `let` chain and the combined slide was at the JavaScript stack limit (jsoo overflows node's default stack at roughly 100 nested lets).~~
  Replacement: the split now happens in #2493 (Modules: sections A and B; Module Signatures: C and D) and #2494 (Module Sealing: E). This part appends section F to the Module Sealing slide; #2498 moves F into its own Module Abstract Types slide together with G. The commit that did the split here was dropped as superseded.
- **An abstract member cannot be named through a signature alias.** `type S = { type T; … } in let y : S.T = …` used to type `S.T` as `?`, so any value passed and the inspector called `?` an abstract type. The alias names no module, so there is no particular `T` to name: it is now an error on the label (`TypAbstractMemberOfSignature`), with the dot explaining that a module `M` of the signature names it as `M.T`; TyDi no longer suggests such members. Manifest members through an alias, and abstract members through a module path, are unchanged. A type path rooted at a base type name (`Nat.T`) now says the name is a base type, not a module. Slide item 36a.


matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**. Per-part detail: [comment](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653053948).

**This part has the one confirmed bug from the pass — filed separately as #2545.**

`Typ.strengthen` selfifies a signature's abstract type members at the top level only and never descends into a `SigModule` member, so aliasing a module loses abstract-type identity for its sub-modules:

```
module M : {module P : {type T; let z : T}} = {module P = {type T = Int; let z = 0}} in
module N = M in
let f = fun (a : N.P.T) -> a in
f(M.P.z)
-- error: Expecting type N.P.T but got inconsistent type M.P.T
```

`N` is `M`, so those are the same type. Top-level abstract passes and nested *manifest* passes, so it needs both nesting and abstractness. `strengthen`'s own comment names `M.P` as a path it's meant to handle, so this reads as a bug against stated intent rather than a design call. #2545 has the controls and a suggested direction.

**Other part 3 spots:**

| | Spot | Status |
|---|---|---|
| 3.1 | `~self` omitted at 5 projection sites; abstract members become `?` | ✅ largely withdrawn — sealing and manifest-mismatch detection both verified working |
| 3.3 | Manifest and abstract type members share a namespace; order decides | open |
| 3.4 | `extend_sig_item` keys value members on the item id, not the pattern id | open — cheap consistency check |
| 3.5 | `is_stuck_path_term` accepts any `ProdProjection(_, Label(_))` | open — worth pinning with a test |

To be fair to this part: the abstract-types machinery works well in the cases I threw at it. `Counter.zero + 1` correctly reports `Counter.T`, and a manifest mismatch reports `Type member T is Bool but its signature declares Int`. The nesting gap is the exception, not the pattern.

src/language/term/Typ.re
```diff
@@ -708,6 +770,42 @@ and as_sig = (~rec_counter, ctx: Ctx.t, ty: t): option(list(Sig.t)) => {
   };
 };
 
+/* Selfification: seen through the path [path] (`M`, `M.P`), a signature's
+   abstract type members are the manifest paths `M.T`, so that an alias of M
+   shares M's abstract types (`module N = M` gives `N.T = M.T`) and a member
+   `x : T` projects to `M.T`. Identity on types without abstract members. */
+let strengthen = (ctx: Ctx.t, ty: t, ~path: t): t => {
```
matthewhammer, line 777: **Spot 3.2 — CONFIRMED BUG, filed as #2545.**

`strengthen` maps over the signature's items at the **top level only** and never descends into a `SigModule` member, so a sub-module's abstract types are never selfified. Repro on the composed tree:

```
module M : {module P : {type T; let z : T}} = {module P = {type T = Int; let z = 0}} in
module N = M in
let f = fun (a : N.P.T) -> a in
f(M.P.z)
-- error: Expecting type N.P.T but got inconsistent type M.P.T
```

`N` *is* `M`. Top-level abstract passes and nested *manifest* passes, so it needs both nesting and abstractness. This comment's own text names `M.P` as a path it's meant to handle, so it reads as a bug against stated intent. All three call sites also pass `~path=Var(name)`, so `strengthen` only ever sees bare-variable modules.

Suggested direction in #2545: recurse through `SigModule` members at the extended path `path.p`, and check the shape agrees with what `sig_members_closed` / `member_replacement` already build.

src/language/term/Typ.re
```diff
@@ -465,13 +472,27 @@ let apply_sig_subst = (sigma: list((Var.t, t)), ty: t): t =>
     sigma,
   );
 
+/* What an abstract type member [name] stands for outside its signature:
+   the path `self.T` when the signature is that of a module path, `?` when it
+   is not (a non-path expression's abstract types cannot be named), or its own
+   bare name when [keep_local] (the enclosing module body binds it). */
+let abstract_replacement =
```
matthewhammer, line 479: **Spot 3.1** (#2544) — with no path (`~self=None`) an abstract member becomes `Unknown`, i.e. `?`, which is consistent with everything. `~self` is omitted at several call sites (`ModuleHelpers.re:176`, `:179`, `:528`, `Typ.re:815`, `TyDiCtx.re:119`).

✅ **Largely withdrawn after testing**: sealing works (`Counter.zero + 1` correctly reports `Counter.T`) and manifest mismatches are caught (`Type member T is Bool but its signature declares Int`). The spot narrows to whether those `~self`-less sites are reachable with an abstract member at all — possibly none are, making this tidiness rather than correctness.

src/language/term/Sig.re
```diff
@@ -230,25 +244,34 @@ let value_names = (ms: list(member)): list(Var.t) =>
...
+/* Value members and type members live in separate namespaces; manifest and
+   abstract type members share one. */
 let same_kind = (a: member, b: member) =>
   switch (a, b) {
   | (Val(x, _), Val(y, _))
-  | (TypeManifest(x, _), TypeManifest(y, _)) => x == y
-  | (Val(_), TypeManifest(_))
-  | (TypeManifest(_), Val(_)) => false
+  | (
+      TypeManifest(x, _) | TypeAbstract(x),
```
matthewhammer, line 269: **Spot 3.3** (#2544) — manifest and abstract type members now share one namespace and `dedup_last` keeps the **last**, so `type T = Int` followed by an abstract `type T` silently turns the member abstract (and `find_type_def` then returns `None`), with the reverse in the other order. Order-dependent; see spot 7.2 on the four independent duplicate-resolution sites.

src/language/statics/Ctx.re
```diff
@@ -83,15 +83,39 @@ let extend_dummy_tvar = (ctx: t, tvar: TPat.t) =>
...
+      ctx,
+      {
+        name,
+        id: IdTagged.rep_id(tp),
+        kind: Abstract,
+      },
+    )
   | SigLet(_)
-  | SigModule(_)
+  | SigModule(_) =>
```
matthewhammer, line 103: **Spot 3.4** (#2544) — value members now bind into the context (needed for `Inner.T`), keyed on `IdTagged.rep_id(item)` — the *item's* id — whereas the type-member branch just above uses `rep_id(tp)`, the pattern's id. Inconsistent id sourcing tends to matter for the inspector and incremental reuse keys. Also note an unascribed `let x;` binds `x : ?` for later items.

src/language/term/Typ.re
```diff
@@ -506,18 +533,37 @@ let sig_project_value =
...
+let sig_project_type =
+    (~self=?, ~keep_local=_ => false, items: list(Sig.t), name: Var.t)
+    : option(t) =>
+  sig_project_type_member(~self?, ~keep_local, items, name)
+  |> Option.map(snd);
+
+/* An abstract type member projected out of a module path, `M.T`, does not
+   reduce: weak_head_normalize returns it as it is. Only meaningful on a
+   weak-head-normalized type. */
+let is_stuck_path_term = (ty: t): bool =>
```
matthewhammer, line 561: **Spot 3.5** (#2544) — this accepts *any* `ProdProjection(_, Label(_))`. I convinced myself it's narrower than it looks, since a failed projection reduces to `Unknown` before `normalize` tests it — but that's an invariant of `weak_head_normalize`, not of this predicate, so it's one change away from also swallowing failed labeled-tuple projections. Cheap to pin with a test.

matthewhammer (commented):

PR #2498, opened 2026-09-05
Modules II, part 4: module-typed functions and generative escape

7h3kk1d:
Fourth of the stack, on top of #2495 (parts 5 to 9 follow). Functions take and return modules with abstract types; a path cannot escape the scope of its root.

- **Avoidance.** `Typ.avoid` runs at function bodies, `let` bodies (so `module … in` too) and each `case` arm. A module path rooted at the binder going out of scope is first reduced in the body's context, so a manifest member simply expands. A path that is still abstract stays abstract as an escaped type (see below), except that a signature member defined as that path becomes abstract and later members refer to it by name. So `fun (m : S) -> m` has type `S -> S`, `fun (m : S) -> m.x` has an escaped abstract type as its result, and each call of a function returning a module with abstract members yields a fresh type (generativity). Inside the scope `m.x : m.T` is unchanged; the part-3 tests that pinned root types now read the in-scope type from the info map. `let f : S -> S = fun m -> m` checks because #2494 seals a function literal's body to the expected codomain.
- **Sibling paths.** The signature walk binds value member names as well, so a member typed by a sibling module's abstract type projects as `M.Inner.T` instead of silently becoming `?`. A written-out signature left behind by a type alias (`{ type T = Int }.T`) also projects.
- **Runtime.** A cast to an abstract path type is a no-op: elaboration normalizes manifest paths away, so a path reaching the runtime is abstract. Before, `let z : m.U = m.y` left the stuck value `1:m.U`.
- **Editor.** Deleting the `=` of a signature `type T = Int` downgrades the tile to the bare abstract form (the inverse of part 3's upgrade), leaving the old definition dangling as a stray type to remove or reattach by retyping `=`. `Editing.SigAbstract` grows to 34 cases: typing the bare form, adding `=` in every position, removing it backward, forward and by selection, round trips, and a MakeTerm-level check of the resulting item kinds.
- **Property test.** The QCheck AST generator emits signature types (with a shrinker), and a new property checks that no binder inside a random program leaks into its root type.

**Doc slide.** ~~The split into `modules.hz` and `module-signatures.hz` now lands in part 3; this part adds section G (module-typed functions) to the second slide. Background: adding that section to the combined slide overflowed the JavaScript stack in the CLI. It is not a loop: statics in jsoo overflows node's default stack at roughly 100 plain nested `let`s (about 10 to 16KB per level), and the combined slide had about 70 nesting levels plus 30 `module` bindings. The C3 build needed a stack between 950 and 984KB for it and this one between 984 and 1000KB, against a default of 984KB. The slide is now two: `modules.hz` (construction, qualified types, signatures, errors) and `module-signatures.hz` (width subtyping, abstract types, module-typed functions), both registered in `Slides.re`.~~
Replacement: the slides are split lower in the stack (#2493: Modules and Module Signatures; #2494: Module Sealing). This part moves part 3's section F out of the sealing slide into a new Module Abstract Types slide and adds section G to it, since G's examples use F's `Counter`. The measured reason for keeping slides small is the Web Worker: it evaluates programs with about half the main thread's stack, so a slide whose evaluation needs more than roughly 450KB under node overflows on the first edit. Each module slide now needs 370KB to 445KB. The stack depth itself deserves an issue.

## Avoidance keeps the type abstract

Where a path outlives its root, avoidance used to fall back to unknown, and
unknown is consistent with everything, so the loss was silent: nothing told
the author, and every later use checked against nothing. It now mints an
escaped abstract type (`Typ.Escaped`), which is consistent with `?` and with
itself and inconsistent with everything else. That is the relation an
in-scope abstract member already has, so no new rule is invented; what is new
is a representation that survives the loss of its root, generalizing the
clause avoidance already had for signature members.

```
let get = fun (m : BOX) -> m.x in get(B) + 1
-- This has the abstract type m.T, which escaped the scope of its module,
   so it cannot be used where a concrete type is required

let dup = fun (m : SHOW, x : m.T) -> (x, x) in dup(ShowInt, "s")
-- m.T escaped the scope of its module, so nothing here can have that type;
   mark the parameter implicit, or annotate it with a manifest signature,
   to keep the caller's type
```

Binding an escaped value, passing it to an unannotated function and giving it
to a hole all still work: the error appears where the author wrote something
concrete, which is the gradual half of the bargain.

- **The parameter is closed too.** The function arm avoided only its body, so
  a component annotated with an earlier component's member left the type
  mentioning a binder no caller can name, and every argument passed, since
  normalization maps a path with an unbound root to unknown. Only the copy
  that becomes the function's type is closed; the pattern's own type and the
  elaborated ascription keep the precise path, so runtime casts and
  incremental-evaluation keys do not move.
- **Generative per application.** `Typ.freshen_escaped` gives each call its
  own identity, so two calls of one function escape different types and
  joining them is an error, while a value bound once and used twice is fine.
  Identities are derived from the application's id rather than generated, so
  re-checking an elaboration still yields an equal type.
- **`Mark.EscapedType` says which side escaped**, so the advice fits the
  case, and an escaped type displays as its path plus a short tag of its
  identity, since two escapes of one path are different types. A failed
  implicit resolution keeps erasing to unknown (`Typ.escape_to`): that
  failure is already marked, and opacity there would only cascade.
- **Not fixed here:** what the caller knows, that the parameter is a
  particular module, still cannot be written in an ordinary arrow type. An
  implicit binder can hold it (#2508); an ML sharing constraint for ordinary
  module parameters is the missing feature. A tuple argument also still
  reports the component and the tuple, which is the pre-existing cascade
  fixed off-stack in #2502.

Two commits: the type form, inert, with its `meet` rules and traversals; then
minting, the closed parameter, freshening and the diagnostics. The escape
tests in the modules suite now assert the escaped type where they asserted
unknown, five new cases cover the behavior, slide item 42 keeps its abstract
type and 43 shows the error.

## Fixes the fuzz properties turned up

Three defects, each found only through a Slow-tagged QCheck property and so only on seeds that generated the shape — parts 5-8 went red while this part stayed green on the very same bugs. Each now has a deterministic test, so the seeds no longer decide.

- **A cyclic alias crashed the analysis.** `type x = x.a in ?` aborted with `weak_head_normalize exceeded 1000 recursive calls`. `TyAlias` already guards self-reference by wrapping the definition in `Rec`, and normalization leaves a `Rec` alone — but from this part `x.a` is a module member projection, so normalizing it asks `path_sig`, and `as_sig` unrolls that `Rec` looking for a signature, handing the projection straight back. Out of fuel the type now comes back unreduced, stuck, which is what the function already returns for everything else it cannot reduce; the program reports `Expected a module or tuple type, found type rec x -> x.a`. `Typ.CyclicAlias` pins it.
- **A printed `:` glued to a leading-plus sum.** `Menhir and maketerm are equivalent` failed on `(()) @< ({ let _:+ Adid }) >`: `:` is an operator character and `Token.is_potential_operator` matches any run of them, so `:+` lexes as one token that is no form at all, while Menhir's separate lexer splits it. `should_add_space` already knew this hazard, but as a list of two characters (`$`, `!`) rather than the property they share; it now asks `Token.begins_with_potential_operator`, which covers `!` along with every other operator character. `$` stays listed, being a name character. An unparenthesized sum only ever reaches a `:` as a sig member's ascription, which is why this appears here.
- **Printing was not idempotent.** `type T = { … }` printed as `({ … })`, then `(({ … }))`, a layer per trip. `parenthesize_typ`'s `Parens` case ran its own content through `paren_typ_at(Precedence.min)`; precedences count down from `max = 0`, so that test wraps exactly the types whose precedence *is* `min` — a Sig, a bare sum, a multihole — and re-wrapped them every pass. The wrapper being emitted is already the protection: nothing outside can bind into it. **Not specific to modules** — `dev` grows the same layers on a parenthesized bare sum (`type T = (+ Adid) in 7`), so this fixes that too and carries it along with the stack.

## Known gaps

- **No dependent domain for ordinary module parameters.** An explicit
  parameter cannot be named later in its own parameter list or in the result
  type, which is exactly the equation this PR reports the loss of. An
  implicit binder (#2508) is the only construct that can hold it, so marking
  the parameter implicit is the current substitute for a sharing constraint.
- **No existentials with an unpacking form.** An escaped abstract type is now
  a real type rather than unknown, but there is no way to give one a name and
  use it, so such a value stays inert. Storing modules with differing
  abstract members in a list, or returning one from a conditional, therefore
  still loses the shared type.

## Verification

`make dev` clean · `make test-quick` green (~~3658~~ ~~3809~~ 3851 tests) · `Statics.Properties` (crash and closedness properties, with signatures generated) and `MenhirFuzz` pass · ~~both slides~~ all four module slides analyze with errors only under their "Error" items · the Menhir and elaboration properties also pass at the five seeds that had been failing (803422318, 57351729, 931683130, 541095201, 92308926) · `make ci-check` clean under the release profile.


7h3kk1d:
Note on the red `Full tests` here: neither failure is this PR's.

- `Menhir and maketerm are equivalent` fails per-seed because this PR's QCheck generator started emitting signature types, which lets it reach a printer divergence on `:+` that exists on dev. Diagnosed, pinned and explained in #2493: https://github.com/hazelgrove/hazel/pull/2493#issuecomment-5575182157
- `Elaboration does not crash` fails per-seed on a cyclic type alias through a projection (`type x = x.a in ?`), which crashes normalization with `weak_head_normalize exceeded 1000 recursive calls` on dev as well.

Both want separate changes against dev.

matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**. Per-part detail: [comment](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653060694).

Worth saying first: **this part diagnoses and fixes the problem I'd been flagging across parts 1–3.** `Typ.Escaped` exists precisely because "unknown is consistent with everything, so the loss was silent" — which is the same critique I'd arrived at independently, and the answer here is the right one. The escaped-type diagnostic is genuinely good in practice:

```
let leak = fun (c : {type T; let zero : T}) -> c.zero in ... leak(C) + 1
-- error: This has the abstract type c.T, which escaped the scope of its module,
   so it cannot be used where a concrete type is required
```

**Spots, all about edges rather than the core:**

| | Spot | Status |
|---|---|---|
| 4.1 | `Id.invalid` is a wildcard in `escaped_equal`; `Factory.escaped` defaults to it | open — needs a code-level guarantee, not a test |
| 4.2 | Escaped identity is a string hash of site + path label | open — also `path_label` returns literal `"?"` for unprintable paths |
| 4.3 | Which type `avoid` runs on depends on part 2's `coercible` flag | open |
| 4.4 | `avoid`'s escape test defers to the type namespace | ✅ verified handled — same diagnostic as 1.2 |
| 4.5 | Property test's power: how often does the generator emit escaping paths? | open |

On **4.1**: `escaped_equal` treats `Id.invalid` as a wildcard matched by label, and the comment says only hand-written test types carry it — but `Factory.escaped` *defaults* `~id=Id.invalid`, and `Id.invalid` already means other things in this stack (part 2's `sig_sub` aliases, the builtin-module marker in #2526). If any non-test path ever builds one without `~id`, it becomes a type equal to every escaped type sharing its label, which is exactly what generativity is meant to prevent. Either a grep-level guarantee or dropping the wildcard and making tests pass real ids.

On **4.5**: adding the property is the right instinct and I'd rather have it than not — but its value is entirely in how often the generator emits module-typed binders with abstract members. A property that passes because it rarely generates the relevant shape still reads as coverage. Worth instrumenting once to see what fraction of generated programs contain a path that *could* escape.

src/language/term/Grammar.re
```diff
@@ -12,6 +12,29 @@ type deferral_position_t =
...
+
+/* How an escaped type shows in a type: the path it came from plus a short
+   tag of its identity, since two escapes of the same path are different
+   types and would otherwise print alike. */
+let escaped_label = (e: escaped): string => e.label ++ "~" ++ Id.str3(e.id);
+
+/* Two escaped types are the same type when their ids agree. Id.invalid is a
+   wildcard matched by label, which only a hand-written (test) type uses:
+   avoidance always derives its ids from the term whose scope closed. */
+let escaped_equal = (a: escaped, b: escaped): bool =>
```
matthewhammer, line 34: **Spot 4.1** (#2544) — `Id.invalid` is a wildcard matched by label here. The comment says only hand-written test types carry it, and in `Typ.avoid` that holds — but `Factory.escaped` **defaults** `~id=Id.invalid`, and `Id.invalid` already means other things in this stack (part 2's `sig_sub` aliases, the builtin-module marker in #2526). If any non-test path ever builds one without `~id`, it becomes a type equal to every escaped type sharing its label — exactly what generativity is meant to prevent. Either a grep-level guarantee, or drop the wildcard and make tests pass real ids.

src/language/term/Typ.re
```diff
@@ -806,6 +827,212 @@ let strengthen = (ctx: Ctx.t, ty: t, ~path: t): t => {
...
+           with `?`, and no longer consistent with a concrete type. The id is
+           derived from [site] and the path, so the same escape at the same
+           site is the same type on every pass. */
+        | None =>
+          switch (escape_to) {
+          | ErasesToUnknown => Unknown(Internal) |> temp
+          | EscapesAt(site) =>
+            let label = path_label(w);
+            Escaped({
+              id: Id.mk_str(Id.to_string(site) ++ "!" ++ label),
```
matthewhammer, line 951: **Spot 4.2** (#2544) — identity is a string hash of site + path label. Two questions: (a) granularity — the PR body says "each call ... yields a fresh type", but this keys on the syntactic *site*, so a function applied twice at one site shares a type; worth confirming "call" means call-site there. (b) `path_label` returns the literal `"?"` for anything it can't print, so two unprintable paths at one site hash equal.

src/language/statics/Statics.re
```diff
@@ -1998,12 +2004,25 @@ and uexp_to_info_map =
...
           ? fresh_ascription(ctx, e_elab, e.elab_syn_ty, Some(e.ty)) : e_elab;
+      let avoid =
+        Typ.avoid(
+          p'.ctx,
+          ~escape_to=EscapesAt(Exp.rep_id(uexp)),
+          ~escaping,
+        );
       let syn_ty_fun =
-        Arrow(p.ty, coercible ? e.ty : e.elab_syn_ty) |> Typ.temp;
+        Arrow(avoid(p.ty), avoid(coercible ? e.ty : e.elab_syn_ty))
```
matthewhammer, line 2024: **Spot 4.3** (#2544) — which type avoidance runs on depends on part 2's `coercible` flag. That couples the two features: a mistake in threading `~coercible` (spot 2.2) no longer just loses sealing, it changes whether a path escapes. Worth knowing before reviewing either in isolation.

src/language/term/Typ.re
```diff
@@ -806,6 +827,212 @@ let strengthen = (ctx: Ctx.t, ty: t, ~path: t): t => {
...
+  | Projector(_, ty) => path_label(ty)
+  | ProdProjection(p, {term: Label(l), _}) => path_label(p) ++ "." ++ l
+  | _ => "?"
+  };
+
+let avoid =
+    (ctx: Ctx.t, ~escape_to: escape_to, ~escaping: list(Var.t), ty: t): t => {
+  let escapes = (ty: t) =>
+    switch (path_root(ty)) {
+    | Some(x) => List.mem(x, escaping) && Ctx.lookup_tvar(ctx, x) == None
```
matthewhammer, line 916: **Spot 4.4** (#2544) — if the path's root also exists as a type variable, the path is treated as *not* escaping and avoidance skips it. ✅ **Verified handled**: the shadowing case produces the same purpose-built diagnostic as spot 1.2 rather than a silent leak. Withdrawing.

Separately — the escaped-type machinery in this part works well in practice. `fun (c : {type T; let zero : T}) -> c.zero` applied and then used concretely gives a genuinely good message.

matthewhammer (commented):

PR #2500, opened 2026-09-05
Modules II, part 5: tuple label errors like modules, named missing members, CLI spans

7h3kk1d:
Fifth of the stack, on top of #2498. Improvements to error reporting that reach beyond modules, kept apart from the module fixes in parts 1 and 3.

- **Labeled tuples follow the module design for bad projections.** `t.y` with no label `y` now marks the label (`LabelNotFound`, now worded "Tuple has no label y; its labels are x") while the dot carries only a message (`Message.TupleLabelNotFound`), the same shape part 1 gave `m.y` on modules and the type-level `M.Fake` always had. Applies to tuples and lists of tuples.
- **Missing members are named for any signature-typed expression.** When an expression of signature type is analyzed against a signature that declares members it lacks, `StaticsBase.expectation_mismatch_mark` reports `ModuleMissingMembers` instead of a generic type mismatch, so a module variable, a function argument, or a function result names what is missing, as a module literal already did. The computation is shared through `Sig.missing_members`. Slide item 26a (in the Module Sealing slide) shows `greet(Carol)` with `Carol` lacking `name`.
- **CLI carets span the whole diagnosed term.** `hazel analyze` used to underline only the tile carrying the term's id, so `M.x + 1`'s error pointed at a single `.`. The caret now covers every measured id of the term (`m.x`, `a + b`, a module literal on one line), matching what the editor highlights; terms inside folded projectors keep the location-less fallback without spurious lookup warnings.

## Verification

`make test-quick` green (3855 tests) · the module slides and the tuples slide analyze with the expected errors only.


matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**. Per-part detail: [comment](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653065672).

Small and mostly diagnostics, so only three spots — and one housekeeping note.

**CI:** this PR's red check is a runner hang, not a test failure. The job produced no test output at all after menhir finished, then `The action 'Run tests' has timed out after 60 minutes` with orphan `make`/`dune`/`node` killed and no junit XML. All nine jobs in the stack started in the same 18:10–18:15 window; the other seven finished in 12–24 minutes, and parts 7–9 — which sit *on top of* this one and contain its code — are green. I've re-run it.

| | Spot | |
|---|---|---|
| 5.1 | CLI caret spans the term's whole subtree | the `m.x` / `a + b` intent is right, but for a module literal or a long `let` chain this underlines everything from first token to last — worth seeing what it prints for one of the module slides |
| 5.2 | "Label not found" is now a mark on one path and a message on two others | `Statics.re:1326` adds a `Mark.LabelNotFound` while `:1416` and `:1547` replace theirs with a `Message` and pass `~marks=[]`; worth confirming the editor and `hazel analyze` agree, and that this plays well with the dedup from #2489 |
| 5.3 | Label lookup reaches through a `List` of labeled tuples | the new guard matches `List({term: Prod(ts), _})`, so `t.y` is label-checked on a *list of* labeled tuples — presumably deliberate for the table work, but it's the one semantic change hiding in an otherwise diagnostic PR and deserves a sentence saying why |

`ids_of_info` calls `map_any_annotation` for its side effect and discards the result, which works but is worth a comment for the next reader.

src/CLI/Diagnostic.re
```diff
@@ -52,6 +52,55 @@ let warning_string = (item: Language.Warning.list_item): string =>
    context. `severity` is the header prefix ("error" / "warning"). `fallback`
    is appended after the header when no source position can be resolved
    (used by errors to print the offending term). */
+/* Every id of the diagnosed term, so the caret spans the whole term (an
+   infix form such as `m.x` or `a + b` includes its operands) rather than
+   only the tile carrying the term's own id. */
+let ids_of_info = (info: Language.Info.t): list(Util.Id.t) =>
```
matthewhammer, line 58: **Spot 5.1** (#2544) — the `m.x` / `a + b` intent is right and matches the editor, but the span is over the term's **entire subtree**, so diagnosing a module literal or a long `let` chain underlines everything from first token to last. Worth seeing what it prints for one of the module slides. Minor: `map_any_annotation` is called for its side effect and the result discarded — worth a comment for the next reader.

src/language/statics/Statics.re
```diff
@@ -1326,6 +1326,12 @@ and uexp_to_info_map =
                 List.mem(name, Sig.type_names(Sig.members(items))),
             }),
           ]
+        /* Labeled tuples follow the same design. */
+        | (Prod(ts) | List({term: Prod(ts), _}), Label(name))
```
matthewhammer, line 1330: **Spot 5.3** (#2544) — this guard matches `List({term: Prod(ts), _})`, so `t.y` is label-checked when `t` is a *list of* labeled tuples, not only a labeled tuple. Presumably deliberate for the table work, but it's the one semantic change in an otherwise diagnostic PR and deserves a sentence saying why.

src/language/statics/Statics.re
```diff
@@ -1410,7 +1416,14 @@ and uexp_to_info_map =
             add(
               ~elab_term=dot_elab,
               ~elab_syn_ty=Unknown(Internal) |> Typ.temp,
-              ~marks=[LabelNotFound(name, labels)],
+              ~marks=[],
+              ~message=
+                Message.Exp(
+                  TupleLabelNotFound({
```
matthewhammer, line 1422: **Spot 5.2** (#2544) — this path replaces its `Mark.LabelNotFound` with a `Message` and passes `~marks=[]`, while `Statics.re:1330` *adds* a `Mark.LabelNotFound` for the `Dot` case. So one user-visible condition is a mark on one path and a message on two others. Worth confirming the editor and `hazel analyze` agree, and that it plays well with the dedup from #2489.

matthewhammer (commented):

matthewhammer:
Follow-up: the re-run came back **green in 22 minutes** ([run](https://github.com/hazelgrove/hazel/actions/runs/34631789123)), so the flake diagnosis holds for this PR. Nothing further needed on CI here.

PR #2501, opened 2026-09-05
Modules II, part 6: a sub-module's wider annotation is kept and sealed

7h3kk1d:
Sixth of the stack, on top of #2500.

~~The signature's expectations now reach into sub-modules and destructuring patterns, so a problem is reported where it is.~~
~~- **Sub-modules.** `module Inner = …` items receive the sub-signature the outer signature declares for them. A member the sub-module lacks is `Module is missing members: y` on the inner literal; a differing type member lands on its `type` item; a wrong definition lands on the definition. Before, all three surfaced as one generic mismatch on the outer module. A sub-module the user annotated keeps its own annotation.~~
~~- **Destructuring.** Variables inside tuple, labeled tuple, nested tuple, list and constructor patterns are annotated individually (`let (a, b) = …` is checked as `(a : Int, b : Int)`), so `let (a, b) = (1, "s")` against `a : Int; b : Int` reports on `"s"`. The tuple literal is also marked, exactly as a top-level `let (a : Int, b : Int) = (1, "s")` is today; that cascade belongs to the generic expected/got reporting and is left for a later change.~~
~~- The elaboration refold strips the synthetic ascriptions by walking the user's pattern and its elaboration side by side.~~
~~Slide items 20a and 20b show both cases.~~

Replacement: that feature moved down to #2493 when the stack was restructured for standalone merging ("Expectations reach sub-modules and destructured binders", slide items 20a and 20b). This PR now carries only the one test that needs sealing to state: a sub-module annotated with a signature wider than the one the outer signature declares for it keeps its own annotation. Its definition is checked against the wider annotation, and the sub-module is sealed to the outer declaration at its binder, so `M.Inner.x` types as `Int` with no error.

## Verification

`make test-quick` green.


matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**.

**No spots for this one.** After the stack was restructured for standalone merging, this PR carries a single test (+21/−0, one file) and no source change, so there's nothing for a review to flag. The test itself — a sub-module annotated wider than the outer signature declares keeps its own annotation and is sealed at its binder — is a good thing to pin.

**CI:** the red check here is a runner hang, not a test failure: no test output at all after menhir, then `The action 'Run tests' has timed out after 60 minutes`, with orphan processes killed and no junit XML written. Since this PR changes no source, that's about as clean a demonstration of an infra flake as the stack offers. I've re-run it.

matthewhammer:
Correcting my earlier comment: I said this was a runner flake and that I'd re-run it. **The re-run timed out again**, identical signature ([run](https://github.com/hazelgrove/hazel/actions/runs/34631788944)) — so "flake" was too strong, and re-running demonstrably does not clear this one.

It still isn't the code, on three pieces of evidence:

- **Passes locally.** Full suite on `d1fc5370ea`: `Test Successful in 572.819s. 4113 tests run.` — 9m36s, no hang.
- **The same code passes in CI** as part of #2503, #2508 and #2510. I checked their green runs tested heads that genuinely contain this branch's head (`6f9fe7143b`, `5ed7503941`, `df149f4689` each have `d1fc5370ea` as an ancestor), finishing in 22–34 minutes.
- **Not caching.** Both re-runs hit the same opam cache keys and reached menhir within six seconds of each other; part 5 then finished while this one sat for an hour with no test output at all.

So the job is the anomaly rather than the branch: one test file, passing locally and passing under everything stacked above it, yet hanging before a single test runs, twice on different days. Flagging rather than claiming a cause — someone who knows the workflow should probably look.

**Unrelated, in the same file:** this diff re-registers ten already-registered tests under the `/* Nested expectations */` comment — `test_nested_missing_member_localized`, `test_nested_type_member_mismatch_localized`, `test_nested_definition_mismatch_localized`, `test_nested_user_annotation_kept`, `test_error_nested_module_sealed`, `test_tuple_binder_mismatch_localized`, `test_tuple_binder_ok`, `test_labeled_tuple_binder_ok`, `test_nested_tuple_binder_mismatch_localized`, `test_cons_binder_mismatch_localized`. Looks like leftover from the restructuring the description mentions. #2503 and above inherit them, so they aren't the hang — but they should probably come out.

PR #2503, opened 2026-09-05
Modules II, part 7: a signature member declared twice is an error

7h3kk1d:
Seventh of the stack, on top of #2501. A signature that declares the same member twice is now an error.

- The later declaration is marked with the new `SigDuplicateMember`: `type S = { let x : Int; let x : Bool }` reports `Member x is declared more than once in this signature` on `let x : Bool`. Value and module members share a namespace (`let m` and `module m` clash); type members, manifest or abstract, have their own, so `let x : Int; type x = Int` is fine.
- The mark lives on the signature item itself. `Info.sig_` gains `marks`, which `Info.marks_of` now reads, so the cursor inspector, the problem sidebar and `hazel analyze` all report it; the CLI caret covers the whole item.
- Module bodies are unchanged: a repeated binding there is shadowing, last wins, as before.

Slide item 23a shows the error. Tests cover value, manifest type, abstract type, manifest-then-abstract, module, and value-versus-module duplicates, the shared and separate namespaces, and that the later item carries the mark.

## Verification

`make test-quick` green · the Modules slide analyzes with errors only under its "Error" items.


matthewhammer (commented):
Part of a review pass over parts 1–7; the whole pass lives in **#2544**. Per-part detail: [comment](https://github.com/hazelgrove/hazel/issues/2544#issuecomment-5653065672).

Verified working on the composed tree — `type S = {let x : Int; let x : Bool}` marks the later item, as designed:

```
error: Member x is declared more than once in this signature
```

| | Spot | |
|---|---|---|
| 7.1 | `InfoSig` can now carry marks, and `marks_of` reads them for the first time | anything that relied on signature items never producing marks — error counting, the problem sidebar, the dedup from #2489, CI's error assertions on the doc slides — now sees a new source of them. Cheap to check, easy to miss |
| 7.2 | The item that gets the error is also the item that wins | probably the right pairing, but worth stating explicitly |

On **7.2**: the mark lands on the later declaration while `extend_sig_item` still extends with it and `Sig.dedup_last` keeps the last of each name — so the marked item is the one whose type is used. That's defensible (mark the offender, keep last-wins to match module bodies). What makes it worth writing down is that this is the **fourth** place in the stack where "which duplicate wins" is decided independently, alongside runtime `add_modval` (part 1), `dedup_last` / `dedup_last_items` (part 2), and `same_kind` (part 3).

I checked the obvious case on the composed tree and they agree — `{let x = 1; let x = 2}` gives no static error and evaluates to `2`. So this isn't a bug report, it's a suggestion: one documented rule that all four sites reference would retire the whole category before it can drift.

src/language/statics/Info.re
```diff
@@ -265,10 +266,10 @@ let marks_of: t => list(Mark.t) =
   | InfoExp({marks, _})
   | InfoPat({marks, _}) => marks
   | InfoTyp({marks, _})
-  | InfoTPat({marks, _}) => marks
+  | InfoTPat({marks, _})
+  | InfoSig({marks, _}) => marks
```
matthewhammer, line 270: **Spot 7.1** (#2544) — `InfoSig` moves out of the "always `[]`" group, so anything that relied on signature items never producing marks — error counting, the problem sidebar, the dedup from #2489, CI's error assertions on the doc slides — now sees a new source of them. Cheap to check, easy to miss.

src/language/statics/Statics.re
```diff
@@ -4324,20 +4324,44 @@ and utyp_to_info_map =
...
+       A member declared twice is an error on the later declaration; values
+       and modules share a namespace, type members have their own. */
+    let (_, _, m) =
       List.fold_left(
-        ((ctx, m), item: Sig.t) => {
+        ((ctx, (seen_values, seen_types), m), item: Sig.t) => {
+          let members = Sig.members([item]);
+          let values = Sig.value_names(members);
+          let types = Sig.type_names(members);
+          let duplicates = (seen, type_member) =>
```
matthewhammer, line 4336: **Spot 7.2** (#2544) — ✅ the feature itself verified working: `{let x : Int; let x : Bool}` marks the later item as designed.

The note is about the pairing: the mark lands on the later declaration while `extend_sig_item` still extends with it and `Sig.dedup_last` keeps the last — so the marked item is the one whose type is used. That's defensible, and I checked statics and runtime agree on the obvious case.

Worth writing down because this is the **fourth** place in the stack that independently decides which duplicate wins, alongside runtime `add_modval` (part 1), `dedup_last`/`dedup_last_items` (part 2) and `same_kind` (part 3). One documented rule all four reference would retire the category before it can drift.

matthewhammer (commented):

PR #2506, opened 2026-09-06
Statics: a capitalized name resolves to its most recent binding (module vs constructor)

disconcision:
```
type T = + Text + Other in
module Text = { let f = fun n -> "x" } in
Text.f(0) == "x"
```
This reports "Expected a tuple" at `Text.f`. A capitalized name parses as a constructor and statics looked it up constructor-first, falling back to a variable binding only when no constructor of that name existed anywhere. So a module named like any constructor in scope lost to it, no matter which was bound later.

This PR makes the most recent binding of the name win, whichever kind it is (`Ctx.newest_var_or_ctr`), which is what shadowing means for every other name. Expression position only; constructor patterns are unchanged. Draft because the rule deserves a careful look: it changes what a capitalized identifier means when a module and a constructor share a name.


7h3kk1d:
Reminder for me to look at this with modules. My 🤖 went the opposite way and picked based off of Ap/Projection

PR #2513, opened 2026-09-08
Folder-style tutorial lessons

7h3kk1d:
Stacked on #2485. **All three PRs in this stack are meant to be merged together** — they are split for review only.

## Summary

Documentation mode already renders `/`-separated slide names as nested breadcrumb dropdowns (`"B2T2 / Table API / Constructors / vcat"`). Tutorial mode has 25 lessons in one flat list. This PR names that hierarchy and then gives tutorial mode folders on top of it.

### `SlidePath`

The hierarchy was an ad-hoc convention spread across three unrelated places: a local `split_filepath` plus inline `take`/`nth`/`dedup_f` in `EditorModeView.indicator_select`, ~60 hardcoded `" / "` literals in `src/b2t2/Slides.re` and `src/docslides/Slides.re`, and `"Derivations / " ++ spec.title` in `Init.re`.

`src/web/util/SlidePath.re` makes it a type:

```reason
type t = {
  folders: list(string),
  leaf: string,
};
```

and owns the operations both modes need: `of_string`/`to_string`, `folder`/`same_folder`, the folder-scoped index arithmetic the navigation arrows use (`folder_indices`, `folder_position`, and `step_in_folder`, which clamps), and `breadcrumb` — the sibling-dropdown decomposition lifted out of the view.

`indicator_select` is now a thin render over `breadcrumb` and takes `list(SlidePath.t)`, which dropped 69 lines from it along with two latent crashes it had been carrying:

- `List.nth(parts', prefix_depth)` raised `Failure("nth")` whenever one slide name was an exact proper prefix of another — a slide named `Tables` alongside `Tables / Tables` would have taken down the whole top bar. Such a path is now skipped, making it merely unreachable from the deeper dropdown.
- `Option.get` on the changed option's index lookup.

Slide names stay plain strings at the persistence boundary — they are IndexedDB keys (`"doc:" ++ name`) and `.hzt` `@title` text — so nothing about the stored format changes; callers convert at the view/navigation boundary.

### Tutorial folders

A lesson's `@title` is now a `SlidePath`, and both arrow surfaces — the in-slide prev/next pair and the top-bar arrows — **step within the current lesson's folder**, clamping at its edges, so the last lesson of a folder shows `Done! 🎉` instead of a next arrow. Crossing folders goes through the folder dropdown. The slide heading shows just the leaf, since the breadcrumb above it already names the folder.

| Folder | Lessons |
|---|---|
| `Basics` | 19 — Holes through Labeled Tuple Projection |
| `Tuple Structural Operations` | 3 — Extension, Omission, List Conversions |
| `Tables` | 6 — Tables, Table Column Projection, Rich Probes, and the three Gradebook tasks below |

Rich Probes sits under Tables because it is a rich probe *for* tables.

### Three tasks recovered from `tables-study`

`Tables` also picks up Gradebook Midterm Mean, Gradebook Overall Grade, and Gradebook Tidy Term — study tasks that had never been ported off the `tables-study` branch, where they were `Ta_*.ml` zipper dumps of 0.5–0.9 MB each. They keep their original ids, so they are the same lessons the study ran.

They were converted by adapting the dumps to the current types (the drift was two things per file: `selection` gained `anchor_caret`/`smart_rounded`, and `refractors` dropped `autos`/`sample_cursor` with `manuals` unchanged), compiling them in alongside `TaskRefDocs`, and running `./hazel tutorial-decode` — which also expanded each task's composed `TaskRefDocs` sections into a literal `@reference`. None of that scaffolding is in the tree.

Two content fixes on top of the decode: Midterm Mean's table held 20 literals like `85.000000`, and past six significant digits a slide falls onto the quadratic parser (slowness being the only symptom), so those are normalized to `85.0`; and its hidden tests did not round-trip, because an implicit `¿` hole collapses to nothing before a `,` inside a labeled tuple, so those twelve are now the explicit `?` token.

Not ported, deliberately: the two Bug Identification tasks, which depend on live typing; the 27 pre-placed probe refractors these three carried (`Triggers.refractor_seg_to_seg` only emits `^^probe(...)` for a refractor whose id is a term root in the current parse, and these ids came from the older grammar, so 26 of 27 dropped); and the study's `setting_overrides` condition flags and `?treatment` A/B ordering, none of which has a `.hzt` marker. The commit records what each task's flags were.

All three start as real tasks — a hole in Midterm Mean, an identity-stub `tidy_term` — so their hidden tests fail until solved.

Recategorizing is safe: lesson identity is the `id=<uuid>` flag introduced in #2485, not the title, so moving a lesson between folders does not reset anyone's saved progress.

`hazel-programs/tutorial/README.md` documents the authoring rules — one folder segment per lesson, no title a proper prefix of another, a folder's lessons contiguous in `src/tutorialslides/Slides.re`, and `Basics / Holes` first because it is the lesson a fresh profile opens on.

## Deliberately not done

A ✔ on a folder's dropdown option when the whole folder is complete. It cannot be built from current data: `TutorialsMode.Update.calculate` evaluates only the current lesson and `TutorialMode.Model.persist` saves only zippers, so `all_tests_passed` is `false` for every non-current lesson — which is also why today's per-lesson ✔ from `return_title` can only ever show for the lesson you are standing on. A real folder ✔ needs persisted per-lesson completion, which is its own change.

## Test plan

- [x] `make dev` builds clean, no new warnings
- [x] `make test-quick` — 3501/3501
- [x] `./hazel tutorial-verify` — 28/28 slides round-trip cleanly
- [x] `Test_SlidePath` (23 cases) — parsing, accessors, folder navigation, and breadcrumb, including the proper-prefix path that used to raise and an out-of-range index
- [x] `Test_Tutorial` (7 cases) — ratchets the authored lessons: folder set and sizes (`Basics x19`, `Tuple Structural Operations x3`, `Tables x6`), contiguity, `Basics / Holes` first as the default lesson, no title a proper prefix of another, and that repeated next/prev traverses exactly one folder and then stops
- [x] Documentation mode — `B2T2 / …` breadcrumbs and `Derivations / …` still render and switch slides
- [x] Tutorial mode — opens on `Basics / Holes`; arrows walk one folder and stop at its ends; the dropdown crosses folders
- [x] `./hazel analyze --warnings` reports nothing on all three ported tasks' starting programs

### DIFF
--- a/hazel-programs/tutorial/README.md
+++ b/hazel-programs/tutorial/README.md
@@ -28,7 +28,7 @@ Marker lines are *exactly*:
 
 | marker | maps to | notes |
 |---|---|---|
-| `@title` | `title` | `"<Folder> / <Lesson>"` (a `SlidePath`, see Folders); defaults to one derived from the filename |
+| `@title` | `title` | the lesson name; defaults to one derived from the filename |
 | `@prompt` | `prompt` | markdown for the instructions panel |
 | `@code` | `your_impl` | editor contents (parsed with `MarkerParse.of_text`) |
 | `@test` | `hidden_tests.tests` | defaults to `test true end` |
@@ -58,41 +58,6 @@ Marker lines are *exactly*:
   impl introduces is in scope for them. That is how a lesson whose tests
   reference `first_four` works — and it is why a lesson needing more than one
   name must leave `wrapper` off.
-
-## Folders
-
-A lesson's `@title` is a **`SlidePath`**: a `/`-separated hierarchical name, the
-same convention Documentation-mode slide names use (`"B2T2 / Table API / ..."`).
-The leading segments are folders, the last is the lesson. `src/web/util/SlidePath.re`
-owns that encoding.
-
-```
-Basics / Holes
-Tuple Structural Operations / Labeled Tuple Omission
-```
-
-The folder drives two things: the top bar renders one dropdown per segment
-(`EditorModeView.indicator_select`), and the prev/next arrows — both the
-in-slide pair and the top-bar pair — **iterate only within the current lesson's
-folder**, clamping at its edges. The last lesson of a folder shows "Done! 🎉"
-instead of a next arrow. Crossing folders is done with the folder dropdown. The
-slide's own heading shows just the leaf, since the breadcrumb already names the
-folder.
-
-Authoring rules, pinned by `test/Test_Tutorial.re`:
-
-- Give every lesson **exactly one** folder segment.
-- Never let one title be a proper prefix of another (`Tables` alongside
-  `Tables / Tables`): the shorter one becomes unreachable from the deeper
-  dropdown.
-- Keep a folder's lessons **contiguous** in `src/tutorialslides/Slides.re` —
-  dropdown options appear in first-appearance order.
-
-Whichever lesson is first in `src/tutorialslides/Slides.re` is the one a fresh
-profile opens on (`TutorialsMode.StoreTutorialKey.default`).
-
-Retitling a lesson is safe: identity comes from the `id=<uuid>` flag, not the
-title, so moving a lesson between folders does not reset anyone's progress.
 
 ## Decoding lessons → text
 
--- a/hazel-programs/tutorial/README.md
+++ b/hazel-programs/tutorial/README.md
@@ -87,8 +87,9 @@ Authoring rules, pinned by `test/Test_Tutorial.re`:
   dropdown.
 - Keep a folder's lessons **contiguous** in `src/tutorialslides/Slides.re` —
   dropdown options appear in first-appearance order.
-- Keep `Basics / Holes` first in that list. It is the lesson a fresh profile
-  opens on (`TutorialsMode.StoreTutorialKey.default`).
+
+Whichever lesson is first in `src/tutorialslides/Slides.re` is the one a fresh
+profile opens on (`TutorialsMode.StoreTutorialKey.default`).
 
 Retitling a lesson is safe: identity comes from the `id=<uuid>` flag, not the
 title, so moving a lesson between folders does not reset anyone's progress.
--- a/hazel-programs/tutorial/README.md
+++ b/hazel-programs/tutorial/README.md
@@ -28,7 +28,7 @@ Marker lines are *exactly*:
 
 | marker | maps to | notes |
 |---|---|---|
-| `@title` | `title` | the lesson name; defaults to one derived from the filename |
+| `@title` | `title` | `"<Folder> / <Lesson>"` (a `SlidePath`, see Folders); defaults to one derived from the filename |
 | `@prompt` | `prompt` | markdown for the instructions panel |
 | `@code` | `your_impl` | editor contents (parsed with `MarkerParse.of_text`) |
 | `@test` | `hidden_tests.tests` | defaults to `test true end` |
@@ -58,6 +58,40 @@ Marker lines are *exactly*:
   impl introduces is in scope for them. That is how a lesson whose tests
   reference `first_four` works — and it is why a lesson needing more than one
   name must leave `wrapper` off.
+
+## Folders
+
+A lesson's `@title` is a **`SlidePath`**: a `/`-separated hierarchical name, the
+same convention Documentation-mode slide names use (`"B2T2 / Table API / ..."`).
+The leading segments are folders, the last is the lesson. `src/web/util/SlidePath.re`
+owns that encoding.
+
+```
+Basics / Holes
+Tuple Structural Operations / Labeled Tuple Omission
+```
+
+The folder drives two things: the top bar renders one dropdown per segment
+(`EditorModeView.indicator_select`), and the prev/next arrows — both the
+in-slide pair and the top-bar pair — **iterate only within the current lesson's
+folder**, clamping at its edges. The last lesson of a folder shows "Done! 🎉"
+instead of a next arrow. Crossing folders is done with the folder dropdown. The
+slide's own heading shows just the leaf, since the breadcrumb already names the
+folder.
+
+Authoring rules, pinned by `test/Test_Tutorial.re`:
+
+- Give every lesson **exactly one** folder segment.
+- Never let one title be a proper prefix of another (`Tables` alongside
+  `Tables / Tables`): the shorter one becomes unreachable from the deeper
+  dropdown.
+- Keep a folder's lessons **contiguous** in `src/tutorialslides/Slides.re` —
+  dropdown options appear in first-appearance order.
+- Keep `Basics / Holes` first in that list. It is the lesson a fresh profile
+  opens on (`TutorialsMode.StoreTutorialKey.default`).
+
+Retitling a lesson is safe: identity comes from the `id=<uuid>` flag, not the
+title, so moving a lesson between folders does not reset anyone's progress.
 
 ## Decoding lessons → text
 
--- a/src/web/app/editors/mode/TutorialsMode.re
+++ b/src/web/app/editors/mode/TutorialsMode.re
@@ -33,11 +33,24 @@ module Model = {
     };
   };
   let get_current = (m: t) => List.nth(m.exercises, m.current);
+  /* Lesson titles are SlidePaths, so the lessons form folders. Navigation is
+     scoped to the current lesson's folder; crossing folders goes through the
+     breadcrumb dropdown. Read the raw title, never return_title -- that one
+     appends " ✔". */
+  let paths = (m: t): list(SlidePath.t) =>
+    List.map(
+      (e: TutorialMode.Model.t) => Tutorial.path_of(e.editors),
+      m.exercises,
+    );
 };
 module StoreTutorialKey =
   Store.F({
     [@deriving (show({with_path: false}), sexp, yojson)]
     type t = Haz3lcore.Id.t;
+    /* The lesson opened on a fresh profile: lesson 0, which is
+       "Basics / Holes" because src/tutorialslides/Slides.re lists
+       01-holes.hzt first. Keep the Basics folder first in that list, and
+       01-holes.hzt first within it. */
     let default = () =>
       List.nth(TutorialSettings.lessons, 0) |> Tutorial.id_of;
     let key = Store.CurrentTutorial;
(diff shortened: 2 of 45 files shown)


7h3kk1d:
@disconcision there's folders now if you want to add your stuff

hazel-programs/tutorial/README.md
```diff
@@ -45,6 +45,40 @@ Marker lines are *exactly*:
...
+- Never let one title be a proper prefix of another (`Tables` alongside
+  `Tables / Tables`): the shorter one becomes unreachable from the deeper
+  dropdown.
+- Keep a folder's lessons **contiguous** in `src/tutorialslides/Slides.re` —
+  dropdown options appear in first-appearance order.
+- Keep `Basics / Holes` first in that list. It is the lesson a fresh profile
+  opens on (`TutorialsMode.StoreTutorialKey.default`).
+
+Retitling a lesson is safe: identity comes from the `id=<uuid>` flag, not the
+title, so moving a lesson between folders does not reset anyone's progress.
```
7h3kk1d: If we want these to be true there should be a test

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+ * and Tutorial-mode lesson titles ("Basics / Holes"). The leading segments are
+ * folders, the last is the leaf.
+ *
+ * Modes keep storing these names as plain strings — they are IndexedDB keys
+ * ("doc:" ++ name, see ScratchMode.Persist) and .hzt `@title` text — so this
+ * module is the one place that knows the encoding. It owns both the breadcrumb
+ * decomposition the top bar renders (EditorModeView.indicator_select) and the
+ * folder-scoped arithmetic the navigation arrows use (TutorialsMode). */
+
+/* `show` is used by Test_SlidePath's Alcotest printer. */
```
7h3kk1d: We don't need to justify why show exists.

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+[@deriving show({with_path: false})]
+type t = {
+  folders: list(string),
+  leaf: string,
+};
+
+/* Written between segments by to_string. of_string trims each segment, so any
+   amount of surrounding whitespace parses to the same path. */
```
7h3kk1d: Write comments that are descriptive of what something is and how it should be used not what current call sites are.

src/web/util/SlidePath.re
7h3kk1d, line 1: Please add an rei file so we have an explicit api. Think about what should just be internal.

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+let separator = " / ";
+
+let of_string = (s: string): t => {
+  let segs = String.split_on_char('/', s) |> List.map(String.trim);
```
7h3kk1d: Repeating the separator here. If we changed it it would have to be in two places.

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+  switch (List.rev(segs)) {
+  | [] => {
+      folders: [],
+      leaf: "",
+    } /* split_on_char never returns [] */
+  | [leaf, ...rev_folders] => {
+      folders: List.rev(rev_folders),
+      leaf,
+    }
+  };
```
7h3kk1d: This seems complicated. Maybe we're missing some ListUtil helpers. Either way these lists are small

src/web/util/SlidePath.re
7h3kk1d: Think about where we should be using Seq instead of list.

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+let segments = (p: t): list(string) => p.folders @ [p.leaf];
+
+let to_string = (p: t): string => segments(p) |> String.concat(separator);
+
+let depth = (p: t): int => List.length(p.folders) + 1;
+
+/* The folder a path sits in, "" for a top-level path. */
+let folder = (p: t): string => String.concat(separator, p.folders);
+
+let same_folder = (a: t, b: t): bool => a.folders == b.folders;
```
7h3kk1d: Let's not use poly eq.

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+  let idxs = folder_indices(~current, paths);
+  let pos =
+    ListUtil.findi_opt(i => i == current, idxs)
+    |> Option.map(fst)
+    |> Option.value(~default=0);
+  (pos, List.length(idxs));
+};
+
+/* The index `by` steps from `current` within its folder, clamped at the
+   folder's edges: returns `current` when the step would leave the folder. */
```
7h3kk1d: what does this mean

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+  | [leaf, ...rev_folders] => {
+      folders: List.rev(rev_folders),
+      leaf,
+    }
+  };
+};
+
+let segments = (p: t): list(string) => p.folders @ [p.leaf];
+
+let to_string = (p: t): string => segments(p) |> String.concat(separator);
```
7h3kk1d, line 31: If this is specifically for persistence we should word it as such.

test/Test_SlidePath.re
```diff
@@ -0,0 +1,321 @@
...
+  ),
+  test_case("out-of-range current yields no crumbs", `Quick, () =>
+    check(crumbs, "empty", [], crumb_options(nav_paths, 99))
+  ),
+];
+
+let tests = [
+  ("SlidePath.of_string", of_string_tests),
+  ("SlidePath accessors", accessor_tests),
+  ("SlidePath navigation", nav_tests),
```
7h3kk1d: I like the grouping but I think having the function name in the test name is useful. Think about more hiearchical grouping.

test/Test_SlidePath.re
```diff
@@ -0,0 +1,321 @@
...
+      check(int, "depth", 3, SlidePath.depth(p));
+      check(string, "folder", "a / b", SlidePath.folder(p));
+    },
+  ),
+  test_case("top-level path has no folder", `Quick, () =>
+    check(
+      string,
+      "folder",
+      "",
+      SlidePath.folder(SlidePath.of_string("Holes")),
```
7h3kk1d: I'm surprised this isn't an option as opposed to an empty string. Is there motivation for this?

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,120 @@
...
+  options: list((int, string)),
+};
+
+/* Decompose a path list into the breadcrumb of sibling-pickers for `current`.
+   Options keep first-appearance order (dedup_f), so a folder option jumps to
+   the first path in that folder. A path too short to have a segment at a given
+   depth is skipped rather than raising, so a name that is a proper prefix of
+   another (e.g. "Tables" alongside "Tables / Tables") cannot crash the top bar
+   — it is merely unreachable from the deeper dropdown. */
+let breadcrumb = (~current: int, paths: list(t)): list(crumb) => {
```
7h3kk1d, line 82: Out of curisoity would it be helpful to have a type that represents all of the slide paths as a tree and gave navigation options on the tree as opposed to this current version that just uses the slidepath and names. I'm worried about complexity and not runtime as this is all fast enough.

test/Test_SlidePath.re
```diff
@@ -0,0 +1,321 @@
...
+        ints,
+        "Tables",
+        [2, 3],
+        SlidePath.folder_indices(~current=3, nav_paths),
+      );
+      check(
+        ints,
+        "singleton",
+        [4],
+        SlidePath.folder_indices(~current=4, nav_paths),
```
7h3kk1d: I don't understand this test when singleton isn't in the nav_paths

test/Test_SlidePath.re
```diff
@@ -0,0 +1,321 @@
...
+      );
+      check(
+        pos_pair,
+        "only one",
+        (0, 1),
+        SlidePath.folder_position(~current=4, nav_paths),
+      );
+    },
+  ),
+  test_case(
```
7h3kk1d: All of these index tests that step are kind of hard to see. It would be great if they had the name of current.

src/web/app/editors/mode/TutorialsMode.re
```diff
@@ -33,11 +33,24 @@ module Model = {
...
+    );
 };
 module StoreTutorialKey =
   Store.F({
     [@deriving (show({with_path: false}), sexp, yojson)]
     type t = Haz3lcore.Id.t;
+    /* The lesson opened on a fresh profile: lesson 0, which is
+       "Basics / Holes" because src/tutorialslides/Slides.re lists
+       01-holes.hzt first. Keep the Basics folder first in that list, and
+       01-holes.hzt first within it. */
```
7h3kk1d: this feels noisy. Shorter comment plz

src/web/util/SlidePath.re
```diff
@@ -0,0 +1,98 @@
...
+
+[@deriving (show({with_path: false}), eq)]
+type t = {
+  folders: list(string),
+  leaf: string,
+};
+
+/* Rendered between segments, and split on to recover them. */
+let separator = " / ";
+let separator_char = '/';
```
7h3kk1d: Shouldn't one of these use the other

src/web/util/SlidePath.rei
```diff
@@ -0,0 +1,98 @@
...
+
+   folder(of_string("a / b / c")) == Some("a / b")
+   folder(of_string("Holes"))     == None */
+let folder: t => option(string);
+
+/* (position of `current` within its folder, that folder's size).
+
+   folder_position(~current=0, space) == (0, 2)
+   folder_position(~current=3, space) == (1, 2) */
+let folder_position: (~current: int, list(t)) => (int, int);
```
7h3kk1d: All of these int returning functions are unclear what the ints mean. This just says "position". Expand the comment a bit to be clearer but don't be verbose. Or consider making a record type that's more informative.

src/web/util/SlidePath.rei
```diff
@@ -0,0 +1,98 @@
...
+ * operations at the bottom answer the two questions a slide picker asks of
+ * it: where the arrows may move within a folder, and which sibling segments
+ * to offer at each depth of a breadcrumb. Positions are indices into that
+ * list, since that is how a picker identifies a slide. Each of those
+ * operations is illustrated against one example space:
+ *
+ *     0  "Basics / Holes"
+ *     1  "Basics / Functions"
+ *     2  "Tables / Tables"
+ *     3  "Tables / Column Projection"
```
7h3kk1d, line 16: I think this would be easier if your example didn't reuse folder name as slide name

PR #2525, opened 2026-09-10
Blackboard MVP.

matthewhammer:
## Blackboard MVP: a proof assistant's core logic inside Hazel

An experiment: an MVP of **Blackboard**, a proposed "clean slate" proof assistant whose core logic has a type of all types, an internal typing relation, and no lambda. Three milestones are here: the kernel, the editor sort, and live checking.

**Try it:** Documentation → Blackboard → 0. Overview, at https://hazel.org/build/blackboard-mvp/

**What it does and does not do,** in one line each: the checker is real and finds the paper's own slips; `by proof` is still a placeholder; and there is no conversion, because the logic has no reduction. The long form is in `docs/blackboard.md`, added by this PR, and the open questions are collected in [#2543, what is implemented / what is a placeholder / five open questions for Hazel](https://github.com/hazelgrove/hazel/issues/2543).

### M0 — the kernel

`src/language/blackboard/` — `BbTerm` (terms `x`, `t : T`, `type`, `(x : T₁) → T₂`, `t₁ t₂`; signatures, blocks, documents; substitution, α-equivalence, printing), `BbCheck` (the decidable fragment the paper calls `tychk`, without conversion), `BbError` (errors as data, located by signature entry), `BbParse` (the paper's concrete syntax, for fixtures). No dependency on tiles or the web layer.

Tests use the paper's own signatures as fixtures. The checker finds the paper's two slips in `cong-ap` (a name bound twice; a family used as a type) and the two in `int-elim` (missing arguments), and records where the paper's examples lean on equations the fragment cannot use.

### M1 — the editor sort

`Bb(BbSort.Term)` is a nested sort, following the ALFA derivation sub-language in `src/language/derivation/`: forms in `Form.re`, molding, `MakeTerm`, a pretty printer, and `blackboard ... end` to embed a document in an expression. `BbGrammar` / `BbTermBase` / `Bb` carry editor terms and read them back as kernel terms, entries, blocks and documents.

Two decisions worth review:

- **The colon binds tighter than the arrow**, so `(x : A) -> x : B` reads as the paper writes it. In a signature entry the colon is a declaration instead, and is loosest; `Bb.rotate_entry` rotates the spine back, so `f : A -> B` declares `f` of type `A -> B`.
- **`Insert.effective_sort` no longer falls back to the enclosing sort inside Blackboard.** It is a closed sub-language, so a token with no Blackboard expansion stays a monotile. Without this, typing `type` inside a block expands into Hazel's `type _ = _ in`. This is the one change outside Blackboard-specific code that affects existing behaviour; it is guarded by `Sort.is_bb`.

### M2 — live checking, and a tint per modality

`Statics.bb_to_info_map` gives every Blackboard node an info entry and runs the checker as you type; `BbInfo` carries the status and the block modality; `BbCursorInspector` shows the message; the problem count and sidebar come free from `Info.is_error`.

The checker runs **per signature entry, not per document**, each in the context of the entries before it. An entry that cannot be read at all reports a syntax error and still enters the context, so its neighbours are still checked — an all-or-nothing check goes silent exactly when the user has just made a mistake. `Test_BlackboardStatics` pins this, and pins that the inspector actually says something: a silent editor fails the suite.

Errors land on the name the user got wrong where one is identifiable, and on the entry otherwise. Since kernel terms deliberately carry no ids, an error carries the subterm it is about and the node is found by re-reading; the consequences of that are written up in `docs/blackboard.md` and item 3 of #2543.

`BbSort` also gains display-only `Assumed` / `Constructed` sub-sorts, refined by `Info.refine_sort_from_mold` exactly as the `Drv` sub-sorts are, so a block is tinted by what it owes: an `assume` block is postulated, a `construct` block is a conservative extension that still owes a witness.

### Documentation slides

Six slides under **Documentation → Blackboard**: Overview, Terms, Signatures, Checking, Metatheory, and Case Studies, the last transcribing all four of the paper's Section 4 examples. They load through the typing parser, since menhir has no Blackboard productions (as with the derivation sub-language). Both corpus ratchets ledger them with a reason, and `DocSlides.ReparseBackuptext` pins their text fidelity.

The `Bb` sort also needed a colour: with no entry in the token palette, Blackboard code rendered in the fallback colour and was effectively unreadable. Measured against the code background, it is now 6.49:1, where `Exp` is 5.33, `Typ` 4.19, `Drv` 3.73 and `Pat` 3.32.

`Test_BlackboardSlides` requires every shipped slide to read back as a document. It earns its place: three separate content bugs made part of a slide silently become a hole while the text still round-tripped. Hazel names cannot contain a hyphen, so `falsity-eq` was three tokens; `false` is a boolean literal, not a name; and `witness : a : A` associates as `(witness : a) : A`. Slides therefore use underscores where the paper uses hyphens, noted on the Terms slide. Behind two of them was a reader bug: a signature of three or more entries nests to the right, and only the top level was being read.

### `docs/blackboard.md`

Written to answer two questions from the paper's author — *are the constructions real?* and *how does it typecheck?* — it records, with a line-level citation on every claim: what the checker really implements; that `by proof`, `by definition`, `by tychk` and `by quotient` are all parsed as bare names and nothing discharges them, so a `construct` block is an `assume` block plus a tint plus an undischarged obligation; that substitution on types happens at exactly one site, the `ap` rule; that α-equivalence is the only notion of "same type", deliberately, since the logic has no reduction; and that the membership-hypothesis lookup in `has_type` is the paper's own escape hatch, being rules `hyp` and `in-` rather than conversion in disguise.

It closes with twelve open questions, grouped by who can settle them: five for the paper, three for this implementation, four for Hazel. The Hazel ones are #2543.

### `mech/`

The Rocq mechanization of the core logic (`Blackboard.v`, Rocq 9.2, no axioms) and its compile log, as the reference semantics: consistency of the bare logic, conservativity of construction blocks, and a five-point model. It covers both the rules as first printed and the author's corrected `arrow+`. Not part of the build.

Theorem 2 is what makes `by proof` a placeholder with a precise meaning rather than a gap: a construction block's obligation is the term `(M : type) → ((x₁ : T₁) → … → (xₙ : Tₙ) → M) → M`, and meeting it makes the block a conservative extension for goals that are types. Nothing in the implementation produces or checks such a witness.

### Testing

```
dune build test/idb_stub.js test/haz3ltest.bc.js --profile dev
bash test/run_node.sh test Blackboard
```

34 Blackboard tests pass, across four groups: `Blackboard` (kernel), `Blackboard editor` (reading editor terms), `Blackboard slides`, `Blackboard statics`. The full suite was last run at M2 (9c5a880790) and was green, 3746 tests; the two commits after it are a documentation slide and a reformat, and this one adds only a markdown file.

### Not in this PR

Implicit arguments, user-defined syntax, tactics, equational rewriting, discharging any construction obligation, dynamics for an embedded document, and menhir support.

### DIFF
--- a/docs/blackboard.md
+++ b/docs/blackboard.md
@@ -1,0 +1,244 @@
+# Blackboard in Hazel
+
+An MVP of **Blackboard: A Clean Slate Proof Assistant** — a core logic with a
+type of all types, an internal typing relation, and no lambda — as a
+sub-language of Hazel. Branch `blackboard-mvp`, PR
+[#2525](https://github.com/hazelgrove/hazel/pull/2525).
+
+This document records what is actually implemented, what is deliberately a
+placeholder, and what is open. It was written to answer two questions from the
+paper's author, which are the two questions anyone should ask of a proof
+assistant MVP: *are the constructions real?* and *how does it typecheck?*
+
+| Where | What |
+|---|---|
+| `src/language/blackboard/` | the kernel (`BbTerm`, `BbCheck`, `BbError`, `BbParse`) and the editor sort (`BbSort`, `BbGrammar`, `BbTermBase`, `Bb`, `BbInfo`) |
+| `src/language/statics/Statics.re` | `bb_to_info_map`: checking as the editor runs it |
+| `src/web/blackboard/BbCursorInspector.re` | what the caret says |
+| `mech/Blackboard.v` | the Rocq reference semantics |
+| `hazel-programs/docs/blackboard/` | six documentation slides |
+| `test/blackboard/` | 34 tests |
+
+## What is real
+
+The checker. `BbCheck` is a genuine bidirectional implementation of the
+decidable fragment the paper calls `tychk`: `synth` finds a term's type,
+`is_type` checks formation, `has_type` compares
+([`BbCheck.re:52`](../src/language/blackboard/BbCheck.re#L52)). Every step is
+an instance of a rule of the corrected Figure 1 — `hyp`, `type`, `ap`,
+`arrow`, `in`, and `in-` for hypotheses — so whatever passes is derivable.
+Much that is derivable does not pass, and the paper says so.
+
+It is not a stub, and the evidence is that it finds the paper's own slips:
+
+| Fixture | What the checker reports |
+|---|---|
+| `cong-ap` as printed | `B : A -> type` used bare where `B a` is meant; `a1` bound twice, leaving `a2` unbound |
+| `cong-ap` repaired | exactly one error: `f2 a2 : B a2` against an expected `B a1` — the step needing `a1 = a2` |
+| `int-elim` as printed | two mismatches, `h : M x` where `s` and `p` expect an `int` |
+| `quot` | `eq-rel` is never defined |
+| `arrow-small` | `A : U` is not a type without the coercion the paper leaves implicit |
+
+These are pinned as tests in `test/blackboard/Test_Blackboard.re`, and they run
+on every keystroke in the editor, not only from tests.
+
+## What is a placeholder
+
+**`by proof` is still a placeholder, exactly as in the paper.** So is
+`by definition`, `by tychk` and `by quotient`. A tactic is parsed as a bare
+name ([`Bb.re:221`](../src/language/blackboard/Bb.re#L221), "a tactic is a
+name") and kept as a string.
+
+**There is no definitional mechanism at all** — no defined constants, no
+unfolding, no δ.
+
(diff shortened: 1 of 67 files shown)


matthewhammer:
Closing: not in scope.

The critical path right now is user-defined livelits (`integration/livelits`, #2576, and the splicerefs variant #2580). Blackboard is not on it, so this PR is being closed rather than left open as a signal of active work.

The branch `blackboard-mvp` is untouched and the work is preserved in its history — reopen or branch from it whenever this comes back into scope.

Issue #2522, opened 2026-09-10
Projector `dynamics` flag has had no reader since the MakeTerm probe rework


7h3kk1d:
`Projector.dynamics: bool` (`ProjectorBase.re:209`, forwarded by `Cook` at `:271`) is documented as "this projector will be instrumented with a probe to collect dynamic information during evaluation", but nothing reads it.

It was added in 99b100f72f (2024-12-29, "projector API flag for dynamics collection") with a real consumer: `MakeTerm.should_instrument` looked it up to decide whether a probe-wrapped term actually received a `Probe` wrapper at parse time —

```reason
| (label, [Exp(body)]) when is_probe_wrap(label) =>
  ret(should_instrument(Exp.rep_id(body)) ? Parens(body, Probe(...)) : body.term)
```

That consumer was removed in 9a8cd47887 (2026-01-08, "rm legacy probe code in maketerm") and nothing replaced it. `ProjectorInfo.mk_info` now populates `info.dynamics` from the dynamics map regardless of the flag, so a projector's declared value has no effect either way.

Consequences: the declaration is misleading — someone tracing where a projector's samples come from is sent to a dead end, and setting it `true` on a new projector looks like opting in when it does nothing. Ten implementations plus the signature carry a value nobody consumes.

Either give it a consumer (if per-projector gating of sample collection is still wanted) or delete it from the signature, `Cook`, and the ten implementations. Removal touches 11 files for zero behaviour change, which is why it wants its own PR rather than riding along with feature work.

matthewhammer:
Flagging a use case that appeared *after* this survey, on
`experimental-lang-integration`. Not an objection to the removal — the
premise here was true of `dev`, and I re-checked it: `MakeTerm.should_instrument`
went in `9a8cd47887` and nothing on `dev` replaced it, so ten implementations
really were carrying a value nobody read.

**What now reads it, on our branch.** `CachedStatics.projector_probe_ids`
(`src/haz3lcore/derived/CachedStatics.re:107-116`):

```reason
let projector_probe_ids = (projectors) =>
  Id.Map.fold((id, p, acc) => {
    let (module P) = ProjectorInit.to_module(p.kind);
    P.dynamics ? Id.Map.add(id, (), acc) : acc;
  }, projectors, Id.Map.empty);
```

Those ids join the probe set, and probing a projector's term id is what
populates its `info.dynamics`. So the flag is the opt-in for "this projector
[...]

PR #2526, opened 2026-09-11
user-defined livelits and modules, integrated.

matthewhammer:
**Draft, for inspection only. Not meant to merge:** this branch is a hypothetical integration built to answer one question, and it contains merges of four other open PRs.

## What this is

dev + Modules II parts 1 and 2 (#2493 signature types and module values, #2494 width subtyping) + the HTML substrate (#2115) + the livelit stack (#2411 user-defined livelits, #2488 expansion typing, #2489 error dedup), with two follow-ups on top:

1. **The livelit code ported to real modules.** A livelit definition's `Model`, `Action` and `Expansion` are read as manifest type members of the signature `ModuleHelpers.module_sig_type` synthesizes; the livelit branch's own `collect_type_exports` and its labeled-tuple type-export injection in the `Let` case are gone. At runtime `LivelitProj.record_field` reads `ModVal` bindings by name from the evaluated `Module`. Two small changes; they are what user-livelits will need once Modules II lands.

2. **Prototype: the HTML substrate as opt-in modules.** hazel-html's ~120 global constructor names and seven global type aliases (`HTML`, `Attr`, `Cmd`, `Sub`, `App`, `KeyEvent`, `MouseEvent`) become four builtin module values `Html`, `Attr`, `Cmd`, `Sub` (`BuiltinsADT.HtmlModules`). Members are the constructors as snake_case functions, types are `Html.T`, `Attr.T`, `Cmd.T`, `Sub.T`, `Html.App`, `Attr.KeyEvent`, `Attr.MouseEvent`:

   ```
   Html.div([Attr.style([("color", "red")])], [Html.text("hi")])
   let view = fun m : Model -> ... : Html.T
   ```

   Constructor names are unchanged, so the renderer and MVU runtime are untouched. Motivation: a user module named `App` lost its type members to the builtin alias `App`, because `Typ.path_sig` tries type aliases before variables. With the modules it works.

## Performance

The naive module version was 3-4x slower than the global one. Four fixes, each at its source, bring it within ~15%:

- `sig_project_value` / `sig_project_type` substitute into the requested member only, instead of closing the whole signature to read one member.
- For a builtin module, a member's type keeps `T` as the path `Html.T` (`Typ.sig_project_value_along`), and `meet` accepts two equal paths without normalizing either.
- `normalize_ctr_type` keeps a builtin-module path compact, as it kept `Var("HTML")`.
- A builtin module's member elaborates directly to the constructor it denotes; otherwise the evaluator substitutes the whole module value into every closure that names `Html` (80 copies in one program's output).

| nutrient-rotation.hz | baseline (global names) | naive modules | final |
|---|---|---|---|
| `cli analyze` | 2.3 s | 8.1 s | 2.8 s |
| `cli run` | 2.75 s | 11 s | 3.1 s |
| printed output | 181 KB | 1.3 MB | 181 KB |

## Verified

- All 14 MVU and livelit example programs type-check in the qualified form (ported, along with the docs, `Test_MVU` and `Test_UserLivelits`).
- 1370 tests across MVU, UserLivelits, Modules, Statics, Evaluator, Equality, Elaboration, Menhir, FastParse, Unicode pass. The three tests pinning the compact `Var("HTML")` annotation are removed: that fast path applied only to global aliases.
- In the browser: the MVU counter renders and counts on clicks, the slider livelit renders and updates with its expansion type check passing, and the Modules slide's item 9 (module named `App`) no longer errors.

## Known gaps

- The projectors doc slide reports type errors around the builtin `^^sliderf` and `^^text` livelits; those lines were not touched here and the errors predate the prototype (they belong to the merge, not the modules design).
- Bare `Int(x)` still type-checks because JSON's constructor `Int` is still global, so a missed port is silent. Same collision in miniature.
- Path-preserving projection and constructor elaboration apply only to builtin modules (variables with `Id.invalid`); user modules keep Modules II's manifest substitution.


matthewhammer:
Merged `dev` into this branch (b6bfd58b6d) to bring it back to mergeable. Only two files conflicted, but one of them needs a decision above my pay grade.

@cyrus- — this one is yours: **this branch now deliberately carries back a cleanup that `dev` already merged**, and I'd like a ruling on whether that is the right long-term shape.

### What happened

#2522 observed that `Projector.dynamics: bool` had no reader — its last consumer, `MakeTerm.should_instrument`, went in 9a8cd47887 — and 209132714b deleted it from the signature, from `Cook`, and from the ten implementations as a zero-behaviour-change cleanup.

That is accurate for `dev`. It is not accurate here. This branch grew a reader in e896245aba (2026-07-24), about seven weeks before the issue was filed, so an audit of `dev` could not have seen it:

```reason
let projector_probe_ids =
    (projectors: Id.Map.t(Base.projector)): Id.Map.t(unit) =>
  Id.Map.fold(
    (id, p: Base.projector, acc) => {
      let (module P) = ProjectorInit.to_module(p.kind);
      P.dynamics ? Id.Map.add(id, (), acc) : acc;
    },
    projectors,
    Id.Map.empty,
  );
```

That set is unioned into `probe_ids_of_zipper`, which reaches `Statics.mk(~probe_ids)` and sets `~is_probed` on the term, which is what puts samples at the id for `ProjectorInfo.mk_info`'s `Dynamics.Map.lookup(p.id, dynamics)` to find. So the issue's "`mk_info` populates regardless" stays literally true — it does not branch on the flag — but here the lookup only *finds* anything when the flag put a probe at that id first. The gating moved upstream of `mk_info` rather than disappearing.

It is load-bearing, not vestigial: the merge does not compile without it (`Unbound value "P.dynamics"` at `CachedStatics.re:112`), and `HTMLProj` / `LivelitProj` need a probe at their own id to have any live value to render.

### What I did, and the alternative I did not take

I restored the signature entry, the `Cook` forwarding and the ten values. `dune build @check --profile release` passes, the quick suite is green (3898 tests), and in the browser the user-defined slider livelit renders and updates: dragging one took it 25 -> 88 with the program result following 100 -> 163, expansion type check passing.

The alternative was to keep the deletion and gate probe installation some other way — realistically by probing every projector unconditionally. That is a behaviour and performance change affecting every projector, and burying it in a merge commit seemed worse than carrying the revert visibly.

### The question

Neither option is obviously right, and it is really a question about `dev`, not about this draft:

1. **Per-projector probe gating is a feature we want.** Then the flag should come back to `dev` with its doc comment naming the real reader, and setting it `true` on a new projector means something again — it is what gets that projector sampled at all.
2. **Every projector should just be probed.** Then this branch should drop `projector_probe_ids` and eat the cost, and the #2522 deletion stands unqualified.

I've left a longer writeup on the issue itself: https://github.com/hazelgrove/hazel/issues/2522#issuecomment-5685825043. Nothing needs undoing on `dev` either way — this is about which shape the livelit work should land in.

(The other conflict was routine: `dev` wrapped `Editor.Update.update` in `PerfMetrics.time_perform` for the profiling sidebar while this branch computes `is_edit` up front for `probe_all`; I composed both.)

matthewhammer:
Retired in favour of #2576.

This branch stops at the `dev` merge. The work continues on `integration/livelits`, which is this plus Modules II part 3, the builtin `Livelit` signature type, and the definition-site check that `expand : Model -> Expansion`.

The active livelit line is now a pair: `integration/livelits` (#2576) without SpliceRefs, and `integration/livelits-splicerefs` with them. This PR stays as the record of the dev-merge state and the `Projector.dynamics` question raised on #2522.

matthewhammer:
Closing: this branch has been shadowed and is no longer active.

The livelit work continues on a pair of branches, both descended from this one:

- **`integration/livelits`** — #2576. This branch plus Modules II part 3, the builtin `Livelit` module signature type, the definition-site check that `expand : Model -> Expansion`, and the livelits documentation deck. **Without** SpliceRefs.
- **`integration/livelits-splicerefs`** — branch only, no PR yet. **With** SpliceRefs, once Matt Keenan's `splices-11` (#2415) is merged and the `SpliceRef` type is built. Currently still identical to `integration/livelits`.

Nothing here is lost: the commits are ancestors of both. This PR stays as the record of the `dev` merge, and of the `Projector.dynamics` question it turned up (#2522), where the flag it removed as unread turned out to have a reader on this branch.

PR #2528, opened 2026-09-11
Printer and incremental-eval fixes surfaced by the modules stack

7h3kk1d:
Three fixes to code that is on `dev` today, surfaced while building the modules stack (#2496). They were originally committed inside parts 2 and 4 of that stack; they are lifted out here so they can land on their own and the stack can sit on top. Each commit carries a test that fails without its fix on `dev`.

## Incremental eval: a binding's ascription is part of its provenance

Editing a binding's ascription in place left its dependents reusing values computed under the old type: the reuse provenance recorded only *that* the binding went through an ascription, not which type. An ascription changes the bound value (a cast; further up the stack, sealing a module), so the type now travels in the provenance as `Ascribed(Typ.t)`, compared with `Typ.fast_equal`.

Pinned by retyping `let x : Int = 1 in x` as `let x : Bool = 1 in x` in place: fresh evaluation reports the inconsistency, the stale reuse handed back the old `1`. (Retyping `Int` as `?`, or relabeling a tuple ascription, is *not* observable on `dev` — casts are stripped by the comparison and a projection's index comes from the new static type — which is why the test uses the inconsistent form.)

## Keep a printed `:` apart from a following operator

`:` is one of `Token.ascii_operator_chars`, and `Token.is_potential_operator` matches any run of them, so a `:` printed flush against a leading-plus sum lexed as the single operator token `:+`, which is no form at all; Menhir has its own lexer and split them, so the two parsers disagreed. `should_add_space` already kept `:` apart from `$` and `!`, as a list of two characters; it now asks `Token.begins_with_potential_operator` for the property they share (`$` stays listed: it is a name character). Pinned at the `should_add_space` level; reaching a `:` with an unparenthesized sum in a real program takes a signature member, so that form stays in the stack.

## Stop re-parenthesizing the content of a Parens type

Printing a parenthesized min-precedence type was not idempotent: `type T = + Adid` came out as `type T = (+ Adid)` and each further trip added a layer. `parenthesize_typ`'s `Parens` case ran its content through `paren_typ_at(Precedence.min)`, which wraps exactly the types whose precedence *is* min — a bare sum, a multihole, and up the stack a Sig — on every pass. The wrapper being emitted is already the protection, so the content needs none; `already_paren` is still passed down. Pinned by printing `type T = + Adid in ?` twice.

The last two were found by the Slow `Menhir and maketerm are equivalent` fuzz property on the stack's CI; the non-idempotence is why its counterexamples could not be pinned with the round-trip helpers.

https://claude.ai/code/session_01UQDwH6tMjSMGSSykhjgQUP

### DIFF
--- a/src/haz3lcore/pretty/ExpToSegment.re
+++ b/src/haz3lcore/pretty/ExpToSegment.re
@@ -727,11 +727,11 @@ and parenthesize_typ =
 
   // Other forms
   | Parens(t) =>
-    Parens(
-      parenthesize_typ(~already_paren=true, t)
-      |> paren_typ_at(Precedence.min),
-    )
-    |> rewrap
+    /* No defensive parens on the content: the wrapper we are emitting is
+       already the protection, and adding another made printing
+       non-idempotent for every type whose precedence IS min -- a Sig, a bare
+       sum, a multihole -- which gained a paren layer on every trip. */
+    Parens(parenthesize_typ(~already_paren=true, t)) |> rewrap
   | Projector(data, t) =>
     Projector(data, parenthesize_typ(t) |> paren_typ_at(Precedence.min))
     |> rewrap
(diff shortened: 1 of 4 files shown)


7h3kk1d:
Folded into #2493: these three commits now sit at the bottom of that PR's history and its description covers them, so the modules stack stays nine PRs on GitHub.

PR #2529, opened 2026-09-11
Fumola with Hazel editor tiles.

matthewhammer:
Integrates Fumola with Hazel as a **sort in the tile grammar** rather than as
a livelit holding an opaque string. The design document is
[`docs/fumola-tiles-design.md`](docs/fumola-tiles-design.md); the design as
first written is kept below, with progress and plans above it.

```
fumola $graphical as store in
  { let p = 0 := 41; @ p }
end
```

## Progress

| | | |
|---|---|---|
| **M0** | AST and printer | ✅ `FumolaGrammar`, `FumolaTermBase`, `FumolaPrint`, checked against the real `fumola_parser` |
| **M1** | the sort and its tiles | ✅ `Fumola(Exp)` / `Fumola(Name)`, forms, precedence, `MakeTerm`, `ExpToSegment`, closed sort |
| **M2** | running, and the value bridge | ✅ runs against the instance it names; `FumolaValue` / `FumolaSource` ported |
| **M3** | the event list | ✅ a panel beside errors and probes, showing the instance's adapton events |
| **M4** | reading Fumola back | ✅ `FumolaParse`, for the round trip that checks structure |
| **M5** | reaching the library | ✅ strings, `import`, `.` projection and `()`, so a tile program can name and call one of the 18 shipped modules |
| **M6** | the cursor inspector | ✅ every Fumola subterm names its form, where the panel used to report whitespace |

Verified in the running app, not only in tests: a program evaluates and its
result comes back as a Hazel value, the event list shows what the instance
recorded, and the documentation deck loads and runs. Full suite: 3917 tests,
no failures.

M5 note: the runtime always shipped `fumola/collections/levelTree` and the
other 17 modules; what was missing was the syntax to say so. Four forms close
that, and they are only useful together — a string literal for the path,
`import` to bind the module, `.` to reach into what it bound, and `()` for
the no-argument call. Checked against the runtime itself, not only in tests:
with the wasm loaded, `do { import Seq = "fumola/collections/levelTree";
Seq.testFromList(()) }` returns an adapton pointer to a real level tree.

None of the four needed a new spelling. `$tag` exists because `#` is Hazel's
comment delimiter; the double quote collides with nothing, and Fumola's own
`LetImport` production marks the `=` in `import P = "path"` as sugar it
accepts either way.

**The syntax**, after moving the binding where `as` belongs:

- `fumola <mode> as <instance> in <program> end` — the instance names the VM
  the program runs against, and is claimed by that name, so the adapton store
  survives every edit that leaves the name alone.
- `hazel … end` inside a program — a Hazel subtree, edited as Hazel, rendered
  as Fumola source on the way in. The livelit had one `input` slot; this can
  stand anywhere a Fumola term can, as many times as the program wants.
- `$tag` for Fumola's `#tag`, because `#` is Hazel's comment delimiter and
  cannot be tokenized. The printer puts the `#` back.

**Documentation**: a *Fumola (Tiles)* deck — Overview, Instance and mode, A
level tree, Hazel inside — named apart from the livelit branch's deck.

## Outstanding, for whoever picks this up

Short notes, not filed issues. Each says where the gap is, so it can be
found without re-deriving it.

- **A `Tile.reassemble` assertion, hit in the app and not reproduced.**
  `assert(List.sort(Int.compare, shards) == shards)` at
  `src/haz3lcore/tiles/Tile.re:117` — a tile's shards out of order after
  being separated and put back. The full suite is green and so is a test
  that parses **every prefix of every tile program** (typing passes through
  each one), so it is not on the insertion path; `reassemble` runs on
  deletion and movement. No repro yet. If you hit it, note what you typed.
- **`ExplainThis` has no entries for Fumola forms.** The cursor inspector
  now names every form (below), but the docs panel says only the form's
  name and *no documentation available*. One entry per form is the work.

Closed since this section was written:

- ~~Nothing inside a Fumola program reports in the cursor inspector.~~
  Done. The *Whitespace or Comment* the panel showed was the **no-info
  fallback** in `CursorInspector.view`, not a claim about the token:
  `Statics.any_to_info_map` filed no entries at all for Fumola subterms.
  `FumolaCls`, `FumolaInfo` and `InfoFumola` now do, and two id bugs that
  hid entries from the lookup are fixed (MakeTerm consumed `$tag` and inner
  `;` blocks without keeping their ids; a `let` shared its body's id).
- **Renderable-but-unbuildable forms.** `ExpToSegment.fumola_to_pretty` ends
  in an `unbuildable()` list — arrays, `?e`, unary, `not`, `~`, `e!`,
  `assert`, `ignore`, `return`, backtick names, `prim`, `if`, `switch`,
  `do @`, `do goto`/`do within` — each in the AST with no tile. They render
  as a hole on purpose: rendering the nearest printable part instead was
  tried on this branch and reverted, because it drops structure silently.
  The `do` family is the notable absence, since it is adapton navigation.
- **`var` and `func` declarations render as a hole** for the same reason:
  they have no tile. `let` and `import` do, and now render with their
  binders — before this PR's last commits a `let x = e` came back as bare
  `e`. Binder patterns cover `PVar` and `PWild` only.
- **No menhir productions for `fumola … end`,** as with the ALFA derivation
  sub-language, so the four doc slides are ledgered in
  `test/Test_FastParseCorpus.re` (MenhirCorpus inherits that ledger). Remove
  the entries if Fumola gains menhir support.
- **Every keystroke re-runs the program,** because elaboration is where it
  runs. Defensible for a cheap program and not for a library call; there is
  no on-demand path. Open question 2 in the design doc.
- **The in-app editor could not reliably be driven from the
  browser-automation pane** (synthetic keystrokes do not reach Hazel's
  hidden textarea, and clicks land only intermittently), so the M5 forms
  were checked through the shim and the test suite rather than by typing
  them into a running editor. The cursor inspector *was* confirmed live.
  Worth a human pass either way.
- **A documentation slide for the round trip** now ships with its probes
  already in it — `^^probe(...)` is a textual projector invocation, so a
  slide can carry probes rather than asking the reader to add them.

## Open, and planned

- **M5, scenes.** 3D output from a Fumola program, shown in Hazel. A scene is
  already plain JSON and reachable through the existing shim, so no new wasm
  export. Measured: this browser allows **16** live WebGL contexts, and
  quietly kills the oldest beyond that — so SVG projection is the option that
  has no ceiling, and an iframe around the existing replayground is the
  fastest demo.
- **An "All Fumolas" PR**, merging this branch with `fumola-livelit-mvp` so
  both integrations and both decks read as one thing. Not started.
- **A `hazel … end` carries a value written in place, not a bound variable.**
  Lifting that means running during evaluation rather than elaboration, which
  means the shim in the web worker — measured: the worker is where a cell's
  result is computed, and it has no `window.fumola`. That would move the
  adapton store out of reach of the event panel, so the runtime stays on the
  main thread for now.
- **Opening the shipped `.fumola` corpus** is not a parser problem: all
  fourteen files are `module { … }` of `public func`s with type annotations,
  and the type sublanguage is deliberately absent. It waits on types.

---

### Where this comes from

Two recent efforts, combined:

- **#2509 / `fumola-livelit-mvp`** — the wasm integration: Fumola source goes
  to the runtime through the `window.fumola` shim, values come back as Hazel
  values. That stays.
- **`blackboard-mvp`** — a bespoke AST hanging off `Grammar.any_t`, with its
  own tiles, its own closed sort, and syntax errors reported the way Hazel
  reports its own. That technique gets repurposed, now for Fumola.

### What the tile route buys

Unmatched delimiters become Hazel's business — `thunk {` with no `}` is a
tile in the backpack with the editor guiding the fix, not a `ParseError`
reported after the fact. Errors get locations, because every node carries an
`Id`. The cursor inspector can say what a Fumola form is. Programs can be
built by inserting forms, not only by typing characters that happen to lex.

### Three findings that shaped the design

1. **There is no parser to write.** Hazel's tile engine is the parser. What
   gets authored is the form table, a closed sort, and `MakeTerm` /
   `ExpToSegment` cases. Blackboard's `BbParse.re` is not on the editor path —
   it feeds the kernel tests.
2. **The obligation moves from parsing to printing.** This route never sends
   text that could fail to parse, so the printer is what has to be right,
   under a contract the livelit route never had: everything Hazel prints must
   be accepted by `fumola_parser` and mean the same thing there. That wants a
   golden corpus run through the real Rust parser in CI.
3. **A tile tree has no model, and the Fumola VM instance needs one.** The
   livelit model carried `instance_id` and `thunk_name` across edits, which is
   what let a thunk keep its history. Blackboard never hit this, because
   `blackboard … end` is inert. Deriving the key from the expression's `Id`
   reintroduces the failure the livelit's own comment warns about.

   **Decided: the program names its instance**, as
   `fumola <instance> in … end`. The name is text the programmer wrote, so it
   is stable by construction and survives a reload, and two blocks can share
   one instance. "Instance" is the wasm API's own word — `FumolaInstanceId`,
   `fumola_create`, `fumola_drop`. The runtime is the wasm module; an instance
   is one VM inside it.

### What the Rust toolchain costs to iterate on

Measured to inform the decision, not to justify it — the case for tiles is
the editing affordances above. Same laptop, caches warm; a no-op build of the
wasm target is 0.1 s, so each figure is work the edit caused.

| Change | Wall |
|---|---|
| lalrpop regen, `fumola_parser` only (debug) | 89.7 s |
| grammar edit → release wasm artifact | 104.1 s |
| **a new prim** — one arm in `PrimFunction::resolve`, grammar untouched | **109.5 s** |
| `Form.re` change rippling downstream (`dune build src`) | **21.6 s** |
| `Form.re` change confined to the module body | **0.66 s** |

The prim row is the one that matters, because it is the edit that actually
happens — roughly weekly — while Fumola's *syntax* stays mostly put. A prim
never touches `parser.lalrpop`, and lalrpop does not regenerate its table
(`build.rs` sets `rerun-if-changed` on the grammar file only). But
`fumola_parser` depends on `fumola_syntax` and so recompiles the generated LR
code for an 834-line grammar anyway. A one-line prim addition pays
essentially the full grammar-edit price.

The cost is a human one. Two minutes is long enough to lose the thread and go
read something else, short enough that you sit there; paid on every iteration
of a prim you are designing by trying, it quietly decides how many variations
you try. Sub-second turnaround does not make the total smaller so much as
keep it under the threshold where attention survives. For agent-driven work
the calculus is different — a two-minute build is just two minutes.

Not claimed: anything about *running* programs. Both routes need the wasm
runtime, and that cost does not go away.

### Scope

Fumola's surface is Motoko-derived — 834 lines of lalrpop covering objects,
classes, actors, generics, candid, async, quoted ASTs. Tiles will not cover
that. Three bits are actively hostile to tiles and shape the subset: `#` is
overloaded three ways (`#tag`, concat, `#[`), `<`/`>` are both relops and
type brackets, and `" << "` is a token with literal spaces in it.

M1 centers on the adapton core — `thunk`, `force`, `:=`, `@`, `do @`,
`do goto`, `do within`, backtick names — and defers modules, types and
classes.

### Milestones

- **M0** — `FumolaGrammar`, `FumolaTermBase`, `FumolaPrint`; round-trip tests
  against the real `fumola_parser`.
- **M1** — the sort and its tiles; `fumola … end` as an expression form.
- **M2** — evaluation and the value bridge (`FumolaValue.re` and
  `FumolaSource.re` cherry-pick over from `fumola-livelit-mvp`; they carry no
  livelit dependency). Statics, which is where Blackboard stopped.
- **M3** — `FumolaParse.re`, so the existing `.fumola` corpus opens as tiles.

Branched off `dev` rather than off `fumola-livelit-mvp`: that branch carries
~950 lines of livelit changes this feature does not want, and the pieces that
are wanted are self-contained.


matthewhammer:
Closing: superseded by `experimental-lang-integration` (#2536), which carries this work forward.

The branch `fumola-tiles-mvp` stays as-is and its history is preserved; nothing here is lost.

Also for the record: Fumola-in-Hazel is shelved for now, and its issues are being filed in `Adapton/fumola` (umbrella Adapton/fumola#135) rather than here. The critical path is user-defined livelits — #2576 and #2580.

PR #2530, opened 2026-09-11
Keep delegation keys by stack reference (#2524)

### LINKED ISSUE #2524: Ascription steps make evaluation quadratic in call-stack depth (regression in Probes IV)
## Summary

An ascription evaluated inside a recursive call now does work proportional to the current call-stack depth. A linear program becomes quadratic, and past roughly 10k frames it dies with a JS heap OOM rather than merely running slowly. This regressed in Probes IV (#2363) — dev immediately before that merge is linear.

## Repro

```hazel
let sum : Int -> Int =
  fun n -> if n == 0 then 0 else n + (sum(n - 1) : ?) in
sum(12800)
```

`./hazel run`, wall clock, baseline startup ~1.1s:

| n | `ac4df9feef` (before Probes IV) | `a816823aee` (Probes IV) |
|---|---|---|
| 3200 | 1.20 | 1.50 |
| 6400 | 1.29 | 2.62 |
| 12800 | 1.49 | **OOM** |
| 25600 | 2.04 | **OOM** |

OOM is `FATAL ERROR: Ineffective mark-compacts near heap limit — JavaScript heap out of memory`. Every pre-merge run completes with the correct answer.

Delete the `: ?` and dev is linear again — 1.21 / 1.29 / 1.47 at 3200 / 6400 / 12800 — so it's the ascription, not the recursion.

## Cause

`Evaluator.re:246`, in the delegation bookkeeping Probes IV introduced:

```reason
(DHExp.rep_id(exp), CallStack.ids_of_stack(call_stack)),
```

`ids_of_stack` is `List.map((f: frame) => f.id, cs)` (`CallStack.re:29`) — O(stack depth), allocating a fresh list. It runs on every step whose kind is `may_delegate: true`, and per `Transition.re:186–188` those are exactly `Ascription`, `AscriptionTypAp` and `AscriptionAp`. One ascription step per frame at depth N gives Θ(N²) work, and since the entries are retained in `delegations`, Θ(N²) live memory — which is what actually kills it.

Counting the steps and summing the stack depth at each confirms the shape: 2000 steps → 1,997,003; 4000 → 7,994,003; 6000 → 17,991,003, i.e. k²/2.

Patching out only that `ids_of_stack` call restores linearity exactly: 1.23 / 1.32 / 1.52 / 2.02 at 3200 → 25600.

## Possible fixes

1. **Gate the insert on the redex id being a probe target.** The list is only consulted inside the `Some` branch of `switch (is_target)` at `Evaluator.re:298–302`, so an entry whose id isn't a target can never match. Semantics-preserving, and makes every unprobed program free. It does not fix a probed ascription inside deep recursion.
2. **Replace the materialized id list with a cheaper stack identity** — an incrementally maintained fingerprint, or keying on the stack value itself. This fixes it in general.

## Notes

Measured via the CLI on two worktrees at the commits above; the evaluator is untouched by any branch involved. `delegations` / `ids_of_stack` do not exist pre-merge (0 occurrences in `Evaluator.re` vs 26 after).

disconcision:
Fixes #2524.

`sum(n - 1) : ?` inside a recursive call was quadratic in depth and died with a JS heap OOM at 12800 frames; without the ascription it was linear. Every delegating ascription step projected the whole call stack to an id list and kept it in `delegations`.

Keep the stack by reference instead (the continuation runs on the very value that was declared, so physical equality hits first), and only declare delegations for probe targets, since no other key is ever looked up.

`./hazel run`, wall clock, this machine (dev → fixed): 3200: 3.2 s → 2.0 s · 6400: 7.0 s → 1.8 s · 12800: OOM → 2.1 s · 25600: — → 4.3 s.

Evaluator, probe and trace test groups green (560 tests).

### DIFF
--- a/src/language/dynamics/Evaluator.re
+++ b/src/language/dynamics/Evaluator.re
@@ -70,9 +70,24 @@ module Eval = Transition(EvaluatorEVMode);
  * Transition.provenance_of_kind) pushes the redex's span key here, and a
  * nested evaluation matching it CONTINUES the enclosing observation span
  * instead of opening its own (the delegation law; consumed in
- * eval_3_record_probe_sample). Keys are (syntax id, call-stack instance
- * as ids); genuine re-entry (recursion) always differs in stack. */
-type delegation = (Id.t, list(Id.t));
+ * eval_3_record_probe_sample). Keys are (syntax id, call-stack instance);
+ * genuine re-entry (recursion) always differs in stack. The stack is
+ * kept by reference, not projected to ids: projecting is O(depth) per
+ * delegating step, which made an ascription inside deep recursion
+ * quadratic (#2524). Only probe targets are declared, since only a
+ * target's key is ever looked up. */
+type delegation = (Id.t, CallStack.t);
+
+let continues_delegation =
+    (expr_id: Id.t, call_stack: CallStack.t, delegations: list(delegation))
+    : bool =>
+  List.exists(
+    ((id, stack): delegation) =>
+      Id.equal(id, expr_id)
+      /* the continuation runs on the very stack value that was declared */
+      && (stack === call_stack || CallStack.equal(stack, call_stack)),
+    delegations,
+  );
 
 let rec evaluate =
         // Constants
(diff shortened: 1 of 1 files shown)


src/language/dynamics/Evaluator.re
```diff
@@ -70,9 +70,24 @@ module Eval = Transition(EvaluatorEVMode);
...
+type delegation = (Id.t, CallStack.t);
+
+let continues_delegation =
+    (expr_id: Id.t, call_stack: CallStack.t, delegations: list(delegation))
+    : bool =>
+  List.exists(
+    ((id, stack): delegation) =>
+      Id.equal(id, expr_id)
+      /* the continuation runs on the very stack value that was declared */
+      && (stack === call_stack || CallStack.equal(stack, call_stack)),
```
7h3kk1d, line 88: Not advocating we change it now but this seems prone to issues in refactors if we're not careful.

7h3kk1d (approved):
Not as part of this PR but we really need regression tests for the linear functions.

PR #2532, opened 2026-09-11
Swap ¿ markers for Grout in place (#2518)

### LINKED ISSUE #2518: A lone ¿ inside a container is dropped by the text round-trip: [¿] reloads as []
`MarkerParse.of_text` recovers Grout by *deleting* the `¿` marker tile and letting the load-time regrout put Grout back — `strip_implicit_holes` does `Select.tile` then `Destruct.go` ([`src/haz3lcore/zipper/MarkerParse.re:94-111`](https://github.com/hazelgrove/hazel/blob/dev/src/haz3lcore/zipper/MarkerParse.re#L94-L111)). That only recovers a hole where regrout would independently place one, so a marker whose removal leaves a **complete** term is lost.

The reachable case is a hole as a container's sole element:

| authored | after one round-trip | |
|---|---|---|
| `[¿]` | `[]` | ✗ dropped |
| `[¿, 1]` | `[¿, 1]` | ✓ |
| `[1, ¿]` | `[1, ¿]` | ✓ |
| `(¿, 1)` | `(¿, 1)` | ✓ |
| `let x = ¿ in x` | `let x = ¿ in x` | ✓ |

Delete the marker in `[¿]` and `[]` is already a well-formed empty list, so regrout has no reason to insert anything.

This is a persistence bug, not just a notational one: grout inside `[]` is a state the editor reaches by typing — inserting a space in `[|]` gives `[ <>|]`, per example III of #1077 — and `to_text` prints it as `[¿]`, so the document does not survive a reload. Same shape as #2450 (a dangling `;` dropping every probe pin from the round-trip), which was fixed.

## Repro

Put `let xs : [Int] = [¿] in xs` in a tutorial lesson's `@code` and run `./hazel tutorial-verify --verbose`:

```
List Literals                      impl:MISMATCH  tests:OK
  --- impl before roundtrip ---
let xs : [Int] = [¿] in
xs
  --- impl after roundtrip ---
let xs : [Int] = [] in
xs
```

## Fix direction

Replace the marker tile with a Grout piece directly instead of destruct-then-regrout, so an authored hole is preserved wherever it was written rather than only where regrout independently agrees one belongs.

## Related

- **#2479** adds the right gate for this class — `TextRoundtrip.EditedGrout` requires grout to survive the load-time regrout and reprint as a fixed point — but its corpus is grout left by *simulated typing* and doesn't cover the sole-element case. It keeps the destruct mechanism and adds a second pass for `⧖`, so it does not fix this.
- **#2406** (grout-free edit state, holes as derived display cells) would obsolete the problem entirely.
- **#2455** is a third `¿` gap: markers inside projector syntax are never found or destructed.
- **#1077** documents that grout inside `[]`/`()` is reachable by typing.

disconcision:
Fixes #2518.

`let xs : [Int] = [¿] in xs` reloaded as `[]` on the slow parse path: `MarkerParse.of_text` destructed the marker tile and relied on the edit-time regrout to put Grout back, which only happens where regrout independently wants a hole. `(¿)` and `f(¿)` were dropped the same way.

Replace each marker tile with a convex Grout of the same id, so an authored hole stays wherever it was written. The new `TextRoundtrip.SoleHoles` cases also check the hole survives `remold_regrout` (the fast-path load runs one); five of six fail on dev.

Note for #2479: the concave marker can take the same swap with `shape: Concave` instead of a second destruct pass.

### DIFF
--- a/src/haz3lcore/zipper/MarkerParse.re
+++ b/src/haz3lcore/zipper/MarkerParse.re
@@ -11,15 +11,15 @@
    round-trip distinct from implicit Grout.
 
    Reparsing marked text yields literal `¿` TILES, which the original
-   program never had. [of_text] parses, then Destructs each marker
-   tile; the remold/regrout pass that runs on every edit re-inserts
-   Grout wherever shape requires it, reconstructing the original
-   zipper. This is the RECOVERING-PARSER half: the fast path reads the
+   program never had. [of_text] parses, then swaps each marker tile for
+   a Grout piece in place. (It used to Destruct the tile and let the
+   edit-time regrout put Grout back, but that only recovers a hole
+   where regrout independently wants one: `[¿]` came back as `[]`,
+   #2518.) This is the RECOVERING-PARSER half: the fast path reads the
    same markers structurally during its weave (see FastParse), but the
    recovering parser has no notion of `¿`, and incomplete programs —
    the grout-heavy ones — are exactly what falls back to it. Sits
    below PersistentZipper so persistence loading can fall back to it. */
-open Util;
 open Base;
 
 let default_implicit_hole = Token.implicit_hole_marker;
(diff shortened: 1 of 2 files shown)


7h3kk1d (approved):
Nice, the code looks cleaner too.

PR #2533, opened 2026-09-11
Anchor the table column menu to its header cell (#2521)

### LINKED ISSUE #2521: Rich table probe: the column (⋮) menu opens at the bottom of the drawer instead of under the column header
Since #2363 — specifically `7428e0e88b`, which retired the anchored modal.

### Repro

Open a table rich probe in the drawer with few enough rows that the drawer doesn't scroll, then click a column's ⋮. The menu appears below the bottom edge of the table rather than under that column's header.

https://github.com/user-attachments/assets/0dbcc198-08bf-4da4-bcd6-6a41198c4d60

### Why

- The `.cm-*` rules place the menu with `top: 100%; left: 0` and are documented as anchoring "inside the `<th>` cell" (`proj-table-probe.css:34-68`). Those offsets resolve against the nearest **positioned** ancestor.
- The only rule that ever made the `th` positioned is `.modal .table th { position: relative }` (`proj-table-probe.css:138-146`), and `.modal` is no longer rendered: `let modal_nodes = []; let _unused_modal = modal_overlay;` (`ProbeProj.re:2489-2492`).
- So the containing block is now `.below-wrapper .rich-drawer { position: relative }` (`proj-probe.css:822-824`) — and `top: 100%` is the bottom of the whole drawer.
- Corroboration: a **tall** table gets `.below-wrapper .rich-drawer.overflowing .table th { position: sticky }` (`proj-probe.css:844-847`), which does establish a containing block, so those tables place correctly. Only short / non-overflowing drawer tables are wrong.
- `Menu.direction_from_id` (`src/util/Menu.re:538-546`) measures the ⋮ button against `#main` and only picks a class name; placement is entirely CSS. The flip decision is honest; the anchor is wrong.

### Suggested fix

Give the header cell a containing block that doesn't depend on the retired modal — e.g. `.table th:has(> .column-menu) { position: relative }`, or hang it off the `menu-open` class the `th` already gets while its menu is open (`TableRenderer.re:433-441`). An unconditional `.table th { position: relative }` would fight the two `position: sticky` rules (`proj-probe.css:844`, `proj-table.css:63`).

### Cleanup that belongs with the fix

Retiring the modal left dead code behind, and that is what broke this anchor:

- `modal_overlay` is unreachable, kept alive only by `let _unused_modal = modal_overlay;` (`ProbeProj.re:2370-2414`, `:2489-2492`).
- The whole `.modal …` block in `proj-table-probe.css:80-169` is orphaned.
- The `nut-menu` class in the column menu's class list is inert — every `nut-menu.css` rule is scoped to `#top-bar`.

disconcision:
Fixes #2521.

In a short (non-scrolling) drawer, a column's ⋮ menu opened at the bottom of the table: retiring the anchored modal removed the only rule that positioned the `th`, so `top: 100%` resolved against the whole drawer. The cell with an open menu is now its own containing block (`.table th.menu-open { position: relative }`); tall drawers already had one via sticky headers, and the standalone table projector renders no column menus, so neither changes.

Also drops what the retirement left behind: `modal_overlay`, the `.modal` CSS blocks (table and card), and the inert `nut-menu` class on the column menu.

Verified in headless Chrome: menu top = header bottom in both the 5-row and 20-row (sticky) cases.

### DIFF
--- a/src/haz3lcore/projectors/implementations/ProbeProj.re
+++ b/src/haz3lcore/projectors/implementations/ProbeProj.re
@@ -2367,52 +2367,6 @@ module M: Projector = {
     };
   };
 
-  let modal_overlay =
-      (
-        ~settings,
-        model,
-        info,
-        ~local: action => Ui_effect.t(unit),
-        ~parent,
-        ~view_seg,
-        ~sort,
-      )
-      : list(Node.t) => {
-    switch (
-      rich_content(~settings, model, info, ~local, ~parent, ~view_seg, ~sort)
-    ) {
-    | None => []
-    | Some(content) => [
-        div(
-          ~attrs=[Attr.classes(["modal-backdrop", "live-offside"])],
-          [
-            div(
-              ~attrs=[
-                Attr.classes(["modal"]),
-                Attr.on_click(_ => Effect.Stop_propagation),
-              ],
-              (
-                model.active_renderer != None
-                  ? [
-                    div(
-                      ~attrs=[
-                        Attr.classes(["modal-close-btn"]),
-                        Attr.title("Close"),
-                        Attr.on_click(_ => local(ToggleModal(None))),
-                      ],
-                      [text("×")],
-                    ),
-                  ]
-                  : []
-              )
-              @ [content],
-            ),
-          ],
-        ),
-      ]
-    };
-  };
-
   let error = (_, _): option(ProjectorBase.error) => None;
 
   let view =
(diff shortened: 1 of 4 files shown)


7h3kk1d (changes requested):
For small tables they still render offside (albeit with some unrelated odd styling), the column is entirely broken then.


https://github.com/user-attachments/assets/74353665-de1d-45d9-b903-62f3c3392a64

PR #2536, opened 2026-09-12
non-`dev` (`integration/modules-livelits`): experimental language integrations

matthewhammer:
### Try it — six links straight into the decks

This branch carries the deep links from #2583, so a URL can say which slide, which panel and where the cursor is. All six are on this branch's build — `https://hazel.org/build/experimental-lang-integration/`

**Fumola, as tiles**

1. **The whole idea in three lines** — a Fumola program inside a Hazel expression, with the instance store showing what it recorded: two cells, a thunk, and the force that read them.
   https://hazel.org/build/experimental-lang-integration/?slide=fumola-tiles-overview&panel=fumola&tab=events&select=6-10
2. **The name is not a variable** — cursor inside a second program naming the same instance, on the Edges tab with the editor's own rows shown, so you can see what a put and a read each recorded.
   https://hazel.org/build/experimental-lang-integration/?slide=fumola-tiles-instance-and-mode&panel=fumola&tab=edges&editor=show&caret=31,12
3. **A force is a traversal** — cursor on a thunk; the event list shows forcing it as a Begin and an End, with an edge to each cell it read.
   https://hazel.org/build/experimental-lang-integration/?slide=fumola-tiles-a-level-tree&panel=fumola&tab=events&caret=37,20

**User-defined livelits**

4. **Define a slider, use it twice** — `^pct(25) + ^pct(75)`, each use a live widget sitting where its value goes.
   https://hazel.org/build/experimental-lang-integration/?slide=livelits-define-a-slider&panel=none&select=51-51
5. **A livelit whose model is a data structure** — grow the tree, trim a branch, walk the seasons; everything the widget knows is written into the program text.
   https://hazel.org/build/experimental-lang-integration/?slide=livelits-tree-care&panel=none&select=269-272
6. **JavaScript in a livelit** — the view runs JS, and the expansion carries the result back into the program.
   https://hazel.org/build/experimental-lang-integration/?slide=livelits-javascript-advanced&panel=none&select=31-31

Right-click any term and choose **Copy URL** to get one of these back for wherever you are. `panel=none` closes the sidebar; `select=12-14` is three whole lines; lines and columns are 1-based, as the gutter numbers them.

The line numbers track the slide text as committed here, so if a slide is edited, re-copy the link rather than hand-editing the numbers.

---

Builds on **#2526 ("user-defined livelits and modules, integrated.")** and adds the
three experimental language integrations we have been building separately, so all of
them run in **one build — one live Hazel notebook**.

The split is deliberate, and the dependency only goes one way:

- **#2526 is the mainstream product.** User-defined livelits, Modules II parts 1–2,
  and the hazel-html substrate as opt-in `Html` / `Attr` / `Cmd` / `Sub` modules. It
  contains **no** experimental language integration — audited, zero Fumola or
  Blackboard files among its 186. That is the branch to read if you want to
  experience user-defined livelits with the new module design.
- **This branch predicts that one's success.** It takes #2526 as given and asks
  whether three independent sub-language integrations survive on top of it. If they
  do, that is evidence the module/livelit design holds up under load it was not
  written for.

## What the integration is for

A sub-language is worth embedding when its data can come back and be *looked at*.
The shape the work has settled into, and the thing to watch for on the Fumola
tiles deck:

1. **Declare the module's types in Hazel.** Not a summary of them — the actual
   types an interesting module traffics in, written in Hazel's own type language,
   on the slide beside the program.
2. **Run the module in Fumola and read the results back as those types.** The
   annotation is what decides the shape a value takes crossing the boundary, so
   the declaration is doing work rather than documenting.
3. **Probe them in place, while editing.** The value sits next to the expression
   that made it, and the instance is named in the program text, so its store
   survives every edit that leaves the name alone. Edit, and watch what the
   numbers do.

The `Fumola (Tiles) / Node info` slide is the worked example, and the module it
does this to is Adapton itself. `Adapton.peekInfo` answers what one node of the
dependency graph knows about itself: its code, its result, and both directions of
its edges. The slide declares `NodeId`, `EdgeId`, `Edge`, `Node`, `ThunkNode` and
`NodeInfo` in Hazel, and the runtime's answer arrives as Hazel values of exactly
those types —

```
(EdgeId(1006),
 (source = (Symbol(Num(3)), Now, 3), target = (Symbol(Num(2)), Now, 6),
  action = Get(1), metaTimes = (6, 6), align = Aligned))
```

— which a probe then shows in place. `Self-inspection` does the same for
`peekEvents`, the store's history in order, and exports it into a table livelit.

Two fields are deliberately `?`, and the reason is the point in miniature: a cell
may hold anything, and an `Action` constructed with `/` takes whatever its node
was called with. A closed type there would be a claim about the runtime that is
not true. The types say what is known and stop.

**What this buys that a printout does not.** The dependency graph is explicit in
Fumola and the probes are live in Hazel, and the two are better together than
either is alone: the graph says what depends on what, the probe says what the
values became, and an edit moves both at once. That synergy is the thesis of the
work, not a side effect of it.

**Where it is still short.** Editing a type annotation does not re-run the
program, so the boundary keeps the shape the previous annotation asked for
(#2557). Loading a program with the annotation already written is correct. Until
that is fixed, the "declare the type and watch the shape change" move has to be
shown by loading, not by typing.

## What is in this build

| | what it is | where |
|---|---|---|
| **user-defined livelits + modules** | the substrate, inherited from #2526 | — |
| **Blackboard MVP** | the core logic of the Blackboard proof assistant as its own tile sort | `Blackboard / 0..5` |
| **Fumola (Livelits)** | Fumola through four builtin livelits | `Fumola (Livelits) / 0..8` |
| **Fumola (Tiles)** | Fumola as a sort in the tile grammar | `Fumola (Tiles) / 0..6` |

Twenty-two documentation slides in one slide list: Blackboard 6, Fumola (Livelits) 9, Fumola (Tiles) 7.

## What the merges actually cost

Two of the three merges were not textual, and the differences are the interesting
part of this branch.

**The Fumola livelit MVP was written against the design #2526 replaces.** `Livelit.re`
diverged in opposite directions: #2526 **deleted** the `Slider` and `Emotion` builtins
(306 lines) because they are user-defined livelits now, shipped as `.hz` programs; the
Fumola branch kept them and added ~900 lines of Fumola builtins beside them.

Resolved as **`Js` plus the four Fumola livelits, and no `Slider` or `Emotion`
builtin** — a language integration is builtin, a widget is user-defined. Keeping
`Slider` as a builtin here would have quietly routed around the thing this branch
exists to test.

Three contracts widened on the way through, each one the Fumola side's need reaching
code that predates it:

- `expand` now takes `~id`, `~ana`, `~tools`. A user-defined livelit needs none of the
  three — it applies its own `expand` member — so it accepts and ignores them. That is
  what lets both kinds share one table.
- `view` now takes `~id`, so a livelit can tell two live occurrences apart. The Fumola
  livelits use it to name their VM instance.
- `requires_annotation` is false for user-defined livelits and true for the Fumola
  ones, which cannot know what to produce without the type the value is read at.

**Blackboard was almost a pure union.** Thirty-six conflicts, nearly all of them both
sides appending adjacent cases to the same exhaustive matches, form table, and
per-sort colour block — because Blackboard embeds its logic as its own sort exactly
the way Fumola does, and the ALFA derivation language did before either. The one place
that is *not* a union is `Precedence.re`, where associativity is keyed by level
**number** and shared across every sort: a level one sort marks left-associative is
left-associative for any other sort reusing it. Both ladders were checked against each
other rather than merged on faith.

## `hazel x end` carries what x is bound to

The escape that takes Hazel into a Fumola program could carry a value *written in
place* and not a variable bound outside it. A slide taught that as a limitation.
It is gone:

```
let m = Graphical in fumola hazel m end as store in 1 + 1 end     -->  2

let leaf : Tree = fumola $graphical as trip in
  { let p = 2 := $leaf(1); @ p } end in
... fumola $graphical as trip in { let p = 3 := hazel leaf end; @ p } end
```

**A Fumola program now runs during evaluation rather than elaboration**, so by the
time it is printed its escapes have been reduced to the values they name.

What blocked this was a measurement read one step too far. The probe said a cell's
result is computed in a worker and the worker has no `window.fumola`, and that was
taken to mean the worker *must* compute it. `EvalResult.calculate` has always taken
`~queue_worker: option(...)`, and `None` routes the cell through
`WorkerServer.evaluate_sync` on the main thread — `ExplainThis` already passed
`None`:

```
start_evaluation; runtime=absent  ctx=worker
evaluate_sync;    runtime=present ctx=main-thread ms=0.7
```

A cell whose statics pass saw a Fumola term is routed to the main thread; every
other cell keeps the worker. **Two things travel that way for the same reason** —
the runtime is a property of `window`, and the typing context is full of OCaml
closures since `Ctx` holds `LivelitCtx` entries. Neither crosses `postMessage`, so
`FumolaCtx` hands the context from the pass that had it to the pass that needs it,
scoped to one evaluation.

**What it costs is slicing.** The worker evaluates in 5000-step slices with
streaming updates and a 20 s timeout; `evaluate_sync` runs to completion in one go.
Measured on the shipped slides: 0.0–4.9 ms of Hazel evaluation, against ~0.4 ms
median for a Fumola run itself and ~65 ms each for `claim` and `ensureMode` the
first time an instance is named — costs elaboration was already paying on this
thread. A Fumola program sharing a cell with a heavy Hazel computation would block
the UI for all of it. That is the case to watch.

**This is where the tile route does better than the livelit one**, and the deck now
says so having checked rather than assumed. Livelit expansion still happens during
elaboration, so `^fumola_with`'s input slot holds an expression too. With no runtime
loaded a program reports which of the two steps failed, and the two answers differ:

```
^fumola_with(0, "`t", "input + input", 21)  -->  no Fumola runtime available
^fumola_with(0, "`t", "input + input", v)   -->  no Fumola source for this expression
```

**And pointers now go both ways.** A Fumola pointer arrives in Hazel as a
`FumolaPeek` — the value form carrying the reading that names the cell alongside the
value in it — because `FumolaValue` is shared between the two routes in this tree and
the term is available to both. (On `fumola-tiles-mvp` alone it does not exist, and a
pointer yields only the value it points at.) It can now go **back**, as the pointer
it is rather than the value it was holding:

```
let cell = fumola $graphical as demo in { let p = 18 := 41; p } end in

fumola $graphical as demo in { @ hazel cell end }                      -->  41
fumola $graphical as demo in { let q = hazel cell end := 123; @ q }    -->  123
fumola $graphical as demo in
  { let t = 19 := thunk { (@ hazel cell end) + 1 }; force t }          -->  124
```

The third is the one worth having: the read inside the thunk records the dependency,
so a pointer that went out through Hazel and came back is as incremental as one that
never left.

What crosses is the symbol turned back into a pointer with the prelude's `pointer` —
a bare `` `x `` is a *symbol*, and `@` on a symbol is a value of the wrong kind. The
parentheses are load bearing rather than defensive: ``@ pointer(`x)`` is a syntax error
(`@` takes an atom and reaches `pointer` before its argument) and ``@ (pointer(`x))``
is the read.

## A slide for the library the packaging already ships (#2540)

Part 2 of #2540. The runtime loads **eighteen Fumola modules** and nothing in the
deck said so, so as far as a reader could tell the only Fumola available was what
they typed themselves. Three tile forms are all it takes — a string, `import`, and
the `.` that reaches into what was imported:

```
import Seq = "fumola/collections/levelTree"; Seq.testFromList(())!
```

which builds a level tree in the store and answers a pointer to its root. That
pointer lands in Hazel as a reference, and the probe shows

```
peek(3-`binary) = Some(Binary(left = peek(2-`binary), level = 1581569,
                              right = peek(5-`binary), symbol = 3))
```

— a node of a real library structure, with its children as live references, none of
which the slide wrote. The reference goes back in through `hazel … end`, and `get`
and `peek` read through it (both take either a symbol or a pointer).

The slide also says what is **not** reachable, which is the honest half: a node is a
variant wrapping a record, taking one apart needs `switch`, and `switch` has a
printer and a parser here but no tile. So a tile program can read a library structure
and not yet destructure one.

Checked against the live runtime while writing it, and worth recording: a composite
symbol round-trips (``pointer(3-`binary)`` rebuilds the pointer the library made);
`array.popBack` does not run (`there is no prim called "arrayPopBack"` in this wasm
build) and `List.exampleList` answers a value of the wrong kind — neither is on the
slide.

Part 1 of #2540, a slide stating the supported surface syntax, is still open.

## Notes for anyone working here

- **Slide filenames are global.** Every `.hz` is copied into one flat directory and
  blobbed by basename, so `overview.hz` is a name, not a path. The Blackboard slides
  are now `blackboard-*.hz` and the Fumola tiles slides `fumola-tiles-*.hz` for that
  reason. Prefix anything you add.
- **A Hazel comment is delimited by `#`**, so a Fumola tag cannot be written inside
  slide prose. It has cost two test runs so far.
- **Doc slides are cached in IndexedDB by name.** A new slide will not appear for a
  returning browser until the `doc:` keys — including `doc:_meta` — are deleted.
- **Neither sub-language has menhir productions**, so their slides are ledgered in
  `test/Test_FastParseCorpus.re` and load through the typing parser. Remove those
  entries if either gains menhir support.

## Still to do

- [ ] A cohesive pass over the two Fumola decks: they were written independently and
      neither points at the other. Where one approach has a limitation the other does
      not, the slide should say so and name the other deck. (In flight. The tiles
      round-trip slide now names the livelit input slide and states the difference,
      with the two runtime answers it rests on; the livelit deck does not yet point
      back.)
- [ ] The Blackboard deck has not been read against the Fumola decks for overlap.
- [ ] `Tile.reassemble` assertion seen once in the app on the tiles branch and not yet
      reproduced; every prefix of every tile program parses clean, so it is on the
      deletion or movement path, not insertion.


matthewhammer:

matthewhammer:
Keeping this open as a draft.

It is not on the critical path — that is user-defined livelits (#2576, and the splicerefs variant #2580) — but it stays useful as a place to learn more about Hazel while working on other official tasks, so it is deliberately not being closed.

Draft status is the signal: not up for review, not where the focus is.

PR #2559, opened 2026-09-14
docs: write down the illustration house style

matthewhammer:
Adds `docs/illustration-style.md`: a written statement of the visual style Hazel's artwork already has.

This came out of redrawing the `mustardWatch` easter egg. Rather than invent a look for it, I went and read the work that already exists — [Week in Hazel](https://andrewblinn.com/hazel/weekly/) (covers 001, −060, −100 and the interior plates), the Noun-Project silhouettes in `Icons.re`, and the palette in `variables.css` — and wrote down what they appear to have in common, so the next piece of artwork has something to match.

**The doc is explicitly a draft reading, not a ruling.** The style it describes is mostly @disconcision's, and the point of the PR is to have him correct it. There are five open questions at the end, which are the places I am least sure of.

What it covers:

- where the reading comes from, and the happy accident that `--ui-bkg` is already within a hair of the Weekly's paper colour
- the palette, with the two rules that matter more than the hex values (no pure black or white; red is rationed)
- technique — why tone has to come from hatching rather than fills, why shading crescents beat clipped rectangles, and the `pattern`/`clipPath`/`stroke-dasharray` mechanics that make it cheap in SVG
- typography and composition — the plate, the ground rule, the small mono caps
- the two-distances rule: what has to survive at 140px, and what is allowed to be a reward for looking closer
- the viewer constraints that decide how in-app art gets built — Vdom nodes rather than markdown images (the translator has no `Omd.Image` case), `dune build @fmt`, namespaced `id`s, and the glyphs that fonts lose

No code changes; documentation only.

The redrawn mustard watch is deliberately **not** in this PR — it is on `experimental-lang-integration`, which has a demo riding on it — so this can be reviewed on its own.


matthewhammer:

matthewhammer:
Closing: not in focus.

The current critical path is user-defined livelits -- `integration/livelits` (#2576) and `integration/livelits-splicerefs` (#2580). The branch stays on origin and this can be reopened whenever the illustration work comes back around.

PR #2565, opened 2026-09-14
Notebook semantics, archivist reflection, and the measured effect schedule

matthewhammer:
> Hazel is Fumola's Strange Observer (and editor).
> Fumola is Hazel's Strange Outsider (and doer/archivist).
>
> They make a strange loop that's very exciting to me.
>
> — Matthew Hammer

Prose only — three new files under `docs/`, no code, nothing to build. They
cross-reference each other with relative links, so they land together or the
links break; that is the only reason this is one PR rather than three.

Two design notes and, now, the first one's first result.

## Why two documents

They fused in the first draft and they answer to different readers.

**[`docs/notebook-semantics.md`](https://github.com/hazelgrove/hazel/blob/docs/programmable-probes/docs/notebook-semantics.md) — the instrument.** An empirical result, which stands whether or not anyone
ever writes a probe with it.

> A Fumola instance is the first state reachable from a Hazel program that
> Hazel's own evaluator does not control.

Hazel's evaluator is deterministic and its language pure, so running an
expression twice is *by construction* indistinguishable from running it once.
That is what the live-programming story rests on, and it is exactly what makes
Hazel's own effect schedule unobservable from inside — re-running is free
because it is invisible, and invisible because it is free. Probes can't read it
(their samples are an output of the evaluation being measured). Livelits can't
(expansion is elaboration-time). `print_endline` can, and costs a rebuild per
question.

A Fumola instance can, because it outlives the evaluation. **A Hazel probe tells
you what your program computed; a Fumola instance tells you what Hazel did.**

Contains the five things it has already read — the worker routing, the
elaboration-vs-evaluation difference, the cache meaning *fewer* effects than
assumed, the display and the store disagreeing, and one slide where the editor
caused 33 of 34 events — each cited to where it was measured. Three of the five
corrected a belief held confidently first. Then the composition law as an open
question with three candidates and a stated preference, and N1/N2 to settle it.

**[`docs/archivist-reflection.md`](https://github.com/hazelgrove/hazel/blob/docs/programmable-probes/docs/archivist-reflection.md) — the capability.** A design argument, which stands whether or not
the schedule turns out to be surprising.

Hazel has two programmable surfaces and neither covers observation: livelits are
programmable and about input; rich probes are about observation and take an
OCaml module plus a mandatory `.rei` plus a registry line plus a rebuild. Both
are transient in the dimension that matters — "what values has this expression
taken *while I have been editing it*" is not a question Hazel can answer at all.

Contains the two-halves-of-a-spreadsheet framing (Hazel has liveness without
memory, Fumola has dependency without a surface), the correspondence table
between the probe system and the store, the design question of what a sample
series is keyed by, and AR1–AR4.

**[`docs/hazel-effect-schedule.md`](https://github.com/hazelgrove/hazel/blob/docs/programmable-probes/docs/hazel-effect-schedule.md) — N1's result.** The schedule, measured.

> On `experimental-lang-integration` as committed, an edit inside a
> `fumola … end` form runs the program **three times**. Not once.

Three runs, within about five milliseconds, for every keystroke.
[`65c0e889c3`](https://github.com/hazelgrove/hazel/commit/65c0e889c3) takes it to
one, and the two that disappear are `ReusePass` and `StreamCollector` —
identified by what removing them removes, since those are the only two that
commit withholds. Each wants the *shape* of a rule and pays for it by firing the
rule: free for every pure rule, a `put` for the one rule whose step reaches a
store Hazel does not own.

Two independent instruments, because agreeing with ourselves is not evidence: a
counter wrapped around `window.fumola`, and the store's own `put` count via
`prim "adaptonPeekHistory" ()`. Twelve and twelve on the baseline instance; 6/6,
2/2 and 1/1 on the other build. Established first that a peek changes nothing.

It also settles N2's premise by measurement — every edge, in every instance, on
both builds, is sourced at `Here/Now/0`: 16 of 16, 33 of 33, 2 of 2 — and gives
the idempotence prediction its first test, which it passes with a correction: a
re-run that signals nobody still *appends* two edges. Idempotent in alignment,
not in history.

## Notes for review

- **`notebook-semantics.md` loses a claim in the opposite direction.** Its draft
  said Hazel's cache means *fewer* effects than assumed. That is true of undo
  alone, measured at zero, and was the wrong generalisation — the instrument was
  built to settle it and settled it the other way. The correction is in the file.
- **The stepper was not measured.** The branch's own claim that stepping puts
  twice is *not* reproduced here and should not be cited to these documents. It
  is the highest-value row still missing.
- **The branch moved while the measurement ran.** The doc pins exactly which
  commits each column was built from, and flags that `ce5c0e5870` (12:06) has
  since changed `ensure_mode` so the `claim`→`ensureMode`→`evalTop` triple no
  longer holds. Run counts are unaffected.
- **The 33-of-34 number is called a demonstration, not a measurement.** One
  small slide is precisely where that ratio should be expected to be extreme.
- **Milestones are N1–N2 and AR1–AR4, not a shared M1–M6**, so each document reads
  standalone. AR1 states its dependency on N1 in prose rather than by numbering.
- **N2 spends a field that is currently wasted.** A node id is
  `(space, time, int)`; `source_is_here` destructures it as `[space, ..._]` and
  already ignores the time, so the time on a Here-rooted node is free — and the
  editor spends every one of them on `Now`. Stamping a distinct time per pass
  gets attribution out of ids already in the store, with no new channel, no wasm
  export and no Rust build. Because times are ordered, it yields the sequence as
  well as the labels; and because an edge spans a `metaTimes` pair, an edge
  straddling two passes is a dependency the cache carried across a boundary.
- **The composition law is not settled by fiat.** Three candidates are written
  down; the preferred one is that it is a discipline on the Fumola program
  (archivist-style ops are already idempotent, so arbitrary re-running is safe)
  rather than a change to Hazel's evaluator — and it is now stated as a
  falsifiable prediction: **a re-run that changes nothing should add no
  `Signaled` align.** The evidence in favour is labelled weak on purpose (across
  six instances the only align seen was `aligned`, which is six demo slides, not
  a sample).
- **Three links point at `experimental-lang-integration` by full URL**, since
  those files are not on `dev`. Everything else is relative and resolves.
- The branch chain is explicitly *not* what this proposes merging; see "What
  this does not propose" in either document.


matthewhammer:
Superseded by #2567, which is the same three documents on the renamed branch
`docs/archivist-reflection`.

This PR was closed by GitHub, not by a person: renaming the head branch deletes
the old ref, and a pull request whose head ref is gone cannot be reopened. No
review or comment was lost — there were none — and the commits are byte-identical.

PR #2572, opened 2026-09-15
non-`dev` (`experimental-lang-integration`): The panel header names the instance, and finds it when you point at its name

matthewhammer:
Two changes to the Fumola panel's header, found while walking the demo deck.

### Pointing at an instance's name emptied its panel

The panel said **"No Fumola instance in focus"** while the caret sat on the
very word it titles itself after. On `Fumola (Tiles) / Instance and mode`,
clicking `other` in `fumola $graphical as other in … end` — status bar reading
`FUMOLA / Instance Name`, so demonstrably inside — blanked it.

Each instance claims a set of ids and the cursor is looked up in that set.
`2077ca0567` fixed the first half of this: `Exp.map_term` walks *Hazel* terms
and never reaches the Fumola program, so the ids stopped at the boundary. It
collected the **body** and stopped there — but a `FumolaQuote` has three
Fumola children, and the other two are the **name** and the **mode**.

So the id set reached everything between `in` and `end` and neither of the two
words before `in`. The name is the likeliest thing in the whole form to point
at, being what the panel is named after and what you point at while explaining
which instance you mean.

Verified at the two carets that failed — on `other` (`FUMOLA / Instance Name`)
and on `$graphical` (`FUMOLA / Variant`) — the panel now shows
"Fumola VM instance other".

### The header now names what it shows

"Fumola VM instance" is identical on every one of them; the name is the half
that tells one from another, and the whole header was bold, so both read with
equal weight.

The label drops to normal weight and the name gets a box in
`--token-fumola-name` — the colour the instance name already reads in **as a
token in the program**, declared there as "a name rather than a keyword". So
the header does not merely spell the name out, it wears the same mark the
thing itself wears a few lines away.

Faint fill, thin border: this is a label, and the two things beside it that
*are* controls — the tab strip and the reset letters — should stay the ones
that look pressable.

`dune build @check --profile release` and `dune fmt` both clean.


matthewhammer:

matthewhammer:
Closing: the Fumola-in-Hazel effort is shelved for now.

The Fumola tiles slide deck (#2529) already demonstrates everything we wanted to show, so this refinement is not worth carrying as an open PR. The current critical path is user-defined livelits, on `integration/livelits` (#2576) and `integration/livelits-splicerefs` (#2580).

Nothing is deleted: the branch `n2/panel-title-weight` stays on origin and this PR can be reopened if the effort resumes. Fumola issues now live in the Adapton/fumola tracker under the umbrella issue Adapton/fumola#135 rather than here.

PR #2576, opened 2026-09-16
user-defined livelits, modules, and the livelit type

matthewhammer:
**Succeeds #2526**, which stops here: this branch is that one plus Modules II part 3 and the first of Cyrus's follow-ups. The pair `integration/livelits` (this) and `integration/livelits-splicerefs` are the active livelit line; everything else of mine is shelved.

## Three places to look

Deep links into this branch's deployed build, each landing on a slide with the
cursor already at the thing worth seeing. (They work because this branch
cherry-picks `07ae7f2645` from #2583, which adds `?slide=`/`?caret=`; the commit
is self-contained and touches nothing livelit-related.)

**1. `^good` and where a wrong `expand` is caught** — [The Expansion](https://hazel.org/build/integration/livelits/index.html?slide=livelits-the-expansion&caret=189,3)

Shows that this is **module typing**, not something ad hoc for livelits. The
definition is checked against the builtin `Livelit` signature with its own
`Model`/`Action`/`Expansion` realized; change `^good`'s `Expansion` to `String`
and the error appears on the definition immediately, as well as on each use.

The Expansion slide, scrolled to the ^good livelit and the commentary on where a wrong expand is caught

**2. `^mood` — the widget, the probe and the greeting together** — [Emotion](https://hazel.org/build/integration/livelits/index.html?slide=livelits-emotion&caret=104,1)

The most interesting `Action` and `Expansion` types in the deck, both with real
**algebraic structure**: `Action = + Down(Int) + MoveTo(Int) + Up` drives a
`Model = (Int, Drag)` whose `Drag = + Idle + Held(Int, Int)` is the drag state
machine, and the expansion is
`Mood = + Happy(MoodIntensity) + Sad(MoodIntensity) + Neutral`. The last line
shows all three at once — drag the face, watch the probe's `Mood` change, and
watch `greeting` turn it into a `String`.

The Emotion slide, showing the draggable face, the probe reading Happy(Very), and greeting turning it into a String

**3. `^tree` at the end of the file** — [Tree Care](https://hazel.org/build/integration/livelits/index.html?slide=livelits-tree-care&caret=274,1)

The Tree Care slide at end of file: the tree widget with season buttons, the census probe, and the result (3, 2, 0)

Goes further, and in doing so shows **the limits of what Hazel's view/update
cycle can handle today**: this one is very slow and unresponsive to interact
with. That is the honest state of it, and the work to face next — tracked in the
non-`dev` issues filed off this branch. The livelit itself is the deck's largest:
a `Model` that is a data structure rather than a number, three sub-languages in
one `Action`, and an `Expansion` that is a census rather than a picture.

One caveat on link 3: the startup scroll is computed before the projector
reserves its block rows, so it lands about seven rows short — the top of the
tree widget is visible and one small scroll brings up the rest. Its screenshot
is taken from just past that point, so it shows the livelit rather than the
bug; the bug is reported on #2583.

## What this is

dev + Modules II parts 1-3 (#2493, #2494, #2495) + the HTML substrate (#2115) + the livelit stack (#2411, #2488, #2489), with the livelit code ported to real modules and the HTML substrate as opt-in builtin modules. See #2526 for that history.

New here, from Cyrus's feedback of 2026-09-15:

**The livelit type is in the system.** `BuiltinsADT.livelit_sig`, registered as the builtin type alias `Livelit` beside `LivelitShape`:

```
{ type Model; type Action; type Expansion;
  let init : Model;
  let update : (Model, Action) -> Model;
  let view : Model -> Html.T;
  let expand : Model -> Expansion }
```

This is what needed part 3: `type Model;` with no definition is its `SigTypeAbstract`, so the signature can leave Model, Action and Expansion to each livelit and still say the four members agree about them. `shape` and helpers are absent on purpose — optional, and extra members are fine under width subtyping.

**The expansion is checked at the definition site.** `UserLivelit.detect` reads that signature back out of the context and realizes its abstract members by the definition's own manifest types, so `expand` is checked as `Model -> Expansion` with this livelit's actual types. New mark `DefMemberMismatch({name, expected, actual})`, rendered in the cursor inspector and `ErrorPrint`.

Two things it deliberately does NOT do:

- **It does not seal.** Sealing a livelit with the signature would make `Expansion` abstract, and a use of `^name` has to keep synthesizing it concretely — that is the paper's abstract reasoning principle. The check is a check, not an ascription.
- **It does not replace the use-site check.** A use-site expansion whose type depends on the model VALUE is invisible at the definition. PLDI 2021 §3.2.5 validates per use; this adds the definition-site half rather than trading one for the other.

## Verified

- `dune build @check --profile release` clean; `dune fmt` clean.
- Three new tests pin the check: a wrong `expand`, a wrong `update`, and a well-typed definition raising no mismatch. `UserLivelits` 37 tests pass; `Typ` 111, `Statics` 471, `MVU` 69.
- Twenty test livelits declared `let view = fun m -> 0`. A view returning `Int` can never render, so they were never valid — nothing had checked. They now return `Html.text("")`. No program in `hazel-programs` is affected.

## Known gaps

- **Two performance regressions from part 3, unmeasured.** `sig_project_value` now closes every member to read one — the whole-signature pass this branch previously blamed for a 3-4x slowdown — and `meet`'s `ty1 === ty2` fast path is dropped. The un-thunked `IdTag.temp` and the `subst`/`normalize` short-circuits are kept. Owed: a run of `cli analyze` against the baseline.
- **Not yet re-checked in the browser** since the part 3 merge.
- Part 3 removes the `_pat` direction variants (`ana_meet(~flipped)` becomes `subsume(~coercible)`); this branch follows part 3's own statics at the pattern site.


matthewhammer:
Status, and a correction to my own process.

This and #2580 are the two active livelit branches — this one without splices, that one with. Both have been ready for review since 2026-09-17 15:45 UTC, a little over 36 hours, with no review on either.

I went looking for why, assuming a review had been requested and missed. It had not: **neither PR has ever had a reviewer assigned**, so there was no notification for anyone to act on. That is on me, not on anyone who has not looked. Fixing it now rather than waiting longer.

@cyrus- — requesting your review here. This is the branch carrying the four items from your feedback of 2026-09-15: the `Livelit` signature as a builtin type, the definition-site check of `expand` against it, the introduction slide, and Modules II part 3 underneath. The "Three places to look" section at the top has three deep links straight to the relevant spots, each with a screenshot beside it in case the link does not cooperate.

Since it went ready for review it has also gained a **Timings** slide: a livelit that measures the responsiveness of the editor it is running in, which is the beginning of an answer to why Tree Care feels slow. It needed two new fields on the builtin `MouseEvent`, so it is a language-surface change and not only a slide — worth a look on that basis alone.

PR #2579, opened 2026-09-17
A module body is a list of items, not a nest

7h3kk1d:
A module or signature body is a list of items. It was rendered as if each item nested the ones after it, so the body arrived on a staircase:

```
{
    let a
    =
        1;
        let b
        =
            a
            * 2;
            let c
            = b
            + 1
}
```

Every binding starts two columns deeper than the last, and each one is split across three lines. This is what the stepper shows for a module at every step, and what the editor auto-indents to as you type one.

## Two causes

**`PrettySegment` has no rule for the item forms.** `build_tile_doc` switches on a tile's label — `[_, "=", "in"]`, `["if", "then", "else"]`, and so on. A module item's tile is `["let", "="]` (`Form.re:528`), which matched no case and fell through to `fallback()`, so it broke after every shard.

**`Indentation.is_incrementor` counts each item as a nesting form.** It returns true for any tile with two or more tokens that is concave on the right, which an item is. The exemption for definition forms is spelled as `ends_with_in`, a test for a trailing `in` token — an item has no `in`, so it is not exempt. Since the fold's `level` is only reset by specific cases, each item then started from the previous item's level.

## The change

- `Sort.is_item` — `Mod` and `Sig` are the sorts whose terms are body items. One predicate, used by both files below.
- `PrettySegment`: a doc rule for `[_, "="]` tiles in those sorts. Unlike a `let … in` binding, an item's definition is not a child of the tile; it is the segment up to the item's `;`. The group that keeps the item on one line is built around that split, which needed a split at a semicolon — so `split_at_comma` is generalized to `split_at(sep)` rather than growing a near-duplicate beside it.
- `Indentation`: an item's `;` resets the level to the body's own, mirroring the comma rule directly above it. A definition that genuinely has to break keeps its hanging indent, and the next item still starts level with its siblings.

```
{                            {
  let a = 1;                   let a = 1;
  let b = a * 2;               let wide =
  let c = b + 1                  [100000, 200000, 300000, 400000];
}                              let b = 2
                             }
```

## Tests

Four, each confirmed to fail with its own change reverted:

- `PrettyPrint.Modules` — the flat case, a broken definition not indenting the next item, and a nested module body.
- `Editing.Indentation.Modules` — the reset through the typing path, since that is where `is_item_sep` lives and auto-indent-on-entry is a separate user-visible behavior from rendering a result.

`make test-quick` green (3632 on this branch), `dune build @check --profile release` clean.

## Scope

Nothing here touches what a module means — only how one is laid out. It was found while stepping a module on the `modules-ii` stack, but the item syntax, `Sort.Mod`/`Sort.Sig`, and all five files this touches are already on dev and identical there, so it is independent of that work and lands on its own.

https://claude.ai/code/session_01VDmhiB9nov659nJzsBKiNc


cyrus-:
Just merged #2374 which fixed this

PR #2580, opened 2026-09-17
user-defined livelits, with splices and SpliceRefs

matthewhammer:
User-defined livelits whose **bounds are the client's own code**, edited in place inside the widget.

```
let baseline = 10 in
let shown = ^^livelit(^between((pct=50, lo=(baseline), hi=(baseline * 2)))) in
```

renders as a slider between two live sub-editors:

```
let shown = [ baseline ——●—— baseline * 2  = 15 ] in
```

Click into either bound and type. The widget reads the value while rendering the code.

## The pair

The livelit critical path runs on two branches:

- **`integration/livelits`** (#2576) — **without** splices. Modules II parts 1-3, the HTML substrate, the livelit stack, the builtin livelit signature type, the definition-site expansion check, and the livelits documentation deck.
- **`integration/livelits-splicerefs`** (this) — a strict superset, with splices.

Both descend from #2526, now closed.

## What has landed

**`splices-11` is merged** (@Negabinary, #2415) — 44 conflicted regions. That branch supplies the editing machinery: `Splice` in the grammar in all three sorts, `SubEditor.re`, and the projector `splice_view` / `splice_size` API.

On top of it, the parts that let a *user-defined* livelit reach that machinery:

- **`Html.splice(n)`** — a livelit's view, written in Hazel, can place one of its own splices. Until now `TableProj` was the only thing in the repo that had ever used a splice, and it is written in ReasonML.
- **Splices get created.** An author marks a model field by parenthesising its value: `let init : Model = (pct=50, lo=(0), hi=(100))`. The parens have to be the marker, because a splice piece prints as nothing but its content and the deck loads from text — so `init` rebuilds the splices from the parens on every load.
- **They survive interaction.** A spliced field is client-owned: `update` may not write it, so on commit it keeps the term in the syntax while unspliced fields take the new value. `SetTerm(~preserve_splices=true)` then re-attaches by id. Everything here is gated on the model actually having splices, so the other eight deck slides take none of it.
- **The caret survives an edit.** Editing a splice rewrites the program, so the code line holding the projector is rebuilt and DOM focus falls to `BODY`. `FocusEffect` now restores it, following the pattern already used for probes.
- **A bound with no value draws a hole** rather than the view's own unreduced source.
- **The widget grows** when a splice needs a second row, instead of clipping itself out of existence.

Measured over 16 edit scenarios (`scripts/deck_regress.py`, which drives a real browser): splices survive **16/16** — including empty content, type errors, unbound names and undo — with 0 residual dumps and 13/16 keeping focus.

## Two signatures

`expand` was doing two incompatible jobs under one name. It now splits:

```
LivelitFun:  let expand_fun : Model -> Expansion              a VALUE
LivelitMac:  let expand_mac : Model -> (Exp, List(SpliceRef)) a PROGRAM  (Fig. 3)
```

A definition answers to exactly one. Two *signatures* rather than one signature with two optional members, because a signature here can only say a member is **required** — optional members are expressed by omission, since a definition may always carry extra ones (width subtyping). So "exactly one of `expand_fun` / `expand_mac`" is not sayable inside a single signature; it is a question about which signature you answer to, which the module system already knows how to ask.

The typing rule for `expand_mac`, when it exists: the expansion, **abstracted over its splices**, checks at `τ₁ → ⋯ → τₙ → Expansion`, with each `τᵢ` declared at `new_splice`. That shape does two jobs at once — a splice passed as an argument is evaluated outside the expansion, so a binder inside cannot capture it, *and* the expansion is checkable without knowing splice contents. Capture avoidance and compositional typing are the same decision.

## What this is NOT

`SpliceRef` still does not exist, and the slide says so at length rather than implying otherwise:

- **The model still declares `Int`, not `SpliceRef`.** The splice is a syntactic wrapper around the Int, not a reference. "Half data and half pointers" does not hold yet — there are no pointers.
- **There is no `new_splice`.** Nothing types the hole at a declared type; it is typed only by where it sits in the model tuple, which happens to be right for `^between` and is not a mechanism. Splice count is fixed by the program text, so a livelit cannot make one at runtime and an editable *list* of splices is not expressible.
- **`expand_mac` is uninhabitable.** `LivelitMac` is registered, and `Exp` and `SpliceRef` are empty sums — in scope as types, with no values. Making them real needs a quoted-code type, quotation syntax, and `new_splice`: language work, not a member rename.
- **Splice content is lost across a save/load round trip.** Only the parens survive, which is why `init` reconstructs.

`docs/livelits-pldi2021-status.md` tracks the same gap and calls it load-bearing.

## Why the pair

The expansion story is coherent without splices — a use of `^name` has type `Expansion` however `expand_fun` is written, and that is checked at both the definition and each use. Splices change what an expansion may *contain*, not what it means. Keeping a branch without them means the simpler story stays demonstrable while the harder one is built.


matthewhammer:
Status, and a correction to my own process.

This and #2576 are the two active livelit branches — that one without splices, this one with. Both have been ready for review since 2026-09-17 15:45 UTC, a little over 36 hours, with no review on either.

I went looking for why, assuming a review had been requested and missed. It had not: **neither PR has ever had a reviewer assigned**, so there was no notification for anyone to act on. That is on me. Fixing it now.

@cyrus- — requesting your review. Read #2576 first if you are picking one: it is the same livelit work without the splice machinery, and it is the more finished of the two.

Honest state of this branch: the merge of Matt Keenan's `splices-11` onto current dev is done and green (4034 tests), and merging that into this branch is in progress — 34 conflict regions across 15 files, of which `LivelitProj` and `IncrEval` are where the livelit model and the splice model actually have to agree rather than merely coexist. `SpliceRef` as a type does not exist yet; the design for it is written down on this branch but not built. So this is worth reviewing for direction, not for completeness.

matthewhammer:
> **Superseded — read the PR description instead.** Written before the work landed, and stale in two ways: `expand` is now `expand_fun` (with `expand_mac` the macro counterpart, see below), and several gaps named here have since been closed. Kept for the reasoning, not the status.

## Splices MVP: what's missing, and the three tiers

Notes for tomorrow's 1-1. This is the status of the `splices-mvp.hz` slide, which was cut from yesterday's team demo because it isn't finished.

### The constraint the slide is about

Every livelit in the deck so far holds **plain data** in its model — Tree Care's model is a tree, the color picker's is three numbers. That isn't a stylistic choice: a model is serialized into the program text, so it can only hold serializable data.

> **Correction, added after testing.** An earlier version of this comment (and the slide itself) claimed you *cannot* write a bound that depends on the client's code. That is wrong, and the correction sharpens the argument rather than weakening it. Details below.

The `^between` slider on this slide has bounds `lo` and `hi`. The question is whether a bound can *depend on the client's code*:

```
let baseline = 10 in
^between((pct=50, lo=baseline, hi=baseline * 2))
```

**This already works today.** Run through the CLI on this branch it evaluates to `15` — correctly, 50% of the way between 10 and 20. The livelit's model argument is an ordinary expression in the client's scope, so `baseline` resolves, and the expansion sees the values. The widget's own rendering goes through the live sample, which likewise carries the model argument's evaluated value.

So the gap is *not* expressiveness. It is narrower and more interesting than that:

- The bound is ordinary program text **outside** the widget. You edit it before or around the widget, as code, the way you'd edit any argument.
- The widget cannot offer it as an **editable region inside its own view**. It has no way to say "the low bound goes here, and you can type in it."
- The widget cannot control how that code is **presented** — labelled, positioned, or styled as part of the interface.

That is what Figure 3 of the livelits paper (Omar et al., PLDI 2021) answers with `SpliceRef`, and what the paper actually claims: a splice is an expression the client wrote, **edited in place inside the GUI**, typed in the client's scope. The novelty is the in-place editing and the widget's control over presentation, not the mere ability to mention `baseline`.

Worth being precise about this, since overstating it would claim the paper solves a problem that isn't there.

### The finding: most of this already exists here

The useful surprise from digging in today is that **the hard part is already on this branch**, from the editor-splices work merged in earlier.

- Splices are a first-class piece (`Piece.mk_splice`), Convex/Convex from the outside — atomic to the parent segment, exactly like a projector.
- They get caret movement, selection, vertical navigation between splices, and splice-local context menus (`Perform.re`, `Move.re`, `ContextMenu.re`).
- **Splices are transparent to statics.** A splice's contents are typed in the *surrounding* scope. That is precisely the client-scope property `SpliceRef` needs, and it's already true.

`TableProj` is a working precedent for all of it. Its `init` returns an `init_override` of `Syntax(seg)` — the projector rewrites its own syntax at creation — and wraps each table cell in a splice. `TableCore.hole_splice()` is `Piece.mk_splice([Piece.mk_grout(Convex)])`: an empty splice, which is `new_splice(τ, None)` from Figure 3, already implemented under a different name.

So this is not "build SpliceRefs from nothing." It is "let user-defined livelits reach machinery that `TableProj` already uses."

### What's actually missing

Three small things, all in the user-defined path:

1. `LivelitProj.init` returns `Some(((), None))` — no syntax override, so a livelit's syntax never contains splices.
2. `LivelitProj.splice_rows` returns `Id.Map.empty`, and `placeholder` ignores `splice_size`, so no space is reserved.
3. The Hazel-level `Html` vocabulary has no way for a view to say *"splice n goes here."* `HazelDOM.render_elem` dispatches on tag name and already has a `view_term` path that renders a Hazel term inside a livelit view — but read-only. The interactive version of that path does not exist.

And one structural constraint that shapes the whole design, found by reading `ExpToSegment.reuse_splices`: when a projector rewrites its syntax, splices survive **only if the regenerated term still contains a `Splice` node with the same id**. Matching is by id, not by position.

That matters because `LivelitProj.commit_model` currently commits through `SetSyntax`, which replaces the projector's whole segment — so every slider drag would destroy the splices it just rendered. The existing `SetTerm(_, ~preserve_splices=true)` path fixes the mechanism, but only if the new model term still carries the splice nodes, and a livelit's `update` returns plain data with no splices in it.

The clean resolution is a rule that turns out to be exactly right semantically: **a spliced field is client-owned, and the livelit's `update` cannot overwrite it.** On commit, any field that was a splice keeps its splice; only unspliced fields take the new value. That's the same boundary the paper draws — a livelit may *read* a splice, but the code in it belongs to the client.

### The tiers

**Tier 0 — make the slide cosmetically safe.** Replace the two `<input type="number">` bounds with click-driven controls, so nothing on screen is visibly broken. The inputs can't be typed into today because `LivelitProj` declares `focusable.keyboard = None`, so keystrokes fall through to the editor and land in the program text. No other deck slide hits this, because every other livelit is driven by clicks and drags — this is the first one that asks you to type.

**Deliberately declined.** It papers over the problem without teaching anything, and the broken input is arguably the honest version of the argument: an HTML input is the *wrong* place for a bound, and the fact that you can't type into one is a symptom of that, not the disease.

**Tier 1 — `Html.splice(n)`: real editable splices in a user-defined livelit.** The three missing pieces above. `^between`'s bounds become actual holes in the program, edited in place inside the widget, typed in the client's scope. This is the target for tomorrow.

What it demonstrates, stated accurately given the correction above: not that `baseline * 2` becomes *possible*, but that it moves **inside the widget** — a region of the client's program that the livelit lays out, labels, and hands back as editable, instead of a bare argument sitting outside the interface.

A pleasant consequence of splices being transparent: because the model argument still evaluates normally, the widget can render the splice *and* use its value in the same view. So some of what I'd filed under Tier 2 may come along for free — the track could reflect the bounds without a separate `eval_splice`. I'd rather find that out by building it than promise it here.

**Tier 2 — `eval_splice`: the widget reads the client's code.** Figure 3's `eval_splice : SpliceRef -> ViewCmd(Maybe(Result))`, with `Result = Val(Exp) | Indet(Exp)`. This is the part a list-of-holes example would never motivate, and the reason `^between` was chosen over one: to draw the track, the widget must know where the track *starts*, and that's an expression the client wrote. It cannot render itself without evaluating someone else's code.

`Indet` is the honest case — blank a bound, or write one that won't reduce, and there's no track to draw, so the widget should say so rather than guess. Figure 3 disables the color preview for exactly this reason.

Not attempted tonight. Named here so it's clearly out of scope rather than forgotten.

### Where this sits on the critical path

Tier 1 is the first point where a **user-defined** livelit — one written in Hazel, in a `.hz` file, by someone who is not modifying the editor — can hold a piece of the client's program. Every splice in the codebase today belongs to a projector implemented in ReasonML.

Open question I'd like your read on: whether the splice count should be declared by the livelit definition (fixed at init, as `^between` needs — two bounds, always) or created dynamically at runtime, which is what an editable list of splices would need. Figure 3 supports the dynamic case via `new_splice` in `UpdateCmd`. The fixed case is a strict subset and is all this example requires, so that's the direction here — but it's worth knowing early if that's the wrong shape to build first.

matthewhammer:
> **Superseded — read the PR description instead.** Written before the work landed, and stale in two ways: `expand` is now `expand_fun` (with `expand_mac` the macro counterpart, see below), and several gaps named here have since been closed. Kept for the reasoning, not the status.

## Tier 1 landed: a livelit can hold a region of the client's program

Pushed as `12a057053f` and `0e9bc601da`. Full suite green locally (4532 tests). This is the follow-up to the tiers comment above — Tier 0 was deliberately skipped, and Tier 2 turned out to come mostly for free.

### What you can do now

On the rewritten Splices MVP slide:

```
let baseline = 10 in
let shown = ^^livelit(^between((pct=50, lo=(baseline), hi=(baseline * 2)))) in
```

renders as

```
let shown = [ baseline ——●—— baseline * 2  = 15 ] in
```

where both bounds are **live sub-editors holding your own code**, placed by the livelit's view with `Html.splice(0)` and `Html.splice(1)`. Click into one and type.

### Measured, cold profile, no saved state

| | splices | contents | shown |
|---|---|---|---|
| loaded | 2 | `baseline`, `baseline * 2` | `= 15` |
| drag slider to 80 | 2 | unchanged | `= 18` |
| type `1` into the right bound | 2 | `baseline * 21` | `= 110` |

Row 2 is splices surviving a model commit. Row 3 is in-place editing, with the widget reading the edited code in the same frame. (`lo=10`, `hi=210`, 50% → 110.)

### How it works

An author marks a model field by **parenthesizing its value**: `let init : Model = (pct=50, lo=(0), hi=(100))`.

The parens have to be the marker rather than a nicety. A splice piece prints as nothing but its content, and the deck loads from text, so a splice cannot round-trip — `init` rebuilds it from the parens on every load. The client's code stays inside the parens, which is what makes the rewrap idempotent.

Keeping splices alive across an interaction was the part the branch's own `TODO(splicerefs)` underestimated. `SetTerm(~preserve_splices=true)` is necessary but **not sufficient**: `reuse_splices` matches by id, and the committed model is freshly evaluated and carries no `Splice` nodes at all. Both ends had to change.

The resolution is a rule that is right rather than merely convenient: **a spliced field is client-owned, and `update` may not write it.** On commit, a spliced field keeps the term in the syntax; only unspliced fields take the new value. That is why `^between`'s `Action` has no `SetLo`/`SetHi` — it *cannot* have them. The update redex is suppressed for the same reason: it is a fresh term with no splices, and (per review) it wins on every committable interaction, not just on failure.

`placeholder` now widens by what the splices actually hold. Not cosmetic — `.livelit > *:not(svg)` is `overflow: hidden`, so an undersized box does not overlap, it clips: the splice goes invisible and unclickable while keyboard navigation still enters it.

**Every path is gated on the model actually containing splices**, so the other eight slides in the deck take none of this. Verified: no invocation in the shipped deck writes a parenthesized labeled field value.

### eval_splice came free

`at` uses `m.lo` and `m.hi` as numbers while the same view renders them as editors. Figure 3 needs a separate `eval_splice` for that. Here it falls out of splices being transparent to statics — the model argument never stopped evaluating.

### What this is NOT

The slide now says this explicitly, at length, because overstating it would be worse than not building it:

- **The model still declares `Int`, not `SpliceRef`.** The splice is a syntactic wrapper around the Int, not a reference. So "half data and half pointers" does not hold yet — there are no pointers.
- **There is no `new_splice`.** Nothing types the hole at a declared type; it is typed only by where it sits in the model tuple. For `between` that happens to give `Int`, but that is luck of position, not a mechanism. Consequence: splice count is fixed by the text, so an editable *list* of splices — the obvious next example — is not expressible.
- **`expand` returns a value, not `(Exp, List(SpliceRef))`.** Capture avoidance is sidestepped rather than solved; nothing here demonstrates the paper's argument for passing splices as arguments.

Two known walls, both written into the slide: splice content is lost across a save/load round trip (only the parens survive), and a widget action evicts the caret from a splice you are editing, because committing rebuilds the zipper from the root.

### The correction, restated

The slide previously claimed a bound could not depend on the client's code. It could, and it always could. What splices change is **location, presentation and ownership** — not expressiveness. The one-line version the slide now builds to:

> A livelit could always COMPUTE with the client's code. It could not HOLD any.

### One loose thread worth a look

`Statics.re:748` does not forward `~coercible` into a splice body, while the structurally identical `Parens` (`:703`) and `Projector` (`:713`) cases both do. Nothing in this work depends on it and I have not tried to construct a case where it bites, but the asymmetry looks unintended, and module sealing is where it would show up.

matthewhammer:
## Two livelit signatures, and the typing rule for the macro one

Design settled tonight. Pushed; the PR description now reflects it. This is the reasoning behind it.

### The problem

`expand` was doing two incompatible jobs under one name:

- `Model -> Expansion` — computes a **value**. A livelit use is a value abbreviation, and `Expansion` is an ordinary type. This is the whole deck today, and it is what makes the module-typing story legible: a use of `^name` has type `Expansion` however the livelit is written.
- Figure 3's `Model -> (Exp, List(SpliceRef))` — writes a **program**.

Those are different kinds of thing, and collapsing them loses the simple case. Making everything a macro would force quotation on every author and give up the ordinary-type story for no gain.

### The split

```
LivelitFun:  let expand_fun : Model -> Expansion               a VALUE
LivelitMac:  let expand_mac : Model -> (Exp, List(SpliceRef))  a PROGRAM
```

A definition answers to exactly one.

**Why two signatures rather than one signature with two optional members.** A signature here can only say a member is *required* — optional members are expressed by leaving them out, since a definition may always carry extra ones by width subtyping. (The comment above the old signature already said this about `shape`.) So "exactly one of `expand_fun` / `expand_mac`" is not sayable inside a single signature. It is a question about which signature a definition answers to, which the module system already knows how to ask — and Modules II is what makes asking it cheap. The `^good` slide argues that livelits use module typing rather than something ad hoc; this is that argument paying off on a real design question rather than as a slogan.

Implementation note for whoever builds it: dispatch on which expansion member is *present*, then check fully against that signature. Otherwise a typo in `expand_mac` is reported as "missing `expand_fun`", which sends the author the wrong way.

### The typing rule

Given splices with types `τ₁ … τₙ` **declared** at `new_splice(τ, _)`:

- the expansion, **abstracted over its splices**, checks at `τ₁ → ⋯ → τₙ → Expansion`
- at each use, splice *i*'s content checks at `τᵢ` in the **client's** context
- the use itself has type `Expansion`

The part worth saying out loud: **capture avoidance and compositional typing are the same design decision.** The paper passes splices as arguments so a binder inside the expansion cannot capture them — and that same function shape is what makes the expansion checkable once, against declared splice types, without knowing splice contents. One choice, two payoffs. That is a stronger argument for Figure 3's shape than capture avoidance alone.

It also fixes a wart in what is currently on this branch: today a splice's type comes from *where it sits in the model tuple*. For `^between` that accidentally gives `Int`. Under `new_splice` the hole is typed by declaration.

### Concrete syntax

One new thing, and it is not indexed by livelit kind.

- **Definitions**: no second form. Which kind a livelit is follows from which member it defines — that is what having two signatures buys. A second definition form would encode the same fact twice and let the two disagree.
- **Uses**: macro uses must carry their splices, e.g. `^^livelit(^between([baseline], [baseline * 2]))`. So one **splice delimiter**, which a `LivelitFun` use simply never contains.

That delimiter is also the fix for a real defect on this branch: a splice piece prints as nothing but its content, so splices do not survive a save/load, which is why `init` currently rebuilds them from a parenthesis marker. Give splices concrete syntax and both problems go away at once.

### Prerequisites, in order

1. `Exp` — a quoted-code type, and quotation syntax
2. `SpliceRef`, and `new_splice : (Typ, Maybe(Exp)) -> UpdateCmd(SpliceRef)`
3. statics dispatch on the expansion member present
4. the typing rule above

1 and 2 are language work. 3 and 4 are small once they exist. `LivelitMac` is registered today and deliberately uninhabitable — `Exp` and `SpliceRef` are empty sums, so `fun (r : SpliceRef) -> r` checks while `let r : SpliceRef = 0` does not. An alias to something convenient would let a definition appear to answer to `LivelitMac` while doing nothing of the sort.

matthewhammer:
**Closed by a branch rename, not by a decision — continued at #2595.**

This branch was renamed `integration/livelits-splicerefs` → `integration/livelits-splices`, and GitHub closes a PR permanently when its *head* branch is renamed (the old ref is deleted; reopening returns 422). That was known and accepted before the rename.

**Why the rename:** the branch has splices and was calling itself SpliceRefs. It does not have SpliceRefs — `Exp` and `SpliceRef` are registered but uninhabited, and there is no `new_splice`. The name now says what the branch is, and the `splicerefs` name is freed for the work that actually implements Figure 3.

Nothing is lost. #2595 has the same commits (`33da2e3df3`), the same description, and the design comment (two signatures, the typing rule, concrete syntax) carried over. This thread stays readable for its history.

Issue #2279, opened 2026-05-07
plain text URL-based loading of scratchpads


cyrus-:
I think currently we only support segment based URL sharing

PR #601, opened 2022-06-22
Playground-like permalink sharing

mirryi:
Proof-of-concept for playground-like code sharing via permalinks. Currently, it adds a "Load permalink" option to the action menu, which serializes (via the textual code representation) the current program into the URL as the hash argument.

Now we won't have to post pictures of Hazel code in issues.

- [x] Basic saving/loading functionality
- [x] Add line to undo history
- [x] Add keyboard shortcut (`CTRL-y`)

Leaving out for now:

- String compression for serialization
- Saving/loading settings

[Example fibonnaci program](https://hazel.org/build/playground-permalink/#let%25C2%25A0fib%25C2%25A0%253A%25C2%25A0Int%25C2%25A0%25E2%2586%2592%25C2%25A0Int%25C2%25A0%253D%250A%25C2%25A0%25C2%25A0fun%25C2%25A0n%25C2%25A0%257B%250A%2520%2520%2520%2520%25C2%25A0%25C2%25A0case%25C2%25A0n%250A%2520%2520%2520%2520%2520%2520%2520%2520%257C%25C2%25A00%25C2%25A0%253D%253E%25C2%25A00%250A%2520%2520%2520%2520%2520%2520%2520%2520%257C%25C2%25A01%25C2%25A0%253D%253E%25C2%25A01%250A%2520%2520%2520%2520%2520%2520%2520%2520%257C%25C2%25A0n%25C2%25A0%253D%253E%25C2%25A0fib%25C2%25A0%28n%25C2%25A0-%25C2%25A01%29%25C2%25A0%252B%25C2%25A0fib%25C2%25A0%28n%25C2%25A0-%25C2%25A02%29%250A%2520%2520%2520%2520%2520%2520%2520%2520end%250A%2520%2520%2520%2520%257D%250Ain%250Afib%25C2%25A07)

PR #2583, opened 2026-09-17
Deep links: the URL can say which slide, which panel, and where the cursor is

matthewhammer:
### Try it

Live on this branch's build — `https://hazel.org/build/deep-links/`

- **A slide, a selection and a panel at once** — the polymorphism slide, its first three lines selected, probes panel open.
  https://hazel.org/build/deep-links/?slide=polymorphism&panel=probes&select=3-5
- **One term, nothing else on screen** — the `id` binding selected, sidebar closed.
  https://hazel.org/build/deep-links/?slide=polymorphism&panel=none&select=5,5-5,7
- **A panel without a cursor** — the probes slide with the probes panel open.
  https://hazel.org/build/deep-links/?slide=probes&panel=probes&select=1-3

Right-click any term and choose **Copy URL** to get one of these back for wherever you happen to be.

Richer examples — Fumola's instance store and the user-defined livelit widgets — are at the top of #2536, which has the decks for them.

The line numbers track the slide text as committed here, so if you edit a slide, re-copy the link rather than hand-editing the numbers.

---

Hazel opens wherever your browser was last, out of IndexedDB. That is right for someone returning and wrong for someone arriving, and until now there was nothing a link could say about it: the URL carried `#debug` and the scratch share pair, and nothing else. A link meant to show someone a documentation slide landed them in the tutorial, or in a scratchpad, or on whichever slide they left open months ago.

This adds five parameters, read once at load, and a "Copy URL" row in the right-click menu that writes one for the term under the cursor.

### Try it

On your own branch's build, or locally:

```
?slide=livelits-builtins
?slide=polymorphism&panel=probes&select=3-5
?slide=tables&panel=none&caret=12,4
```

| param | example | what it does |
|---|---|---|
| `slide` | `livelits-builtins` | opens that documentation slide |
| `panel` | `probes`, `problems`, `docs`, `none` | which sidebar panel is open, or closed |
| `caret` | `12,4` | caret at line 12, column 4 |
| `select` | `12,1-14,20`, or `12-14` for whole lines | selects that span |

Right-click a term → **Copy URL** gives you the link back to it, with the sidebar as you have it:

```
http://localhost:8031/?slide=polymorphism&panel=probes&select=5,5-5,7
```

### Why

Sharing is the obvious use, but the one that pulled this out is the edit–reload loop. While working on a slide you reload it over and over, and every reload costs the same interactive navigation to get back to what you were looking at. A URL in the address bar survives the reload, and brings the panel back with it.

### Details worth knowing

- Names match on letters and digits alone, so a link can carry `livelits-builtins` rather than `Livelits%20%2F%20Builtins`. The deck is matched **as loaded** rather than as shipped, so a slide renamed or removed since that browser last visited does not throw the index off.
- A `?slide=` naming no slide is ignored rather than opening the deck somewhere arbitrary — a link gone stale should be no worse than no link.
- Panels are matched against their own constructor names (`enumerate` on the panel type), so a panel added later is linkable without touching `DeepLink`.
- A closed sidebar is stated rather than omitted (`panel=none`): a missing parameter means "leave it alone", which would hand the follower whichever panel they already had open.
- Lines and columns are 1-based, as the gutter numbers them. A column past the end of its line lands at the end of it, so `select=12-14` is three whole lines.
- The copied span is found through the same `select_enclosing_term` that "Select term" uses, so a link selects what you would have got by asking for it. Somewhere with no term — in a comment, or between two of them — falls back to the caret.
- Nothing writes to the URL. A deep link sets where you are; from then on it is ordinary state and persists like any other.

Where each parameter is applied is set by when it can be: the slide before Bonsai starts, the panel over the stored settings, and the caret from `on_startup`, the first moment there is an editor to act on.

### Not covered

`?slide=` addresses the documentation deck only. Tutorial slides carry per-slide editor state under a different model shape and are not wired up.

Nothing tells you the copy happened. Hazel's only toast is a hard-coded node inside the agent chat, and plain Copy is equally silent, so this matches it rather than inventing a second mechanism; filed separately as a use case for a real notification affordance.

### Prior art

The URL work in this repo so far carries *content*, not *location*: #601 put the program text in the hash ("now we won't have to post pictures of Hazel code in issues"), later reworked into today's `?name=&share=`; #2279 asks for that payload to be plain text. Nothing has proposed carrying which slide, which panel, or where the cursor is — this is complementary to #2279 on the same URL surface.

### Testing

`dune build @check` clean, `@test-quick` green (3687 tests). Verified in the browser on this branch: every parameter above, the copied link round-tripping back to the same term, and a stale `?slide=` leaving the browser where it was.


matthewhammer:
@cyrus-  I added you as the reviewer, but everyone is welcome to comment.

I think this feature would make live testing more streamlined, both 
 - inside the one-person dev cycle of one person working with Claude and testing its work, 
 - and across distinct people checking each other's work in other builds.

cc @7h3kk1d @Negabinary

matthewhammer:
Cherry-picked `07ae7f2645` onto `integration/livelits` (now `e322257f96`) and used it to build three curated links into the livelits deck, which are now in #2576's description. It works well. One conflict, `Globals.re`, both-added record field (`slide_name` vs that branch's `apps`) — kept both, nothing to decide.

Verified against a local build of that branch: `?slide=` loose matching resolves all seven livelit slides (`livelits-the-expansion`, `livelits-emotion`, `livelits-tree-care`, …) against their `"Livelits / <Name>"` registrations, and `?caret=L,C` lands and scrolls.

**One bug worth fixing:** the startup caret scroll is computed before projectors reserve their block rows, so on a slide with a `Block(w, h)` projector above the target, the scroll lands short by exactly those rows.

Concretely, Tree Care at `?caret=274,1` (the file's last line) settles at `scrollTop` 6144.8 against a max of 6350 — 205px, about 7 rows, which is exactly that livelit's `shape = Block(20, 7)`. `?select=269-274` lands identically, so it is the scroll computation rather than the caret path. Any slide whose target sits below a block projector will be off by that projector's height.

Not fatal — the widget's top edge is visible and one small scroll brings up the rest — but it is the difference between a link that lands and one that nearly lands. Likely the same ordering hazard as the `hydrate_current` note in the commit message: the deep link acts before the layout it is aiming at exists.

7h3kk1d:
The selection in url seems useful. We already have a way to encode the slides contents into the url. I think doing it via name or via content have different semantics depending on the use case so it's quite hard to think though who would want what/when. This is probably more useful than just screenshots though. If we ever switch to cloudflare and can afford per-build deploys I think it would be great to encode commit sha, selection, and slide content into a url as that's actually reproducible.

matthewhammer:
Agreed on the semantics, and I think the split is sharper than "hard to say who wants what".

To be sure we mean the same mechanism: the existing one is `?name=<title>&share=<compressed text>`, read by `ScratchMode.integrate_share` through `StringUtil.decompress`, and it lands the content in **Scratch**. So it is a *snapshot* — a copy of the text, detached from the deck, reproducible forever and unaffected by anything we do to the slide afterwards.

`?slide=` is a *pointer*. It names a slide and you get whatever that build ships, so it follows edits.

Which you want falls out of that:

- **Pointer** for "look at the deck as it stands" — review, teaching, a deck under active revision. Following later edits is the feature; a snapshot would go stale silently and show a reviewer the wrong thing.
- **Snapshot** for "look at exactly this" — bug reports, regressions, anything where the content is the evidence.

On reproducibility you are right, and I have a fresh demonstration. I sent three `?slide=` links into #2576 yesterday. The Deploy check went green, the commit landed in `hazelgrove/build` at 20:17:59Z — and the links still served a **two-hour-old build**, because GitHub Pages failed six consecutive times from 20:15:56 and nothing on the PR reflects that. A cold-profile browser was still landing on the old slide half an hour later. That is exactly the failure you are pointing at: a pointer link silently resolved against the wrong version, and the only reason I noticed is that I checked the served bytes rather than trusting the checkmark.

So yes to commit sha. Worth noting the blocker is the hosting, not the URL format: `hazelgrove/build` is ~5.7 GB against a 1 GB Pages cap, and every branch deploy rebuilds the whole site — which is plausibly why it started erroring during a burst of twelve deploys in fourteen minutes.

One thing that may reconcile the two rather than choosing: `caret`/`select` say **where**, `slide`/`share` say **what**, and they are independent — `DeepLink.action` just builds an ordinary editor action. So the positional params should ride on a `share` link too and give you a reproducible link that also points at a span. I have not tested that combination, but nothing in the code argues against it.

And not either/or with screenshots: #2576 carries both now, because they fail differently. A screenshot cannot go stale but cannot be clicked; a link is live but depends on a deploy chain that, as above, can quietly let you down.

matthewhammer:
Worth saying plainly, since it bears on how you read the reply above and the branch work around it: I am driving this through Claude Code. The deep-link cherry-pick onto `integration/livelits`, the scroll-bug diagnosis, and that comparison of snapshot-versus-pointer semantics were all produced that way, with me steering and checking. Same for the slide I just pushed to #2576.

I mention it for two reasons beyond disclosure.

First, it is the use case in your point about reproducible links. A one-person loop of me plus an agent generates a lot of "look at this exact thing" moments, and the failure I described — a green Deploy check over a two-hour-old build — is much easier to walk into when an agent hands you a link it has not itself opened. Content-addressed links would remove a whole class of that. I now have it check the served bytes rather than the checkmark, but that is a workaround for a URL that cannot say which build it means.

Second, it is why I would rather you push back hard on this design than wave it through. Volume is cheap for me right now; judgement is not. Your read that the two addressing schemes have genuinely different semantics is the kind of thing that is easy to paper over when work is coming out quickly, and I would rather resolve it than ship both and leave it ambiguous.

matthewhammer:
> The selection in url seems useful. We already have a way to encode the slides contents into the url. I think doing it via name or via content have different semantics depending on the use case so it's quite hard to think though who would want what/when. This is probably more useful than just screenshots though. If we ever switch to cloudflare and can afford per-build deploys I think it would be great to encode commit sha, selection, and slide content into a url as that's actually reproducible.

Hey @7h3kk1d -- this is Matthew the human, not the agent extension now ---

In the spirit of merging features that don't upset other features we have, and refining them, I advocate for 
  (1) merging this to `dev`
  (2) continuing to refine this feature --- how to "point deeply" into a specific view of a specific slide, that should be on a build, for a review or some bot interactions, etc, is a very specific use case for dev ops and agent interactions.

I apologize for the AI-generated, overly verbose messages above --- I didn't realize it was writing these, sorry

My POV right now -- I see some value in having an incomplete feature that can be refined with use, rather than waiting for some totally perfect design to emerge fully formed.  There are some URL features, but not this one, in dev.

It _almost_ seems like the agent is trying to argue that we have the perfect design here.  Clearly we don't.  But it's a useful starting point for a "deep links" feature that I think we are missing, and could use going forward.  It's been useful for me, so automate putting screenshots into PRs, or for getting the agent to understand my focus from a URL it can reproduce rather than just a screenshot that I cut and paste.

BTW, 
I'll try to minimize the bot comments going forward unless they are very short, or I've edited them myself.

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+module Sidebar = SidebarModel.Settings;
+
+/* Where a link lands.
+ *
+ * Hazel opens wherever this browser was last: the mode and the slide come
+ * back out of IndexedDB, and so does the sidebar. That is right for someone
+ * returning and wrong for someone arriving, who follows a link about one
+ * particular thing and lands in the tutorial, or in a scratchpad, or on
+ * whichever slide they happened to leave open months ago. These parameters
+ * say where to land instead. They are read once, at load.
```
7h3kk1d: This feels like I'm reading a poem or ancient translation of something

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * Nothing here writes to the URL, and nothing here is a mode of its own: a
+ * deep link sets where you are, and from then on it is ordinary state, saved
+ * like any other. Reloading without the parameters leaves you where the link
+ * put you, not where you were before it.
+ */
+
+/* Letters and digits, lowercased. Both sides of every name comparison go
+   through this, so spacing, case and punctuation cost nothing. */
+let key = (s: string): string =>
```
7h3kk1d: I don't think it's very clear what `key` is supposed to be for. Based on the comment above it seems like slide names are fuzzy matched. Not sure if it's just for that.

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+   as loaded -- not against the shipped list -- is what makes this survive a
+   slide being added, removed or renamed in a browser that has been here
+   before. */
+let slide_index = (names: list(string)): option(int) => {
+  let* wanted = slide() |> Option.map(key);
+  names
+  |> List.mapi((i, name) => (i, key(name)))
+  |> List.find_opt(((_, name)) => name == wanted)
+  |> Option.map(fst);
+};
```
7h3kk1d: I think we have List.find_index as of ocaml 5.1

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+
+/* --- The slide --- */
+
+let slide = (): option(string) => param("slide");
+
+/* Which of `names` the `?slide=` names, if any. Matching against the deck
+   as loaded -- not against the shipped list -- is what makes this survive a
+   slide being added, removed or renamed in a browser that has been here
+   before. */
+let slide_index = (names: list(string)): option(int) => {
```
7h3kk1d: In general nothing in this file is super testable. I would probably pass in the query param as an argument here or figure out some other way this can be tested without the browser.

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+
+let panel = (): option(panel_request) =>
+  switch (param("panel") |> Option.map(key)) {
+  | None => None
+  | Some("none" | "off" | "closed") => Some(Closed)
+  | Some(_) =>
+    variant(
+      ~all=Sidebar.all_of_panel,
+      ~show=Sidebar.show_panel,
+      ~aliases=panel_aliases @ [("documentation", LanguageDocumentation)],
```
7h3kk1d: It's not a big deal but I think this would be simpler if you just added the enumerate/show mapping here rather than having  to deal with aliases and the default show inside variant.

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+      "panel",
+    )
+    |> Option.map(p => Open(p))
+  };
+
+/* The settings a link asks for, over the ones that were stored. */
+let settings = (settings: Settings.Model.t): Settings.Model.t => {
```
7h3kk1d: panel request should be passed into this function

src/web/DeepLink.re
7h3kk1d: The PanelRequest type in this file is the best pattern for this. I think we'd benefit from a type that represents the entire `DeepLink` request. Then separate out the "parsing"/"fetching" from the browser's query params from the logic that actually takes this and applies it to settings and the editor.

Ideally I think we would have this somewhat subsume the scratchpad url encoding so there would be a variant that either points to an existing slide by name/index or a new scratchpad by contents. Then the selection/sidebar stuff works as described.

7h3kk1d: If we did this we should be able to test the validation/parsing logic, the editor actions, and the url generation all separately.

src/web/DeepLink.re
```diff
@@ -0,0 +1,270 @@
...
+    s,
+  );
+  let out = Buffer.contents(out);
+  /* A trailing separator comes from trailing punctuation, and says nothing. */
+  String.length(out) > 0 && out.[String.length(out) - 1] == '-'
+    ? String.sub(out, 0, String.length(out) - 1) : out;
+};
+
+/* The shortest spelling that reads back as this panel. */
+let panel_key = (p: Sidebar.panel): string =>
```
7h3kk1d: We could probably just define this function in Sidebar.re as a switch to override the default show for the panel key.

src/web/app/Page.re
```diff
@@ -846,6 +846,12 @@ module View = {
       get_log_count: _ =>
         failwith("get_log_count is deprecated, use Log.get_count_sync"),
       export_all: Export.export_all,
+      slide_name:
+        switch (editors) {
+        | Documentation(m) =>
```
7h3kk1d: Do we want to support tutorial sharing?

src/web/app/globals/Globals.re
```diff
@@ -129,7 +135,10 @@ module Model = {
   };
 
   let load = () => {
-    let settings = Settings.Store.load();
+    /* A link's `?panel=` and the rest sit over the stored settings: the
+       reader arriving is shown what the link is about, and everything they
+       have not been sent to stays as they left it. */
+    let settings = Settings.Store.load() |> DeepLink.settings;
```
7h3kk1d: I wonder if this should just be handled via an action in Main.re like the other actions are

src/web/app/globals/Globals.re
```diff
@@ -82,6 +82,11 @@ module Model = {
     font_metrics: FontMetrics.t,
     meta_down: bool,
     visible_rows: option(VisibleRows.t),
+    /* The documentation slide showing, when one is: what a copied link names,
+       and the only deck `?slide=` can address. Assembled per frame in
+       `Page.main_view`, which is the first place that knows both the mode and
+       the slide -- an editor does not know it is on a slide at all. */
+    slide_name: option(string),
```
7h3kk1d: Not blocking, but I'm not sure `slide_name` belongs on `Globals`. It's Documentation-mode state, derived per frame in `Page.main_view`, living on an app-wide record so that one editor feature can read it. (It's also under `// State:` rather than `// Calculated:`.)

An alternative: `CodeEditable` only knows the span, so it could raise a global action carrying just that, e.g. `inject_global(CopyDeepLink(span))`, and let the Page-level update, which already knows the mode, the slide and the sidebar, build the URL and write it to the clipboard. That keeps the logic where the data lives and leaves `Globals.Model.t` unchanged.

It would also take care of `~linkable: bool`, which is currently `globals.slide_name != None` computed twice in `CodeEditable` and threaded through `get_sections` / `get_all_items` / `handle_listener_key` / `view`. If the menu took the `option` (or the Page level decided whether the row appears), the bare bool goes away.

src/web/app/editors/code/CodeEditable.re
```diff
@@ -669,6 +669,49 @@ module View = {
...
+      let z = model.editor.state.zipper;
+      let syntax = model.editor.syntax;
+      let selected =
+        Haz3lcore.Selection.is_empty(z.selection)
+          ? Select.select_enclosing_term(
+              syntax.term_data,
+              syntax.measured,
+              model.statics.info_map,
+              z,
+            )
```
7h3kk1d: It's not clear to me why we default select the enclosing term as opposed to just sharing the caret position. I'm not opposed to it just unclear why

7h3kk1d (changes requested):
Feature works. Most feedback here is surface level but I think it deserves a once over before I approve. Biggest conceptual issue is we now have the scratch pad url encoding and this name-based documentation slide url encoding and they're completely separate and neither work in tutorial or other modes. Outside of that tests would be nice but possibly not necessary.

Issue #2588, opened 2026-09-19
MakeTerm is quadratic in program size: the backpack-dump caret walk, plus a second quadratic inside MakeTerm.go


Negabinary:
`MakeTerm.from_zip_for_sem` is quadratic in the size of the program. On a list
literal it dominates everything else in the front end by two orders of
magnitude: at 2000 elements it costs seconds, while statics on the same program
costs tens of milliseconds.

There are **two independent quadratics**. One is identified and has a fix; the
other is confirmed but not yet localized.

## Where the time goes

Per-stage timings, upstream `dev` @ `420ec35a`, list literal `let xs = [1, 2,
..., n] in 0`, median of 5 trials, via a standalone harness (see bottom).

| n | parse | **MakeTerm** | ·Dump.to_segment | ··Dump.to_zipper | ··unselect_and_zip | ·Segment.skel | **Statics.mk** | EvalInfo |
|---|---|---|---|---|---|---|---|---|
| 100 | 2.2 | 30.2 | 20.8 | 22.5 | 0.0 | 0.0 | 1.6 | 0.1 |
| 200 | 2.2 | 98.7 | 74.6 | 73.9 | 0.0 | 0.0 | 1.9 | 0.1 |
| 500 | 8.7 | 660.6 | 473.4 | 460.6 | 0.0 | 0.0 | 7.4 | 0.2 |
| 1000 | 30.9 | 2757.0 | 2219.3 | 1979.7 | 0.1 | 0.2 | 22.5 | 0.6 |
| 2000 | 23.6 | 11294.1 | 6507.8 | 6485.0 | 0.0 | 0.1 | 55.7 | 0.7 |
| **scaling exponent** | **0.98** | **2.00** | **1.96** | **1.92** | 0.43 | 0.71 | **1.27** | 0.97 |

(Exponent = least-squares slope of log t against log n. 1 is linear, 2 quadratic.)

So:

- **Parsing is linear** (~1.0). The menhir fast path handles a plain list
  literal fine. This is *not* the quadratic fallback parser of #2541.
- **Statics is essentially linear** and costs 56 ms where MakeTerm costs 11 s.
  Its exponent sits a little above 1 mainly because the small-n samples
  (1.6 ms) are at the timer's noise floor. Note `Statics.mk` returns
  `(info_map, elab)` from one pass, so this column is statics *and*
  elaboration together — they can't be separated without changing `Statics`.
[...]

PR #1892, opened 2025-08-21
Backpack dump tuneup

disconcision:
Issues due to Dump backpack / Backpack II interaction:
Closes #1852
Closes #1887

Issues due to TermData / Singleton labelled tuples interaction:
Closes #1888
Closes #1893
Closes #1894

Unrelated fixes: 
Closes #1868

Issue #1887, opened 2025-08-20
cursor inspector shows "whitespace or comment" for a hole when applying a function


cyrus-:
<img width="438" height="410" alt="Image" src="https://github.com/user-attachments/assets/bc5ce9a1-a707-4cad-a8e5-33f3b6d1dadc" />

(doesn't seem to always happen)

Issue #1852, opened 2025-07-31
Flickering static errors on escaping selection


7h3kk1d:
In certain contexts when you escape selection with the mouse static errors flicker. They seem to persist when evaluation is disabled.

Minimal reproduction:
```
[(true, false)]: [(Bool, Bool)]
```
Select from after the right paren to the right of `true` then click on the comma.

https://github.com/user-attachments/assets/af9c1cde-7add-4e8a-945c-58cb17e9e9d9

PR #2589, opened 2026-09-19
Dump: skip the backpack-dump caret walk when the backpack is empty

Negabinary:
Skips the backpack-dump caret walk in `Dump.to_segment` when the local backpack
is empty. Closes part of #2588.

## Why

`Dump.to_zipper` exists to put the backpack down before zipping. Its guard is
on the *selection*, so for any program with an empty selection — every batch
call, and the editor at rest — it runs a caret walk across the whole program.
Each single step costs Θ(program length): `Zipper.move` ends in
`Relatives.reassemble` → `Segment.incomplete_tiles` over the right siblings,
and `can_put_down` → `local_missing_shards` scans both sibling lists. N moves ×
Θ(N) = Θ(N²).

When the backpack is empty the walk puts nothing down. All it does is move the
caret, and `unselect_and_zip` discards the caret. So the loop provably cannot
affect the result, and this skips it.

## What this deliberately does *not* do

**It does not revert #1892.** The empty-selection guard inside `to_zipper`
stays exactly as it is, so the behaviour #1892 was fixing is untouched. This
adds an *additional* early exit for a case where the walk is a no-op.

The skip is in `to_segment`, not `to_zipper`, on purpose. `to_zipper` has other
callers — the `Dump` editor action (`Perform.re`), `AgentToolCallHandler`,
`CompositionGo` — where the resulting caret position is user-visible. Those are
untouched. `to_segment` has exactly one caller, `MakeTerm.from_zip_for_sem`,
which throws the caret away.

Note also that whenever the selection is non-empty, `to_zipper` returns its
argument unchanged, so skipping the call is the same thing. #1852's fix is
preserved by construction, not by measurement.

## Effect

Measured in the same process (so the ratio isn't exposed to machine drift) on
`let xs = [1, 2, ..., n] in 0`, median of 5 trials:

| n | MakeTerm before | MakeTerm after | speedup |
|---|---|---|---|
| 100 | 15.9 ms | 4.5 ms | 3.5× |
| 200 | 105.5 ms | 26.3 ms | 4.0× |
| 500 | 575.1 ms | 180.8 ms | 3.2× |
| 1000 | 3264.5 ms | 667.2 ms | 4.9× |
| 2000 | 8686.5 ms | 1720.0 ms | 5.1× |
| **scaling exponent** | **2.11** | **2.00** | |

**This is a 3–5× constant-factor win, not a complexity fix.** The exponent
stays at 2.0 because `MakeTerm.go` is separately quadratic; that second
quadratic is described in #2588 and is not addressed here. Absolute
milliseconds were taken on a busy shared machine and are worth ±2×; the ratio
and the exponents are the stable part.

## Correctness checking

- **Differential test over real zipper states.** For each of the ten
  documentation slides, unpersisted and sampled at every 3rd caret position
  plus a range of selection states — 2891 states, of which 2879 take the new
  skip — the resulting `Segment` is bit-identical with and without this change.
  **0 mismatches.** The comparison was validated: a sanity check on the same
  run detects 9 differences across the corpus, exactly the 9 slide boundaries,
  so it is not silently comparing everything as equal.
- **#1852 ("Flickering static errors on escaping selection")** — the issue's
  own repro program `[(true, false)]: [(Bool, Bool)]`, over 190 selection
  states: segment unchanged by this patch in all 190, and 0 states report
  static errors.
- **#1887 ("cursor inspector shows whitespace or comment for a hole")** — I
  could **not** build a reproduction. The issue is a screenshot with no
  deterministic steps and says "doesn't seem to always happen", so I cannot
  claim to have re-checked it directly. What I can say is that the parts of
  #1892 it was fixed by (`put_down_as_much_as_possible`, the linebreak branch)
  are inside `to_zipper` and are not touched here, and that the segment
  handed to `MakeTerm` is unchanged across the 2891 states above.

## Tests run

Green: `MakeTerm` (44), `TextRoundtrip` (27), `FastParse` (30), `Menhir` (114),
`Editing` (691), `Indication` (162), `Introduce` (17), `Undo` (6),
`ExpToSegment` (33), `TermData` (2), `Statics` (337), `Elaboration` (38),
`Derivation` (65), `AgentTools` (154), `ShardCrashRepro` (14), `Reassociate`
(29), `Abbreviate` (75), `Indentation` (35), `Tutorial` (11), `Autoprobe` (32),
`MultiProbe` (44).

`make test-quick`: **3628 tests, all green.**

**A full `make test` has NOT been run** — the machine was shared at the time.
That still needs doing before merge; in particular `DocSlides` and the
property-based tests are outside everything above.

## Benchmark harness

The per-stage harness behind these numbers is a small standalone
js_of_ocaml executable. It is deliberately **not** in this diff so the review
stays on the one-line change. Happy to contribute it separately as a `bench/`
target if that's useful.


disconcision (approved):
lgtm. completion-provenance removes this entire system fyi. the new system is more performant in all realistic cases

PR #2595, opened 2026-09-23
user-defined livelits, with splices

matthewhammer:
> **Replaces #2580**, which GitHub closed permanently when this branch was
> renamed from `integration/livelits-splicerefs` to `integration/livelits-splices`.
> Renaming a branch that is a PR's *head* deletes the old ref and the PR cannot be
> reopened. Nothing was lost: the description and the design comment are carried
> over here, and the commits are identical (`33da2e3df3`).
>
> **Why the rename.** The branch had splices and called itself SpliceRefs. It does
> not have SpliceRefs — `Exp` and `SpliceRef` are registered but uninhabited, and
> there is no `new_splice`. The name now says what the branch is. The real Figure 3
> work gets a fresh branch that earns the `splicerefs` name.

User-defined livelits whose **bounds are the client's own code**, edited in place inside the widget.

```
let baseline = 10 in
let shown = ^^livelit(^between((pct=50, lo=(baseline), hi=(baseline * 2)))) in
```

renders as a slider between two live sub-editors:

```
let shown = [ baseline ——●—— baseline * 2  = 15 ] in
```

Click into either bound and type. The widget reads the value while rendering the code.

## The pair

The livelit critical path runs on two branches:

- **`integration/livelits`** (#2576) — **without** splices. Modules II parts 1-3, the HTML substrate, the livelit stack, the builtin livelit signature type, the definition-site expansion check, and the livelits documentation deck.
- **`integration/livelits-splices`** (this) — a strict superset, with splices.

Both descend from #2526, now closed.

## What has landed

**`splices-11` is merged** (@Negabinary, #2415) — 44 conflicted regions. That branch supplies the editing machinery: `Splice` in the grammar in all three sorts, `SubEditor.re`, and the projector `splice_view` / `splice_size` API.

On top of it, the parts that let a *user-defined* livelit reach that machinery:

- **`Html.splice(n)`** — a livelit's view, written in Hazel, can place one of its own splices. Until now `TableProj` was the only thing in the repo that had ever used a splice, and it is written in ReasonML.
- **Splices get created.** An author marks a model field by parenthesising its value: `let init : Model = (pct=50, lo=(0), hi=(100))`. The parens have to be the marker, because a splice piece prints as nothing but its content and the deck loads from text — so `init` rebuilds the splices from the parens on every load.
- **They survive interaction.** A spliced field is client-owned: `update` may not write it, so on commit it keeps the term in the syntax while unspliced fields take the new value. `SetTerm(~preserve_splices=true)` then re-attaches by id. Everything here is gated on the model actually having splices, so the other eight deck slides take none of it.
- **The caret survives an edit.** Editing a splice rewrites the program, so the code line holding the projector is rebuilt and DOM focus falls to `BODY`. `FocusEffect` now restores it, following the pattern already used for probes.
- **A bound with no value draws a hole** rather than the view's own unreduced source.
- **The widget grows** when a splice needs a second row, instead of clipping itself out of existence.

Measured over 16 edit scenarios (`scripts/deck_regress.py`, which drives a real browser): splices survive **16/16** — including empty content, type errors, unbound names and undo — with 0 residual dumps and 13/16 keeping focus.

## Two signatures

`expand` was doing two incompatible jobs under one name. It now splits:

```
LivelitFun:  let expand_fun : Model -> Expansion              a VALUE
LivelitMac:  let expand_mac : Model -> (Exp, List(SpliceRef)) a PROGRAM  (Fig. 3)
```

A definition answers to exactly one. Two *signatures* rather than one signature with two optional members, because a signature here can only say a member is **required** — optional members are expressed by omission, since a definition may always carry extra ones (width subtyping). So "exactly one of `expand_fun` / `expand_mac`" is not sayable inside a single signature; it is a question about which signature you answer to, which the module system already knows how to ask.

The typing rule for `expand_mac`, when it exists: the expansion, **abstracted over its splices**, checks at `τ₁ → ⋯ → τₙ → Expansion`, with each `τᵢ` declared at `new_splice`. That shape does two jobs at once — a splice passed as an argument is evaluated outside the expansion, so a binder inside cannot capture it, *and* the expansion is checkable without knowing splice contents. Capture avoidance and compositional typing are the same decision.

## What this is NOT

`SpliceRef` still does not exist, and the slide says so at length rather than implying otherwise:

- **The model still declares `Int`, not `SpliceRef`.** The splice is a syntactic wrapper around the Int, not a reference. "Half data and half pointers" does not hold yet — there are no pointers.
- **There is no `new_splice`.** Nothing types the hole at a declared type; it is typed only by where it sits in the model tuple, which happens to be right for `^between` and is not a mechanism. Splice count is fixed by the program text, so a livelit cannot make one at runtime and an editable *list* of splices is not expressible.
- **`expand_mac` is uninhabitable.** `LivelitMac` is registered, and `Exp` and `SpliceRef` are empty sums — in scope as types, with no values. Making them real needs a quoted-code type, quotation syntax, and `new_splice`: language work, not a member rename.
- **Splice content is lost across a save/load round trip.** Only the parens survive, which is why `init` reconstructs.

`docs/livelits-pldi2021-status.md` tracks the same gap and calls it load-bearing.

## Why the pair

The expansion story is coherent without splices — a use of `^name` has type `Expansion` however `expand_fun` is written, and that is checked at both the definition and each use. Splices change what an expansion may *contain*, not what it means. Keeping a branch without them means the simpler story stays demonstrable while the harder one is built.


matthewhammer:
## Two livelit signatures, and the typing rule for the macro one

Design settled tonight. Pushed; the PR description now reflects it. This is the reasoning behind it.

### The problem

`expand` was doing two incompatible jobs under one name:

- `Model -> Expansion` — computes a **value**. A livelit use is a value abbreviation, and `Expansion` is an ordinary type. This is the whole deck today, and it is what makes the module-typing story legible: a use of `^name` has type `Expansion` however the livelit is written.
- Figure 3's `Model -> (Exp, List(SpliceRef))` — writes a **program**.

Those are different kinds of thing, and collapsing them loses the simple case. Making everything a macro would force quotation on every author and give up the ordinary-type story for no gain.

### The split

```
LivelitFun:  let expand_fun : Model -> Expansion               a VALUE
LivelitMac:  let expand_mac : Model -> (Exp, List(SpliceRef))  a PROGRAM
```

A definition answers to exactly one.

**Why two signatures rather than one signature with two optional members.** A signature here can only say a member is *required* — optional members are expressed by leaving them out, since a definition may always carry extra ones by width subtyping. (The comment above the old signature already said this about `shape`.) So "exactly one of `expand_fun` / `expand_mac`" is not sayable inside a single signature. It is a question about which signature a definition answers to, which the module system already knows how to ask — and Modules II is what makes asking it cheap. The `^good` slide argues that livelits use module typing rather than something ad hoc; this is that argument paying off on a real design question rather than as a slogan.

Implementation note for whoever builds it: dispatch on which expansion member is *present*, then check fully against that signature. Otherwise a typo in `expand_mac` is reported as "missing `expand_fun`", which sends the author the wrong way.

### The typing rule

Given splices with types `τ₁ … τₙ` **declared** at `new_splice(τ, _)`:

- the expansion, **abstracted over its splices**, checks at `τ₁ → ⋯ → τₙ → Expansion`
- at each use, splice *i*'s content checks at `τᵢ` in the **client's** context
- the use itself has type `Expansion`

The part worth saying out loud: **capture avoidance and compositional typing are the same design decision.** The paper passes splices as arguments so a binder inside the expansion cannot capture them — and that same function shape is what makes the expansion checkable once, against declared splice types, without knowing splice contents. One choice, two payoffs. That is a stronger argument for Figure 3's shape than capture avoidance alone.

It also fixes a wart in what is currently on this branch: today a splice's type comes from *where it sits in the model tuple*. For `^between` that accidentally gives `Int`. Under `new_splice` the hole is typed by declaration.

### Concrete syntax

One new thing, and it is not indexed by livelit kind.

- **Definitions**: no second form. Which kind a livelit is follows from which member it defines — that is what having two signatures buys. A second definition form would encode the same fact twice and let the two disagree.
- **Uses**: macro uses must carry their splices, e.g. `^^livelit(^between([baseline], [baseline * 2]))`. So one **splice delimiter**, which a `LivelitFun` use simply never contains.

That delimiter is also the fix for a real defect on this branch: a splice piece prints as nothing but its content, so splices do not survive a save/load, which is why `init` currently rebuilds them from a parenthesis marker. Give splices concrete syntax and both problems go away at once.

### Prerequisites, in order

1. `Exp` — a quoted-code type, and quotation syntax
2. `SpliceRef`, and `new_splice : (Typ, Maybe(Exp)) -> UpdateCmd(SpliceRef)`
3. statics dispatch on the expansion member present
4. the typing rule above

1 and 2 are language work. 3 and 4 are small once they exist. `LivelitMac` is registered today and deliberately uninhabitable — `Exp` and `SpliceRef` are empty sums, so `fun (r : SpliceRef) -> r` checks while `let r : SpliceRef = 0` does not. An alias to something convenient would let a definition appear to answer to `LivelitMac` while doing nothing of the sort.

